Issue #148, opened 2014-12-15
Path error compiling


malkavi:
I am compiling the github code using build.opensuse.org and I have this problem:

```
[   55s] running install_data
[   55s] creating /home/abuild/rpmbuild/BUILDROOT/lutris-unstable-0.3.x0.3.6.3-11.1.x86_64/pixmaps
[   55s] copying share/pixmaps/lutris.png -> /home/abuild/rpmbuild/BUILDROOT/lutris-unstable-0.3.x0.3.6.3-11.1.x86_64/pixmaps
[   55s] creating /home/abuild/rpmbuild/BUILDROOT/lutris-unstable-0.3.x0.3.6.3-11.1.x86_64/icons
[   55s] creating /home/abuild/rpmbuild/BUILDROOT/lutris-unstable-0.3.x0.3.6.3-11.1.x86_64/icons/hicolor
[   55s] creating /home/abuild/rpmbuild/BUILDROOT/lutris-unstable-0.3.x0.3.6.3-11.1.x86_64/icons/hicolor/scalable
[   55s] creating /home/abuild/rpmbuild/BUILDROOT/lutris-unstable-0.3.x0.3.6.3-11.1.x86_64/icons/hicolor/scalable/apps
[   55s] copying share/icons/hicolor/scalable/apps/lutris.svg -> /home/abuild/rpmbuild/BUILDROOT/lutris-unstable-0.3.x0.3.6.3-11.1.x86_64/icons/hicolor/scalable/apps
```

Example: The icons folder is being copied to the wrong path. 'share' folder is missing for all the data files. See lutris.svg file line above.

malkavi:
Fixed with commit "Remove slash at beginning of data path files"

PR #150, opened 2014-12-17
Fix path error

malkavi:
os.path.join doesn't allow the second parameter to star with '/', we need to remove the first character ('/'). Fix for #148


malkavi:
Fixed with commit "Remove slash at beginning of data path files"

PR #156, opened 2015-02-12
Support file ids in args parameter of execute directive

ezeenova:
This makes it possible to pass file ids as arguments to an arbitrary executable via the args parameter of the execute directive. This is useful when the executable in question is meant to do something with a user provided file.


strycore:
Perfect, thanks!

PR #159, opened 2015-03-03
 fix(readme): format 'reach us' block to please github parser 

edouard-lopez:
otherwise it look like Godzilla scramble the area…


strycore:
Godzilla not welcome, only Gamera allowed

PR #168, opened 2015-03-25
Dosbox fullscreen & None's

daniel-j:
Allow None choices to be selected
Added fullscreen option for Dosbox
No change -> Auto for default output screen/monitor


Xodetaetl:
But yeah, I'll merge it and we can see about that later. As for the fullscreen option, we'll need to do what I've been talking about before 0.3.7 because it's only partially working in that state.

PR #169, opened 2015-03-26
Set column width for list view

daniel-j:
TODO: save column values so they are restored when restarting lutris


Xodetaetl:
That's all about personal taste, I prefer the thin columns (although a little more padding would be better).
So your TODO task is what we need.

daniel-j:
Alright, I'll open an issue instead :)

Xodetaetl:
Thanks :)

PR #183, opened 2015-04-30
Minor fixes to documentation of "files" and "extract".

ERIIX:
Specific changes:

A file in the section "files" accepts "url" as an argument.  Previously
documented as "uri".

The "extract" directive's "file" argument can be a "file id" (as
previously documented) or a pathname.


Xodetaetl:
Committed on next, closing this.

PR #188, opened 2015-05-18
Start modernizing Gtk/GLib usage

TingPing:
As is lutris is designed very much like a Gtk2 application so I thought I would bring it up to modern Gnome standards. This work is far from done but I'd like to actually start a discussion over my changes before I rewrite the entire application :P.

Note that this branch requires Gtk 3.16.


Xodetaetl:
Yes, sorry about the delay! I'll try to review your work tomorrow

TingPing:
> Yes, sorry about the delay! I'll try to review your work tomorrow

:smile: 

> Edit: Oh wait, I think I mistook it for another pull request, no delay here so I pull back my apology. ^^

:frowning:

Xodetaetl:
So, GTK version... just checked and sadly 3.16 is much too recent. We're currently requiring v3.4 which allows support for distros as old as Debian Wheezy and Ubuntu 12.04. Which isn't that old for lambda Linux users. We should keep support for the oldest Ubuntu version supported by canonical IMO.

Hey but, perhaps we could... run Lutris itself in the runtime with the latest gtk ? :D
@strycore, you think it could work ?

Feedback on your changes after I've installed a recent distro in a VM. :)

TingPing:
>  We should keep support for the oldest Ubuntu version supported by canonical IMO.

That's just far too old for anything reasonable. Gtk3 is extremely fast moving with very good improvements each release and the design language changing drastically.

Xodetaetl:
The thing is, due to the dual client-server nature of Lutris, we'll have to have an update mechanism integrated to the client to keep it up to date indendently from the distro's packaging system. Which means that the latest version of the client has to support old distros.
That's how I see it at least.

strycore:
We do not support 'oldstable' distributions, what we do support are Ubuntu 14.04 and Debian Jessie which means gtk 3.10

strycore:
> perhaps we could... run Lutris itself in the runtime with the latest gtk ?

I wouldn't risk it, it seems extremely prone to problems

Xodetaetl:
I cherry-picked and committed "Remove dependency on pyxdg", which was not dependent on Gtk 3.16. https://github.com/lutris/lutris/commit/5718e1a7edaa623cdc6b166257e81c0b59a3f388
Also fixed some of the deprecation warnings.

RobLoach:
Might be good to check out the work over at https://github.com/lutris/lutris/pull/281

TingPing:
Closed by 1d397c19a24b454e8162a84f12d30a7ea8b12683

PR #205, opened 2015-10-14
Fix installation

malkavi:
Fix for missing module after installation


malkavi:
Sorry, this is for "next" branch

PR #210, opened 2015-11-02
Update Wine to the latest development version

RobLoach:
1.7.54


strycore:
Hopefully, this will be made obsolete soon. Just a couple things to change and we can use the 'default' field made available by the API (https://lutris.net/api/runners/wine). Still, thanks for the update :)

PR #225, opened 2015-12-12
Some spec file improvements

akien-mga:
Those changes should be pretty harmless, but I have no real experience with OBS or cross-distro RPM packaging, so you should probably try them first :)

### DIFF
--- a/lutris.spec
+++ b/lutris.spec
@@ -31,15 +31,14 @@ Requires:       python-gobject, python-gtk, python-PyYAML
 
 
 %description
-Install and play any video game easily
- Lutris is a gaming platform for GNU/Linux. Its goal is to make
- gaming on Linux as easy as possible by taking care of installing
- and setting up the game for the user. The only thing you have to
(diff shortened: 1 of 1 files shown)


akien-mga:
Also I have a few more Requires in my Mageia package so that it runs properly out of GNOME, so I wonder if your RPMs might not be lacking some stuff: http://svnweb.mageia.org/packages/cauldron/lutris/current/SPECS/lutris.spec?view=markup&pathrev=909684
I'm not 100% sure if all those deps are still required though, but I had issues at some time with 0.3.6.x on Mageia and had needed to add them (e.g. gvfs or glib-networking).

strycore:
Thanks a lot! I'll see if this fixes the Fedora builds. We dropped the dependencies on gvfs and glib-networking so you can safely remove them on your side.

strycore:
Seems like something in your pull request broke the builds for openSUSE :/
https://build.opensuse.org/package/show/home:strycore/lutris#

Issue #227, opened 2015-12-12
Error  building on Debian 8 / SteamOS


ProfessorKaos64:
Error building on Debian 8 with `dpkg-buildpackage -rfakeroot -us -uc` Seems the sdl library isn't spec'd with a symbols file to that binary. How is your team building packages? I am building for SteamOS mainly, and yes I am aware you provide deb packages on your site. I wish to build this package form the master tree though. I don't want to use a hacky LD_LIBRARY path or override_dh_shlibdeps if I can help it.

error:
```
dh_shlibdeps -plutris    
dpkg-shlibdeps: error: couldn't find library libSDL2-2.0.so.0 needed by debian/lutris/usr/share/lutris/lib/koku-xinput-wine.so (ELF format: 'elf32-i386'; RPATH: '')
dpkg-shlibdeps: error: cannot continue due to the error above
Note: libraries are not searched in other binary packages that do not have any shlibs or symbols file.
To help dpkg-shlibdeps find private libraries, you might need to use -l.
dh_shlibdeps: dpkg-shlibdeps -Tdebian/lutris.substvars debian/lutris/usr/share/lutris/lib/koku-xinput-wine.so returned exit code 2
/usr/share/cdbs/1/rules/debhelper.mk:277: recipe for target 'binary-predeb-IMPL/lutris' failed
make: *** [binary-predeb-IMPL/lutris] Error 2
dpkg-buildpackage: error: fakeroot debian/rules binary gave error exit status 2
```

library check:
```
desktop@steamos:~/build-lutris-temp/lutris$ ldd share/lutris/lib/koku-xinput-wine.so 
        linux-gate.so.1 (0xf7736000)
        libSDL2-2.0.so.0 => not found
        libstdc++.so.6 => /usr/lib/i386-linux-gnu/libstdc++.so.6 (0xf7614000)
        libgcc_s.so.1 => /lib/i386-linux-gnu/libgcc_s.so.1 (0xf75f6000)
        libc.so.6 => /lib/i386-linux-gnu/libc.so.6 (0xf7484000)
        libm.so.6 => /lib/i386-linux-gnu/libm.so.6 (0xf743f000)
        /lib/ld-linux.so.2 (0x565f8000)
```

ProfessorKaos64:
Update, libsdl2-2.0-0:i386 was a missing build dep in my case. Can this be added to the control file? If not, can you please tell me host best to build? Thank you.

strycore:
[...]

PR #228, opened 2015-12-12
correct build dependency for koku-xinput-wine.so

ProfessorKaos64:
- koku-xinput-wine.so library symbols check requires the 32-bit library for libsdl2-2.0-0
- See: https://github.com/lutris/lutris/issues/227


strycore:
Thanks!

PR #241, opened 2015-12-29
Add ZDoom Runner

RobLoach:
Could add a few different command line arguments in there:
http://zdoom.org/wiki/Command_line_parameters

Added [The Adventures of Square](https://lutris.net/games/the-adventures-of-square/) as a demonstration of using the ZDoom runner.


RobLoach:
I understand this is embedded in the Game Launchers themselves currently, but having ZDoom as a runner will make DOOM launchers easier to use and create.

PR #256, opened 2016-02-23
update debian/ to use cleaner pybuild system

ProfessorKaos64:
* Uses [pybuild](https://wiki.debian.org/Python/LibraryStyleGuide) build system for clean builds
* Makes use of debhelper (python), `dh-python`

This is just a thought, and I recall you use OpenSuSe to build, so I understand if this isn't accepted. Tested with clean-room building utilizing pbuilder.


debian/control
```diff
@@ -1,15 +1,18 @@
...
+ libc6:i386,
  python,
  dh-python,
  gir1.2-gtk-3.0,
  gir1.2-glib-2.0,
  python-gi,
- libgirepository1.0-dev,
-Maintainer: Mathieu Comandon <strider@strycore.com>
+ libgirepository1.0-dev
+Maintainer: Michael DeGuzis <mdeguzis@gmail.com>
```
RobLoach: Maintainer change?

ProfessorKaos64: Oh shoot, sorry. you can change that if you accept this or I can commit again/squash.

strycore: yep, can you add another commit with the original maintainer address and with the i386 libs removed (OBS doesn't provide multi-arch). Other than that it looks cool, I'll test in for next release.

ProfessorKaos64: I don't think it will build without them. The culprit is koku-xinput-wine, which I am now familiar with, as I have a Debian package for it now. I'll make sure though. That is why they are there. 

On February 24, 2016 6:09:44 PM EST, Mathieu Comandon <notifications@github.com> wrote:
>>   python,
>>   dh-python,
>>   gir1.2-gtk-3.0,
>>   gir1.2-glib-2.0,
>>   python-gi,
>> - libgirepository1.0-dev,
>> -Maintainer: Mathieu Comandon <strider@strycore.com>
>> + libgirepository1.0-dev
>> +Maintainer: Michael DeGuzis <mdeguzis@gmail.com>
>
>yep, can you add another commit with the original maintainer address
>and with the i386 libs removed (OBS doesn't provide multi-arch). Other
>than that it looks cool, I'll test in for next release. 
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/lutris/lutris/pull/256/files#r54023796

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

strycore: Ok, so I'm going to release 0.3.7.3 today (using the older packaging system, the packages are already built) and then remove koku-xinput-wine from the project and move it to the lutris runtime, where it belongs. This should solve the issue. Keep an eye for activity on the Next branch in the next few hours.

ProfessorKaos64: @strycore , ok will do, i'll try a build later tonight or tomorrow morning. Thank you

strycore: All the dependencies listed are included in the runtime (in both 32bit and 64bit) and they shouldn't affect the packaging process in any way.

Issue #274, opened 2016-04-22
Use a dark theme by default ?


MrTimscampi:
This is a proposal, not really an issue.

I made a quick change to run Lutris with the dark variant of the user's GTK theme by default.
Imho, it looks much better with the current interface and makes it feel a little bit more like Steam.

It's completely standard, so will work for all users.

Here is how it looks with the default Adwaita theme:
![screenshot from 2016-04-22 20-26-55](https://cloud.githubusercontent.com/assets/1087450/14751108/97128d2e-08c8-11e6-8c5e-5b868ec4601d.png)

And with the latest version of Arc:
![screenshot from 2016-04-22 20-27-42](https://cloud.githubusercontent.com/assets/1087450/14751126/aeac8dea-08c8-11e6-94e5-eb1ef61e8566.png)

If there is interest, I can send a pull request (It's already up in the "dark-theme" branch of my fork) and re-locate the code if need be (It's currently in lutriswindow.py, right after the regular settings)

strycore:
I'm so used of seeing Lutris with a dark theme (currently using Arc dark) that I didn't think about it! ^^ 
Please submit a pull request, I'll gladly accept it

MrTimscampi:
Done :)

PR #275, opened 2016-04-22
Set GTK to use the dark variant of the user's theme by default

MrTimscampi:
As mentioned in #274, this tells GTK to use the dark variant of the user's theme by default for Lutris.


strycore:
Thanks!

PR #281, opened 2016-04-23
[WIP] Update to GTK3.14+

MrTimscampi:
_**Don't merge this while there is a [WIP] prefix to it. It will break things.**_ (You probably know that, but just to be sure :p )

This is a pull request to track my update of the UI to GTK 3.14+ and a general cleanup of the whole UI.

As discussed with strycore on IRC, when practical or needed, I'll implement a version check for higher GTK versions and handle things differently for Ubuntu 16.04 (GTK 3.18) and newer distros, in order to keep deprecated things to a minimum and making the update easier once SteamOS updates from 3.14 to a newer version.

## Todo list

- [x] **04/23/2016** - Initial porting to Gtk.Application()
- [x] **04/23/2016** - Redesign the main interface
    - [ ] Fix major breakages

### Initial porting to Gtk.Application()

This takes the patch from TingPing and updates it to work with the current code.
It doesn't use anything more recent than GTK 3.0 at the moment, so it shouldn't pose any problem.

### Redesign the main interface

screenshot from 2016-04-23 19-55-21
screenshot from 2016-04-23 20-42-50

Another (partial) port of a patch by TingPing.
This changes most of the main UI to implement modern GTK interface design.
It also manages to break most things, due to being such a huge change. I somehow managed to break the settings writing, so pretty much nothing works properly. All of this will be fixed in future commits.
This is currently untested on GTK 3.14. I need to spin up a SteamOS VM and see how it handles things, but I want to fix basic operation first.

### DIFF
--- a/bin/lutris
+++ b/bin/lutris
@@ -12,29 +12,37 @@
 #
 # You should have received a copy of the GNU General Public License along
 # with this program.  If not, see <http://www.gnu.org/licenses/>.
-
+'''
 import dbus
 import dbus.service
+'''
 import os
 import sys
+'''
 import logging
 import optparse
+'''
 import signal
+'''
 import time
 import json
+'''
 
 # pylint: disable=E0611
-import gi
-gi.require_version('Gdk', '3.0')
-gi.require_version('Gtk', '3.0')
 
+import gi
+# gi.require_version('Gdk', '3.0')
+gi.require_version('Gtk', '3.0')
+'''
 from dbus.mainloop.glib import DBusGMainLoop
 DBusGMainLoop(set_as_default=True)
 
 from gi.repository import Gdk, Gtk, GObject, GLib
+'''
+from gi.repository import Gtk
 
 from os.path import realpath, dirname, normpath
-
+'''
 LAUNCH_PATH = dirname(realpath(__file__))
 if LAUNCH_PATH != "/usr/bin":
     SOURCE_PATH = normpath(os.path.join(LAUNCH_PATH, '..'))
@@ -50,7 +58,9 @@ from lutris.game import Game
 from lutris.gui.installgamedialog import InstallerDialog
 from lutris.settings import VERSION
 from lutris.util import service
-
+'''
+from lutris.gui.application import Application
+'''
 # Support for command line options.
 parser = optparse.OptionParser(version="%prog " + VERSION)
 parser.add_option("-v", "--verbose", action="store_true", dest="verbose",
@@ -112,9 +122,12 @@ check_config(force_wipe=False)
 
 installer = False
 game = None
+'''
 
+app = Application()
 signal.signal(signal.SIGINT, signal.SIG_DFL)
 
+'''
 # D-Bus init
 
 bus = dbus.SessionBus()
@@ -197,3 +210,6 @@ if game_slug or installer:
 lutris.run(int(time.time()))
 if lutris.is_running():
     Gdk.notify_startup_complete()
+'''
+
+app.run(sys.argv)
--- a/lutris/gui/application.py
+++ b/lutris/gui/application.py
@@ -1,0 +1,117 @@
+#!/usr/bin/python2
+# -*- Mode: Python; coding: utf-8; indent-tabs-mode: nil; tab-width: 4 -*-
+
+import logging
+import json
+
+from gi.repository import GLib, Gio, Gtk
+
+from lutris.util.log import logger
+from lutris.config import check_config
+from lutris.game import Game
+from lutris import pga
+from lutris.settings import VERSION, GAME_CONFIG_DIR
+from lutris.gui.installgamedialog import InstallerDialog
+from lutris.gui.lutriswindow import LutrisWindow
+
+
+class Application(Gtk.Application):
+    def __init__(self):
+        Gtk.Application.__init__(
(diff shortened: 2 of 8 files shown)


RobLoach:
Looks beautiful! Well done :+1: . I'm getting this error when running it:

```
  File "bin/lutris", line 64, in <module>
    from lutris.gui.application import Application
ImportError: No module named application
```

MrTimscampi:
Hm, weird, I didn't get that error at all, even with a clean install...
I did get a few other errors that I tamed down, but not that one...
Was application.py properly copied by setup.py for you ?
I'll have to make a few VMs for testing on something other than Arch, when I'm a bit further along.

I've had to change the connection menu (which didn't work anyway by a "profile"/connection toggle button and a refresh button, like this:
screenshot from 2016-04-24 01-09-45
TingPing's patch used a 3.16 feature (PopoverMenu) so I had to fall back to a Popover, but I get warnings if I spawn more than one. However, looking at the documentation, it may be a misuse on my end. I'll have to investigate further. But I think that refresh button makes more sense there than hidden in a menu somewhere, so it'll stay.

My current plan is to get the current remade UI working, then hunt down bugs before going further and doing testing on other things.
But that PopoverMenu is the kind of thing where I'll most likely do two implementations with a version check, one for 3.14 and one for 3.16+, depending on what PopoverMenus bring to the table (I'm still relying heavily on the doc, as I'm learning Gtk while doing this)

TingPing:
On a non-technical note, I would like credit for work since so far you are just rebasing it.

TingPing:
I probably should have put a license header on files I created, like for `application.py`:

```python
# -*- Mode: Python; coding: utf-8; indent-tabs-mode: nil; tab-width: 4 -*-
# Copyright (C) 2015 Patrick Griffis <tingping@tingping.se>
# This program is free software: you can redistribute it and/or modify it
# under the terms of the GNU General Public License version 3, as published
# by the Free Software Foundation.
#
# This program is distributed in the hope that it will be useful, but
# WITHOUT ANY WARRANTY; without even the implied warranties of
# MERCHANTABILITY, SATISFACTORY QUALITY, or FITNESS FOR A PARTICULAR
# PURPOSE.  See the GNU General Public License for more details.
#
# You should have received a copy of the GNU General Public License along
# with this program.  If not, see <http://www.gnu.org/licenses/>.
```

Xodetaetl:
TingPing, I've added your name to the [global copyright notice](https://github.com/lutris/lutris/blob/master/AUTHORS).

I prefer centralized copyrights than having them in every file, maintaining them would be a drag. Git is normally here to keep track of the specifics of who did what... But I don't know what to do in this case. Git doesn't allow multiple authors for a single commit.

MrTimscampi:
I took over on this as there wasn't any activity on the other PR for months and it targeted 3.16, which is 2 version to high from where support will be for 0.4.0 (Ubuntu 16.04 ships with 3.18, but SteamOS is stuck on 3.14).
I haven't done any compatibility checking yet, so those two commits are basically copy-paste of TingPing's code in the new source and trying to make it work (A lot has changed since that PR was made)
I'm starting to change things a bit, mainly with the popovers for now, as PopoverMenu was introduced in 3.16, as were quite a few things used in those popovers :)
The interface .ui itself was also redone in Glade from scratch, based on TingPing's.

However, if TingPing wants to resume his work, I'll gladly close this PR and work on something else, no issues on my end :) You'll most likely do it much faster than me, as well :p (I'm still learning GTK and this is the first open source project I contribute code to)

Xodetaetl:
> so those two commits are basically copy-paste of TingPing's code in the new source

If so, you could try to prepend his actual commits to yours (on a clean branch, using cherry-pick to add his commit then yours).

TingPing:
>  Git is normally here to keep track of the specifics of who did what... But I don't know what to do in this case. Git doesn't allow multiple authors for a single commit.

Yea just unfortunate they couldn't be clean rebases.

> However, if TingPing wants to resume his work, I'll gladly close this PR and work on something else, no issues on my end :) You'll most likely do it much faster than me, as well :p (I'm still learning GTK and this is the first open source project I contribute code to)

I didn't want to do the extra work [and worse results] that came with compromising on the version, that is why I let that branch sit until they could target a new version. That said I might monitor your progress here or idle on #lutris if you have any questions.

MrTimscampi:
@Xodetaetl: What I meant is that the code he added was transposed into the new source (No all of it). It's not a direct copy-paste of the actual commit.
The files have changed a LOT over those past few months, it seems, so I don't really know how to that another way than how I did it...

Aside from the UI design, which I redid from scratch using his original design as a guide and the .ui files for the popovers, which I'm going to rewrite from scratch due to them using 3.16 concepts, the rest can't really be done any other way, as I understand it (A Gtk.Application() will always be created the same way, for example).   
Now, of course, credit where credit is due. But as I said, I don't see how to prepend TingPing's commits and not remove months of works from the original source.
Then again, I'm not that good at Git. Actually, I might've chose too big of a fish to fry, as this is what I use to learn GTK and it's actually the first big code base that I haven't written which I try to contribute to.

I'll close this PR and reassess the situation.

RobLoach:
Closed the issue, and deleted his own user account?

strycore:
Now, that's very weird!

Hey, come back! You were doing great!

Issue #292, opened 2016-06-09
Suggestion: Allow custom icon / banner


althonos:
Hi ! Title says it all, I've been using Lutris with ROM-hacked games and as it is just the different version of the same game, it shares banners and make it a bit confusing...
So why not look for banners and icons named .icon and .banner in the game folder, only falling back to default ones if those files are absent / not working ?

jtrees:
I would also be interested in being able to set custom banners for games that were added manually.

strycore:
The current implementation available in the master branch allows to set custom banners for all games :)

jtrees:
Oh sweet. I look forward to the release.

PR #294, opened 2016-06-12
Tweaked the gameviews.

althonos:
## Overview

I'm using **Lutris** as a frontend for emulation, and I enjoy playing Rom Hacks, so I was quite pleased with the "install (add) another version" option. Yet I ended up having a library full of odd-looking duplicated entries which were not the best (see #292), so i forked and did the following : 
#### Added custom banner & icon

Now the gameviews.py pixbuf loading function will load in priority files called **banner.jpg** or **banner.png** (or whatever format Gtk can load) located in the same folder as the 'main file' / 'iso' (I registered only those two cases, but I don't know how other runners game_config may provide the directory of the main file...).
capture d ecran_2016-06-11_23-47-21
#### Added more size options

As the largest repository for game banners is probably [Steam Banners](http://steambanners.booru.org/), I thought it would be a good idea to provide support for those banners in Lutris so that they wouldn't appear stretched or anything ([lutris.net](http://lutris.net) and Steam using banners of different aspect ratio). To do so, I added more size options in the Icon style (renamed Game Art Style) menu :

| Size | Aspect ratio | Width | Height |
| --- | --- | --- | --- |
| Banner (Large) | 2,14 | 460 | 215 |
| Banner (Normal) | 2,14 | 230 | 108 |
| Banner (Small) | 2,14 | 185 | 86 |
| Banner (Tiny) | 2,14 | 121 | 56 |
| Wide Banner (Small) | 2,66 | 184 | 69 |
| Wide Banner (Tiny) | 2,66 | 120 | 45 |
| Icon (Normal) |  | 32 | 32 |
| Icon (Large) |  | 64 | 64 |

This allows for an enhanced compatibility with high resolution screens where art of higher resolution can be displayed, while legacy lutris users who wish to continue using the current banner format can use the _wide banner_ options.
On my 1366x768 screen, **Banner(Large)** looks like the following : 
capture d ecran_2016-06-12_14-48-56
And **Banner (Normal)** like so : 
vuuus
#### Corrected column not resizing bug

In List view, column size did not update after changing the size of banners (see below)
before - on large
before - on small
Workaround was to pop and add a new colum to force the game view to update.

### DIFF
--- a/lutris/gui/config_dialogs.py
+++ b/lutris/gui/config_dialogs.py
@@ -45,10 +45,7 @@ class GameDialogCommon(object):
         if config_level == 'game':
             self._build_info_tab()
             self._build_game_tab()
-            
-            #if self.game is not None:
             self._build_art_tab()
-
         self._build_runner_tab(config_level)
         self._build_system_tab(config_level)
 
@@ -78,9 +75,6 @@ class GameDialogCommon(object):
         self._add_notebook_tab(info_sw, "Game info")
 
     def _build_art_tab(self):
-
-
-
         
         # Art box: main container of the notebook page
         self.art_box = Gtk.Grid(column_homogeneous=True)
@@ -124,11 +118,8 @@ class GameDialogCommon(object):
             self.banner_folder_btn = Gtk.FileChooserButton()     
             self.banner_folder_btn.set_action(Gtk.FileChooserAction(2)) #Equivalent to SELECT_FOLDER
             self.banner_folder_btn.set_margin_right(20)
-            
-            #self.banner_folder_btn.connect('update-preview', self.on_banner_folder_changed)
-            #self.banner_folder_btn.connect('current-folder-changed', self.on_banner_folder_changed)
             self.banner_folder_btn.connect('selection-changed', self.on_banner_folder_changed)
-            #self.banner_folder_btn.connect('file-set', self.on_banner_folder_changed)
+            
 
             self.banner_default_btn = Gtk.Button(label="Default")
             self.banner_default_btn.set_margin_left(20)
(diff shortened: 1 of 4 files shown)


strycore:
This is really awesome and goes in the direction of 2 very requested features (bigger banners and custom banners). Since this is a pretty big change, I would make this a feature for Lutris 0.4 so you should prepare to send a Python 3 compatible pull request at some point ;) 
There have been some prior contributions heading in that direction but so far @Xodetaetl has been managing most of it.
I plan to write a scrapper for http://steambanners.booru.org/ so we should have them all available on our website at some point in the future (with the addition of Steam banners).
The only detail I disagree with is using a banner.jpg in the game's folder. In the case of ROMs, it is often the case to have all games for a system in the same folder so this wouldn't work. Also, I don't like the idea of cluttering the user game files with app specific stuff (Thumbs.db and .DS_Store comes to mind) We haven't decided yet how to implement custom banner and icons but as I said, it's an implementation detail, we'll find a way.
Also, I'm looking into the possibility of adding overlays on banners, which would be useful for ROM hack or for differentiating several games with the same name.
Don't worry if your Pull request takes a bit of time to get integrated into the master branch, it *will*, this is definitely something we want in Lutris! Thanks again for your contribution, it is very valuable!

althonos:
Hi there !
I'm glad to see I went towards the good way, I had feared I'd spent an afternoon for nothing. Python3 is no prob for me, I'm a Py3 user so it's more Py2 that hurts when I've to use it.

As for the custom banner, I just wanted to provide the user an easy way to add new banners without the need to go to data files... So to change that I see the following design choice: 
- **Store the path** to the custom banner & the path custom icon, leaving the user to manage their fanart collection the way the want, 
- **Provide a way to change** / remove that path (probably in the gamedialog _Game Info_ tab) 
- Make the default folder to **look into the folder** where the scraper would download the files for that particular game (like instead of having .local/share/lutris/banners/whatever-game.jpg, maybe have .local/share/lutris/banners/whatever-game/*.jpg with all files downloaded by the scraper, lutris selecting the first one of those by default if no others are provided).
- Instead of just having a FileChooser widget, **have a GtkIconView** inside the _Game Info_ tab to display the scraped banner directly from the directory, and add a "+ Add your own.." button or something like that to select real custom ones. (A bit like the wallpaper config screen if you see what I mean :)
capture d ecran_2016-06-13_09-44-12

For overlays, I guess an idea would be to add a wrapper (maybe decorator) around `get_pixbuf_for_image` to add an overlay after loading whatever image is selected, and have a .local/share/lutris/overlays directory to store templates in...

I can continue working onto my branch 'til we reach a satisfying point of merging, if that's ok with you :)

althonos:
Hi there ! 
So after what you told me I did the following:

### Changed the way custom icon & banner are store
Now, instead of registering an `asset_directory` for every game when writing them with `pga`, games can now have two optional attributes `banner` and `icon` which are the path to custom banner and custom icon, and is set to None if default banner / icon is to be used.

### Added and **Artwork** tab to the configuration edit notebook
I added an artwork tab which is rendered only if GameDialog is loaded with a game value (so it will be hidden with Add Game Dialog except if using `install another version`). Showing it for any AddGameDialog would probably confuse the user, as he'd think that setting a banner is mandatory, whereas lutris actually provides banners itself.

I used a Gtk.Grid item as I saw Gtk.Boxes were deprecated, so the day you want to migrate Boxes to Grids you won't have to worry about that tab.

Here's how it looks like right now:
artwork
You can also expand the window in width to expand the banner view:
artwork large

Only a row and half of banners is visible currently as I wanted to leave some space for the IconView below without raising the DIALOG_HEIGHT, but depending on how it looks after I add IconView i'll tweak that so it looks nicer.


### Default banner needs a format change
Instead of having default banner in `.local/share/lutris/banners/game_slug.jpg`, it would be better to have default banners `.local/share/lutris/banners/game_slug/default.jpg`. That way, by default the artwork banner view will show content of `.local/share/lutris/banners/game_slug`, allowing user to choose other banners in that folder, folder we could populate with a Steam Banners scraper.

### Icon path
I'll add the same kind of window for icon path, maybe smaller to only show 64x64 icons.

althonos:
Closing since this is now obsolete and far behind the current state of the repo.

PR #312, opened 2016-07-10
Fix zdoom file and executable args

RobLoach:
Adds the `-file` argument to load multiple WAD files, as well as fixes the runner location.


RobLoach:
@strycore I've added an installer using this for Doom: https://lutris.net/games/the-ultimate-doom/

Know why it is not functioning correctly? I'm finding the working directory is not taking effect, which makes it not able to find the WAD file correctly.

PR #333, opened 2016-08-24
Add retroarch.play()

RobLoach:
This will append all the arguments together to play RetroArch with the given core and file.


strycore:
Nice, thanks! Btw, it should probably throw an error when no core or ROM is provided.

RobLoach:
> Btw, it should probably throw an error when no core or ROM is provided.

Technically, you can run RetroArch without a core or a ROM. Some games are just the core (like Dinothawr).

strycore:
Yes, there is an attribute to set in runners to specify which runners can be used as a standalone app, this does not use the play() method. As for Dinothawr, it seems unlikely that this core / game will be part of the distrbuted cores in Lutris.

PR #383, opened 2016-10-07
[wine] Add run EXE inside prefix option. Fixes #381

### LINKED ISSUE #381: [WINE Runner] --> run exe
It should be possible to run an EXE/MSI in this winebottle/gamedir to install a patch, update or language fix(eg add it to the right click menu)

daniel-j:
I've also missed this.


strycore:
That will be very handy :D

PR #453, opened 2016-12-04
Properly use Gtk.Application, kill custom DBus service

TingPing:
Part 1 of rebasing #188

Note that this has **not** been throughly tested, just directly ported. Please try all the various command line arguments (both on main instance and secondary) to ensure everything is functional.


TingPing:
Oh sorry I forgot for master we'll have to use `optparse` still instead of `Gio.Application.add_main_option()`.

strycore:
That's awesome, thanks a lot!

TingPing:
To be clear this did bump the GLib dependency to 2.42.

strycore:
I'll test it against SteamOS and Debian 8 and if it runs, I'm good with it.

PR #475, opened 2017-01-26
Migrate to GAction and GMenu

TingPing:
This is another step towards a new UI decoupling actions from the widgets that display them. The same still needs to be done for the right click menu.

Testing as always is appreciated.


strycore:
Awesome! I'll try to test and review this by the end of the weekend!

Issue #430, opened 2016-11-08
[steam] Game installed but not shown


telanus:
Saw this, this morning, if you get a new game on steam, and install it, then launch Lutris, without adding it to your library via the website, lutris will state, in the terminal: 

`[steam]:Setting You Have to Win the Game as installed`

but won't show in the front-end, the first time you run lutris, only shows up on the second launch.

RobLoach:
Confirmed...

1. Quit Lutris
1. Make sure https://lutris.net/games/you-have-to-win-the-game/ is part of your lutris.net library
1. Install http://store.steampowered.com/app/286100 through Steam
1. Run Lutris
1. See `Setting You Have to Win the Game as installed` appear in the console
1. You Have To Win The Game is not in the Lutris window
1. Quit Lutris
1. Re-run it
1. See the game now in the Lutris window

daniel-j:
Fixed in #493

telanus:
one problem I've seen, when you remove a game IE QMC2 (mame frontend) and restart Lutris it will be readded to Lutris,

daniel-j:
Right now there is a game/executable blacklist in desktopapp.py. I want this to be configurable in the future, but for now you can submit a PR that adds MAME to the blacklist :)

PR #493, opened 2017-03-01
[winesteam] Add 64-bit prefix support

daniel-j:
With this patch, you can set the architecture in winesteam lutris installer scripts for games that require 64-bit.
```
game:
  appid: 12345
  arch: win64
```
It will create the win64 wineprefix `$HOME/.local/share/lutris/runners/winesteam/prefix64` and run steam from the 32-bit install, so you don't have to have two steam installations on disk.

This seem to require a system-install of Wine (64-bit obviously), because else Lutris tries to run it with it's default 32-bit wine, but that's related to the Wine runner.

Shoutout to **[DK] derklempner** who brought this to attention!

### DIFF
--- a/lutris/runners/winesteam.py
+++ b/lutris/runners/winesteam.py
@@ -128,7 +128,11 @@ class winesteam(wine.wine):
 
     @property
     def prefix_path(self):
-        _prefix = self.game_config.get('prefix') or self.get_default_prefix()
+        _prefix = \
+            self.game_config.get('prefix') or \
+            self.get_or_create_default_prefix(
+                arch=self.game_config.get('arch')
+            )
         return os.path.expanduser(_prefix)
 
     @property
@@ -300,13 +304,13 @@ class winesteam(wine.wine):
         if steamapps_paths:
             return steamapps_paths[0]
 
-    def create_prefix(self, prefix_dir):
+    def create_prefix(self, prefix_dir, arch='win32'):
         logger.debug("Creating default winesteam prefix")
         wine_path = self.get_executable()
 
         if not os.path.exists(os.path.dirname(prefix_dir)):
             os.makedirs(os.path.dirname(prefix_dir))
-        create_prefix(prefix_dir, arch=self.wine_arch, wine_path=wine_path)
+        create_prefix(prefix_dir, arch=arch, wine_path=wine_path)
 
         # Fix steam text display
         set_regedit("HKEY_CURRENT_USER\Software\Valve\Steam",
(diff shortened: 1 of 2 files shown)


lutris/runners/winesteam.py
```diff
@@ -128,7 +128,7 @@ def appid(self):
 
     @property
     def prefix_path(self):
-        _prefix = self.game_config.get('prefix') or self.get_default_prefix()
+        _prefix = self.game_config.get('prefix') or self.get_or_create_default_prefix(arch=self.game_config.get('arch'))
```
Freso: Does Lutris (aspire to) adhere to the PEP 8 line length limit? If so, this line needs to be split.

lutris/runners/winesteam.py
```diff
@@ -156,7 +156,8 @@ def launch_args(self):
         args = [self.get_executable(), self.get_steam_path()]
 
         # Fix invisible text in Steam
-        args.append('-no-dwrite')
+        # Since Wine 1.7.50, DirectWrite has been fixed
+        # args.append('-no-dwrite')
```
Freso: If so, just remove the code entirely. (This should be separate commit though, so you can explain the reasoning properly.) The code will still be in the git repository, so there's no need to leave commented out stale code in the codebase.

lutris/runners/winesteam.py
```diff
@@ -300,30 +301,34 @@ def get_default_steamapps_path(self):
...
         # Fix steam text display
+        # Since Wine 1.7.50, DirectWrite has been fixed.
         set_regedit("HKEY_CURRENT_USER\Software\Valve\Steam",
-                    'DWriteEnable', '0', 'REG_DWORD',
+                    'DWriteEnable', '1', 'REG_DWORD',
                     wine_path=self.get_executable(),
                     prefix=prefix_dir)
 
-    def get_default_prefix(self):
+    def get_default_prefix(self, arch=None):
```
Freso: Maybe explicitly set `'win32'` as the default (like for `get_or_create_default_prefix()`)? 64-bit is the default for more and more things, so explicitly declaring the current default here and handling both/all cases explicitly in the code would mean that whenever Lutris decides to switch to 64-bit Wine by default, it'll only have to update this line. (Though the argument is transferred over straight from `get_or_create_default_prefix()`, so this would also mean maintained the default both places…)

daniel-j:
This fixes #430! Yay

Freso:
I would [squash](https://duckduckgo.com/?q=git+squash) the two commits and add what you wrote in the description here to the new commit, as well as a line akin to "`Fixes https://github.com/lutris/lutris/issues/430`". This will assure that the arguments and reasoning for the changes survive the git repository, even if GitHub disappears. :) (The the "Fixes" line will make the issue auto-close when the PR gets merged into master.)

(See also https://chris.beams.io/posts/git-commit/ )

strycore:
> This seem to require a system-install of Wine (64-bit obviously), because else Lutris tries to run it with it's default 32-bit wine, but that's related to the Wine runner.

We ship 64bit builds of every Wine version (including Staging), the Wine version can be specified in the install scripts in the runner section.

PR #494, opened 2017-03-01
Auto-import system installed games on startup

daniel-j:
This PR adds support for auto-importing system installed games. For example it detects installed from your package manager or installed GOG games. It adds applications from the Game category, with some exceptions that aren't games (such as Lutris itself, Steam, PlayOnLinux..)
For some reason you have to restart Lutris an extra time before the games appear.


lutris/util/desktopapps.py
```diff
@@ -0,0 +1,127 @@
...
+def sync_with_lutris():
+    apps = get_games()
+    desktop_games_in_lutris = pga.get_desktop_games()
+    slugs_in_lutris = set([str(game['slug']) for game in desktop_games_in_lutris])
+
+    seen_slugs = set()
+    for app in apps:
+        game_info = None
+        name = app[0]
+        slug = slugify(name)
```
Freso: I will not be surprised when some games exist in Lutris and have a different slug than what this will generate, but not sure there's a good way around that without either 1) lobby for an extension to .desktop for games to use a globally unique identifier (say, an UUID like [MusicBrainz](https://musicbrainz.org/doc/MusicBrainz_Identifier) (which reminds me I still need to make that forum post)), or 2) maintain some sort of alias list in Lutris. (The latter could also be helpful if we need/want to "rename" games in Lutris but avoid breaking old links/URLs/slug references.)

Not really a problem with the PR per se, more an underlying issue that again should maybe be considered. :)

lutris/util/desktopapps.py
```diff
@@ -0,0 +1,127 @@
...
+
+        # app is blacklisted
+        if appid in IGNORED_APPS:
+            continue
+
+        # executable is blacklisted
+        if app.get_executable().lower() in IGNORED_EXECUTABLES:
+            continue
+
+        # must be in Game category
```
Freso: Maybe make this the first check to make it bail out a few cycles earlier? Far most .desktop entries will not be `Game`s, so weeding these out before checking for executable or `appid` sound sensible to me.

lutris/xdgmenu.py
```diff
@@ -1,30 +0,0 @@
-"""Get games available from the XDG menu."""
```
Freso: Was this file actually used anywhere? I don't see any removals of or updates for references to this.

lutris/gui/lutriswindow.py
```diff
@@ -74,6 +75,11 @@ def __init__(self, application, **kwargs):
             settings.read_setting('sidebar_visible') in ['true', None]
         self.use_dark_theme = settings.read_setting('dark_theme') == 'true'
 
+        # Sync local lutris library with current Steam games and desktop games
+        # before setting up game list and view
+        steam.sync_with_lutris()
+        desktopapps.sync_with_lutris()
```
daniel-j: It doesn't today, no. Yeah, should be in a separate thread.

lutris/util/desktopapps.py
```diff
@@ -0,0 +1,127 @@
...
+from gi.repository import Gio
+from lutris import pga
+from lutris.util.log import logger
+from lutris.util.strings import slugify
+from lutris.config import make_game_config_id, LutrisConfig
+
+IGNORED_APPS = (
+    "lutris", "mame", "dosbox", "playonlinux", "org.gnome.Games", "retroarch",
+    "steam", "steam-runtime", "steam-valve", "steam-native"
+)
```
daniel-j: GNOME Games also use [blacklists](https://github.com/GNOME/gnome-games/tree/master/plugins/desktop/data/blacklists), I'm not sure how this could work otherwise.
Maybe we can have a menu item Tools > Import games > Installed and present a list of games and let the user choose what to not import.

lutris/util/desktopapps.py
daniel-j: Yeah, for example Xonotic adds two .desktop files: Xonotic (SDL) and Xonotic (GLX), and both gets added to Lutris as they are. Lutris don't detect that they are the game Xonotic, and show no banner etc.

lutris/xdgmenu.py
daniel-j: It was unused.

lutris/util/desktopapps.py
strycore: I don't consider this to be our problem. The slug of any game can be changed and the banners as well. The recommended (and supported) way to install Xonotic is with Lutris. We don't need think about every edge case with .desktop files.

strycore:
Looks great!

PR #495, opened 2017-03-02
Remember Lutris window maximized state

daniel-j:
Maximized state is stored in the lutris.conf file. Now it will only store the window size when not maximized, as expected. Before it stored the size when maximized too, so when you launched Lutris again it would cover the screen but unmaximized state.
Also makes the side panel to not change width when the window is resized/maximized.

### DIFF
--- a/lutris/gui/lutriswindow.py
+++ b/lutris/gui/lutriswindow.py
@@ -66,6 +66,8 @@ class LutrisWindow(Gtk.ApplicationWindow):
         width = int(settings.read_setting('width') or 800)
         height = int(settings.read_setting('height') or 600)
         self.window_size = (width, height)
+        self.maximized  = settings.read_setting('maximized') == 'True'
+
         view_type = self.get_view_type()
         self.load_icon_type_from_settings(view_type)
         self.filter_installed = \
@@ -83,6 +85,8 @@ class LutrisWindow(Gtk.ApplicationWindow):
                          icon_name='lutris',
                          application=application,
                          **kwargs)
+        if self.maximized:
+            self.maximize()
         self.init_template()
         self._init_actions()
 
@@ -439,8 +443,15 @@ class LutrisWindow(Gtk.ApplicationWindow):
 
     @GtkTemplate.Callback
     def on_resize(self, widget, *args):
-        """WTF is this doing?"""
-        self.window_size = widget.get_size()
+        """Size-allocate signal.
+
+        Updates stored window size and maximized state.
+        """
+        if not widget.get_window():
+            return
+        self.maximized = widget.is_maximized()
+        if not self.maximized:
+            self.window_size = widget.get_size()
 
     @GtkTemplate.Callback
     def on_destroy(self, *args):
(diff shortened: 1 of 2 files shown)


lutris/gui/lutriswindow.py
```diff
@@ -66,6 +66,8 @@ def __init__(self, application, **kwargs):
         width = int(settings.read_setting('width') or 800)
         height = int(settings.read_setting('height') or 600)
         self.window_size = (width, height)
+        self.maximized  = settings.read_setting('maximized') == 'True'
```
Freso: `True` as a string? Is this correct? (This might be correct if it's some (GUI) framework thing, but in general Python `True` is its [its own value](https://docs.python.org/3/library/stdtypes.html#boolean-values).)

(Also, there should only be one space before the `=`.)

lutris/gui/lutriswindow.py
```diff
@@ -439,8 +443,11 @@ def _open_browser(url):
 
     @GtkTemplate.Callback
     def on_resize(self, widget, *args):
-        """WTF is this doing?"""
-        self.window_size = widget.get_size()
+        """ size-allocate signal - updates stored window size and maximized state """
```
Freso: This is a better description of what the method does, but the formatting is worse. 🙈

Check out https://www.python.org/dev/peps/pep-0008/#documentation-strings / https://www.python.org/dev/peps/pep-0257/

lutris/gui/lutriswindow.py
daniel-j: When saving a bool to a file, it gets retrieved as string. Running `bool('False')` returns True, so `bool()` won't work.

lutris/gui/lutriswindow.py
```diff
@@ -439,8 +443,11 @@ def _open_browser(url):
...
+        if not widget.get_window(): return
```
Freso: Similar `if` clauses elsewhere in the codebase has the `return` on its own line. This should happen here too.

Freso:
Oh, and another comment. If you write "Also makes …" in the description, that's a good pointer that you have code that should be in its own pull request—or at the very least [a separate](https://chris.beams.io/posts/git-commit/#limit-50) [commit](https://www.freshconsulting.com/atomic-commits/).

lutris/gui/lutriswindow.py
Freso: Ah, ok. Thanks. :)

lutris/gui/lutriswindow.py
```diff
@@ -439,8 +443,14 @@ def _open_browser(url):
 
     @GtkTemplate.Callback
     def on_resize(self, widget, *args):
-        """WTF is this doing?"""
-        self.window_size = widget.get_size()
+        """size-allocate signal
```
Freso: Docstrings should use proper grammar. Ie., this should be capitalised and use punctuation: "Size-allocate signal."

lutris/gui/lutriswindow.py
```diff
@@ -439,8 +443,14 @@ def _open_browser(url):
...
+        Updates stored window size and maximized state
```
Freso: Should have a blank line separating this from the method "title". Also, proper grammar: end with a period/`.`.

https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

PR #496, opened 2017-03-02
Add support to remove games with delete key

daniel-j:
It's the simple things in life.

### DIFF
--- a/lutris/gui/gameviews.py
+++ b/lutris/gui/gameviews.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-from gi.repository import Gtk, GObject, Pango, GLib
+from gi.repository import Gtk, Gdk, GObject, Pango, GLib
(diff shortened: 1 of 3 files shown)


lutris/gui/gameviews.py
```diff
@@ -137,6 +137,7 @@ class GameView(object):
         "game-selected": (GObject.SIGNAL_RUN_FIRST, None, ()),
         "game-activated": (GObject.SIGNAL_RUN_FIRST, None, ()),
         "game-installed": (GObject.SIGNAL_RUN_FIRST, None, (int,)),
+        "remove-game": (GObject.SIGNAL_RUN_FIRST, None, ())
```
Freso: Include a trailing comma, like the line above this, since this is the code style of the surrounding code. (See also, [PEP 8: When to use trailing commas](https://www.python.org/dev/peps/pep-0008/#when-to-use-trailing-commas).)

PR #497, opened 2017-03-02
Fix issues when switching between grid/list views

daniel-j:
I think this was the thing that annoyed me the most when using Lutris, the inconsistency when switching between grid and list views.


Freso:
My only comment here is to suggest that you rewrite the commit message. 😇

https://chris.beams.io/posts/git-commit/

PR #498, opened 2017-03-02
Add "View on Lutris.net" context menu item to games

daniel-j:
It won't work for games that aren't on the Lutris website yet, but better than nothing!
Works for both installed and uninstalled games.

### DIFF
--- a/lutris/gui/lutriswindow.py
+++ b/lutris/gui/lutriswindow.py
@@ -111,6 +111,7 @@ class LutrisWindow(Gtk.ApplicationWindow):
              self.remove_menu_shortcut),
             ('install_more', "Install (add) another version", self.on_install_clicked),
             ('remove', "Remove", self.on_remove_game),
+            ('view', "View on Lutris.net", self.on_view_game),
         ]
         self.menu = ContextualMenu(main_entries)
         self.view.contextual_menu = self.menu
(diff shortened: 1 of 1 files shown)


lutris/gui/lutriswindow.py
```diff
@@ -111,6 +111,7 @@ def __init__(self, application, **kwargs):
              self.remove_menu_shortcut),
             ('install_more', "Install (add) another version", self.on_install_clicked),
             ('remove', "Remove", self.on_remove_game),
+            ('view', "View on Lutris.net", self.on_view_game)
```
Freso: Trailing comma, please. :)

daniel-j:
> Also, I wonder if there's some way that we can tell whether something is on Lutris or not based on the data in the client/client library?

Not sure, the only way I can think if is checking if the game as a year or not, since you can't change release year for a game you added yourself. Another option would be to poll the site and see what games in library are on lutris.net, I think the sync does this.

strycore:
The sync can be used to do that but if you want a simpler solution, you can just filter the game based on whether they have an installer_slug (other than the recent `desktopapp` that is).

daniel-j:
> The sync can be used to do that but if you want a simpler solution, you can just filter the game based on whether they have an installer_slug (other than the recent `desktopapp` that is).

The simple solution wouldn't work if the user adds a game manually that exists on the site, it gets no installer_slug.

PR #499, opened 2017-03-02
[winesteam] Don't disable DirectWrite

daniel-j:
Since Wine 1.7.50, DirectWrite has been fixed.
Enjoy smooth fonts!

Note: This doesn't enable DirectWrite, you have to do that yourself in Steam settings (or reinstall winesteam)


strycore:
Wouldn't it be preferable to remove this entirely?
We could keep the function in the utils/steam module if we want to reintroduce this as a runner option.

PR #507, opened 2017-03-03
Clean up Wine runner. Fixes #398

### LINKED ISSUE #398: "Run EXE inside wine prefix" for a game with disabled Lutris runtime still includes Lutris runtime
If I set a game (not the runner or in the system) to "Disable Lutris Runtime" and then try to "Run EXE inside wine prefix" from the game's right-click menu, Lutris still includes its runtime in the execution.

daniel-j:
Now uses game config to enable lutris runtime for wineexec, winecfg, winetricks, joycpl and regedit.
Before only system- and runner-wide lutris runtime switch was used.


strycore:
@daniel-j  you rule :D

PR #509, opened 2017-03-03
Implement last played

daniel-j:
TODO: Update list view row when game stops.


strycore:
I'm going to merge this for now but I'd like to eventually have something a bit more complex with a record of all playtime (that would allow counting the total number of hours played)

PR #513, opened 2017-03-06
Fix errors when an installed game's runner is not installed

daniel-j:
Got errors when dosbox runner wasn't installed but I had dosbox games installed.


daniel-j:
Don't merge yet, I got another solution in the works!

PR #518, opened 2017-03-07
Clean up README

daniel-j:
README was formatted like Markdown, but it's a [reStructuredText](http://docutils.sourceforge.net/docs/user/rst/quickref.html) file, so I formatted it to use proper rst syntax. Quick preview link: https://github.com/daniel-j/lutris/blob/clean-readme/README.rst

Added missing runners and removed Desura.

Libretro is the proper `human_name` of the libretro runner.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -12,23 +12,24 @@ The games can be installed anywhere you want on your system, the tool
 does not impose anything.
 
 Lutris relies on various programs referenced as 'runners' to provide a
-vast library of games. These runners (with the exception of Steam,Desura
-and Web browsers) are provided by lutris, you don't need to install them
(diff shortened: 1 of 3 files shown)


README.rst
```diff
@@ -12,39 +12,46 @@ The games can be installed anywhere you want on your system, the tool
...
 * Vice
 * Stella
 * Atari800
 * Hatari
 * Virtual Jaguar
+* Snes9x
+* Mupen64Plus
 * Dolphin
-* PPSSPP
+* PCSX-Reloaded
```
strycore: PCSX Reloaded needs to be removed.

PR #538, opened 2017-03-12
Add global setting to set SDL_GAMECONTROLLERCONFIG

Orochimarufan:
SDL_GAMECONTROLLERCONFIG is very useful for using non-standard mappings in any SDL-based games, which includes XInput games on Wine using koku-xinput-wine.

This is one way to achieve it, I'm open to suggestions on better options though.

### DIFF
--- a/lutris/game.py
+++ b/lutris/game.py
@@ -197,6 +197,14 @@ class Game(object):
             self.state = self.STATE_STOPPED
             return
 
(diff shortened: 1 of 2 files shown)


strycore:
That's neat stuff and matches some of the features we want to add in later versions.

Does that enable koku-xinput-wine to be used without putting a gamecontrollerdb.txt inside the game folder? If so, that's very cool, we will ship an up to date version of this file in the client so this can be the default.

PR #555, opened 2017-04-07
Update libretro.py

telanus:
Added a few cores I use, and fixed the platform description for the NEC PC Engine both miss the last )


RobLoach:
Thanks!

strycore:
Not really a fan of having a bunch of emulators for the same machine but, oh well, the website will sort that out with the auto installers. The cores not used for auto-installers will simply not be used as much.

PR #557, opened 2017-04-26
[reicast] Only load joypads once

daniel-j:
This reduces the time it takes to open Lutris from 10 seconds to 3. Also speeds up the time it takes to open the runner manager.

This fix make it so reicast runner only loads the joypad list/evdev devices once.

### DIFF
--- a/lutris/runners/reicast.py
+++ b/lutris/runners/reicast.py
@@ -16,6 +16,8 @@ class reicast(Runner):
     platforms = ['Sega Dreamcast']
     runner_executable = 'reicast/reicast.elf'
 
+    joypads = None
+
     game_options = [{
         'option': 'iso',
         'type': 'file',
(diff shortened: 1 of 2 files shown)


lutris/runners/reicast.py
```diff
@@ -40,28 +65,28 @@ def __init__(self, config=None):
                 'option': 'device_id_1',
                 'type': 'choice',
                 'label': 'Joypad 1',
-                'choices': self.get_joypads(),
+                'choices': reicast.get_joypads(),
```
strycore: Expensive methods should not be called during the init stage, even if this improve the startup time, there is still an underlying issue there. Also there is something weird going on with your setup, get_joypads should be fast (but even when fast, __init__ methods in runners should *not* make any system calls, ever). 
We have to fine a way to lazy load this.

PR #561, opened 2017-04-30
add missing sections for bat and lnk file support

GloriousEggroll:
added missing lowercase conversion to checks, added missing check for .lnk in def wineexec, added missing check for .bat in def play


strycore:
Fixed in 9d1617153da8f9446643c6986ce96f0890c8f987

PR #574, opened 2017-05-27
Modernize UI

TingPing:
Some comments:

```irc
<TingPing> strycore, some design questions, do you think we should just shove things like syncronize library, disconnect, etc into the AppMenu, or like my original mockup should it be its own button, or any other ideas?
<TingPing> strycore, Browse games just opening a web browser seems a bit odd, but i guess as you said you want to move that in app
<TingPing> strycore, perhaps import and add game could be combined somehow
<TingPing> strycore, redid the sidebar, would be good to transition those icons to symbolic ones, also is the platform list even used?
```


TingPing:
TODO based on comments:

- [ ] Make Stop/Logs more discoverable, perhaps an infobar when a game is running
- [ ] Show platforms in sidebar again
- [ ] Start searching on keypress
- [ ] Move account info to own popover
- [ ] Kill statusbar

Issue #582, opened 2017-06-09
Convert runner icons to symbolic


TingPing:
Just making an issue to document that we should do this. Help from anybody willing to make relatively simple svg's would be appreciated. For some of them there may possibly be freely-licensed ones that exist.

The list of icons needed: https://github.com/lutris/lutris/tree/master/share/lutris/media/runner_icons

romlok:
I've started doing a bit of work toward this: https://github.com/romlok/lutris/tree/runnericons/share/lutris/media/runner_icons

So far, I've more or less just added the SVG icons that I could find which were clearly freely-licensed.

I'm not sure in what contexts Lutris uses these, and some of the icons are quite different from the originals, so improvement suggestions are welcome.

TingPing:
@romlok It looks like those svg's aren't symbolic though. An example would be like this for browser: https://github.com/GNOME/adwaita-icon-theme/blob/master/Adwaita/scalable/apps/web-browser-symbolic.svg

romlok:
Noted, but I'm not clear on exactly how symbolic you're wanting. Are, for example, the existing web/HTML5 and XDG icons symbolic enough?

![HTML5 logo](../blob/master/share/lutris/media/runner_icons/web.png?raw=true) ![Freedesktop.org logo](../blob/master/share/lutris/media/runner_icons/xdg.png?raw=true)

TingPing:
Symbolic icons are a gnome spec, they would be one flat color that the icon loader replaces dynamically with your Gtk themes colors.

romlok:
Oh, I see. That should probably be specified and linked to in the issue description then, so it's more clear what you're looking for.

romlok:
![winesteam-symbolic.svg](/romlok/lutris/blob/runnericons/share/lutris/media/runner_icons/winesteam-symbolic.svg)
Howzat?

Is there a quick way to test how well icons work as symbols; showing them recoloured, etc.?

TingPing:
@romlok Looks perfect!

> Is there a quick way to test how well icons work as symbols; showing them recoloured, etc.?

Hmm, I don't know of any tool for that. I might look into making a simple tool for it but I believe things have to be part of a valid icon theme so we will have to change the layout on disk.

romlok:
I added a symbols.html file to that directory in my branch, which lets you see the symbolic icons with different foreground and background colours. It doesn't work directly through github, but it works browsing locally.

[...]

PR #583, opened 2017-06-09
Add Platform icons

RobLoach:
This adds support for Platforms to have icons in the sidebar.

platform_icons


TingPing:
Ideally these would be symbolic SVGs.

RobLoach:
> Ideally these would be symbolic SVGs.

Oh definitely. Let's tackle that over in your other issue https://github.com/lutris/lutris/issues/582 :+1:

strycore:
@RobLoach can you add a spacer or a blank icon for the platforms that don't provide icons? I'll merge your PR after.

RobLoach:
That looks better. Good call.

screenshot at 2017-06-23 16-21-31

PR #591, opened 2017-06-14
Add button to view install script before installing

joonatoona:
Installer scripts can execute arbitrary code, so having the ability to view them would be nice.

Button:
Button
Dialog:
Dialog


RobLoach:
Nice work. This will help with debugging and security to make sure you're getting the right one.

Question: Should this bring up the installer in the Web Browser instead? The editing tools and such are right there.

TingPing:
We could just optionally use `GtkSourceView` for a nicer editor.

Also nitpick, we should be moving away from `GtkBuilderDialog` to normal `GtkTemplate` but since we already extensively use that I'll have to fix that anyway.

lutris/gui/installgamedialog.py
```diff
@@ -529,6 +536,14 @@ def on_cancel_clicked(self, _button):
             self.interpreter.revert()
         self.destroy()
 
+
+    # -------------
+    # View Source
+    # -------------
+
+    def on_view_clicked(self, _button):
+        InstallerSourceDialog(self.interpreter.script_pretty)
```
TingPing: Must set the parent for dialogs.

share/lutris/ui/script-dialog.ui
```diff
@@ -0,0 +1,84 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Generated with glade 3.20.0 -->
+<interface>
+  <requires lib="gtk+" version="3.20"/>
+  <object class="GtkAdjustment" id="adjustment1">
```
TingPing: Unused.

share/lutris/ui/script-dialog.ui
```diff
@@ -0,0 +1,84 @@
...
+    <property name="can_focus">False</property>
+    <property name="title" translatable="yes">View Install Script</property>
+    <property name="type_hint">dialog</property>
+    <property name="has_resize_grip">True</property>
+    <child internal-child="vbox">
+      <object class="GtkBox">
+        <property name="can_focus">False</property>
+        <property name="orientation">vertical</property>
+        <property name="spacing">2</property>
+        <child internal-child="action_area">
```
TingPing: This should be using action widgets rather than hard-coding buttons, and it should be using a headerbar.. but I can deal with the UI tweaks if needed (glade can't do it).

joonatoona:
Yeah, I have no clue what I'm doing, I'm sure there's better ways to do it. I'm completely unfamiliar with PyGtk, I'll look into opening it in browser instead. I didn't use a browser because xdg-open doesn't work on my machine, so I wouldn't be able to test it.

I'll fix it all later, I don't have time for a few days. I'll close this and open another PR once I've done all that?

TingPing:
I personally prefer it being in-client.

joonatoona:
I didn't know GtkSourceView existed, and I have no idea how to pass data between a window and its parent.

I'll look into PyGtk more and try and improve my code :)

TingPing:
Another note, `pygtk` is a project that has been dead for half a decade, avoid that term if you want to find relevant information, this is using `PyGObject` and `Gtk3`.

joonatoona:
I did not know that either, thanks! I'm assuming it's compatible, seeing as I used PyGtk docs :D

TingPing:
No they are not compatible, it just happens most function names in Gtk didn't get randomly renamed.

joonatoona:
Ah, OK. Well thanks a lot! I will read the docs and properly implement this feature :)

TingPing:
Feel free to stop by #lutris on freenode if you have any direct questions, I've been doing the UI redesign for 0.5. You also may want to base your branch on `next` too for that.

joonatoona:
Will do. I was considering asking in IRC earlier, but my irssi wasn't working :(

Yeah, I noticed when I was creating the PR, but it wasn't compatible for an automerge and I didn't want to rewrite everything. I will do that now.

PR #593, opened 2017-06-18
Fix Attribute Error when using lutris --list-games

RobLoach:
When running `--list-games`, there is no `self.window`, so calling destroy would cause an error. Checking if the window is available fixes it.


strycore:
thanks!

PR #604, opened 2017-06-28
Adding a few more cores for new systems

telanus:
Added the PC-98 and the X68000 to the RA runner

### DIFF
--- a/lutris/runners/libretro.py
+++ b/lutris/runners/libretro.py
@@ -11,8 +11,10 @@ def get_core_choices():
     # The order has to be the same!
     return [
         ('4do (3DO)', '4do'),
+        ('Citra (Nintendo 3DS)', 'citra'),
         ('CrocoDS (Amstrad CPC)', 'crocods'),
         ('DesmuME (Nintendo DS)', 'desmume'),
+        ('Dolphin (Nintendo Wii)', 'dolphin'),
         ('FCEUmm (Nintendo Entertainment System)', 'fceumm'),
         ('Fuse (ZX Spectrum)', 'fuse'),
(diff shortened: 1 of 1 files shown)


RobLoach:
`citra` and `dolphin`, while you're at it :wink:

lutris/runners/libretro.py
```diff
@@ -94,6 +99,7 @@ class libretro(Runner):
         'Sharp X68000',
         'Sony PlayStation Portable',
         'Sega Dreamcast',
+        'Sega Dreamcast',
```
TingPing: Duplicate?

lutris/runners/libretro.py
```diff
@@ -11,8 +11,10 @@ def get_core_choices():
     # The order has to be the same!
     return [
         ('4do (3DO)', '4do'),
+        ('Citra (Nintendo 3DS)', 'citra'),
         ('CrocoDS (Amstrad CPC)', 'crocods'),
         ('DesmuME (Nintendo DS)', 'desmume'),
+        ('Dolphin (Nintendo Wii)', 'dolphin'),
```
TingPing: and GameCube?

RobLoach: `Dolphin (Nintendo Wii/Gamecube)`?

lutris/runners/libretro.py
telanus: nope, redream is a dreamcast emu and reicast is a dreamcast emu, and red comes before rei, tried to keep to the alphabetic order that was already there in the core election order, thus here it look like a double entry when it isn't really

PR #632, opened 2017-08-26
Improve installer tasks and installer documentation.

daniel-j:
New tasks:

- `write_file` (create text files with content)
- `write_json` (update or create json files, similar to write_config)

Improved tasks:

- `chmodx` (made it sync, uses python's built-in chmod method)
- `execute` (you can run shell commands easier, set env variables, set processes to include/exclude in the thread monitor)
- `wineexec` (you can set env, dll overrides and processes to include/exclude from the thread monitor)

Misc changes:

- ~~Running winetricks from the context menu runs it without Lutris Runtime, because it uses system's zenity which doesn't play nice with the runtime.~~
- Improved the installer documentation.
- ~~Fix typo `'SteamService.ex'`~~
- You can set `include_processes` and `exclude_processes` in system config.

Example usage of tasks:
```yaml
  - execute:
      args: --data-dir "$GAMEDIR/data" --update
      description: Updating My Little Karaoke
      file: updater
      include_processes: python3
      disable_runtime: true
```
```yaml
  - execute: 'echo Hello World!'
```
```yaml
  - write_json:
      file: $GAMEDIR/drive_c/users/$USER/Application Data/Battle.net/Battle.net.config
      data:
        Client:
          HardwareAcceleration: 'false'
          Sound:
            Enabled: 'false'
```
```yaml
  - task:
      name: wineexec
      arch: win64
      executable: setup
      args: battlenet://Pro/
      exclude_processes: Agent.exe
```
```yaml
  system:
    exclude_processes: Agent.exe
```

### DIFF
--- a/docs/installers.rst
+++ b/docs/installers.rst
@@ -2,6 +2,8 @@
 Writing installers
 ==================
 
+See an example installer at the end of this document.
+
 Fetching required files
 =======================
 
@@ -51,7 +53,7 @@ directly but make the installer extract it from an archive or something, you
 can reference the rom with the ``main_file`` parameter.
 Example: ``main_file: game.rom``
 
-For browser games, specify the game's URL with ``main_file``.
+For web games, specify the game's URL (or filename) with ``main_file``.
 Example: ``main_file: http://www...``
 
 Presetting game parameters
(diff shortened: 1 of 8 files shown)


lutris/runners/wine.py
```diff
@@ -756,7 +766,7 @@ def run_regedit(self, *args):
         wineexec("regedit", wine_path=self.get_executable(), prefix=self.prefix_path, config=self)
 
     def run_winetricks(self, *args):
-        winetricks('', prefix=self.prefix_path, wine_path=self.get_executable(), config=self)
+        winetricks('', prefix=self.prefix_path, wine_path=self.get_executable(), config=self, disable_runtime=True)
```
strycore: Not sure we want to make disabling runtime mandatory

lutris/thread.py
```diff
@@ -27,7 +27,7 @@
     'lutris', 'python', 'python3',
     'bash', 'sh', 'tee', 'tr', 'zenity', 'xkbcomp', 'xboxdrv',
     'steam', 'Steam.exe', 'steamer', 'steamerrorrepor', 'gameoverlayui',
-    'SteamService.ex', 'steamwebhelper', 'steamwebhelper.', 'PnkBstrA.exe',
+    'SteamService.exe', 'steamwebhelper', 'steamwebhelper.', 'PnkBstrA.exe',
```
strycore: Not a typo, the process names in /proc have a limited number of characters.

lutris/thread.py
```diff
@@ -244,7 +245,11 @@ def watch_children(self):
             if child.pid in self.old_pids:
                 processes['external'].append(str(child))
                 continue
-            if child.name in EXCLUDED_PROCESSES and child.name not in self.include_processes:
+
+            if (child.name and
+                (child.name in EXCLUDED_PROCESSES or
+                 child.name in self.exclude_processes) and
```
strycore: We might as well merge EXCLUDED_PROCESSES into self.exclude_processes

lutris/util/system.py
```diff
@@ -113,7 +113,7 @@ def get_md5_hash(filename):
 
 
 def find_executable(exec_name, quiet=False):
-    if not exec_name:
+    if not exec_name and not quiet:
```
strycore: Why? This will crash without an exec_name.

docs/installers.rst
```diff
@@ -379,10 +468,49 @@ Trying the installer locally
...
-``lutris -i /path/to/file``
+the ``name``, ``game_slug``, ``slug`` and ``runner`` directives. The value for
+``runner`` must be the slug name for the runner. (E.g. winesteam for Steam Windows.)
+You can also add ``version``, ``description`` and ``notes`` to the installer file.
+Under ``script``, add ``files``, ``installer``, ``game`` and other installer
+directives. See below for an example.
+Save your script in a .yaml file and use the following command in a terminal:
+``lutris -i /path/to/file.yaml``
+
+Example Linux game:
```
strycore: Not sure I want to advertise the API response format in the installer documentation.

lutris/installer/commands.py
```diff
@@ -16,6 +18,19 @@
 from lutris.thread import LutrisThread
 
 
+def selective_merge(base_obj, delta_obj):
```
strycore: Can this reside in a module in lutris.util?

docs/installers.rst
daniel-j: Why not? This is the only way to run installers locally.

PR #664, opened 2017-10-02
Added a few extra platforms

telanus:


RobLoach:
A few? ahhaha. This is the world of platforms! :smile:

strycore:
This is great, the only bit that worries me is that we still have to maintain those 2 list of platforms, this is silly and shouldn't be too hard to fix.

strycore:
We have to do something about this XD

workspace 2_001

PR #665, opened 2017-10-04
flowbox: Fix ImportError for get_pixbuf_for_game

aszlig:
The error was introduced in 60f90f0c58acc3a17365132c541864e284c011e4 and I've stumbled upon this while trying to run the test suite.


strycore:
Thanks, I've missed that one!

PR #667, opened 2017-10-06
updated the Devices and added the DEC PDP-1

telanus:
Added most of the normal devices used in mame/mess, the more out there ones IE bitbanger and midiin and midiin I didn't add as the systems mostly don't work or so niche that I doubt it'll ever get used


strycore:
Nice!

Issue #573, opened 2017-05-20
Lutris uses 100% of one core after game ends


HamishTPB:
Distro: Antergos
Lutris: 0.4.10 (Arch aur)

Info from Lutris script:
Desktop Info for Lutris

Environment variables:
DESKTOP_SESSION=i3
GDMSESSION=i3
SHELL=/bin/bash

Distribution:
Linux polarbear 4.10.13-1-ARCH #1 SMP PREEMPT Thu Apr 27 12:15:09 CEST 2017 x86_64 GNU/Linux

Running desktop processes:
56 kdevtmpfs
21457 kdeinit5
21466 kded5

OpenGL info:
OpenGL vendor string: NVIDIA Corporation
OpenGL renderer string: GeForce GTX 660/PCIe/SSE2
OpenGL core profile version string: 4.5.0 NVIDIA 381.22
OpenGL core profile shading language version string: 4.50 NVIDIA
OpenGL core profile context flags: (none)
OpenGL core profile profile mask: core profile
OpenGL core profile extensions:
OpenGL version string: 4.5.0 NVIDIA 381.22
OpenGL shading language version string: 4.50 NVIDIA
OpenGL context flags: (none)
OpenGL profile mask: (none)
OpenGL extensions:
OpenGL ES profile version string: OpenGL ES 3.2 NVIDIA 381.22
OpenGL ES profile shading language version string: OpenGL ES GLSL ES 3.20
OpenGL ES profile extensions:

RAM info:
              total        used        free      shared  buff/cache   available
Mem:           7.8G        3.6G        1.1G        191M        3.1G        3.7G
Swap:          9.0G        1.2M        9.0G

CPU info:
model name      : AMD FX(tm)-8350 Eight-Core Processor

Wine info:
Wine installed system wide: /usr/bin/wine
Version: wine-2.8

Symptoms:
When I run a game from Lutris, even  if the game runs fine, once the game ends then Lutris uses 100% of one of my CPU cores until I quit Lutris, sometimes Lutris hangs there and I have to kill it too. CPU usage becomes normal again once Lutris is closed/killed.

HamishTPB:
[...]

PR #669, opened 2017-10-07
Clean up stdout_monitor if pipe is broken

manuelVo:
This PR causes the io watch of the stdout_monitor to be cleaned up if the pipe is broken. This happens for example if a steam game is started and steam is already running in the background. In this case the started process (which will be the `LutrisThread.game_process`) will merely send a signal to the actual steam process and then terminate - causing a broken stdout pipe.

Performing this cleanup fixes several issues:
1. For some reason beyond my knowledge PyGTK causes 100% CPU load on one processor if there is an io watch on a broken pipe. (Which seems to be the cause of #573 )
2. The io watch isn't always cleaned up properly right now. For example if the game process gets killed via the stop button in Lutris no cleanup will be performed. This causes memory and cpu to leak as io watches and zombie child processes will stack up over time (the io watch holds a reference to the game process which prevents it from being cleaned up).

### DIFF
--- a/lutris/thread.py
+++ b/lutris/thread.py
@@ -123,9 +123,12 @@ class LutrisThread(threading.Thread):
 
         if self.watch:
             GLib.timeout_add(HEARTBEAT_DELAY, self.watch_children)
-            self.stdout_monitor = GLib.io_add_watch(self.game_process.stdout, GLib.IO_IN, self.on_stdout_output)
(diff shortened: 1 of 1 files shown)


strycore:
YOU ARE MY HERO

(Seriously, come on IRC or Discord, I'll give you a bunch of Steam keys)

Issue #614, opened 2017-07-13
[installer] execute task in installer scripts are quite broken in later versions


daniel-j:
When using a blacklisted process, for example `bash` or `sh`, to run a quick script, the lutris thread will wait for 5 minutes and 40 seconds before realizing bash is not running anymore. This is calculated by `WARMUP_TIME + MAX_CYCLES_WITHOUT_CHILDREN * HEARTBEAT_DELAY / 1000`
Example task:
```yaml
  - execute:
      file: /bin/bash
      args: -c 'echo hello world'
```
This will output `[thread]:Processes: external: python3 (9167:S) | excluded: bash (9175:Z)` until that timeout. Some game installers are affected by this, most notably Battle.net and Overwatch installers.
If the script launches a sub-process for about 3 seconds, it will be detected by Lutris and flagged as "yes, the process has started, don't wait 5 minutes of warm-up", so it will "only" take 40 seconds before Lutris continue the install process.

daniel-j:
It will output this with the latest Lutris version:
```
[thread]:Processes: excluded: bash (10419:Z)
[thread]:Warming up…
```

PR #670, opened 2017-10-07
Count terminated non child processes to number of terminated_children

manuelVo:
Currently the only way lutris checks if a watched process has terminated is to see whether it has become a zombie process. This however does not work for non child processes like steam games (if steam was already running). These games get cleaned up instantly by their parent processes and thus just disappear.

To improve process monitoring this PR additionally compares the list of currently monitored processes with the previous list of monitored processes and treats processed that have disappeared as terminated as well.

Besides greatly improving Steam game monitoring this could possibly help #614 as well (I cant reproduce that issue though, so I can't verify that).

### DIFF
--- a/lutris/thread.py
+++ b/lutris/thread.py
@@ -267,6 +267,10 @@ class LutrisThread(threading.Thread):
             processes['monitored'].append(str(child))
             if child.state == 'Z':
                 terminated_children += 1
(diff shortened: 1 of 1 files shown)


strycore:
That seems like it will help a lot too, thanks!!

PR #676, opened 2017-10-17
Count processes that were just terminated as children

manuelVo:
This PR fixes a little oversight by me in #670. I forgot to increment one of the process counters. As far as i can tell this should be without any real consequence (in the worst case Lutris tries to kill processes which are already terminated anyway). I think for the sake of clean code this should be done anyway.


strycore:
Nice catch,  thanks!

PR #679, opened 2017-10-24
Better controller support

tammasloughran:
The controller mappings needed to be updated for each machine type.
Rather than manually defining every controller mapping, I used the controller database to create the button mappings and just duplicated them for each machine controller layout.
It might not give the ideal mapping for every game or every machine controller so I added a runner option to use the default mappings from mednafen-09x.cfg so the user can just hit alt+shift+1 in mednafen and redefine the mappings themselves.


strycore:
Thanks a lot, that's very well done :D

PR #688, opened 2017-11-20
Add basic support for DRI_PRIME

sonicpp:
Depending on the number of available graphic cards (provided by xrandr) show option for enabling DRI_PRIME.
Option will run the game with the 'DRI_PRIME=1' environment variable.

This is basic functionality. In future, it would be better to show available graphic cards and simply let the user to choose one. But it will be much harder to implement (determine which one is sink provider, test it on setup with more displays, ...)

### DIFF
--- a/lutris/game.py
+++ b/lutris/game.py
@@ -302,6 +302,12 @@ class Game(object):
         ld_preload = gameplay_info.get('ld_preload') or ''
         env["LD_PRELOAD"] = ld_preload
 
+        dri_prime = system_config.get('dri_prime')
+        if dri_prime:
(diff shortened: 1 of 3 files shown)


strycore:
Thanks!
Please note that we plan getting rid of the xrandr dependency quite soon, mainly to provide proper support for wayland. We plan switching to libgnome-desktop but I'm not sure this library would allow you to list GPUs.  Maybe the output of lspci is enough to achieve that.

sonicpp:
Ok, I will try find solution for Wayland (or display server independent)

PR #710, opened 2017-12-09
docs update round 2

soredake:


docs/installers.rst
```diff
@@ -561,7 +562,7 @@ Trying the installer locally
 ============================
 
 If needed (i.e. you didn't download the installer first from the website), add
-the ``name``, ``game_slug``, ``slug``, ``version`` and ``runner`` directives.
+the ``name`` (if name contains : character surrond name with quotes), ``game_slug``, ``slug``, ``version`` and ``runner`` directives.
```
strycore: typo on surround

docs/installers.rst
```diff
@@ -583,12 +584,241 @@ Example Linux game:
...
       - myfile: http://example.com/mygame.zip
 
       installer:
       - chmodx: $GAMEDIR/mygame
+      system:
+        terminal: true
+        env:
+          SOMEENV: true
+
+Example wine game:
```
strycore: Is it possible to add anchors to markdown? Would be cool to that a list of those example so we can  quickly switch to them.

PR #726, opened 2017-12-28
[jzintv runner] Resolution support

maxlefou:
This mod adds a new configuration option for the runner that lets the user change the size of the game's window.
deepin capture-ecran_20171227224608

jzintv uses the option -z to change resolutions. It either can be used as -z#x#,# to set a completely custom resolution, or (in the actual case here) using presets with -z# (from 0 to 7). the code here adds a dropdown menu that lets the user choose between those presets and launches the game using the choosen preset.

**Things to eventually add in the future**
* Make lutris choose the default option by default instead of using a blank option
* Add the possibility to let the user using completely custom resolution sets (with text boxes?)


strycore:
Perfect, thanks!

PR #740, opened 2018-01-22
Fix a crash when game is added without installer and outside the library

Xenega:
When a game has no installer, you can add manually the game. If the game is not present in library, the dialog crash. Instead, my fix download the game data from remote and add it. If not exists, display a empty dialog.

### DIFF
--- a/lutris/api.py
+++ b/lutris/api.py
@@ -86,4 +86,25 @@ def get_games(game_slugs=None, page=1):
     else:
         payload = None
     response.get(data=payload)
-    return response.json
+    response_data = response.json
+
+    if not response_data:
+        logger.warning('Unable to get games from API')
+        return None
+
+    results = response_data.get('results', [])
+    while response_data.get('next'):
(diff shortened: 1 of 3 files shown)


Xenega:
Note: Travis check fail because it fail to install python.
https://github.com/lutris/buildbot/issues/34

strycore:
@Xenega Thanks for the patch!

Our travis setup isn't comple yet. The Ubuntu version Travis provides is too old so we have to setup a Docker container in order to get tests running.
Meanwhile, just ignore anything Travis related.

lutris/api.py
```diff
@@ -86,4 +86,25 @@ def get_games(game_slugs=None, page=1):
     else:
         payload = None
     response.get(data=payload)
-    return response.json
+    response_data = response.json
+
+    if not response_data:
+        logger.warning('Unable to get games from API')
+        return None
```
strycore: None can be omitted here

PR #744, opened 2018-01-30
Allow to customize wine sandbox

### LINKED ISSUE #596: Make wine desktop integration optional
Currently `WinePrefixManager.setup_defaults()` gets called every time a game is launched at [this location](https://github.com/lutris/lutris/blob/2a7f48108c2f017b2bcbb51f90d2f86f176188ff/lutris/runners/wine.py#L807). This resets any user configuration to these values, in particular the "Desktop Integration" settings for linking the desktop and documents folder. While this can be useful, it should not be forced at every launch. I like to keep my saves in a separate Wine directory, so they aren't deleted when I uninstall a game. It was very confusing to see them disappear every launch. So please don't reset to the default values before running a game, it should be done after prefix creation only.

Xenega:
Added two options to customize the wine sandbox.
First a checkbox to disable it.
Second, a directory chooser to change the default sandbox wine directory.

Resolve #596

Tested, but if others can test it in another setup, it would be good.

### DIFF
--- a/lutris/runners/wine.py
+++ b/lutris/runners/wine.py
@@ -139,7 +139,7 @@ def create_prefix(prefix, wine_path=None, arch='win32', overrides={},
     logger.info('%s Prefix created in %s', arch, prefix)
     prefix_manager = WinePrefixManager(prefix)
     prefix_manager.setup_defaults()
-
+    self.sandbox(prefix_manager)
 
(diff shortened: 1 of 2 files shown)


strycore:
Looks cool, might need a few changes but we'll make those at a later time

PR #746, opened 2018-02-03
Update desktop icons after copy.

### LINKED ISSUE #659: Shortcut icons not updated when changing icons in Lutris
If you configure a game and change its icon to a custom one within Lutris, that icon does not seem to be used when creating a desktop shortcut. Instead the default icon off the website gets used.

Is this expected behavior or a bug?

Xenega:
Force update of icon cache with xdg-icon-resource. I execute "xdg-icon-resource forceupdate" after each copy of a icon. I think xdg-icon-resource is sufficient cross-unix, but surprise can happen.

Fix #659

### DIFF
--- a/lutris/gui/config_dialogs.py
+++ b/lutris/gui/config_dialogs.py
@@ -11,7 +11,7 @@ from lutris.gui.widgets.common import VBox, SlugEntry, NumberEntry
 from lutris.gui.widgets.dialogs import Dialog
 from lutris.gui.widgets.utils import get_pixbuf_for_game, get_pixbuf, BANNER_SIZE, ICON_SIZE
 from lutris.util.strings import slugify
-from lutris.util import datapath
+from lutris.util import datapath, resources
 
 DIALOG_WIDTH = 780
 DIALOG_HEIGHT = 560
@@ -401,6 +401,9 @@ class GameDialogCommon(object):
             pixbuf = get_pixbuf(image_path, None, size)
             pixbuf.savev(dest_path, file_format, [], [])
             self._set_image(image_type)
+
+            if image_type == 'icon':
+                resources.udpate_desktop_icons()
 
         dialog.destroy()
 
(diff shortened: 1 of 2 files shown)


strycore:
Looks cool

TingPing:
Instead of introducing an intermediate dependency you should just call `gtk-update-icon-cache`, this tool adds no value on top.

strycore:
@Xenega can you update that in a new PR? Thanks!

Xenega:
@TingPing @strycore
`gtk-update-icon-cache` only update image cache for gtk. Any other desktop manager cache will not be updated. So no, it cannot replace `xdg-icon-resource`, whose value is to be independent of the specific desktop environment.

TingPing:
That isn't true at all; Firstly there is no such thing as "cache for just gtk", this is all a standard spec. Secondly read the source of `xdg-icon-resource` it literally just calls `gtk-update-icon-cache` as that is the main implementation that exists and every distro uses.

Xenega:
I disagree. `gtk-update-icon-cache` is a specification only for GTK 2.6+. No documentation indicate that `gtk-update-icon-cache` is becoming a standard spec. And some major distribution don't use it. For example, KDE don't use `gtk-update-icon-cache`, but don't need update (kde detect it).

However, you are right, gtk seems the only implementation used on the _official implementation_. But, xdg-utils can be tuned by distros. 

Note : 
KDE => icon-cache.kcache
GTK+ => icon-theme.cache

TingPing:
Ok I didn't realize KDE ignored that cache. The moral of the story still is this script is just a wrapper for `gtk-update-icon-cache` and it is just an extra dependency.

Xenega:
Yes. Do you prefer I use `gtk-update-icon-cache` if it exist ? Since it mostly a gtk problem (other desktop manager seems to do the overheard of detecting icon image change).

TingPing:
I just see no point in using a wrapper script, that is all.

PR #756, opened 2018-02-11
Fix sidebar list item size and icons

daniel-j:


lutris/gui/sidebar.py
```diff
@@ -62,7 +62,6 @@ def _create_button_box(self):
             image.show()
             btn.add(image)
             btn.connect('clicked', entry[2])
-            btn.get_style_context().add_class('sidebar-button')
```
TingPing: To state the obvious, this is a bad solution. Edit the actual CSS file....

PR #760, opened 2018-02-13
Add rpcs3 runner

999gary:
Cleaned Commits


strycore:
We still need to get a build of some sort, the AppImage looks like the simplest option here.

PR #764, opened 2018-02-16
gtk-update-icon-cache instead of xdg-icon-resource

Xenega:
As wanted by @TingPing, Use directly `gtk-update-icon-cache` to update the icon cache on gtk+ desktop manager.
That means for other (or new) desktop manager, we will need to add their cache command in the future. The rational is that for now only gtk+ need manually update, so `xdg-icon-resource` is just a wrapper around `gtk-update-icon-cache`.

Could someone test it on a KDE deistro to check against regression ?

### DIFF
--- a/lutris/util/resources.py
+++ b/lutris/util/resources.py
@@ -1,7 +1,9 @@
 import os
 import re
+import shutil
 import concurrent.futures
 from urllib.parse import urlparse, parse_qsl
+from gi.repository import GLib
 
 from lutris import settings
 from lutris import api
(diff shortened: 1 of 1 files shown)


lutris/util/resources.py
```diff
@@ -73,7 +75,13 @@ def fetch_icons(game_slugs, callback=None):
         callback(updated_slugs)
 
 def udpate_desktop_icons():
-    os.system("xdg-icon-resource forceupdate")
+    # Update Icon for GTK+ desktop manager
+    gtk_update_icon_cache = shutil.which("gtk-update-icon-cache")
+    if (gtk_update_icon_cache):
+        os.system("gtk-update-icon-cache -tf %s"
+        % os.path.join(GLib.get_user_data_dir(), 'icons', 'hicolor'))
```
strycore: I'll fix that later but that is a weird way of cutting lines! X)

strycore:
Thanks!

lutris/util/resources.py
Xenega: Oh sorry. I'm not from a python background.

strycore:
Oh by the way, I have to test something on Mint KDE, I have to remember to test this (it will be a pretty old version)

lutris/util/resources.py
strycore: It's ok, the code is good. I suggest you integrate Flake8 in whatever editor you're using and if you have no Flake8 warning then it's formatted correctly, and will tell you what's wrong if not

TingPing:
> That means for other (or new) desktop manager, we will need to add their cache command in the future. 

The icon theme spec is almost 20 years old and in those 20 years no other tool was made.

PR #768, opened 2018-02-17
Add basic LSI support

999gary:
Requires installation or it wont do anything, snap support is unknown if someone wants to test the snap. Also disables native runtime option in favor of the LSI runtime.

### DIFF
--- a/lutris/runners/steam.py
+++ b/lutris/runners/steam.py
@@ -92,6 +92,16 @@ class steam(Runner):
                      "Make sure you disabled Lutris Runtime and "
                      "have the required libraries installed.")
         },
+	{
+	    'option': 'lsi_steam',
+	    'label': "Start Steam with LSI",
+	    'type': 'bool',
+	    'default': False,
+	    'help': ("Launches steam with LSI patches enabled. "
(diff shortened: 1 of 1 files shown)


999gary:
Just realized I made a mistake checking the LSI runtime option, not sure of the easiest way to fix it without closing the PR.

Xenega:
You can just amend your commit (add `--amend` option to your `git commit` command).

999gary:
fixed, unfortunately does not work with the snap.

lutris/runners/steam.py
```diff
@@ -166,7 +176,10 @@ def steam_data_dir(self):
                 return path.rstrip('sSteamAp')
 
     def get_executable(self):
-        return system.find_executable('steam')
+        if system.find_executable('lsi-steam') and self.runner_config.get('lsi_steam'):
```
strycore: Put the config condition first so we don't do a lookup for the executable if the config isn't set

999gary:
Changed the commit, if I messed up something with git I'm sorry I'm still new to git.

strycore:
Looks cool, thanks!

Issue #728, opened 2017-12-29
i18n is not fully implemented


Sebastian1978:
Lutris doesn't have russian translation.

liberavia:
There are some po-files in po-directory, but they don't seem to be really used. The only position where gettext is included is in lutris/gui/application.py but binded to a common directory (/usr/share/locale)

strycore:
Yeah. I know.

TingPing:
I'd set it up if Lutris ever moves to a real build system where you can reasonably build, extract, and merge translations.

strycore:
Hopefully, we'll migrate to Meson during the 0.5 cycle.

liberavia:
I currently try to test if at least basic string translation with gettext can be implemented with not too much effort 

I began to wrap strings of the code and tried to import and install gettext all over the lutris package. However, it's not yet working.

My first approach is in this commit:
https://github.com/liberavia/lutris/commit/f996fb6d1f205796c92062b1a0a3570439f30ce4

Hints are welcome

TingPing:
- Installing translations in `./locale` is a bit lame
- Storing binary `.mo` files in git is a bad idea
- Data files like desktop file need merging
- Automated way to do all of that and extract translations would be nice

But most of that comes down to having a build system with `gettext` support built in.

liberavia:
Ok, so it would be at least helpful if I do a PR with just the marking stuff then?

TingPing:
I think so yes. Also for text like this: `<i>(Italic indicates that this option is modified in a lower configuration level.)</i>` don't have users translate the markup.

Here are some general guidelines: https://wiki.gnome.org/TranslationProject/DevGuidelines (And the [markup page specificially](https://wiki.gnome.org/TranslationProject/DevGuidelines/Avoid%20markup%20wherever%20possible))

liberavia:
Hi there,

i removed some of the code that i did yesterday and went further searching through files for making them translatable, so that there available in a pot file without markup and whitespaces. I noticed that there is a 80 column limit that is used which often leads to strings wrapped with brackets to go over newlines. These are hard to catch for translation due to `_()` will destroy the structure. This occures for example in sysoption.py where currently I just marked the labels for now:
`    
{
        'option': 'game_path',
        'type': 'directory_chooser',
        'label': _('Main game folder'),
        'default': os.path.expanduser('~/Games'),
        'scope': ['runner', 'system'],
        'help': ("The main folder where you install your games.\n"
[...]

PR #769, opened 2018-02-17
Internationalization: Step 1

liberavia:
This pull is related to issue https://github.com/lutris/lutris/issues/728

It contains marking any found string that seems to be played out in gui, so it is some sort of base for using gettext.


lutris/installer/commands.py
```diff
@@ -173,16 +173,15 @@ def _on_input_menu_validated(self, widget, *args):
...
-            'message',
-            "Insert or mount game disc and click Autodetect or\n"
-            "use Browse if the disc is mounted on a non standard location."
-        )
-        message += (
-            "\n\nLutris is looking for a mounted disk drive or image \n"
-            "containing the following file or folder:\n"
-            "<i>%s</i>" % requires
-        )
+        message_action = _("Insert or mount game disc and click Autodetect or") + "\n"
```
TingPing: This shouldn't be split like this, it turns a normal paragraph into weird partial sentences.

lutris/runners/atari800.py
```diff
@@ -26,14 +26,29 @@ class atari800(Runner):
...
         "5200_rom": ""
     }
+
+    def get_long_help_msg(self,option):
+        help_messages = {}
+
+        # main_file
+        help_messages['main_file'] = _("The game data, commonly called a ROM image.") + " \n"
+        help_messages['main_file'] += _("Supported rom formats: ATR, XFD, DCM, ATR.GZ, XFD.GZ")
+        help_messages['main_file'] += _("and PRO.")
```
TingPing: Same comment about split line.

lutris/runners/atari800.py
```diff
@@ -26,14 +26,29 @@ class atari800(Runner):
...
+
+        # bios_path
+        help_messages['bios_path'] += _("A folder containing the Atari 800 bios files.") + " \n"
+        help_messages['bios_path'] += _("They are provided by Lutris so you shouldn't have to") + " "
+        help_messages['bios_path'] += _("change this") + "\n"
```
TingPing: Same.

lutris/gui/config_dialogs.py
```diff
@@ -452,8 +452,9 @@ def get_config_id(self):
 class EditGameConfigDialog(Dialog, GameDialogCommon):
     """Game config edit dialog."""
     def __init__(self, parent, game, callback):
+        headline_game_config_dialog = _("Configure ") + game.name
         super(EditGameConfigDialog, self).__init__(
-            "Configure %s" % game.name,
+            headline_game_config_dialog,
```
TingPing: So I believe it is best to actually expose `%s` here as it allows RTL languages to be in the correct order. Same for a few others. It does have the downside that bad translations can cause runtime errors but there are tools to validate this.

liberavia:
I changed those lines the way strycore suggested in https://github.com/lutris/lutris/issues/728#issuecomment-360007657

Could someone review this to make sure this is the way you want it to be implemented before I go over all files again?

lutris/gui/installgamedialog.py
```diff
@@ -263,18 +264,18 @@ def prepare_install(self, script_index):
...
             self.set_path_chooser(self.on_target_changed, 'folder',
                                   default_path)
             self.non_empty_label = Gtk.Label()
-            self.non_empty_label.set_markup(
-                "<b>Warning!</b> The selected path "
-                "contains files, installation might not work properly."
-            )
+            label_markup = "<b>" + _("Warning!") + "</b>"
+            label_markup += " " + _("The selected path") + " "
+            label_markup += _("contains files, installation might not work properly.")
```
liberavia: Todo: Remove splitting of parts that belong together

lutris/gui/sync.py
```diff
@@ -68,8 +68,9 @@ def __init__(self, parent=None):
         self.get_content_area().add(box_outer)
 
         description_label = Gtk.Label()
-        description_label.set_markup("You can import games from local game sources, \n"
-                                     "you can also choose to sync everytime Lutris starts")
+        label_markup = _("You can import games from local game sources,") + " \n"
+        label_markup += _("you can also choose to sync everytime Lutris starts")
+        description_label.set_markup(label_markup)
```
liberavia: Todo: line and meaning splitting

lutris/gui/uninstallgamedialog.py
```diff
@@ -82,10 +82,12 @@ def on_apply_button_clicked(self, widget):
         if remove_contents and not hasattr(self.game.runner,
                                            'no_game_remove_warning'):
             game_dir = self.game.directory.replace('&', '&amp;')
+            question =  _("Are you sure you want to delete EVERYTHING under") + " \n<b>"
+            question += game_dir
+            question += "</b>?\n " + _("(This can't be undone)")
+
```
liberavia: Using pattern instead of only concatenation

TingPing:
> There is a problem with resolving the problem not having markup in translations and keeping textblocks together. Any suggestions how to deal with this?

If it comes down to it not splitting up sentences is more important than some markup being leaked IMO.

strycore:
We are not splitting strings for i18n, this is my final design decision on the subject.

TingPing:
Hmm, I'm not sure if using `{runner}` over `%s` is a great idea. It is bound to be translated incorrectly by somebody and also it bypasses the `gettext` tooling which is aware of `%s` style formatting.

liberavia:
You mean I should do it this way?

```
mystring = _(
    "This %s message"
    "goes over"
    "several lines"
) % some_message_type
```

strycore:
This is now too old and too out of sync to be rebased against master. A new PR has to be submitted.

PR #774, opened 2018-02-25
Add higan runner.

FakeShemp:
[_higan_](https://byuu.org/emulation/higan/) is a multi-system emulator with great SNES support. As far as I know it is the only one supporting certain SNES games.

It uses its own [Gamepak](https://byuu.org/emulation/higan/game-paks) system for loading games. Gamepaks are an attempt to simulate physical Nintendo Game Paks in digital form. The idea is that for each game, one folder acts as a gamepak, and contains all data that is specific to said game. _icarus_ is its complimentary program for automagical creation of these Gamepaks, but it is not supported in this runner as the exact same functionality can be written in the install scripts with more versatility.


strycore:
Nice! Didn't know about higan!

PR #795, opened 2018-03-16
accept kitty as terminal

soredake:
@strycore


strycore:
I'm pretty sure having a hard coded list of terminal is the wrong way to do it.

PR #799, opened 2018-03-16
system: Add MATE Terminal

RobLoach:
[MATE Terminal Emulator](https://github.com/mate-desktop/mate-terminal) is a good one. Add that along-side the others.

Inspired by https://github.com/lutris/lutris/pull/795 ... Thanks soredake!

### DIFF
--- a/lutris/util/system.py
+++ b/lutris/util/system.py
@@ -18,6 +18,7 @@ TERMINAL_CANDIDATES = [
     'xfce4-terminal',
     'pantheon-terminal',
     'terminator',
+    'mate-terminal',
     'urxvt',
     'cool-retro-term',
     'Eterm',


strycore:
I'm pretty sure having a hard coded list of terminal is the wrong way to do it.

TingPing:
> I'm pretty sure having a hard coded list of terminal is the wrong way to do it.

It actually isn't, there is no standard method to define a default terminal. XFCE has its own spec which is reasonable but nothing else implements it afaik, Debian has its own defaults mess, etc.

TingPing:
You can however find installed terminals dynamically, that doesn't need to be hardcoded necessarily. The question just becomes preference at that point.

PR #802, opened 2018-03-19
update Indonesian translation

cho2:
update Indonesian translation


strycore:
Thanks! (Just hope that everyone submitting translations realizes we have no i18n support in or code yet)

PR #822, opened 2018-04-04
rpm: correct macro names and dependencies

TomasTomecek:
I recently installed lutris on my Fedora system and realized that
it's not pulling any dependencies.

The reason for that was that you have incorrect macros used. Macros
for Red Hat-based systems don't contain the `_version` suffix.

I'm not sure about OpenSuse since I'm not familiar with it.


strycore:
Thanks! This is a serious issue if the dependencies weren't installing!

PR #828, opened 2018-04-19
Various fixes for winesteam for Linux game data download and wine game execution

Ryochan7:
1. Use proper runner config data when downloading Linux game data from Steam for Windows otherwise no wine prefix is detected (ex. Doom 3 BFG); Lutris was attempting to use linux runner data so no prefix keys were found in the runner_config dict.

https://lutris.net/games/doom-3-bfg-edition/

2. For default winesteam install, use default_arch if auto is set for the prefix architecture when no explicit prefix is set (ex. Warhammer 40K: Dawn of War 1).

https://lutris.net/games/warhammer-40000-dawn-of-war/

3. Steam for Windows install can get stuck when output from stdout is not read. If watch is not set, just close the stdout stream. This problem mainly occurred on my Xubuntu 16.04 and SteamOS installs but I think it has happened sometimes while testing on Xubuntu 17.10. The tweak fixes the problem on all my Linux installations.

### DIFF
--- a/lutris/installer/interpreter.py
+++ b/lutris/installer/interpreter.py
@@ -699,7 +699,7 @@ class ScriptInterpreter(CommandsMixin):
         appid = self.steam_data['appid']
         if not steam_runner.get_game_path_from_appid(appid):
             logger.debug("Installing steam game %s", appid)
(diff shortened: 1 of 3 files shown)


strycore:
Thanks! Those are some very useful fixes!

PR #834, opened 2018-04-23
Use get_env() return as default environments.

Xenega:
Set et_env() return as default environments. Because a lot of play function don't set the environment. Avoid empty environments.


strycore:
thanks!

PR #855, opened 2018-05-08
libretro: Add ChaiLove core

RobLoach:
This change adds the [ChaiLove](https://github.com/libretro/libretro-chailove) libretro core, which is a 2D game framework. It's similar to [Love2D](https://love2d.org/), but using [ChaiScript](http://chaiscript.com/).

Added a game to Lutris that uses it over at: https://lutris.net/games/floppy-bird/


RobLoach:
Uploaded the core to the buildbot.

Issue #869, opened 2018-05-17
Linux Steam no longer launches from Lutris since 0.4.15/0.4.16 update


DaDummy:
Lutris 0.4.16 on Kubuntu Bionic Beaver (18.04) here.

I've got Lutris with wine, wine-steam, steam and some Linux games on my system.

Since 0.4.15 or 0.4.16 I'm having trouble running Steam and Steam games through Lutris - the process dies instantly, which it didn't before. I haven't changed anything except for updating Lutris through apt. Running Steam from Dolphin/bash still works as expected.

As far as I have tested some Linux games still work others do not.
As far as I have tested wine games still work as before.
The wine-steam runner and all installed wine-steam games have disappeared from Lutris with the update, so I cannot test if those still work. (My wine-steam setup was a little unconventional since I did want my wine-steam in a win64 prefix which I didn't know how to configure through Lutris back then.)

Please let me know if there is any way I can narrow down the cause of that problem.

DaDummy:
Some additional info that might be relevant:
I have lutris installed from the Lutris apt repo for Ubuntu 18.04.

wine-development and steam:i386 are installed from Ubuntu repos, wine-stable is not installed.

DaDummy:
I ran a git bisect and this appears to be the commit that broke Linux-Steam for my system:

```
d8a664f444c2d9d9be1c6d74890bf18a93bfbd62 is the first bad commit
commit d8a664f444c2d9d9be1c6d74890bf18a93bfbd62
Author: Travis Nickles <nickles.travis@gmail.com>
Date:   Wed Apr 18 21:39:23 2018 -0500

    Fixed winesteam data download, winesteam auto prefix workaround, winesteam install
    
    Use proper runner config data when downloading game data from Steam
    otherwise no prefix is detected (was attempting to use linux data).
    For default winesteam install, use default_arch if auto is set for
    the prefix architecture. Steam for Windows install can get stuck when
    output from stdout is not read. If watch is not set, just close the
    stdout stream.

:040000 040000 0d50bf3d56711cb6cc77ba7892bd9e0a8592dc76 63ba9b799b84109891fc6894b4ec8aed92941fdd M      lutris
```

DaDummy:
As stated in the title: Linux Steam.

[...]

PR #870, opened 2018-05-17
Fixed launching Steam from Lutris on Ubuntu 18.04

DaDummy:
I ran into the issue that Steam and a couple of other native applications could no longer be launched through Lutris after recent updates. I've identified the commit responsible for my issue using git bisect and then partially reverted a single change from that commit that appeared to be the cause of my trouble. Steam and other apps work as expected again with this change.

The issue is further discussed in #869.

### DIFF
--- a/lutris/thread.py
+++ b/lutris/thread.py
@@ -132,8 +132,6 @@ class LutrisThread(threading.Thread):
             self.stdout_monitor = GLib.io_add_watch(self.game_process.stdout,
                                                     GLib.IO_IN | GLib.IO_HUP,
                                                     self.on_stdout_output)
-        else:
-            self.game_process.stdout.close()
 
     def on_stdout_output(self, fd, condition):
         if condition == GLib.IO_HUP:
--- a/lutris/thread.py
+++ b/lutris/thread.py
@@ -180,8 +180,14 @@ class LutrisThread(threading.Thread):
         try:
             if self.cwd and not system.path_exists(self.cwd):
                 os.makedirs(self.cwd)
+
+            if self.watch:
+                pipe=subprocess.PIPE
+            else:
+                pipe=None
+
             return subprocess.Popen(command, bufsize=1,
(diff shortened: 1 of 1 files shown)


strycore:
I'll let @Ryochan7 decide on that one since this was introduced in a recent PR and I wasn't able to reproduce the issue in #869

DaDummy:
I have extended my fix to include a change that should reinstate the functionality @Ryochan7 intended, but without the negative side effects on my side.

Ryochan7:
I performed a fresh install of Xubuntu 18.04 on a spare partition and I did not experience the problem that were describing regarding Steam. I would have thought that the change would be fairly benign and it did not cause any behavior changes with the other applications that I had installed on Lutris. The change that you have proposed looks like it would work fine for the original problem that I had.

The change was original made due to behavior that I experienced installing Steam for Windows on multiple occasions. Very frequently, Steam would get stuck at one of the last windows at the "Verifying Installation" step. I tested this in the Python interpreter as well and similar steps resulted in Steam getting stuck until the Popen.communicate method was called to read the data from stdout. At that point, Steam would continue as expected and finish the installation process. I will revert the change on my copy and see out of curiosity if this behavior can still happen.

Ryochan7:
I have now verified that the problem can exist in Xubuntu 18.04 without the fix on version 0.4.16. It requires a specific execution path. Install Wine Steam from the Manage Runners window, right-click on the Wine Steam entry in the main window sidebar, then click the Run menu option. Without the stdout workaround, Steam will get stuck while trying to finish the installation. Attempting to finish a Wine Steam install while installing a game seems to have Lutris run Steam from winestream.install_game. That routine does not attempt to open the stdout pipe so it works.

strycore:
Thanks for the heads up @Ryochan7, I'll merge this for now because I'm suspecting this is causing some other issues. I'll see if the PR improves the situation.

Issue #385, opened 2016-10-07
add an option to toggle the composite manager


ghost:
for best performance there should be an easy way to disable the windowmanagers composite.

eg:

-->user want to run a game
-->lutris disables composite if enabled and launches game
-->user exists game
-->lutris re-enables composite

also here some commands to start:

Plasma/kwin:

```
if [ "$(qdbus org.kde.KWin /Compositor org.kde.kwin.Compositing.active)" = "true" ]; then
    qdbus org.kde.KWin /Compositor org.kde.kwin.Compositing.suspend
else
    qdbus org.kde.KWin /Compositor org.kde.kwin.Compositing.resume
fi
```

also same is possibe with xfwm, metacity & co if i remember correctly they just use their own tools xfconf, gsettings/gconf

eg xfwm:

```
xfconf-query --channel=xfwm4 --property=/general/use_compositing --set=false 
```

tannisroot:
Feature was implemented in #881.

PR #881, opened 2018-05-23
Option to disable the compositor for better performance

nastys:
Improves performance and reduces stuttering.
Supported window managers:

- KWin/KDE
- Marco/MATE
- XFWM/XFCE.

Enabled by default.

Implements #385

### DIFF
--- a/lutris/game.py
+++ b/lutris/game.py
@@ -112,6 +112,24 @@ class Game(object):
         else:
             self.runner = runner_class(self.config)
 
+    def desktop_effects(self, enable):
+        if enable:
+            system.execute(self.start_compositor, shell=True)
+        else:
+            if os.environ.get('DESKTOP_SESSION') == "plasma":
+                stop_compositor = "qdbus org.kde.KWin /Compositor org.kde.kwin.Compositing.suspend"
+                self.start_compositor = "qdbus org.kde.KWin /Compositor org.kde.kwin.Compositing.resume"
+            elif os.environ.get('DESKTOP_SESSION') == "mate" and system.execute("gsettings get org.mate.Marco.general compositing-manager", shell=True) == 'true':
+                stop_compositor = "gsettings set org.mate.Marco.general compositing-manager false"
+                self.start_compositor = "gsettings set org.mate.Marco.general compositing-manager true"
+            elif os.environ.get('DESKTOP_SESSION') == "xfce" and system.execute("xfconf-query --channel=xfwm4 --property=/general/use_compositing", shell=True) == 'true':
+                stop_compositor = "xfconf-query --channel=xfwm4 --property=/general/use_compositing --set=false"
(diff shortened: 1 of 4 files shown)


lutris/game.py
```diff
@@ -320,6 +322,21 @@ def do_play(self, prelaunched, _error=None):
             show_obnoxious_process_monitor_message()
         process_watch = not monitoring_disabled
 
+        if self.runner.system_config.get('disable_compositor'):
+            if os.environ.get('DESKTOP_SESSION') == "plasma":
```
strycore: Could the block below be implemented in it's own method?

lutris/game.py
```diff
@@ -403,6 +420,9 @@ def on_game_quit(self):
            or self.runner.system_config.get('reset_desktop'):
             display.change_resolution(self.original_outputs)
 
+        if self.compositor_disabled:
+            subprocess.Popen(self.start_compositor, shell=True)
```
strycore: use `lutris.system.execute` instead of Popen calls

lutris/sysoptions.py
```diff
@@ -205,6 +205,14 @@ def get_dri_prime():
         'advanced': True,
         'help': "Disables process monitoring of games, Lutris won't detect when game quits."
     },
+    {
+        'option': 'disable_compositor',
+        'label': "Disable desktop effects",
+        'type': 'bool',
+        'default': True,
```
strycore: Default to False, this is the current behavior.

lutris/game.py
nastys: Done.

lutris/game.py
nastys: Done. I had to allow system.execute to optionally run the command in a shell (default is still False, so it shouldn't break compatibility).

lutris/sysoptions.py
nastys: I've set it to False, though I'm not sure why it shouldn't be enabled by default.

nastys:
Everything done. By the way, you should be able to commit directly to my fork if you want to make any more edits.

strycore:
Looks all good!
I'm merging it to master after the 0.4.18 release so we can test it more, it will be included in the upcoming 0.5.0 release

daniel-j:
Would this enable `xfwm4` compositor even though I have disabled it before launching the game? I use a custom compton compositor!

nastys:
@daniel-j No, it won't resume xfwm4 if it wasn't previously started.
Unfortunately it doesn't support Compton yet, so it won't make any difference if you enable this option.

yaomtc:
Does this support Compiz?

nastys:
@yaomtc No, it only supports the window managers on the list. I don't think supporting Compiz would be possible, at least not without replacing it with another window manager temporarily (but that may cause some issues).

Issue #913, opened 2018-06-12
error wrong format


sigmaSd:
HI , in interperter.py line 201
shouldnt this
```            
                if  not system.find_executable(dependency):
                    raise ScriptingError(
                        "This installer requires %s on your system" % ' or '.join(dependency)
                    )
```
be
```         
                if  not system.find_executable(dependency):
                    raise ScriptingError(
                        "This installer requires %s on your system" % (dependency)
                    )
``` 
This is what I get for example installing runescape3 without required deb
 [https://postimg.cc/image/jbadq42h9/](https://postimg.cc/image/jbadq42h9/l)

PR #914, opened 2018-06-13
Fix error formating in case of missing dependency

sigmaSd:
Fix https://github.com/lutris/lutris/issues/913

### DIFF
--- a/lutris/installer/interpreter.py
+++ b/lutris/installer/interpreter.py
@@ -200,7 +200,7 @@ class ScriptInterpreter(CommandsMixin):
             else:
                 if not system.find_executable(dependency):
                     raise ScriptingError(
-                        "This installer requires %s on your system" % ' or '.join(dependency)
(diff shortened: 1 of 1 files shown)


strycore:
That just avoids a crash but does not fix the root issue, which is that the dependency is missing.

sigmaSd:
No did you check the image I upload it? it still crashes that's not the point of this patch, the problem it fixes is if you have a missing required binaries the error message have the right format for example without this patch, I got 'requires b or s or d oa or r' instead of 'requires bsdtar' when it's missing.

sigmaSd:
here is the picture I didnt know that github can handle image..
screenshot from 2018-06-12 08-19-58

strycore:
I think I misread what the fix was :sweat_smile:

sigmaSd:
Sorry for the bad title thats what mislead you probably xD

PR #932, opened 2018-06-19
Added TIC-80 Runner

telanus:


lutris/runners/tic80.py
```diff
@@ -0,0 +1,53 @@
...
+        {
+            'option': 'skip',
+            'type': 'bool',
+            'label': 'Skip startup animation',
+            'default': False
+        },
+        {
+            'option': 'sound',
+            'type': 'bool',
+            'label': 'Start in silent mode',
```
RobLoach: Rename to "Mute"?

lutris/runners/tic80.py
```diff
@@ -0,0 +1,53 @@
...
+            'default': False
+        },
+        {
+            'option': 'skip',
+            'type': 'bool',
+            'label': 'Skip startup animation',
+            'default': False
+        },
+        {
+            'option': 'sound',
```
RobLoach: Having the option key match the argument makes the code easier to read.... Rename it to `nosound`?

lutris/runners/tic80.py
```diff
@@ -0,0 +1,53 @@
...
+    ]
+
+    def play(self):
+        """Run the game."""
+        arguments = [self.get_executable()]
+        if self.runner_config.get('surf'):
+            arguments.append('-surf')
+        if self.runner_config.get('skip'):
+            arguments.append('-skip')
+        if self.runner_config.get('sound'):
```
RobLoach: If you rename the option to `nosound`, the logic will be easier to grasp.

lutris/runners/tic80.py
```diff
@@ -0,0 +1,53 @@
...
+            'type': 'bool',
+            'label': 'Start in silent mode',
+            'default': False
+        }
+    ]
+
+    def play(self):
+        """Run the game."""
+        arguments = [self.get_executable()]
+        if self.runner_config.get('surf'):
```
RobLoach: I see a `-fullscreen` option we could use here too! https://github.com/nesbox/TIC-80/wiki#command-line-arguments

RobLoach:
This looks great! Added a few notes of how I think it could improve :+1:

lutris/runners/tic80.py
telanus: Fullscreen doesn't seem to work. I created a issue on their [github](https://github.com/nesbox/TIC-80/issues/626), will add the option as soon as it works

PR #934, opened 2018-06-20
Fix runtime on ubuntu, prefer system libs

daniel-j:
Include `/lib/x86_64-linux-gnu` and `/lib/i386-linux-gnu` if they exist and look for libraries there before Lutris runtime libs.


AccountOneOff:
Tested and working. No longer complains about `libgpg-error.so.0` and `libgcrypt.so.20` when opening winetricks.

PR #941, opened 2018-06-23
Fix hardcoded path

gregoryhugaerts:
the path in the next branch is hardcoded to the devs machine, so it won't run anywhere else


RobLoach:
What is `~/games-list`? Should this be configurable, or moved inside `~/.config/lutris/runners/gog/games-list` or something?

strycore:
Lol, this is a development branch, I'm not done there yet.

PR #953, opened 2018-06-25
[runtime] Add more system paths

daniel-j:
Add some more system library paths for Ubuntu.


duhow:
```bash
echo "/usr/lib/$(uname -m)-linux-gnu"
/usr/lib/x86_64-linux-gnu
```

PR #954, opened 2018-06-26
Improve sorting in gameviews

daniel-j:
Adds sorting options to the menu, uses a single button to toggle between grid and list, sorting is synced with grid and remembered on save, sorting falls back to sort by name and runner if values are equal (if you sort by platform, year etc.), fix zoom/icon size slider.


strycore:
Very cool :D

PR #969, opened 2018-07-15
Update DXVK versions & Fix DXVK URL

FedoraTipper:
Good day, 

There is currently a bug where dxvk.py's self.version returns the full string "0.**  (default)", which is displayed to the user through the UI. This has prevented lutris from downloading the default selected version of dxvk. 

A simple fix which I have made has been to simply regex/substitute the dxvk_url string.

I've also added an update to DXVK versions.


strycore:
This is already fixed in next

PR #971, opened 2018-07-18
prevent buttons from showing null when hovering over them

luthub:
When starting lutris without having any games installed or being logged in, it shows three buttons. When hovering the mouse over two of them a little tooltip appears saying 'null'.
This pull request fixes that by adding an empty tooltip_text property to these buttons so no tooltip appears when the mouse is moved on them, which makes it look a bit better.

before:
before
after:
after


strycore:
Thanks!

PR #972, opened 2018-07-20
Add supporting functionality to fetch and install mod files from https://www.moddb.com/

glitchbunny:
I added a game to the Lutris database that is a Source engine mod that isn't available from Steam but is available from https://www.moddb.com/.  In order to install the mod, the mod files must be downloaded and unpacked in Steam's sourcemods directory.

I attempted to write an installer to do this but found that I needed an extra bit of functionality to make it work, which is included in this pull request.

The additions are:
- A function to follow the ModDB permalink for the download through to an appropriate mirror.
- An additional variable available in the installer script that expands to the Steam data directory.

I'm not sure if this is a good idea for a number of reasons, including;
- It is likely to be fragile, depending on ModDB specifics.
- The addition of third-party-site specific code into Lutris, which would perhaps be better pulled out in some sort of configuration file, addon or similar structure, particularly if there were more similar additions in the future.
- Security issues associated with downloads from random sources.
- Problematic legalities / Mod DB T&Cs.
- I'm new to the Lutris codebase, and have possibly missed existing functionality or established conventions.

I'd welcome any comments/discussion.  Additionally, I would be happy to recreate this pull request as a branch on the main repository if required, in order to avoid any issues when I delete this repository in the future.

For reference, the following is the installer that would make use of these changes;

```
files:
- archive: https://www.moddb.com/downloads/start/126529
installer:
- execute:
        file: p7zip
        args: --decompress $archive
        working_dir: $STEAM_DATA_DIR/steamapps/sourcemods/
- rename:
        src: Portal-StillAlive
        dst: Portal-stillalive
        working_dir: $STEAM_DATA_DIR/steamapps/sourcemods/
game:
        appid: '400'
        args: -game $STEAM_DATA_DIR/steamapps/sourcemods/Portal-stillalive
requires: portal
requires-binaries: p7zip
steam: {}
```

### DIFF
--- a/lutris/installer/interpreter.py
+++ b/lutris/installer/interpreter.py
@@ -293,8 +293,6 @@ class ScriptInterpreter(CommandsMixin):
 
         if file_uri.startswith("/"):
             file_uri = "file://" + file_uri
-        elif file_uri.startswith("https://www.moddb.com/downloads/start/"):
-            file_uri = self._moddb_ultimate_download_uri(file_uri)
         elif file_uri.startswith(("$WINESTEAM", "$STEAM")):
             # Download Steam data
             self._download_steam_data(file_uri, file_id)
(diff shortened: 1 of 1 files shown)


glitchbunny:
Thinking on this a little further, I realise that what I primarily want to encapsulate in the installer are the steps required to make the mod work on Linux, as although very simple (at least for the mod that prompted this PR), I found that information about the steps required for Linux were not easy to discover and it would therefore be valuable to have this information captured in an installer.

The download of the mod can be left to the user (in a similar way to installers for games distributed on disks or that require ROMs to work).

Removing the download aspect of this PR will hopefully remove any potentially contentious part of the PR, allowing it to be accepted more easily.

strycore:
You bring a very interesting problem, downloading files from ModDB has always been a bit of an issue. 

But. I think there's a way to get this working in a more reliable way. We could use something similar to how we log in to GOG, on the `next` branch. If we can just get a link to whatever popup ModDB… and actually show it, in a Webkit2 widget, just like we do for the GOG login form.

Then we let the ModDB do it's thing, with the timer and all and we should get a redirect event, which the Webkit2 widget is able to intercept. From this event, we grab the destination URL and send that to a download progress box and destroy the webkit widget.

Let me know if you're up for that kind of task, it wouldn't rush since all the code is in next, which is far from ready.

glitchbunny:
Yes,  I wouldn't mind looking into that, however, when I searched for a ModDB API that could be used to perform the download (which I didn't find, sadly, as it would obviously be a far nicer way to implement something like this), I found https://mod.io/ - Which appears to be an effort to move ModDB toward an infrastructure that is integrated into the games themselves.

It is early days (the first blog post is dated 6 months ago) but if that effort is successful (and not unreasonably slow to develop) it appears that it would remove the download requirement from Lutris altogether.   It could be worth watching, in any case!

PR #977, opened 2018-07-22
Remove lutris.settings reference from setup.py

RobLoach:
Executing `python setup.py install` gets an error when importing lutris.settings....
```
Running: python setup.py install
  File "setup.py", line 5, in <module>
    from lutris.settings import VERSION
  File "/run/build/lutris/lutris/settings.py", line 4, in <module>
    from gi.repository import GLib
ImportError: No module named gi.repository
```
It's because gi.repository isn't installed, yet. Switching the version out for the actual version fixes it.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,7 +2,6 @@
 import os
 import sys
 from setuptools import setup
-from lutris.settings import VERSION
 
 if sys.version_info < (3, 4):
     sys.exit('Python 3.4 is required to run Lutris')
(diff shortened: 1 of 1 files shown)


strycore:
this is not a proper solution, while importing gi.repositiory in a settings file is not ideal, it uncovers an issue, which is that gi.repository should be available on any system that is going build Lutris.

My guess is that `python setup.py install` ran setup.py with python 2 and that failed.

RobLoach:
After installing PyGObject manually, it worked.

PR #980, opened 2018-07-23
[dxvk] symlink to dxvk dlls instead of copying

FlyingWombat:
In the case of using system installed DXVK, symlinks will ensure the
files in the prefix are up to date with system version.


strycore:
Seem's a very unsafe change, I expect some breakage, but that should be handled during testing.

PR #989, opened 2018-07-29
Change example urls in installer documentation to https and example.com

luthub:
Not sure what to do with line 34, because I don't actually know what a referer is xD


strycore:
thanks!

PR #994, opened 2018-08-03
Auto update latest DXVK version

### LINKED ISSUE #974: Auto update latest DXVK versions
https://github.com/lutris/lutris/blob/525c84d9da173e84fd5585eed6b9d2fef5bef2b2/lutris/util/dxvk.py#L11-L12

Instead of writing static versions, use GitHub API to get the latest versions of DXVK:
```bash
curl -s https://api.github.com/repos/doitsujin/dxvk/tags  | jq '[.[] | .name] | join(", ")'
"v0.62, v0.61, v0.60, v0.54, v0.53, v0.52, v0.51, v0.50, v0.42, v0.41, v0.40, v0.31, v0.30, v0.21, v0.20"
```

matthewkovacs:
resolves #974

### DIFF
--- a/lutris/util/dxvk.py
+++ b/lutris/util/dxvk.py
@@ -3,13 +3,20 @@ import os
 import time
 import shutil
 
+from requests import get
 from lutris.settings import RUNTIME_DIR
 from lutris.util.log import logger
 from lutris.util.extract import extract_archive
 from lutris.util.downloader import Downloader
 
-DXVK_LATEST = "0.52"
-DXVK_PAST_RELEASES = ["0.51", "0.50", "0.42", "0.31", "0.21"]
+# Get latest DXVK versions from GitHub
+dxvk_url = "https://api.github.com/repos/doitsujin/dxvk/tags"
+dxvk_json = get(dxvk_url).json()
+dxvk_versions = [x['name'].replace('v', '') for x in dxvk_json]
+
+DXVK_LATEST = dxvk_versions[0]
+DXVK_PAST_RELEASES = [x['name'].replace('v', '') for x in dxvk_json]
+
 
 
 class DXVKManager:
--- a/lutris/util/dxvk.py
+++ b/lutris/util/dxvk.py
@@ -1,23 +1,41 @@
 """DXVK helper module"""
 import os
+import json
 import time
 import shutil
+import urllib.request
 
-from requests import get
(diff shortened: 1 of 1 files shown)


matthewkovacs:
Thanks for the quick feedback!  

I've attempted to address most of the revisions you mentioned, except I am not sure where a better place would be to write this request call.  If you can name a file specifically I'd be happy to move the request call there instead.

lutris/util/dxvk.py
```diff
@@ -1,23 +1,41 @@
...
 
-DXVK_LATEST = dxvk_versions[0]
-DXVK_PAST_RELEASES = [x['name'].replace('v', '') for x in dxvk_json]
+def get_latest_dxvk_versions():
+    """Get latest DXVK versions from GitHub"""
+    dxvk_url = "https://api.github.com/repos/doitsujin/dxvk/tags"
+    cache = os.path.join(RUNTIME_DIR, 'dxvk', 'dxvk_versions.json')
 
+    # If the DXVK cache does not exist then download it
+    if not os.path.exists("./dxvk_versions.json"):
```
gort818: Why are you not checking the full file path?  Substitue "./dxvk_versions.json" with cache

lutris/util/dxvk.py
```diff
@@ -1,23 +1,41 @@
...
+        urllib.request.urlretrieve(dxvk_url, cache)
 
+    # Re-download DXVK versions cache if more than a day old
+    if os.path.getmtime("dxvk_versions.json")+86400 < time.time():
```
gort818: "dxvk_versions.json" should be cache

lutris/util/dxvk.py
```diff
@@ -1,23 +1,41 @@
...
+        urllib.request.urlretrieve(dxvk_url, cache)
+
+    with open("dxvk_versions.json", "r") as f:
```
gort818: "dxvk_versions.json" should be cache.

lutris/util/dxvk.py
matthewkovacs: Good idea!

gort818:
Awesome I tested this and seems to work great

strycore:
Very nice! Thanks a lot!

steelbrain:
Can we do `mkdir -p $HOME/.local/share/lutris/runtime/dxvk/` before writing to that directory?

PR #999, opened 2018-08-03
residualvm no longer has --soft-renderer option

Qalthos:
It now has a choice of three options and a `--renderer` cli flag.


strycore:
Thanks!

PR #1002, opened 2018-08-04
Try to improve gui behavior for services with connect

NicoVIII:
While experimenting with the development branch of lutris, I noticed a problem.
I connected GOG and activated sync. I disconnected GOG again but forgot to disable sync. Lutris did not start anymore because it tried to sync on startup, but couldn't.

I therefore disabled the sync button and the sync switch, if the service is not connected and disable syncing on disconnect.


strycore:
Yeah, that part still needs a lot of work, thanks for the fixes!

PR #1004, opened 2018-08-05
Hide menu bar at Installer Window

### LINKED ISSUE #1003: Menu bar visible during install
image

While installing a game, the menu bar at the dialog window shouldn't be visible.

duhow:
Fixes #1003


strycore:
That was much simpler than what I thought it would be :D Many thanks!

PR #1011, opened 2018-08-09
Fix detection of xgamma

### LINKED ISSUE #1010: “Restore default gamma....” doesn’t restore gamma after game quits

luthub:
Fixes #1010

### DIFF
--- a/lutris/util/display.py
+++ b/lutris/util/display.py
@@ -12,7 +12,6 @@ from lutris.util.log import logger
 
 XRANDR_CACHE = None
 XRANDR_CACHE_SET_AT = None
-XGAMMA_FOUND = False
 
 
 def cached(function):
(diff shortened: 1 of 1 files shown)


CitrusAlex:
Will gamma restoration run every time any game leaves or only with 'Use XVidMode to change resolutions" (which allows games to change screen gamma) enabled?
Because if it's the former it will reset Night Light mode, which will be extremely annoying if it will happen with every game, even if it doesn't change gamma.

lutris/util/display.py
```diff
@@ -159,11 +158,9 @@ def change_resolution(resolution):
 
 def restore_gamma():
     """Restores gamma to a normal level."""
-    global XGAMMA_FOUND
-    if XGAMMA_FOUND is None:
-        XGAMMA_FOUND = system.find_executable('xgamma')
-    if XGAMMA_FOUND is True:
-        subprocess.Popen(["xgamma", "-gamma", "1.0"])
+    XGAMMA_PATH = system.find_executable('xgamma')
```
strycore: Cool but this removes caching so we make a call to `which xgamma` every single time. And yes, this is way more costly than one would think.

luthub: You're right. I hope it's better now.

luthub:
This PR is rubbish. I opened a new one that hopefully does it right. Sorry for the inconvenience

PR #1027, opened 2018-08-15
Fix detection of xgamma

### LINKED ISSUE #1010: “Restore default gamma....” doesn’t restore gamma after game quits

luthub:
fixes #1010 properly now (I hope)

### DIFF
--- a/lutris/util/display.py
+++ b/lutris/util/display.py
@@ -163,7 +163,7 @@ def restore_gamma():
     if XGAMMA_FOUND is None:
         XGAMMA_FOUND = system.find_executable('xgamma')
     if XGAMMA_FOUND is not None:
-        subprocess.Popen(["XGAMMA_FOUND", "-gamma", "1.0"])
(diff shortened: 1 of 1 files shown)


luthub:
That the caching won't work if xgamma is not found was kind of intended. If a user enables the restore gamma option without having xgamma installed, he is very likely to install it at some point later (maybe after seeing the warning that it isn't installed). If he installs it while lutris is running, that change would be picked up immediately and he won't have to restart lutris. If he doesn't have restore gamma enabled, the existence of xgamma will never be checked, so I personally don't see an issue there. But I'll change it if you want to.

It seems easier to use XGAMMA_FOUND as string (that's why i renamed it to XGAMMA_PATH in my previous PR because it would make more sense that way), but I can also handle it as boolean if you want to.

luthub:
I made a few changes. Now it should be exactly the way you imagined.

luthub:
Ok, this is rubbish as well. I opened a third one that really does it right and the way you wanted it. Sorry for being an idiot

PR #1030, opened 2018-08-16
Do not change configuration with mouse wheel scrolling

### LINKED ISSUE #1024: Do not change configuration with mouse wheel scrolling.
There is no need to change options in 'Configure" menu with mouse wheel scrolling.
Current behavior creates situations when user can mess up their configuration just by scrolling through options and not realize it because UI doesn't make it very obvious.
Plus this duality of the scrolling action is just annoying because you have to dodge the selection boxes to just navigate through the menu.

duhow:
Fixes #1024

### DIFF
--- a/lutris/gui/config_boxes.py
+++ b/lutris/gui/config_boxes.py
@@ -252,12 +252,18 @@ class ConfigBox(VBox):
                 combobox.set_active_id(default)
 
         combobox.connect('changed', self.on_combobox_change, option_name)
+        combobox.connect('scroll-event', self.on_combobox_scroll)
         label = Label(label)
         label.set_alignment(0.5, 0.5)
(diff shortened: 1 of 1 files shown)


tannisroot:
With this hovering over option boxes stops the scrolling.
I don't think this is the way it should be.

strycore:
Can confirm, scrolling through a combo box stops the scrolling altogether. The event has to bubble up to its parent container, we just want to prevent the default handler from running.

strycore:
I fixed this in https://github.com/lutris/lutris/commit/a745d016d30489144c23c14071d4bd50bbbc2088 

It's very close to the original PR but follows this method instead: https://stackoverflow.com/questions/7611813/remove-scroll-ability-from-gtkspinbutton

PR #1031, opened 2018-08-16
Fix detection of xgamma

### LINKED ISSUE #1010: “Restore default gamma....” doesn’t restore gamma after game quits

luthub:
Fixes #1010


tannisroot:
Can this be merged to master too?

strycore:
Very cool! I don't think we want to handle edge cases where someone installs xgamma while lutris is running because that's so unlikely to happen while still inflicting a performance penalty on all user without xgamma installed.

strycore:
@tannisroot yes, I'll merge it

PR #1046, opened 2018-08-19
dxvk: Fix wine_dll_path check

elovin:
When a new wine prefix is being used the original wine_dll_path is moved (because of the optional backup),
so the removal failed.


ibrokemypie:
confirmed to work for me

strycore:
thanks!

PR #1050, opened 2018-08-21
Add ScriptError to game folder creation

### LINKED ISSUE #930: No feedback given if 'main game folder' doesn't exist and lutris has no permissions.
Lutris seems to hang without any kind of error dialog. e.g. my game dir would be `/opt/wine`, and I want to install a game into `/opt/wine/$game`, there's no feedback that not only does `/opt/wine` not exist, but Lutris can't create the game's parent directory. This could be improved.

NicoVIII:
This should close #930 

I simply raised an ScriptError, if lutris does not have the permission to create target path.

Is there a better way to report an error to the user in the installation process?
I'm not happy with the installer not to close completely but show the message and then stay open and stuck.

(I hope this is the right branch to merge to)

### DIFF
--- a/lutris/installer/interpreter.py
+++ b/lutris/installer/interpreter.py
@@ -251,8 +251,12 @@ class ScriptInterpreter(CommandsMixin):
                 os.mkdir(self.cache_path)
 
             if self.target_path and self.should_create_target:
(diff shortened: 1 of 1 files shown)


strycore:
Thanks! That should work nicely!

PR #1054, opened 2018-08-21
fixes "ResourceWarning: unclosed file"

keturn:
read_yaml_from_file: make sure file handle is closed

while I was here I also switched it to using `safe_load`, which isn't strictly necessary since the config file is trusted, but is a good default habit unless we really intend the config file to be able to construct instances of arbitrary Python classes.


strycore:
Nice, thanks!

PR #1069, opened 2018-08-25
 Add 'Primary' option to 'Turn off monitors except'.

openglfreak:
I have a laptop that I use at two locations, one of which has an external monitor.  
I don't want to always change the 'Turn off monitors except' option, so I added the option to turn off all monitors except the primary one.


strycore:
Looks good!

PR #1073, opened 2018-08-27
Add Turkish translation file

yunusem:
Add tr.po file to the source code.

We are planning to use lutris and generate debian package from source. I am creating this request for end-user requirements of us. [Pardus](https://www.pardus.org.tr/) team.


strycore:
Thanks! Please note that i18n is not currently fully implemented.

PR #1076, opened 2018-08-30
Normalized platforms

telanus:


lutris/runners/libretro.py
```diff
@@ -52,7 +52,7 @@
     ('Neko Project 2 (NEC PC-98)', 'nekop2', 'NEC PC-98'),
     ('Neko Project II kai (NEC PC-98)', 'np2kai', 'NEC PC-98'),
     ('O2EM (Magnavox Odyssey²)', 'o2em', 'Magnavox Odyssey²'),
-    ('ParaLLEl N64 (Nintendo 64)', 'parallel_n64', 'Nintendo 64'),
+    ('ParaLLEl N64 (Nintendo 64)', 'parallel_n64', 'Nintendo N64'),
```
RobLoach: [Datomatomatic](http://datomatic.no-intro.org) names it "Nintendo 64". If anything, I'd opt to be even more verbose and roll with the full datomatic name...

```
Nintendo - Nintendo 64
```

PR #1094, opened 2018-09-07
Restore custom refresh rates and modes when restoring resolution (next)

Medath:
Functionally the same as #1049 but rewritten for the next branch to be compatible with #1069.


strycore:
thanks!

Issue #1106, opened 2018-09-13
DXVK versions not updating when dxvk directory is missing


Medath:
When `~/.local/share/lutris/runtime/dxvk/` is not present (because a user never used dxvk before), Lutris fails to create `dxvk_versions.json` because the parent directory does not exist, which leads to Lutris showing the (outdated) fallback DXVK versions. A workaround is to create the missing directory manually and to restart Lutris.

![Missing DXVK versions](https://user-images.githubusercontent.com/40150792/45500181-5c790f00-b77e-11e8-9f09-2cbf351ae5eb.png)

elovin:
1107 does not fix the missing directory, Lutris should create this directory during installation.

Medath:
No, #1107 does not fix this. It just fixed a typo.
I personally think that creating dxvk_versions.json in the the runtime directory rather than the dxvk directory would be easier.

tannisroot:
can you reopen this? it was not solved by that PR

Tom-Todd:
Original fix assumed that lutris and lutris/runtime in local were present.
Changed call time of dxvk version check to ensure lutris has created the folders.
Call to get dxvk versions was occuring prior to application init function.

PR #1107, opened 2018-09-13
update fallback dxvk versions and fix a typo

Medath:
When Lutris gets to show the fallback DXVK versions (Because someone fell victim to #1106 or for some other reason), it shows `0.530.52` as available version which is caused by a missing comma in the code. I also used this opportunity to add 0.70 and 0.71 (and 0.62 to keep the code beautiful) to the list (Which will probably be outdated till the next Lutris release anyway).


strycore:
thanks! nice catch!

Issue #1104, opened 2018-09-13
[Feature] ZDoom command options


bebop350:
I'm on Lutris v0.4.18.1 and notice it seems Lutris can't accommodate command options for ZDoom like specifying the save or config directory. 

For example I run this to launch Brutal Doom:
gzdoom -iwad doom -file bd21testApr25.pk3 DoomMetalVol4.wad -config /Gaming/Linux/Installers/Doom/wads/BrutalDoom/brutaldoom.ini -savedir /Gaming/Linux/Installers/Doom/wads/BrutalDoom

One could use .sh files for the games and add as Linux native titles, but it's not ideal.

Something to consider. Thank you.

RobLoach:
Great idea... The needed changes would go over at:
https://github.com/lutris/lutris/blob/master/lutris/runners/zdoom.py

SenileAnimal:
My apologizes if I gave the wrong impression or tagged this inappropriately, I have only a (very) basic familiarity with Python (and code in general) and Github. In retrospect a [Suggestion] tag would be more fitting?

Regardless, I'll see if it's feasible for me to do. Maybe the args section from the Linux runner can be used.

RobLoach:
No worries! Always looking to improve it, just trying to encourage you to get your hands dirty :wink: 

If you copy some of the options, you'll see how the Lutris Runner API works. You would...

1. You'd copy an options like this:
https://github.com/lutris/lutris/blob/master/lutris/runners/zdoom.py#L46-L51

2. Add it to the command later on:
https://github.com/lutris/lutris/blob/master/lutris/runners/zdoom.py#L104-L107

SenileAnimal:
Thank you.

I'd like to try to add save and config fields where the user browses to the desired directories, plus an arguments field. If I can't manage it, just an arguments field.

I guess the thing to do is get Lutris updated then see about forking zdoom.py.

bebop350:
Resolved with 1113 merge into master

PR #1113, opened 2018-09-14
zdoom.py add config, save, and arguments. Fix skill

bebop350:
Added config and save file selection to the Game Options menu. Allows the user to keep multiple config and save files in different directories for different games.

#1104

### DIFF
--- a/lutris/runners/zdoom.py
+++ b/lutris/runners/zdoom.py
@@ -28,6 +28,19 @@ class zdoom(Runner):
             'type': 'string',
             'label': 'Warp to map',
             'help': ("Starts the game on the given map.")
+        },
+        {
+            'option': 'config',
+            'type': 'file',
+            'label': 'Config file',
+            'help': ("Used to load a user-created configuration file. If specified, "
+                     "the file must exist and contain the wad directory list or launch will fail.")
+        },
+        {
+            'option': 'savedir',
+            'type': 'directory_chooser',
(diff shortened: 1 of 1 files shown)


RobLoach:
Nice work! Should these be Runner options instead of Game options? Or is Game options intentional?

bebop350:
Thank you. I was debating the same thing. I'm open to suggestion. 

For what it's worth, I noticed as far as ZDoom is concerned [https://zdoom.org/wiki/Command_line_parameters#](url) skill and warp selection are considered "loading options". 1 of which is on the game options area, the other the runner options. 

Looking now I see config and save launch options are considered "configuration options". The more I think about it, I would agree it's a bit more relevant to the runner. I'd be happy to add the changes there.

RobLoach:
Hmmm, I see the save is `save0.zds`, it doesn't prefix the game name...... Should "Save Path" be game option? I think Config file makes sense as a Runner.

bebop350:
If I understand you correctly, I  thought that naming scheme was standard behaviour? My saves both in testing with Lutris and in general are all labeled as such, but added to the user defined directory.

Save path might make a bit more sense as a game option, that's a good point.

bebop350:
Game options with added Arguments and Save path fields:
lutrisgameoptionscut

Runner options with added Config file field:
lutrisrunneroptionscut

strycore:
Looks good!

PR #1115, opened 2018-09-15
Added initial implementation for removing runners from GUI (closes #1057)

### LINKED ISSUE #1057: Allow deleting Runners from the UI
I need to because of a bug in the version of steam I am using.

AlexanderRavenheart:
Closes #1057


strycore:
Looks good, although the runners should be checked to see if the current condition can apply to all of them

PR #1118, opened 2018-09-15
Fix wine libs not being preferred for winetricks

### LINKED ISSUE #1100: Winetricks doesn't launch when game's version of wine mismatches system-wide version
If the user has wine installed system-wide, has a game installed in Lutris with Runtime enabled, and that game is using a version of wine that doesn't match the system version (for example,  3.14 with staging patches when on the system there is staging-3.15 and vice versa), the Winetricks will fail to start with the folllowing output

> WINEPREFIX INFO:
> Drive C: total 28
> drwxr-xr-x  6 alex alex 4096 авг 31 15:19 .
> drwxr-xr-x  4 alex alex 4096 сен  4 18:09 ..
> drwxr-xr-x  4 alex alex 4096 сен  4 15:28 ProgramData
> drwxr-xr-x  8 alex alex 4096 авг 31 15:21 Program Files
> drwxr-xr-x  4 alex alex 4096 авг 31 15:19 users
> drwxr-xr-x 16 alex alex 4096 сен  1 08:28 windows
> -rw-r--r--  1 alex alex    9 авг 31 15:19 .windows-serial
> 
> Registry info:
> /home/alex/Games/the-sims/system.reg:#arch=win32
> /home/alex/Games/the-sims/userdef.reg:#arch=win32
> /home/alex/Games/the-sims/user.reg:#arch=win32

But if the users gets the versions match, OR removes wine from their system, OR disables Runtime, the issue will disappear.
Removing the runtime completely (except for winetricks) while Lutris is still running doesn't change anything. Actually, getting rid of winetricks too doesn't change the error either.
This affects a lot of users (as can be seen in #1085), and it's not specific to the platform used.

Medath:
Fixes #1100.

The issue was that when launching winetricks, wine_path pointed to the winetricks binary (instead of wine, which would usually be the case) and the wine libs are only preferred if WINE_DIR is in wine_path which isn't the case in this situation. When winetricks is launched, winetricks_wine is the variable that contains the actual wine path, so I added a check for it.

### DIFF
--- a/lutris/runners/wine.py
+++ b/lutris/runners/wine.py
@@ -256,6 +256,8 @@ def wineexec(executable, args="", wine_path=None, prefix=None, arch=None,
     if use_lutris_runtime(wine_path=wineenv['WINE'], force_disable=disable_runtime):
         if WINE_DIR in wine_path:
             wine_root_path = os.path.dirname(os.path.dirname(wine_path))
+        elif WINE_DIR in winetricks_wine:
+            wine_root_path = os.path.dirname(os.path.dirname(winetricks_wine))
         else:
             wine_root_path = None
         wineenv['LD_LIBRARY_PATH'] = ':'.join(runtime.get_paths(


tannisroot:
Can confirm, the fix works (Arch).

Holzhaus:
I can also confirm that this fixes the winetricks issue for me. Btw I use arch ;-)

strycore:
Thanks!

PR #1136, opened 2018-09-29
runnerinstalldialog.py: Made dialog 60px wider to account for the progress bar.

### LINKED ISSUE #1135: "Manage Versions` window is too narrow
Recent improvements to Wine version managing are currently a bit pointless due to Manage Versions windows being too narrow and not being able to display the status of the downloading process:
peek 2018-09-29 11-37

Tom-Todd:
Fixes #1135


tannisroot:
Hmm. 400 doesn't seem enough. I think having it at 440 would be better:
screenshot from 2018-09-29 12-40-12
screenshot from 2018-09-29 12-42-46

Tom-Todd:
Hm, looked OK on my end, I'll update it

Tom-Todd:
Updates to be larger, now at 460px instead of 400 to account for differences in people's DE / Themes

PR #1138, opened 2018-09-30
Added new esync option, which adds the WINEESYNC environment variable.

### LINKED ISSUE #1131: [Feature Request] Add Esync toggle + check and warn low limits
It would be nice to have a Esync toggle that allows to easily change the value of `WINEESYNC` envar (1 for on and 0 for off) if an esync build (they all have esync in their name) is in use. But to avoid issues on distributions that still use low file descriptor limit, Lutris should check it with `ulimit -r' and if that returns a value that it lower than 1048576, turning the toggle on will display a warning message that will deny turning it on and suggest to increase the limits value (and link https://github.com/lutris/lutris/wiki/How-to:-Esync).

Tom-Todd:
Added checkbox with callback to allow for checks when enabling esync.

The checkbox will not enable, if ulimit -Hn does not return correctly. There is also a warning displayed if the user tries to enable it with a wine version without 'esync' in the title, to remind them to check that it is esync capable. This does not stop them enabling it however, to allow for custom versions.

Fixes #1131

### DIFF
--- a/lutris/gui/config_boxes.py
+++ b/lutris/gui/config_boxes.py
@@ -152,6 +152,9 @@ class ConfigBox(VBox):
         elif option_type == 'bool':
             self.generate_checkbox(option, value)
             self.tooltip_default = 'Enabled' if default else 'Disabled'
+        elif option_type == 'extended_bool':
+            self.generate_checkbox_with_callback(option, value)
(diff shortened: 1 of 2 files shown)


strycore:
nicely done, thanks!

PR #1147, opened 2018-10-03
Improve behaviour of Lutris background process

manuelVo:
This commit adresses the following issues:
 * Closing the Lutris main window kills games started by Lutris
 * Starting a game using `lutris:rungameid` closes the lutris main window
 * Opening the main window after starting a game with `lutris:rungameid` quits Lutris when that game closes.

With this commit Lutris won't shut down anymore when using the close button of the main window. The main window will simply be hidden (it can still be shut down via Lutris->Quit). It can be made visible again by re-launching Lutris. This allows the Game to continue to run and be monitored by Lutris even though that window is gone. If the window is made visible again it will show up in the proper state - showing which game is currently running and able to provide the logs of that game).

Shutting down Lutris now works by polling two conditions: If the game is stopped and if the window is hidden. If both conditions are true Lutris will shut down. For this polling I reused the existing polling mechanism for quitting the background process launched by `lutris:rungameid`.

I skimmed the code of #1117 (tray icon support) and think both PRs should play nicely with each other, however I didn't test that so far.

### DIFF
--- a/lutris/gui/application.py
+++ b/lutris/gui/application.py
@@ -173,7 +173,7 @@ class Application(Gtk.Application):
                 self.css_provider,
                 Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION
             )
-        self.window.present()
+            GLib.timeout_add(300, self.refresh_status)
 
     @staticmethod
     def _print(command_line, string):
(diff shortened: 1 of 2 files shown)


strycore:
That's pretty cool tbh, thanks!

PR #1152, opened 2018-10-05
Improve formatting in the Wine dialog and add a missing dot.

tannisroot:
Before:
screenshot from 2018-10-05 12-12-56
After:
screenshot from 2018-10-05 12-12-37


RobLoach:
Unsure the newline is needed, since it's building on the same idea as the previous sentence.

tannisroot:
You mean like this? I mean, it looks fine but I feel like it's a bit of a waste of space.
image

strycore:
I actually like the last one better!

tannisroot:
Ok, I made it look like the last screenshot.

strycore:
Nice!

Issue #962, opened 2018-07-06
Lutris syncs STEAM games when sync is turned off (library)


BloodyIron:
I didn't turn sync on, and so far as I can tell, it's off (the slider is set to off) to sync with STEAM.

But Lutris keeps syncing my STEAM library with its library, and I don't want it to!

I'm on 0.4.18.

I've heard this happening to other people too. Thanks! :D

strycore:
I think what's happening is that Lutris is syncing your Lutris library, which contains all your Steam games if your have synced your library on the website. The game Lutris sees as installed as likely to be marked as such.

NicoVIII:
As far as I know, Lutris does not delete a game from the library on the lutris homepage, if you delete it in the lutris application. Therefore this game will always show up again, if you don't delete it on the webseite.

You could test, if this is your problem by deleting a game you think is imported through steam on the lutris website and then delete it in the lutris application and restart.

BloodyIron:
I don't have the program logged into my account. It's pulling from the native STEAM library I have on the same system.

KonomiKitten:
> I don't have the program logged into my account. It's pulling from the native STEAM library I have on the same system.

I also have the same problem and I'm also not logged into any account that has to do with Lutris.

keturn:
I was going to take a shot at this one, as I've seen it myself.

After a little testing, it seems like the entry isn't imported when Lutris starts, it's imported when the local Steam client starts.

Ah, okay, I got it to trigger again:

`LutrisWindow.__init__` sets up a `SteamWatcher`, with callback `LutrisWindow.on_steam_game_changed`, and that will do a `mark_as_installed`, and it doesn't seem to be checking any sort of "should I sync?" in the process.

When is it the right time for it to check that? Should it not start the SteamWatcher at all, if it's not set up to sync?

Or does the first part of `on_steam_game_changed`, the part that works with `running_game`, still need to get updated, but it should stop after that if sync isn't on?

NicoVIII:
After @keturn s work I could reproduce as well. If steam and lutris are open and I install a game, lutris adds it.

I guess that lutris removes a game if it is uninstalled from steam is useful even without enabled syncing because it can not be used from lutris anymore. But if live syncing is fixed, we should change the text as well. For now it justs says that libraries are synced when lutris starts.

Just to put my two cents in. :smile:

BloodyIron:
I don't want it to sync with sync set to off, whether I'm logged in or not. It creates too much clutter!

[...]

PR #1160, opened 2018-10-07
LutrisWindow: Added check to see if steam games should be synced when…

Tom-Todd:
… watcher detects change.

Addresses issue #962


strycore:
Cool, Thanks!

moll:
Gotcha. Thanks, @Tom-Todd. I was indeed under the impression the syncing toggle was suppose to disable _all_ Steam syncing, not just new games.

Issue #1122, opened 2018-09-19
Introduce fail message for the installation process


tannisroot:
To avoid stuff like #1121, at least with Wine installers, Lutris should check if the game executable exists after the installer finished and if it doesn’t, show a fail message (instead of a success like it is now).

Kreyren:
Agreed, but i can't imagine an effective way of implementing such feature since every installer is different.

duhow:
Lets say...
```yaml
- task:
    name: wineexec
    prefix: $GAMEDIR
    executable: $installer
- expect:
    file: "$GAMEDIR/Program Files/Game/game.exe"
    status: exists
- expect:
    regkey: "HKLM/....."
    status: exists
```

tannisroot:
Or Lutris could just check for `game: exe:` (game’s executable). 
That would be much easier to implement.

AlexanderRavenheart:
Would be easier to implement but would not cover multiple possible fail scenarios, like corrupt/inexistent registry entries, other files missing, etc.

Tom-Todd:
Currently just checks for the expected executable. Think this is a decent start.

tannisroot:
Closed by https://github.com/lutris/lutris/pull/1172

PR #1172, opened 2018-10-09
installer/interpreter: Add check that expected exe is in the correct …

Tom-Todd:
…location.

Fixes issue #1122

### DIFF
--- a/lutris/installer/interpreter.py
+++ b/lutris/installer/interpreter.py
@@ -532,34 +532,36 @@ class ScriptInterpreter(CommandsMixin):
     # ----------------
 
     def _finish_install(self):
-        exe = self.script.get('game').get('exe')
-        if not exe is None:
-            path = self._substitute(exe)
+        game = self.script.get('game')
+        launcher, launcher_value = self._find_launcher_in_script(game)
+        if game:
+            launcher_value = game.get('exe')
+        if launcher_value:
+            path = self._substitute(launcher_value)
             if not os.path.isabs(path):
-                path = os.path.join(self.target_path, exe)
+                path = os.path.join(self.target_path, launcher_value)
 
-        if not os.path.isfile(path):
+        if path and not os.path.isfile(path):
             self.parent.set_status("Installation didn't complete successfully")
             self.parent.on_install_error("Installation failed (specified"
             " executable not found)!")
-            return
+        else:
+            self.parent.set_status("Writing configuration")
+            self._write_config()
+            self.parent.set_status("Installation finished !")
+            self.parent.on_install_finished()
 
-        self.parent.set_status("Writing configuration")
-        self._write_config()
-        self.parent.set_status("Installation finished !")
-        self.parent.on_install_finished()
-
-    def _get_game_launcher(self):
+    def _get_game_launcher(self, script):
         """Return the key and value of the launcher"""
         launcher_value = None
 
         # exe64 can be provided to specify an executable for 64bit systems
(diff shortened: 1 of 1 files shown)


lutris/installer/interpreter.py
```diff
@@ -532,6 +532,18 @@ def _map_command(self, command_data):
     # ----------------
 
     def _finish_install(self):
+        exe = self.script.get('game').get('exe')
+        if not exe is None:
+            path = self._substitute(exe)
+            if not os.path.isabs(path):
+                path = os.path.join(self.target_path, exe)
+
+        if not os.path.isfile(path):
```
strycore: what if path set to if exe is None?

lutris/installer/interpreter.py
```diff
@@ -532,6 +532,18 @@ def _map_command(self, command_data):
     # ----------------
 
     def _finish_install(self):
+        exe = self.script.get('game').get('exe')
```
strycore: if the script doesn't have a 'game' section, this will crash.

lutris/installer/interpreter.py
```diff
@@ -532,6 +532,18 @@ def _map_command(self, command_data):
...
+        if not exe is None:
```
strycore: if not exe is simpler

lutris/installer/interpreter.py
strycore: there are other ways to provide the 'exe', sometimes it's referred to as 'iso', 'main_file', 'rom'. There is code in the interpreter that handles that, maybe it should be refactored and used here as well.

lutris/installer/interpreter.py
```diff
@@ -532,6 +532,18 @@ def _map_command(self, command_data):
...
+            path = self._substitute(exe)
+            if not os.path.isabs(path):
+                path = os.path.join(self.target_path, exe)
+
+        if not os.path.isfile(path):
+            self.parent.set_status("Installation didn't complete successfully")
+            self.parent.on_install_error("Installation failed (specified"
+            " executable not found)!")
+            return
```
strycore: That's questionable.

lutris/installer/interpreter.py
Tom-Todd: I tested it out with an installer without a game section, this didn't seem to cause a crash, however, I agree it could do with changing. As for the other ways to refer to the executable, I'll take another look and take this in to account

lutris/installer/interpreter.py
```diff
@@ -532,22 +532,35 @@ def _map_command(self, command_data):
...
+            self.parent.on_install_error("Installation failed (specified"
+            " executable not found)!")
+        else:
+            self.parent.set_status("Writing configuration")
+            self._write_config()
+            self.parent.set_status("Installation finished !")
+            self.parent.on_install_finished()
 
-    def _get_game_launcher(self):
+    def _get_game_launcher(self, script):
```
strycore: what did just happen here?

tannisroot:
Works nicely now so far!

lutris/installer/interpreter.py
```diff
@@ -532,31 +553,24 @@ def _map_command(self, command_data):
...
-            launcher = None
-        return (launcher, launcher_value)
+        game = self.script.get('game')
+        if game:
+            launcher, launcher_value = _get_game_launcher(game)
+        if launcher_value:
+            path = self._substitute(launcher_value)
+            if not os.path.isabs(path):
+                path = os.path.join(self.target_path, launcher_value)
+            print(launcher_value)
```
strycore: you left a print here

PR #1176, opened 2018-10-10
config_boxes: fix issues with file chooser, ensure parent exists to s…

Tom-Todd:
…top error and account for file choosers

using '~'

Fixes #1175

PR #1178, opened 2018-10-11
Fix rpcs3 runner to run games on launch

Roxxers:
The previous fix to this was slightly incorrect. This changes the main file argument from the game folder to the **full** path of the EBOOT.BIN file. 

Previously lutris would run the command `rpcs3 "/dir/to/gamefolder"`
This is now `rpcs3 "/full/dir/to/EBOOT.BIN"`

This now launches the game when the emulator launchs.

[Changes in action](https://gfycat.com/gifs/detail/OldShyChimpanzee)

If you want to keep the game folder solution, the change needs to add "USRDIR/EBOOT.BIN" to the end of the directory and check for it to be a full directory path but this seemed a simpler solution.

### DIFF
--- a/lutris/runners/rpcs3.py
+++ b/lutris/runners/rpcs3.py
@@ -12,7 +12,7 @@ class rpcs3(Runner):
             "option": "main_file",
             "type": "file",
             "default_path": "game_path",
-            "label": "Full path to EBOOT.BIN (/home/...)"
+            "label": "Path to EBOOT.BIN"
         }
     ]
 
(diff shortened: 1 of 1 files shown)


Roxxers:
#1176 actually solves the issue I had with full directory paths so I have changed the wording to better fit; since Lutris is meant to convert ~ to the full path but it is just bugged atm. I have kept the EBOOT file selector instead of the directory approach. Its probably better to point to the file rather than assume the file structure of the user. It isn't that much extra work for the end user with a standard directory tree and allows those with non-standard tree's to still run the games.

PR #1186, opened 2018-10-13
Add vulkan check

### LINKED ISSUE #1038: Have Lutris check if Vulkan is present on system
Automatically check if vulkan drivers are installed in the system
    or if vulkan even just plain works(edited)
    and if it doesn't, just grey out DXVK option

Per @tannisroot on Discord

Tom-Todd:
Fixes #1038
Add warning messages to be shown to user, if vulkan libraries are not found in the expected locations. This warning is dismissible (don't show again option) so that advanced users or users with unusual system setups can get rid of it.

### DIFF
--- a/lutris/runners/wine.py
+++ b/lutris/runners/wine.py
@@ -522,11 +522,12 @@ def build_vulkan_error(option, on_launch):
         checkbox_message = "Enable anyway and do not show this message again."
 
     DontShowAgainDialog('hide-no-vulkan-warning',
-        message,
-        secondary_message="Please follow the installation procedures as described in\n"
-        "<a href='https://github.com/lutris/lutris/wiki/How-to:-DXVK'>"
-        "How-to:-DXVK(https://github.com/lutris/lutris/wiki/How-to:-DXVK)</a>",
-        checkbox_message = checkbox_message)
+                        message,
+                        secondary_message="Please follow the installation "
+                        "procedures as described in\n"
+                        "<a href='https://github.com/lutris/lutris/wiki/How-to:-DXVK'>"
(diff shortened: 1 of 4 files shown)


lutris/runners/wine.py
```diff
@@ -507,6 +508,39 @@ def get_real_executable(windows_executable, working_dir=None):
...
+
+    DontShowAgainDialog('hide-no-vulkan-warning',
+                        message,
+                        secondary_message="Please follow the installation "
+                        "procedures as described in\n"
+                        "<a href='https://github.com/lutris/lutris/wiki/How-to:-DXVK'>"
+                        "How-to:-DXVK(https://github.com/lutris/lutris/wiki/How-to:-DXVK)</a>",
+                        checkbox_message=checkbox_message)
+
+def dxvk_vulkan_check(on_launch):
```
strycore: this has nothing to do with wine

lutris/util/vulkan.py
```diff
@@ -0,0 +1,19 @@
+"""Vulkan helper module"""
+import os
+from enum import Enum
+
+def vulkan_check():
+    vulkan_lib = os.path.isfile("/usr/lib/libvulkan.so")
+    vulkan_lib32 = os.path.isfile("/usr/lib32/libvulkan.so")
+    vulkan_lib_multi = os.path.isfile("/usr/lib/x86_64-linux-gnu/libvulkan.so")
```
Vexatos: Note that this probably won't work on certain distributions, where the file, for instance, is actually called `/usr/lib/x86_64-linux-gnu/libvulkan.so.1`.

PR #1187, opened 2018-10-13
Add missing statement to set value if no callback_on is set or when n…

Tom-Todd:
…ot when active is the opposite of

callback_on

Fixes issue from PR 1138


Tom-Todd:
Note, this is also fixed as a side effect in PR 1186, but in case that one doesn't pass review, I've submitted this one too.

PR #1190, opened 2018-10-14
Fix check for vulkan on launch.

Tom-Todd:
Previous change to vulkan check refactored the checking code to move it in to utill/vulkan missed changing the call to this in the launching code. Fixed in this PR


lutris/runners/wine.py
```diff
@@ -517,7 +517,7 @@ def display_vulkan_error(option, on_launch):
     if option == 2:
         message = "64-bit Vulkan loader was not detected."
 
-    if message:
+    if option != 3:
```
strycore: wth does 3 mean?!!!

PR #1192, opened 2018-10-14
Refactor for vulkan util and check in wine runner for readability.

Tom-Todd:


strycore:
Ah cool, much more understandable now!

PR #1199, opened 2018-10-16
Deleting an unnecessary line

LeandroStanger:


LeandroStanger:
@cho2 You could translate appdata and desktop to Indonesian?

cho2:
@LeandroStanger
 https://github.com/lutris/lutris/blob/master/share/appdata/lutris.appdata.xml and https://github.com/lutris/lutris/blob/master/share/applications/lutris.desktop ?

LeandroStanger:
@cho2 Yes

Tom-Todd:
The line you removed should now end with a />

PR #1200, opened 2018-10-16
forbid overwriting prefixes and fix a bug

sigmaSd:
1- I think overwriting pfx should be not be allowed at all, I cant think of a good use case where its necessary to overwrite a pfx and on the other-hand if a user installs a game on the root of a windows partition and deletes it he will delete the whole partition, obviously there are a lot of warnings but its a big deal and it happened [https://forums.lutris.net/t/lutris-deleted-my-whole-partition-data/3853](https://forums.lutris.net/t/lutris-deleted-my-whole-partition-data/3853) and I don't see the benefit.
2-  os.path.exists(path) and os.listdir(path) cant detect paths with "~" correctly they will always output false , we have to expand username first

### DIFF
--- a/lutris/gui/installerwindow.py
+++ b/lutris/gui/installerwindow.py
@@ -309,7 +309,7 @@ class InstallerWindow(Gtk.ApplicationWindow):
         
         # replace ~ with full path so os.path.exists and listdir work correctly
         if path.startswith("~/"):
-            path = path.replace("~",os.path.expanduser('~'),1)
+            path = path.replace("~", os.path.expanduser('~'), 1)
 
         if os.path.exists(path) and os.listdir(path):
             self.non_empty_label.show()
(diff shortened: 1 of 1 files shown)


tannisroot:
I am against forbidding to overwrite prefixes. Many people use this feature to update their installations by running the updated installer without specifying new prefix path.

sigmaSd:
Interesting, I'll just add a seperate pr for the bug fix, and keep this one open for now to see if someone else has another opinion.

tannisroot:
But what is the path fix for? What issue does your this solve exactly?

sigmaSd:
the warning dosent appear if u change dir to path that contains "~" basically just download a game change the dir to a non empty pfx and u wont see the warning

tannisroot:
What warning? Change path of what? Prefix?

sigmaSd:
the $GAMEDIR path, this warning                 "<b>Warning!</b> The selected path " "contains files, installation might not work properly."

tannisroot:
Huh, the warning really does not appear if you change to a path with ~ during installation. Fair enough.

sigmaSd:
u convinced me about the overwrite issue I'll look for another method to prevent ppl from deleting their partition.

PR #1201, opened 2018-10-16
bugfix

sigmaSd:
os.path.exists(path) and os.listdir(path) cant detect paths with "~" correctly they will always output false , we have to expand username first, discssed here [https://github.com/lutris/lutris/pull/1200](https://github.com/lutris/lutris/pull/1200)


RobLoach:
Would `os.path.abspath()` do it? Would save you some string manipulation.

sigmaSd:
abspath is equivalent to `normpath(join(os.getcwd(), path))`, which basically joins current dir to arg dir, which is not the intended result, plus even if i figure out  something for os.path.exists I still have os.listdir to figure out, so this is the easiest way I found for now

PR #1205, opened 2018-10-17
Sidebar shows number of installed games per runner and platform.

Wcubed:
This adds an extra field to the sidebar, which shows the number of games that are currently installed for that runner/platform.


tannisroot:
How does that look in the UI?

Wcubed:
Like so :)
lutristreecount

tannisroot:
This is pretty neat, I guess?

PR #1206, opened 2018-10-17
Add check on esync at launch, fix limit check condition, change esync…

Tom-Todd:
… message.


Tom-Todd:
Also changed wine version check to look at the return of --version on the wine executable

PR #1214, opened 2018-10-19
[WIP] add playtime to lutris 

sigmaSd:
screenshot from 2018-10-19 12-36-37

1 - it works but the time gets updated only after i close and open lutris not sure how to update it
2 - code is messy now would appreciate some pointers to improve it, in particular this:
```python 
 def on_game_quit(self):
        
        self.timer.end_t()
        if self.playtime == '':
            self.playtime = str(self.timer.duration())
        else:
            self.playtime = (str(datetime.datetime.strptime(self.playtime,"%H:%M:%S") + self.timer.duration())).split()[1]
```


sigmaSd:
Refactored the code , I think its better now, I still didn't find how to to trigger an event to update the window gui to show the new play time.

sigmaSd:
felt like steam style is cooler 
screenshot from 2018-10-19 20-30-50

sigmaSd:
Found out how to trigger row update , its now feature complete xD made a lot of assumptions and guessing about a lot of stuff though so i would appreciate correction. I'm still not sure about design though.

PR #1220, opened 2018-10-20
Don't hide installation notes scrollbar when the notes are longer than the view.

### LINKED ISSUE #959: Always show scrollbar for notes
It's not really evident that there are more notes under these unless you hover over them. Always showing the scrollbar, when it exceeds in length, would be better practice.

**Shown:**

Do not sign-in when prompted. Close Origin and wait for the installation to finish then you can relaunch and sign-in.

Disable Origin In-Game just in case.

Make sure "Disable Lutris runtime" is unchecked in the game's system options tab or else download will fail.

**Hidden:**

Should your download pause at 22% press the <a href="https://i.imgur.com/8Vv0bva.png">play icon</a> until it resumes downloading.

screenshot from 2018-07-01 15-03-23

Wcubed:
This way you can always see when there are more notes to read. Even when the notes have a line break exactly where the scrolled window ends.

scrollbaralwaysvisible

(Closes #959 )

### DIFF
--- a/lutris/gui/installerwindow.py
+++ b/lutris/gui/installerwindow.py
@@ -239,6 +239,7 @@ class InstallerWindow(Gtk.ApplicationWindow):
         except AttributeError:
             logger.debug("set_propagate_natural_height not available")
         notes_scrolled_area.set_min_content_height(100)
+        notes_scrolled_area.set_overlay_scrolling(False)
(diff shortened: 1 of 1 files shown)


tannisroot:
I observer no changes with this (Gnome 3.30):
peek 2018-10-21 01-07

strycore:
Fixed in next

PR #1222, opened 2018-10-20
Make sure the view gets re-sorted on config change, as game gets 'add…

### LINKED ISSUE #1184: Changing configuration with "Installed games first" sorting option puts the game at the bottom
This is quite weird behavior that should be fixed.
peek 2018-10-13 15-50

Tom-Todd:
…ed' again to the store.

Also, change sort_view method, so that it sorts by name first then by installed first if enabled
so that positioning is consistent.

Fixes #1184

### DIFF
--- a/lutris/gui/gameviews.py
+++ b/lutris/gui/gameviews.py
@@ -118,12 +118,11 @@ class GameStore(GObject.Object):
 
     def sort_view(self, show_installed_first=False):
         self.show_installed_first = show_installed_first
+        self.store.set_sort_column_id(COL_NAME, Gtk.SortType.ASCENDING)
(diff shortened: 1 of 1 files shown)


TingPing:
This is a weird workaround, every insert should internally resort in the model.

Tom-Todd:
> This is a weird workaround, every insert should internally resort in the model.

"internally resort in the model" I'm not sure I understand what this means

strycore:
I'm merging this for the release as it seems to work but a proper implementation would be nice.

TingPing:
Glancing at the source of Gtk really quick the actual solution would be to change `self.store.append` to `self.store.insert_with_values` and that will properly resort after inserts.

PR #1223, opened 2018-10-21
Fixing comment syntax of translations.

BunnyApocalypse:
This is a small change to meet what I believe to be the proper syntax for translations of comments in .desktop files. This is currently causing Fedora (and I would guess other rpm based distros) to fail their builds.

-Bunny


strycore:
I've updated the git tag to include this commit and will repackage the tar.xz

BunnyApocalypse:
Sweet, thanks!

PR #1225, opened 2018-10-21
Linter fixes for lutris/util

PythonicChemist:


lutris/util/dxvk.py
```diff
@@ -138,7 +139,7 @@ def enable_dxvk_dll(self, system_dir, dxvk_arch, dll):
                 os.remove(wine_dll_path)
             os.symlink(dxvk_dll_path, wine_dll_path)
 
-    def disable_dxvk_dll(self, system_dir, dxvk_arch, dll):
+    def disable_dxvk_dll(self, system_dir, dll):
```
strycore: I don't think this is correct

PythonicChemist: I've reverted the change.

strycore:
thanks!

PR #1227, opened 2018-10-21
Added regex to check to allow for libvulkan.so.1 and similar as raise…

### LINKED ISSUE #1226: Vulkan check will not work on certain distributions
Note that this probably won't work on certain distributions, where this file, for instance, is actually called `/usr/lib/x86_64-linux-gnu/libvulkan.so.1` or similar. Confirmed not working on Linux Mint 19.

_Originally posted by @Vexatos in https://github.com/lutris/lutris/pull/1186_

Tom-Todd:
…d as an issue on mint

and other distributions.

Fixes #1226

### DIFF
--- a/lutris/util/vulkan.py
+++ b/lutris/util/vulkan.py
@@ -1,5 +1,6 @@
 """Vulkan helper module"""
 import os
+import re
 from enum import Enum
 
 class vulkan_available(Enum):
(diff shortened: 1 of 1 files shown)


Medath:
Can confirm it's getting detected now. But you left a print in the code.

Tom-Todd:
> Can confirm it's getting detected now. But you left a print in the code.

Yeah, fixing that and the failed check currently, but thanks for the heads up

TingPing:
The correct method (that respects ldconfig) is to just use `ctypes.util.find_library()`

PR #1232, opened 2018-10-22
Add call to ldconfig to check for libvulkan, keep file search as backup.

### LINKED ISSUE #1231: latest version 0.4.21 thinks i don't have a vulkan loader
Got the new message pop up to say I don't have a Vulkan loader installed. This is on an Ubuntu 18.10 fresh upgrade.

I do:

> liam@liam-main:~$ sudo apt install libvulkan1 libvulkan1:i386 
[sudo] password for liam: 
Reading package lists... Done
Building dependency tree       
Reading state information... Done
libvulkan1 is already the newest version (1.1.82.0-0ubuntu1).
libvulkan1:i386 is already the newest version (1.1.82.0-0ubuntu1).
0 to upgrade, 0 to newly install, 0 to remove and 5 not to upgrade.

Need more details? Let me know.

Tom-Todd:
As suggested on #1227 added call to ldconfig for more robust check, keeping the search for the files as backup. Tested working on Ubuntu 18.10 and Arch.
Fixes #1231

### DIFF
--- a/lutris/util/vulkan.py
+++ b/lutris/util/vulkan.py
@@ -16,10 +16,10 @@ def vulkan_check():
     for line in subprocess.check_output(["ldconfig", "-p"]).splitlines():
         line = str(line)
         if 'libvulkan' in line:
-            if not 'x86-64' in line:
-                has_32_bit = True
-            else:
+            if 'x86-64' in line:
                 has_64_bit = True
+            else:
+                has_32_bit = True
 
     if not (has_64_bit or has_32_bit):
         return vulkan_available.NONE
--- a/lutris/util/vulkan.py
+++ b/lutris/util/vulkan.py
@@ -1,6 +1,8 @@
 """Vulkan helper module"""
 import os
 import re
+import subprocess
+import io
 from enum import Enum
 
 class vulkan_available(Enum):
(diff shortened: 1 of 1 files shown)


lutris/util/vulkan.py
```diff
@@ -19,12 +21,24 @@ def search_for_file(directory):
...
 def vulkan_check():
-    vulkan_lib = search_for_file("/usr/lib")
-    vulkan_lib32 = search_for_file("/usr/lib32")
-    vulkan_lib_multi = search_for_file("/usr/lib/x86_64-linux-gnu")
-    vulkan_lib32_multi =  search_for_file("/usr/lib32/i386-linux-gnu")
-    has_32_bit = vulkan_lib32 or vulkan_lib32_multi
-    has_64_bit = vulkan_lib or vulkan_lib_multi
+    has_32_bit = False
+    has_64_bit = False
+    pattern = re.compile(r'^libvulkan\.so')
```
TingPing: unused

lutris/util/vulkan.py
```diff
@@ -19,12 +21,24 @@ def search_for_file(directory):
...
+    proc = subprocess.Popen("ldconfig -p", shell=True, stdout=subprocess.PIPE)
```
TingPing: No reason to expose `shell=True`

lutris/util/vulkan.py
```diff
@@ -19,12 +21,24 @@ def search_for_file(directory):
...
+    for line in io.TextIOWrapper(proc.stdout, encoding="utf-8"):
+        if line.find('libvulkan') != -1:
```
TingPing: This is a bit more pythonic:
```python
if 'libvulkan' in line:
    if 'x86-64' in line:
        ...
```

lutris/util/vulkan.py
```diff
@@ -19,12 +21,24 @@ def search_for_file(directory):
...
+            if line.find('x86-64') != -1:
+                has_32_bit = True
+            else:
+                has_64_bit = True
+
+    if not (has_32_bit or has_64_bit):
+        vulkan_lib = search_for_file("/usr/lib")
```
TingPing: I find these fallbacks questionable. If `ldconfig` didn't find vulkan than no application will, so its pointless to find files otherwise.

lutris/util/vulkan.py
TingPing: Honestly just use `subprocess.check_output()` not sure the benefit of using the low level apis.

lutris/util/vulkan.py
```diff
@@ -1,6 +1,8 @@
 """Vulkan helper module"""
 import os
 import re
+import subprocess
+import io
```
TingPing:
```suggestion
```

Tom-Todd:
CodeFactor is complaining about the use of 'ldconfig' rather than an absolute path, not sure of the best way to resolve it, as different distributions can have it in different locations.

Also if 'x86-64' in line: however, I think that it's suggestion would lead to the resulting code being more verbose and less clear.

lutris/util/vulkan.py
```diff
@@ -9,22 +10,16 @@ class vulkan_available(Enum):
...
-    vulkan_lib32_multi =  search_for_file("/usr/lib32/i386-linux-gnu")
-    has_32_bit = vulkan_lib32 or vulkan_lib32_multi
-    has_64_bit = vulkan_lib or vulkan_lib_multi
+    has_64_bit = False
+    has_32_bit = False
+    for line in subprocess.check_output(["ldconfig", "-p"]).splitlines():
+        line = str(line)
+        if 'libvulkan' in line:
+            if 'x86-64' in line:
+                has_32_bit = True
```
strycore: if x86-64 in line then it has 32bit? :thinking:

strycore: Also, with this code, the CodeFactor warning doesn't make sense.

PR #1235, opened 2018-10-23
add playtime to lutris

sigmaSd:
hopefully this is rebased !

PR #1236, opened 2018-10-23
add playtime

sigmaSd:
commit history got really ugly so this was the easiest way 
history:https://github.com/lutris/lutris/pull/1235

### DIFF
--- a/lutris/game.py
+++ b/lutris/game.py
@@ -16,6 +16,7 @@ from lutris.util.log import logger
 from lutris.config import LutrisConfig
 from lutris.thread import LutrisThread, HEARTBEAT_DELAY
 from lutris.gui import dialogs
+from lutris.util.timer import Timer
 
 
 class Game:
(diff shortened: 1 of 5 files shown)


strycore:
It's one way to do it but not the only way, here's a time machine for next time ;) 

https://git-scm.com/book/en/v2/Git-Tools-Rewriting

tannisroot:
@sigmaSd  I don't like that you can only see playtime in a List mode, the feature is useless in the Grid mode, which is default and what most users use.

strycore:
that feature is barely in prototype stage

sigmaSd:
@strycore thanks for the link I'll keep it in mind for next time :) , @tannisroot I agree with  but its like @strycore  said and I'm thinking in the future maybe lutris will get a nice side bar that shows game preview (video image screenshot) and underneath it game info . and anyway gathering playtime info from now is good idea everyone like statics ^^

Issue #831, opened 2018-04-20
[Feature Request] Option to run a script along with the game


diamondburned:
Hello. Just requesting this feature that would launch a script before the start of a game.

RobLoach:
What would you like to accomplish by running a script? What would the script do?

diamondburned:
For example running a script that assists gameplay, or a Discord Rich Presence script, or anything to be fair. Maybe a notification display?

sorsasampo:
I would like to expand on this feature request a bit.

I wrote [`played`](https://gitlab.com/sorsasampo/played), a script to track playtime for games. If Lutris provided some way to wrap all game launches into a custom template script, I could use it to track playtime of all games.

The default could be just to launch the game:

    $COMMAND

To use `played` for tracking playtime, it could be changed to:

    played play "$GAMENAME" $COMMAND

What @diamondburned requested could be written like:

    notify-send "Lutris" "$GAMENAME started"
    $COMMAND

diamondburned:
Awesome idea, I'm up. By the future, if Lutris provides a way to print what game the user's playing I might be able to whip up Rich Presence support with playtime.

Meloncon:
This would also be a great feature for a user to add even simple scripts like `xset r off` to turn off Key Repeat when a game launches. Even a "run script on exit" feature would easily allow a user to toggle Key Repeat back on with very minimal fuss. +1

rubberbandaid:
I agree, this would be great! In my case, to start up a script that presses multiple keys when I hit a single one, which I want to run only while the game runs.
Or, when I start a game I rarely play, I have a script first remind me of various hacks while playing, such as "Hit 'm' to set the FOV to widescreen", as an example.

tannisroot:
The feature has already been implemented.

strycore:
indeed

PR #1242, opened 2018-10-24
Add option to launch script at game start

sigmaSd:
1 - this have some handy side effects, the script actually stops when the game stops so it can be used as one-shot or daemon
2 - Didn't want to add an option to launch script at game stop because it felt like its unnecessary:
if someone for example  wants a function to execute when a game exit he can use the bash trap function

### DIFF
--- a/lutris/game.py
+++ b/lutris/game.py
@@ -4,6 +4,7 @@ import os
 import time
 import shlex
 import subprocess
+from threading import Thread
 
 from gi.repository import GLib, Gtk
 
@@ -63,6 +64,7 @@ class Game:
         self.timer = Timer()
         self.playtime = game_data.get('playtime') or ''
 
+        self.pre_script_thread = None
     def __repr__(self):
         return self.__unicode__()
 
@@ -372,6 +374,9 @@ class Game:
         if self.runner.system_config.get('disable_compositor'):
             self.desktop_effects(False)
 
+        if self.runner.system_config.get('pre_script'):
+            self.pre_script_thread_start()
+
         self.game_thread = LutrisThread(launch_arguments,
                                         runner=self.runner,
                                         env=env,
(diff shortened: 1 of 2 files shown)


sigmaSd:
fix https://github.com/lutris/lutris/issues/831

lutris/game.py
```diff
@@ -503,3 +508,29 @@ def notify_steam_game_changed(self, appmanifest):
             logger.info("Steam game %s updating, setting game thread as not ready",
                         appmanifest.steamid)
             self.game_thread.ready_state = False
+
+    def call_process_with_err(self, arg):
+        """Handle subprocess errors in the new theread"""
```
PythonicChemist: typo "theread"

lutris/game.py
```diff
@@ -372,6 +374,9 @@ def do_play(self, prelaunched, error=None):
         if self.runner.system_config.get('disable_compositor'):
             self.desktop_effects(False)
 
+        if self.runner.system_config.get('pre_script'):
+            self.pre_script_thread_start()
```
PythonicChemist: self.pre_script_thread_start() -> self.pre_script_thread.start() ?

lutris/game.py
```diff
@@ -503,3 +508,29 @@ def notify_steam_game_changed(self, appmanifest):
...
+
+        script = self.runner.system_config.get('pre_script')
+        if os.path.isfile(script):
+            self.pre_script_thread = Thread(target=self.call_process_with_err,
+                                            args=[self.runner.system_config.get('pre_script')])
+
+            self.pre_script_thread.start()
+        else:
+            logger.error(
+                "Script not found at %s" % script)
```
PythonicChemist: Please use lazy evaluation just replace " %" with a comma. :)

lutris/game.py
sigmaSd: thats confusing but its actually the name of the fn `pre_script_thread_start`

PythonicChemist: nvm then

lutris/game.py
sigmaSd: done

sigmaSd:
thanks for the corrections :) if u find other issues I'll be glad to fix them

PythonicChemist:
Maybe put it in its own Preprocess class and then let Game call it via self.preprocess() or something like that? Game class is already quite large and we could benefit from isolating some of its functionalities, I think.

PythonicChemist:
Oh you leave the "%s" in the string for lazy evaluation:

logger.error("Script not found at %s", script)

The logger will do the string manipulation when/if needed.

sigmaSd:
I extracted the logic, indeed it looks nicer ^^

Issue #681, opened 2017-11-02
Open manual/readme


FakeShemp:
Some—especially older—games pretty much require you to read a manual to understand how to operate them. Even more so when they contain crude copy protection. As most digitally released games seem to at least be bundled with a PDF or readme.txt that detail this information, would it be possible to have a button
 somewhere—like in the right-click context menu—that opens up the manual/readme if its location is defined in the config?

strycore:
Yes, that's a good idea. I would make this a bit more generic and add additional launch options for games that have, for example, a map editor or a separate config tool. These additional option would appear in the contextual menu and would be defined in install scripts under a special key for custom actions.

PR #1243, opened 2018-10-24
Add Pre , post  - game and ondemand script fn

sigmaSd:
1 - this have some handy side effects, the script actually stops when the game stops so it can be used as one-shot or daemon
2 - Didn't want to add an option to launch script at game stop because it felt like its unnecessary:
if someone for example wants a function to execute when a game exit he can use the bash trap function

### DIFF
--- a/lutris/gui/lutriswindow.py
+++ b/lutris/gui/lutriswindow.py
@@ -724,8 +724,9 @@ class LutrisWindow(Gtk.ApplicationWindow):
         selected_game = self.view.selected_game
         UninstallGameDialog(game_id=selected_game,
                             callback=self.remove_game_from_view,
-                            parent=self)   
-    
+                            parent=self)
+
+     
     def remove_game_from_view(self, game_id, from_library=False):
         def do_remove_game():
             self.view.remove_game(game_id)
--- a/lutris/gui/lutriswindow.py
+++ b/lutris/gui/lutriswindow.py
@@ -292,7 +292,6 @@ class LutrisWindow(Gtk.ApplicationWindow):
         self.view.connect("remove-game", self.on_remove_game)
         self.view.connect("play-script", self.on_play_script)
         
-
     @staticmethod
     def check_update():
         """Verify availability of client update."""
--- a/lutris/gui/lutriswindow.py
+++ b/lutris/gui/lutriswindow.py
@@ -725,7 +725,8 @@ class LutrisWindow(Gtk.ApplicationWindow):
         UninstallGameDialog(game_id=selected_game,
                             callback=self.remove_game_from_view,
                             parent=self)
-        
+      
+    
     def remove_game_from_view(self, game_id, from_library=False):
         def do_remove_game():
             self.view.remove_game(game_id)
--- a/lutris/gui/lutriswindow.py
+++ b/lutris/gui/lutriswindow.py
@@ -724,8 +724,7 @@ class LutrisWindow(Gtk.ApplicationWindow):
         selected_game = self.view.selected_game
         UninstallGameDialog(game_id=selected_game,
                             callback=self.remove_game_from_view,
-                            parent=self)
-      
+                            parent=self)   
     
     def remove_game_from_view(self, game_id, from_library=False):
         def do_remove_game():
--- a/lutris/gui/lutriswindow.py
+++ b/lutris/gui/lutriswindow.py
@@ -692,12 +692,12 @@ class LutrisWindow(Gtk.ApplicationWindow):
             callback=lambda: self.add_game_to_view(dialog.game.id)
         )
         return True
-    
+        
     def on_play_script(self, *args):
         """Play Script"""
-        game = pga.get_game_by_field(self.view.selected_game, 'id')
-        ScriptThread(game['playscript']).start()
-        
+        game = Game(self.view.selected_game)
+        play_script = game.runner.system_config.get("play_script")
+        ScriptThread(play_script).start()        
         
     def add_game_to_view(self, game_id, is_async=True):
         """Add a given game to the current view
--- a/lutris/gui/lutriswindow.py
+++ b/lutris/gui/lutriswindow.py
@@ -183,7 +183,7 @@ class LutrisWindow(Gtk.ApplicationWindow):
             'stop-game': Action(self.on_game_stop, enabled=False),
             'start-game': Action(self.on_game_run, enabled=False),
             'remove-game': Action(self.on_remove_game, enabled=False),
-            'play-script': Action(self.on_scripts_mgr_btn_clicked, enabled=False),
+            'play-script': Action(self.on_play_script, enabled=False),
 
             'preferences': Action(self.on_preferences_activate),
             'manage-runners': Action(lambda *x: RunnersDialog()),
@@ -290,7 +290,7 @@ class LutrisWindow(Gtk.ApplicationWindow):
         self.view.connect("game-activated", self.on_game_run)
         self.view.connect("game-selected", self.game_selection_changed)
         self.view.connect("remove-game", self.on_remove_game)
-        self.view.connect("play-script", self.on_scripts_mgr_btn_clicked)
+        self.view.connect("play-script", self.on_play_script)
         
 
     @staticmethod
@@ -694,7 +694,7 @@ class LutrisWindow(Gtk.ApplicationWindow):
         return True
     
     @GtkTemplate.Callback
-    def on_scripts_mgr_btn_clicked(self, *args):
+    def on_play_script(self, *args):
         """Play Script"""
         game = pga.get_game_by_field(self.view.selected_game, 'id')
         ScriptThread(game['playscript']).start()
--- a/lutris/gui/lutriswindow.py
+++ b/lutris/gui/lutriswindow.py
@@ -693,7 +693,6 @@ class LutrisWindow(Gtk.ApplicationWindow):
         )
         return True
     
-    @GtkTemplate.Callback
     def on_play_script(self, *args):
         """Play Script"""
         game = pga.get_game_by_field(self.view.selected_game, 'id')
--- a/lutris/gui/lutriswindow.py
+++ b/lutris/gui/lutriswindow.py
@@ -725,7 +725,6 @@ class LutrisWindow(Gtk.ApplicationWindow):
         UninstallGameDialog(game_id=selected_game,
                             callback=self.remove_game_from_view,
                             parent=self)
-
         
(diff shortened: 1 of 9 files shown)


sigmaSd:
fix #831

PythonicChemist:
btw when you have finished squashing and rebasing, you can use the "-f" flag with your push to overwrite your github branch that you already have to get a clean history.

sigmaSd:
Wow finally the info i was looking for, thank u mate that srsly annoyed me and I didn't find the flag xD

tannisroot:
What a great feature! Thanks!

sigmaSd:
Thanks!

strycore:
That's very cool!

Do you think you think you would implement the post-run script option as part of the same PR?

sigmaSd:
I was arguing it is not necessary but here you go :p the post script should be only used as one-shot which makes sense since it would be probably used for cleaning , using it as daemon will cause errors

sigmaSd:
That's a really nice feedback, done!

tannisroot:
Would it be possible to implement an additional functionality for this feature that would delay the start of the actyal game until the script finished?
Also, same question but for a button that runs the script when the users requests it.

sigmaSd:
I think delaying the script is not necessary since you can do it on the script itself (just append sleep for example to the script beginning)  
About the second request it's possible but I'm not sure its worth doing it (since u have to alt tab out off the game than click the button u basically don't gain any automatic function, u can similarly alt tab and start the script) but I'll think about it

sigmaSd:
Actually thinking about it more I like the second suggestion, but It will take a bit of effort to implement, I'll feel motivated if someone has a good use case of this feature to share

tannisroot:
It’s feature that would be extremely useful not exactly for a game, but for buggy game clients like Battle.Net (Blizzard Games) and Origin (EA). 
Battle.net has a nasty issues that causes the agent process to get bugged out sometimes (usually after game updates) promting manual deletion of a certain directory inside the prefix to get it fixed. 
And Origin gets borked when it tried to update itself, but it gets workarounded by copying the contents or a certain dir into the main folder of the program.
If I, as a maintainer of installer scripts, could just write a script that would automate this process and simplify it in just 1 button action, it would be extremely helpful.

sigmaSd:
for feature 1 what about adding a check box to wait till script finishes to start the game?

for feature 2 ur use case is a bit different than what u're suggesting , installer for example only work once and wont help user when he encounter the bug at later time. That said that gives me an idea each game should have a script manager u click the icon u get script list and just hit play the script u want (for example the battle net workaround when issue happen), what do u think? anyhow that's definitely for another pull request its quit a bit of work^^

sigmaSd:
Read the edit, so indeed what ur suggesting is a script manager, hopefully I can give it a shot in another pr

PythonicChemist:
I could see a list of provided helper scripts being useful. Some games require pre-creating a config file and you can make games stop working by misconfiguring them like BlazBlue games and turning on Bloom/Anti-Aliasing. So for these titles having a script fixing issues would be nice.

I am not sure it should be in scope for this very PR though, but would in general be very handy. Maybe create an issue where we could discuss further details of that feature first before trying to wing it in a PR?

sigmaSd:
I'll work on the mockup later and add the issue

sigmaSd:
I add the feature one hopefully that cover ur use case

tannisroot:
Feature 1 - yes, that’s a great idea and solves my usercase.
Feature 2 - I think just a single script that gets triggered by a button in UI (main menu alongside stop, play and remove) would suffice for initial implementation of this. And then you can add what you suggested later in another PR.

tannisroot:
Thank you! This is something I have wanted in Lutris for such a long time, it’s going to improve stuff a lot!

tannisroot:
@PythonChemist you can create configuration files with the current installers. Check documentation on those.

sigmaSd:
I added ur second request, thx to @PythonicChemist  idea to refactor scriptthread it was not that hard to implment

Medath:
> Also, same question but for a button that runs the script when the user requests it.

That could/should be combined with #681, then it would be absolutely perfect. But I think this is a case for another PR. This one is great as it is already.

sigmaSd:
I think this pr handle in part #681 , basicly the installer should configure play script to use `xdg-open` file (pdf, txt,etc)

sigmaSd:
Would someone review this last commit https://github.com/lutris/lutris/pull/1243/commits/cdd1e7d7272115441e7ab19c3342443fbe366b36 is it really needed or there is a way to add play_script config entry for already installed games?

lutris/util/script_thread.py
```diff
@@ -0,0 +1,45 @@
+import os
+import subprocess
+from threading import Thread
+
+from lutris.util.log import logger
+from lutris.gui import dialogs
+
+
+class ScriptThread():
```
strycore: This is called ScriptThread and yet it doesn't subclass Thread ???\!!!

strycore: This class is totally undocumented, what is it supposed to achieve???

lutris/util/script_thread.py
```diff
@@ -0,0 +1,45 @@
...
+
+        try:
+            return subprocess.call(arg)
+
+        except PermissionError:
+            error = "Unable to execute " + arg + ", script is not executable"
+            logger.error(error)
+            dialogs.ErrorDialog(error)
+        except OSError:
+            error = "Unable to execute " + arg + ", script maybe missing Shebang"
```
strycore: String interpolation

lutris/sysoptions.py
```diff
@@ -161,6 +161,34 @@ def get_optirun_choices():
         'advanced': True,
         'help': "Environment variables loaded at run time"
     },
+    {
+        'option': 'play_script',
+        'type': 'file',
+        'label': 'Play Script',
+        'advanced': True,
+        'help': ("Choose a script to execute with script icon")
```
strycore: WAT

lutris/sysoptions.py
```diff
@@ -161,6 +161,34 @@ def get_optirun_choices():
...
+    {
+        'option': 'pre_script',
+        'type': 'file',
+        'label': 'Pre Launch Script',
+        'advanced': True,
+        'help': ("Choose a script to execute at game start")
+    },
+    {
+        'option': 'wait_for_script_completion',
+        'label': "Wait for script completion before starting the game",
```
strycore: WTF

lutris/pga.py
```diff
@@ -90,6 +90,7 @@ def migrate_games():
         {'name': 'has_custom_banner', 'type': 'INTEGER'},
         {'name': 'has_custom_icon', 'type': 'INTEGER'},
         {'name': 'playtime', 'type': 'TEXT'},
+        {'name': 'playscript', 'type': 'TEXT'},
```
strycore: wth is a new field needed in the PGA for that!?!??!!?!?!?

lutris/gui/lutriswindow.py
```diff
@@ -688,7 +692,14 @@ def on_add_game_button_clicked(self, *args):
             callback=lambda: self.add_game_to_view(dialog.game.id)
         )
         return True
-
+    
+    @GtkTemplate.Callback
```
strycore: what the fuck.
why tis there an action for that

strycore:
I think this needs to be redone from scratch, this PR angers me. There is no hope getting it merged in such a state.

lutris/gui/lutriswindow.py
sigmaSd: oups, tbh was just copying add_game and i didn't notice, I'll just remove it

lutris/pga.py
sigmaSd: This and the one after are just guessing, for the pga I thought when i add the play_script to it, when someone makes an installer he can create a script and point  play_script to it,

 for example a battle net installer, with play_script that deletes Application Data/battle.net folder which is a frequent workaround, the installer might say for example if u have problem starting battle net click play script than try again

lutris/sysoptions.py
sigmaSd: this is a tick box to use with pre_launch_script basically when this is checked the game will only start after the pre_launch_script finishes its work, this works by using the script thread join() fn,
as opposed to the default behavior which is the game can be started even if the pre_launch_script is still running

lutris/sysoptions.py
sigmaSd: I taught it would make sense if u saw lutris xD
script
Obviously the wording is bad though, i'll try to figure out something else

lutris/util/script_thread.py
sigmaSd: ScriptThread, executes a given script in another thread `start()` or on the caller thread with `join()` with a wrapper fn over `subprocess.call` to handle the exceptions in the new thread.
If u're talking about lutris Thread class, i didnt need any of its fns so thats why i opted for a new one if u're talking about python one sure that make sense i guess i'll make it a subclass
2 - i'll add the doc according to ur explenation

sigmaSd:
Just to be clear I made a lot of guessing about the code, I can redo it I just have a couple of questions above

lutris/pga.py
sigmaSd: and for things like #681 basically the installer would say if u want to read the manual hit play_script, the script would have for example (xdg-open manual.pdf)

lutris/util/script_thread.py
sigmaSd: Thread sub classed + doc

lutris/gui/lutriswindow.py
sigmaSd: removed

lutris/util/script_thread.py
sigmaSd: I think its resolved ( I googled this and still didn't find a good explanation)

sigmaSd:
Huh, now that i read interpeter.py turns out that I don't need to add anything, the installer script just needs to look like this
 ```yaml
system:
  play_script: $GAMEDIR/script
```
for example

sigmaSd:
I think I addressed most issues this pr is ready for second round of review

strycore:
I still don't understand what a "play script" is and no of that makes sense to me, please clarify the whole thing.

lutris/sysoptions.py
strycore: I'm not adding that button.

strycore:
Closing this. We should have agreed on a design before implementation. I don't want to merge any of the code in that PR.

sigmaSd:
Understandable, at least on the bright side i learned a lot about the internals of lutris doing this ^^

Issue #938, opened 2018-06-21
Disable continue button when asking for setup file


AccountOneOff:
When installers ask for a setup file the option to continue should be greyed out until the user has selected said file. Maybe also add a popup when hovering over telling the user to select the setup file to continue?

duhow:
Testing with [amnesia-gogcom](https://lutris.net/games/install/2696/view), the installer will select a default path, it should be set to **None**, so the user is _forced_ to select a file/folder.

![image](https://user-images.githubusercontent.com/1145001/44371057-4d83a180-a4dd-11e8-90e8-a78adbfc0bf7.png)

The code changes should be close to here, i'll try to do them...
https://github.com/lutris/lutris/blob/d02e42b35413e7bbcce463a749906f80cbedf93c/lutris/gui/installerwindow.py#L315

PR #1244, opened 2018-10-24
Disable continue button when asking for setup file

sigmaSd:
fix https://github.com/lutris/lutris/issues/938


sigmaSd:
I'm not entirely sure how set_path_chooser is used so I may need to add a check before https://github.com/lutris/lutris/blob/1cb7bdae3408d904aa066c24fedf4f4a21494ab5/lutris/gui/installerwindow.py#L355 aka if a game installer comes with  the correct path

sigmaSd:
After more reflection check is not needed xD

PR #1248, opened 2018-10-25
Make file limit and process monitor warnings dismissible

Medath:
Someone on irc complained that Lutris doesn't allow him to launch an esync game because he doesn't want to increase the file descriptor limit. Turns out that the file limit warning was indeed not dismissible so I changed it.
The process monitor warning is dismissible per game. So if someone ticks the box for one game, the warning will still appear for other games. But it's saved in lutris.conf. It would probably be better to make that a game config option, but I'm keeping it that way because it's simpler.

I also removed a double import in dialogs.py

### DIFF
--- a/lutris/util/wine.py
+++ b/lutris/util/wine.py
@@ -19,6 +19,9 @@ WINE_PATHS = {
     'wine-development': '/usr/lib/wine-development/wine',
     'system': 'wine',
 }
+
+ESYNC_LIMIT_CHECK=os.environ.get('ESYNC_LIMIT_CHECK', '').lower()
+
 def get_proton():
     """Get the Folder that contains all the Proton versions. Can probably be improved"""
     for path in [os.path.join(p,'common') for p in steam().get_steamapps_dirs() ]:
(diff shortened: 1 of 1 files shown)


tannisroot:
The limits warning was made dismissable on purpose. He can’t use esync without increasing those file limits. I am against this change.

Medath:
After some discussion on the discord I decided to change this a bit. To launch an esync game with too low fd limits the user now has to launch lutris with the environment variable `ESYNC_LIMIT_CHECK` to 0. That var would need to be documented somewhere in the wiki.
I think bypassing the fd limit check that way is a good challenge for the user to prove that he knows enough to understand the issues with using esync with too low fd limits.

tannisroot:
Nice 👍

Medath:
Changed it to resolve conflicts with master

strycore:
Please remove the dismissible option for the process monitor

lutris/game.py
```diff
@@ -362,11 +362,13 @@ def do_play(self, prelaunched, error=None):
 
         monitoring_disabled = system_config.get('disable_monitoring')
         if monitoring_disabled:
-            dialogs.ErrorDialog("<b>The process monitor is disabled, Lutris "
-                                "won't be able to keep track of the game status. "
-                                "If this game requires the process monitor to be "
-                                "disabled in order to run, please submit an "
-                                "issue.</b>")
+            setting="hide-" + self.slug + "-proc-mon-warning"
```
strycore: This shouldn't be dismissible, not merging that.

PR #1255, opened 2018-10-27
MelonDS Runner

telanus:
melonDS basic runner. Bios files need to reside in ~/.config/melonds

via https://github.com/StapleButter/melonDS/issues/72


strycore:
Don't we have a Nintendo DS runner already?

tannisroot:
@strycore yes but that project seems to be abandoned.

strycore:
So MelonDS would be the better emulator at the moment? Will have to give it a try!

telanus:
@RobLoach the libretro port is currently lagging behind, and doesn't have all the features of the stand-alone

@strycore With the latest release of melonDS, it has overtaken desmume as the better ds emulator. On both my and my wife's pc's it runs a lot faster than the other emu

strycore:
Good to know! We try to ship the best emulators without having too much overlap between them. I've never had a great experience with Desmume so  that can maybe be deprecated.

PR #1259, opened 2018-10-28
Add proton to wine versions

sigmaSd:


tannisroot:
What does this add exactly?

sigmaSd:
It searches for proton version under steamapps/common and add their wine build to the runner tab

tannisroot:
But Proton is highly tied to Steam runtime and usually won’t work without it at all?

sigmaSd:
No it's not it's just a patched wine, and it doesn't depend on steam at all,( apart from its installed from it) this PR is more a future proof thing, what I mean maybe in the future proton and wine will diverge a lot so this just makes it easy to use it

sigmaSd:
Could be wrong though but that's what I know so far

sigmaSd:
Plus alot of ppl already using it with the custom option this just makes life easier for them

tannisroot:
This PR broke lutris, it won’t launch if `~/.local/share/Steam/steamapps/common` doesn't exist.

sigmaSd:
oups, thx for the fix @Medath , just to make sure @tannisroot you don't have steam installed or steam have a diffrent path?

sigmaSd:
https://github.com/lutris/lutris/pull/1262 We decided to use what lutris uses already to find steam path, should be good now to correctly detect the correct path, plus for custom location there is already the custom option ;)

PR #1262, opened 2018-10-29
Don't crash if PROTON_PATH doesn't exist and find Proton in custom locations

Medath:
Fix possible crash introduced by #1259

### DIFF
--- a/lutris/util/wine.py
+++ b/lutris/util/wine.py
@@ -115,12 +115,13 @@ def get_wine_versions():
             if is_version_installed(dirname):
                 versions.append(dirname)
 
-    proton_versions = [p for p in os.listdir(PROTON_PATH) if "Proton" in p]
-
-    for version in proton_versions:
-        proton_path = os.path.join(PROTON_PATH, version, 'dist/bin/wine')
-        if os.path.isfile(proton_path):
-            versions.append(version)
+    if os.path.isdir(PROTON_PATH):
+        proton_versions = [p for p in os.listdir(PROTON_PATH) if "Proton" in p]
+    
+        for version in proton_versions:
+            proton_path = os.path.join(PROTON_PATH, version, 'dist/bin/wine')
+            if os.path.isfile(proton_path):
+                versions.append(version)
     return versions
 
 
--- a/lutris/util/wine.py
+++ b/lutris/util/wine.py
@@ -9,6 +9,7 @@ from lutris.util import system
 from lutris.util.log import logger
 from lutris.util.strings import version_sort
 from lutris.util.vulkan import vulkan_available
+from lutris.runners.steam import steam
 
 MIN_NUMBER_FILES_OPEN = 1048576
 WINE_DIR = os.path.join(settings.RUNNER_DIR, "wine")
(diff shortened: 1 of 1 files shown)


sigmaSd:
What about using `steam_data_dir` function from steam.py for more feature proof

Medath:
I think `get_steamapps_dirs` or `get_game_path_from_appid` would be better. That way it can also find Proton if the default download folder in Steam is custom.

sigmaSd:
Make sense just 1- wasn't sure that proton would also get installed in custom folder and 2 steam_data_dir is the easiest fn to extract out of the class

Medath:
That's how I noticed the crash. My default Steam download directory is a custom one and therefore `common` in `~/.local/share/Steam/steamapps/` didn't exist. Steam downloads the Proton versions to the folder that is set to be the default.

sigmaSd:
Nice find! , so  steam_data_dir is out, this `get_game_path_from_appid` is problematic how would we get the id steamdb, hardcode?  this `get_steamapps_dirs` looks like the best option but it will probably mean we will need to instantiate steam class in wine.py

sigmaSd:
Guess I vote for get_steamapps_dirs, looks the easiest and best solution

sigmaSd:
Here's my suggestion :) 

```python

PROTON_PATH = steam().get_steamapps_dirs()
PROTON_PATH = [p for p in os.join(PROTON_PATH,'common')]
```

```python
    for path in PROTON_PATH:
        proton_versions = [p for p in os.listdir(PROTON_PATH) if "Proton" in p]
        for version in proton_versions:
            proton_path = os.path.join(PROTON_PATH, version, 'dist/bin/wine')
            if os.path.isfile(proton_path):
                versions.append(version)
```

sigmaSd:
wine under util also uses proton_path so my suggestion will result in duplication

sigmaSd:
here is my new suggestion I think this one should do it
```python
def get_proton():
    protons = {}
    steam_dir = [os.path.join(p,'common') for p in steam().get_steamapps_dirs() ]
    for path in steam_dir:
        proton_versions = [p for p in os.listdir(path) if "Proton" in p]
        for version in proton_versions:
            proton_path = os.path.join(path, version, 'dist/bin/wine')
            protons[version] = proton_path
    return protons

PROTON_PATH = get_proton()
```

sigmaSd:
Just a suggestion I think using a dict for proton path is better , u can avoid redundancy at the second call u will just iterate over the keys , and for utilities/wine.py it will be simple as

 if proton in version return proton path[version]

Medath:
Ok I implemented basic functionality to find Proton versions in custom download locations. However, my code assumes that all Proton versions are located in the same directory. I think this will work good enough for most people but could possibly be improved.

sigmaSd:
And with dict u can remove that assumption ^^

sigmaSd:
My bad I meant runner wine u can do this ```if 'Proton' in version:
            return PROTON_PATH[version] ``` can u please try again with dict and post ur problem I don't see the issue,

sigmaSd:
just to be clear( aka proof I'm not just annoying u :p) this is why i want dict 

1- it looks better no no redundancy 

2 - it doesn't have the aforementioned assumption

Medath:
Ooop it got merged already. Whatever, I have tried it with a dict first but I'm kind of a noob so I couldn't really figure out how to do it. And I don't really see where the redundancy is. My only problem is the assumption that all Proton versions are in the same download directory, but I think this is almost always the case.
I guess you can improve it in a new PR now...

sigmaSd:
if strycore like its cool ^^ as long ur assumption its true i guess it doesn't really matter^^

sigmaSd:
thanks for ur work btw !

PR #1267, opened 2018-10-30
fix tray icon

### LINKED ISSUE #1266: Lutris doesn't preserve the setting for tray icon on relaunching
While changing it when Lutris is running has an effect, it doesn't last after a restart of the app:
https://youtu.be/W7J7ZkxN_XM

sigmaSd:


sigmaSd:
fix https://github.com/lutris/lutris/issues/1266

Issue #1014, opened 2018-08-11
make downloading and using lutris provided wine optional


soredake:
![11-08-2018_08 37 55](https://user-images.githubusercontent.com/5204968/43988605-f9f3a508-9d41-11e8-8ce7-4ede36f5b1fa.png)
Downloading is pretty slow sometimes, and i'm using system version anyway.

duhow:
When you install a game using a Lutris install script, it will download its wine package required, although you can change your Wine settings after completing the install.

You can also download the script and change it locally.
(btw, download is not that slow, and it's only done once)

strycore:
The wine builds are moving to Amazon S3 eventually so that we solve the download speed issue. 

Also, what version do you currently have installed on your system?

soredake:
git version, as always
I just want to have an option that will allow me not to download versions from Lutris and not use them, that is, I always use the system version, every time I change the version to the system one in the settings and update the prefix after the change to the system one, rather annoying.

duhow:
You can set the default Wine version in Runner settings, so all the games
that you add manually will have that version. But as I said, the installer
will download and use the Wine version set, as the game might require that
version, or it may have better performance (for old games?)

soredake:
>You can set the default Wine version in Runner settings, so all the games

I know how to use lutris, thanks.

>as the game might require that version, or it may have better performance (for old games?)

I can have build with all fixes i need, so there's no need for lutris-provided versions.

duhow:
No problem. So... Requesting:

`Advanced options for install > Choose Wine version > System Wine`

strycore:
> I can have build with all fixes i need, so there's no need for lutris-provided versions.

Yes, but that's only you. I agree that we should not try downloading wine if no explicit version is set but if there is one, we're not going to make it optional. That just plain wouldn't work and it's too risky.

strycore:
Wine builds are now hosted on DigitalOcean and are downloading fast, marking this as closed

tannisroot:
They are not fast, at least for me, specifically the versions that are over 100 megs. The downloading is actually super slow.
And even then, this is still an issue for people with slow or limited connections. Please re-open and maybe add as a goal for 0.5 instead.

strycore:
This needs to be further designed to provide a proper solution.
I'm mark that for the 0.5 cycle since this is where we'll start letting the user customize the install process.

PR #1277, opened 2018-11-01
Add wine version choice to the installer

sigmaSd:
screenshot from 2018-11-01 13-23-01

### DIFF
--- a/lutris/gui/installerwindow.py
+++ b/lutris/gui/installerwindow.py
@@ -20,7 +20,7 @@ from lutris.util import jobs
 from lutris.util import system
 from lutris.util.log import logger
 from lutris.util.strings import add_url_tags
-
+from lutris.util.wine import get_system_wine_version, get_default_version
 
 class InstallerWindow(Gtk.ApplicationWindow):
     """GUI for the install process."""
@@ -91,10 +91,11 @@ class InstallerWindow(Gtk.ApplicationWindow):
         self.install_button = self.add_button("_Install", self.on_install_clicked)
         self.continue_button = self.add_button("_Continue")
         self.play_button = self.add_button("_Launch game", self.launch_game)
-        self.close_button = self.add_button("_Close", self.close)
+        self.close_button = self.add_button("_Close", self.close)        
+        self.wine_choices = None
 
         self.continue_handler = None
-
+        
         self.get_scripts()
 
         # l33t haxx to make Window.present() actually work.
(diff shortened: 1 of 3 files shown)


sigmaSd:
fix https://github.com/lutris/lutris/issues/1014

tannisroot:
I think “installer-specified wine” is more obvious and sounds better. Should probably have something like “Recommend” near it too.
Also, “Use” before each option doesn’t look very good, users can get confused.
Can it be like this?
Use:
- installer-specified wine
- default wine
- system wine

Also, would it be possible to show what wine version is behind each option? 
Would look kind of like this:
- installer-specified wine (tkg-3.19-86_64) - Recommenced (in bold)

sigmaSd:
1 - I changed the names to what you suggested.
2 - who gets to decide whats recommended , even if the install script specifies it it will just get outdated with time
3 - version added
screenshot from 2018-11-01 14-03-31

PythonicChemist:
Maybe write something like "Select Wine version:" before listing the options, then call the options simply recommended, default, system? Also It would be great to standardize the wine version format. It does look kind of messy right now.

sigmaSd:
making the installer provided one recommended feels weird cause scripts gets outdated quickly,
I prefer not to add "select wine" cause i'll have to account for the case where only one version exist and remove that label, 
I agree it looks messy what about adding version to tooltip instead?

PythonicChemist:
If only one version exists, this feature should be invisible to the user anyways imo.

sigmaSd:
I believe the case where only one version exists is rare plus even with one choice it doesn't look bad visually

sigmaSd:
screenshot from 2018-11-01 15-14-47

Cool info to be able to screenshot the tooltip you can't use the screenshot window option you have to screenshot all the desktop

sigmaSd:
@PythonicChemist added ur ideas :) this feature will be invisible if there is only one version
screenshot from 2018-11-01 15-33-49

sigmaSd:
winesteam
screenshot from 2018-11-01 15-41-40

tannisroot:
Why is Winesteam one different? You can use custom wine versions with those and custom prefixes too.

tannisroot:
I would prefer if it had some sort of indication that installer-specific wine is recommended as usually that’s the version the game was tested with and confirmed working. It’s rare for a game to just stop working with the same wine version after a while, the installers are usually made broken or vandalized.

sigmaSd:
winesteam has the same logic, its just that the install script for that game "advance quest" is that simple
make sense, so would installer-specified (Recommended) be enough?

sigmaSd:
screenshot from 2018-11-01 16-49-18
I like it, it works well with the tooltip

strycore:
This actually offering users to break their games by letting them use another wine version that what is used in the installer. This is a feature that can only be used for doing bad and ugly things, I don't see how can this possibly make the lifes of the users better when we give such ways to harm them.

sigmaSd:
I see your point, will close I guess, I'll reopen if someone have a good argument for it.

tannisroot:
I have 3 arguments for this option:
1) I have no idea why, bu the servers Lutris hosts wine are complete crap and have abysmally slow speed, at least for people from Europe. So I would rather use whatever wine version I have that i know will work just fine with the game I am downloading than to wait half an hour for Lutris to finish downloading Wine and then doing that again 3 more times because it broze after it has downloaded it.
2) Save time/money for people with bad or limited internet connection
3) Avoid downloading ancient version of wine because the installer wasn’t updated in ages. I will switch to something newer after the installation anyway, so why waste time?
If users breaking stuff is such a big concern, the option can hidden under a spoiler of some sort.
So yeah, please reopen, this is a good PR.

strycore:
This is not a proper solution to the problem you are exposing, it cannot get solved by butchering the UI and entrapping users.

tannisroot:
At the end of the day, what should we care about more - the UI or usability of the scripts?

sigmaSd:
I definitely sympathize with tannisroot points , those are actually also my personal reasons for doing this pr, but also I agree with strycore this pr doesn't fit with the lutris philosophy of making installing games on linux as easy as hitting the install button without much thinking.
That being said I think pr likes this will still pop in the future where something must be sacrificed usability or friendliness, so what I suggest is hiding this pr with `LUTRIS_ADVANCED` env and future pr like this

AlexanderRavenheart:
Hiding features behind environment variables is basically invalidating the feature for all but the most technical and curious of users. Environment variables should not be used for user facing options, but for changing features in a development environment.

tannisroot:
So why add this feature to the milestone then? This all just a waste of code and time now.

Issue #1286, opened 2018-11-03
[Regression] Warning about existing prefix doesn't dynamically disappear when the prefix name is changed


tannisroot:
Behavior in 0.4.21.1:
![peek 2018-11-03 10-52](https://user-images.githubusercontent.com/10602045/47949561-a7c9c580-df56-11e8-8117-30e9cc654615.gif)

Behavior in master:
![peek 2018-11-03 10-55](https://user-images.githubusercontent.com/10602045/47949595-faa37d00-df56-11e8-8aaa-8ff5863458a0.gif)

tannisroot:
Actually, this is not just the warning - Lutris just ignores custom paths and uses the default path anyway.

PR #1288, opened 2018-11-03
fix regression

sigmaSd:
fix https://github.com/lutris/lutris/issues/1286, oups bug introduced with https://github.com/lutris/lutris/pull/1244/files


sigmaSd:
My bad xD

PR #1289, opened 2018-11-03
replace os.path.exists with lutris/util/system/path_exists

PythonicChemist:


PythonicChemist:
I will look into the cyclic import issues.

Issue #1287, opened 2018-11-03
Update list of DXVK versions when config menu is opened


tannisroot:
Currently, the user has to wait 24 hours to get it updated or manually remove dxvk_versions.json file to get an up to date list. This is not very user friendly and should be improved by making it update the list every time the configuration menu is opened instead, for example.
Opened to replace https://github.com/lutris/lutris/issues/1189

strycore:
I agree, it could be reloaded once per lutris instance, when the config dialog is first accessed

PR #1291, opened 2018-11-03
update dxvk_versions at the first acess of config dialog once per lutris instance

sigmaSd:
fix https://github.com/lutris/lutris/issues/1287

### DIFF
--- a/lutris/gui/application.py
+++ b/lutris/gui/application.py
@@ -37,7 +37,6 @@ from lutris.services.steam import AppManifest, get_appmanifests, get_steamapps_p
 from lutris.settings import read_setting, VERSION
 from lutris.thread import exec_in_thread
 from lutris.util import datapath
-from lutris.util.dxvk import init_dxvk_versions
 from lutris.util.log import logger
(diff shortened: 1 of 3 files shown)


tannisroot:
What does “once” here mean?

sigmaSd:
was lazy to write :p i meant what strycore said

> I agree, it could be reloaded once per lutris instance, when the config dialog is first accessed

sigmaSd:
I would've preferred it just to update each time lutris open ( means less code) but I thought maybe strycore has reasons for wanting it to update at config dialog access

sigmaSd:
as @PythonicChemist suggested here are  two alternative solutions 

https://pastebin.com/jRBayyzK 
or
```python 
@functools.lru_cache(maxsize=None, typed=False) 
def init_dxvk_versions():
```
so wanted to know what do u prefer @strycore ?

strycore:
The approach can be changed after the release

strycore:
Also, I like the decorator approach more, with proper naming it's easier to see what the purpose is

PR #1294, opened 2018-11-03
Fix appdata file

AsciiWolf:
Fix wrong url type causing issues with appstream generators/parsers.


RobLoach:
Confirmed: https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-url

PR #1296, opened 2018-11-04
Add ondemand script

sigmaSd:
add ondemand script execution fn  to game's contextual menu, obviously I'm not sure about the wording If someone has good suggestion I'll be glad to use it


tannisroot:
I think “Execute script” fits here more.

PythonicChemist:
The LutrisWindow class probably shouldn't have ownership of the logic for ondemand scripts.

strycore:
The whole thing is going to be reworked for 0.5 anyway, this is temporary.

Issue #1304, opened 2018-11-06
Incorrect architecture for GameMode causes warnings in the logs


tannisroot:
Whenever I start any game with Feral Gamemode enabled (which is enabled for all games by default), I get this message spamming the terminal output (when Lutris is launched with `lutris -d`):
```
ERROR: ld.so: object '/usr/lib/libgamemodeauto.so' from LD_PRELOAD cannot be preloaded (wrong ELF class: ELFCLASS64): ignored.
```
This is quite annoying, and may cause some slow downs due to the spam. But disabling Gamemode gets rid of this.

strycore:
I'm not sure this is fixable, the method documented on Feral's repo just doesn't work.

Zeioth:
I was about to open pretty much the same issue:

I installed gamemode from the arch package from AUR.  I can run Lutris games with the checkbox "Enable Feral gamemode" checked.

But when I run any game, it doesn't seem to load it correctly (Lutris -d)
![gamemode](https://user-images.githubusercontent.com/3357792/48113600-ecf23e00-e25b-11e8-81c9-1b55da910dd0.png)

As you can see, I have /usr/lib/libgamemodeauto.so in the right location
![libs](https://user-images.githubusercontent.com/3357792/48113663-2e82e900-e25c-11e8-9e05-ada313689ad2.png)

sigmaSd:
@Zeioth can you test with my PR https://github.com/lutris/lutris/pull/1305 I can't reproduce strycore issue

sigmaSd:
I have another method to fix this but I really want to make sure that simple solution is not working

strycore:
I still see the warnings on the current codebase

tannisroot:
@strycore it's because you reversed that PR.

strycore:
@tannisroot uh, nope, the current code has the `/usr/$LIB/libgamemodeauto.so.0` part, I have both 64 and 32bit builds of gamemode. I still have warnings on wine games.

strycore:
Please send a PR if this still needs fixing.

strycore:
Should be fixed by #2319 , please reopen if not the case.

PR #1305, opened 2018-11-06
fix game_mode

sigmaSd:
preloading https://github.com/lutris/lutris/blob/master/lutris/util/system.py#L50 is based on luck, to get the correct library for the game we can use $LIB see this insightful comment https://github.com/FeralInteractive/gamemode/issues/64#issuecomment-404779034,

### DIFF
--- a/lutris/game.py
+++ b/lutris/game.py
@@ -341,10 +341,9 @@ class Game:
         if gamemode:
             env['LD_PRELOAD'] = ':'.join([
                 path for path in
-                [env.get('LD_PRELOAD'), system.GAMEMODE_PATH]
+                [env.get('LD_PRELOAD'), "/usr/$LIB/libgamemodeauto.so.0"]
                 if path
             ])
-
         # LD_LIBRARY_PATH
         game_ld_libary_path = gameplay_info.get('ld_library_path')
         if game_ld_libary_path:


sigmaSd:
fix https://github.com/lutris/lutris/issues/1304

sigmaSd:
I kept GAMEMODE_PATH to use it for checking if gamemode is installed

strycore:
I'll have to test this. I remember trying to use $LIB (it's what's documented on Feral's repo) and it didn't work for some reason.

strycore:
As expected, this doesn't work and just breaks the gamemode feature, reverted.

sigmaSd:
Weird It works on my end

Zeioth:
I've tried 

    cd /usr/$LIB/"

and it goes to the "/usr/" directory. Probably that alias is exclusive of your distro.

sigmaSd:
$LIB doesn't work like that it only works with ld preload

sigmaSd:
No problem thx for testing

sigmaSd:
tested it on arch solus ubuntu 18.04 and this pr works

Zeioth:
I've reinstalled Lutris from this commit and it still have problems to find gamemode:
notworking

I've cheked and yes, the patch is applied correctly.
checking

sigmaSd:
Allright now I'm convinced that there is an issue for some ppl ^^
what distro are you running?

sigmaSd:
Either way for now till this clears up we should do  it steam way preload both libs that wont get rid of the spam but at least it will fix gamemode

strycore:
I don't think this is an issue with the distribution. I'm on Ubuntu 18.04 and I get the same thing as @Zeioth

sigmaSd:
@Zeioth , @strycore  did more digging, do you have the 32 bit lib of gamemode installed, I  think whats happening is in case of 32bit exe the $LIB is expanding to lib32 which maybe you don't have installed

sigmaSd:
I say this because feral script defaults to compiling only 64bit lib

Zeioth:
* In my last test I installed **gamemode**
* In this test I've installed **gamemode** and **lib32-gamemode** (you need to have multilib enabled in your distro to download it).

Having this two libraries, it seems gamemode is detected correctly (only, after applying this patch).
gamemode

PR #1307, opened 2018-11-07
Make setup.py builds less sensitive to imports

PythonicChemist:
Now setup.py doesn't need to import any business logic at all and have the chance at importing a dependency unintentionally. Additionally this is a very common way to do versioning in python.


strycore:

PR #1317, opened 2018-11-07
forbid ntfs usage

sigmaSd:


lutris/util/system.py
```diff
@@ -451,3 +451,21 @@ def is_running(process):
...
+    # import here to avoid cyclic import with lutris.gui.dialog
+    from lutris.gui.dialogs import ErrorDialog
+
+    def display_ntfs_warning():
+        ErrorDialog(
+            "<b>Using NTFS partition causes major issues!</b>\n")
+    out = None
+
+    if path:
+        out = subprocess.check_output(["df", "-Th", path]).decode("UTF-8")
```
PythonicChemist: Would this work instead: https://docs.python.org/3.7/library/os.html#os.fstatvfs ?

sigmaSd: I'm not sure how to use it the output is cryptic here is an example `os.statvfs_result(f_bsize=4096, f_frsize=4096, f_blocks=0, f_bfree=0, f_bavail=0, f_files=0, f_ffree=0, f_favail=0, f_flag=4106, f_namemax=255)
`

sigmaSd: weird it gives a lot of info about the file system but not what it is or I overlooked something

PythonicChemist: According to os.statvfs the support for filesystem ID field f_fsid was only added in Python 3.7.

sigmaSd: I compared os.statvfs out in ntfs and ext4 unfortunately nothing strikes out

sigmaSd: weird I have python 3.7.1 and I still don't see f_fsid in the output

sigmaSd: either way I guess that's too recent for what lutris targets

sigmaSd:
btw @PythonicChemist I want to ask you this
```python
 if 'main_file' in self.game.config.game_config.keys():
```
is annoying me,I use it to avoid key not found error, would it be acceptable to wrap the next lines in try expect so I can remove this overhead

sigmaSd:
@PythonicChemist  thanks a lot that's what I was looking for!
for the gui code, so you want to move ntfs check to gui dir? I just didn't find a suitable file for it there

PythonicChemist:
No ntfs check is fine where it is. Just don't use gui code for it.

sigmaSd:
Ah I get it I'll fix that^^

strycore:

Issue #1302, opened 2018-11-06
Integrate Libstrangle into Lutris


Zeioth:
There are some wine games that run too fast, or too slow, because the pace is bound to the game's FPS. An option to limit the FPS of a game from the installer, would help to prevent this.

**All this games are fixed by using strangle:**

* [Deltarune](https://lutris.net/games/deltarune/) (The pace is bound to FPS)
* [The Sims 3](https://lutris.net/games/the-sims-3/) (Very bad performance due to unlimited FPS)
* [Crash Bandicoot N Sane Trilogy](https://lutris.net/games/crash-bandicoot-n-sane-trilogy/) (Severe stuttering due to poor ingame FPS limiter implementation)
* [Doom 2016](https://lutris.net/games/doom/) (Severe stuttering due to poor ingame VSync implementation)

**URL of the library:** https://gitlab.com/torkel104/libstrangle

Zeioth:
This issue only affect a few games, so it's probably low priority.

strycore:
That is very cool! I was looking for something like that to fix No One Lives Forever and couldn't find one. (Game runs fine now, so I won't need it for that but still cool)

Zeioth:
Crash Bandicoot running without "strangle 60"
![nostrangle](https://user-images.githubusercontent.com/3357792/48173538-765d4b00-e304-11e8-8ac5-838ed74a5283.png)

Crash Bandicoot running with "strangle 60"
![withstrangle](https://user-images.githubusercontent.com/3357792/48173539-765d4b00-e304-11e8-83ac-97d3305c38d8.png)

Fixes stuttering completely.

Zeioth:
Support for FPS limit implemented

Medath:
As @sigmaSd already said, you'll be able to use
```
system:
  fps_limit: "60"
```
to set the fps in an installer

PR #1319, opened 2018-11-08
add fps limit support

sigmaSd:


sigmaSd:
fix https://github.com/lutris/lutris/issues/1302

sigmaSd:
needs testing

Zeioth:
Tested. It works fine. The FPS limit is applied correctly inside of the game. Only if libstrangle is installed in your system.
oki

Still need two things:

* Distribute libstrangle with Lutris, so it's avaliable for everyone.
* An option to set the number of FPS from a game installer.

And optionally:

* A way to enable the Vsync features of libstrangle

Zeioth:
In the future we could consider to add a way to enable the Vsync features of libstrangle. This would allow the games without VSync, to use it (but most games already have Vsync).

sigmaSd:
for the installer its allready covered you can do
```yaml
system:
  fps_limit: "60"
```
distributing I guess is a good idea, I'll check out the vsync opt

strycore:

sigmaSd:
rebased

PR #1320, opened 2018-11-08
grey out execute script if no script is set and change ondemand to manual

sigmaSd:
the execute script on the contextual menu is not intuitive if one doesn't know about the feature, this will grey it out till a script is set,  this is more like a bug fix till 0.5 revisit the gui stuff


tannisroot:
Can it be greyed-out instead?
Also, can the name be changed to a more uniform “Manual command”?
So in the context menu it would be “Execute command” and in the config menu it would be called “Manual command”.

sigmaSd:
done^^

strycore:

sigmaSd:
rebased

PR #1328, opened 2018-11-10
[WIP] improve playtime and gerneralize timer class

sigmaSd:
1- this make it way less ugly xD 
2- <del>the old implementation actually didn't update the row when the game was stopped with the stop button</del> <del>both implementation have a weird bug: the first time when you stop a game the row doesn't get updated</del>figured it out check last comment


sigmaSd:
3 - fix sorting with playtime, change empty ('') default to '0.0 hrs' default to fix sorting

sigmaSd:
4 -
- changed https://github.com/lutris/lutris/blob/244dee5099b7b24cf6caeff685d891d31fa01427/lutris/gui/gameviews.py#L285 to make life easier
- timer.end() is now called when stop icon is clicked to fix issue in first comment
- Added a cache function to Timer class to avoid calling self.increment twice for the same instance to the game
- refactored stuff

PythonicChemist:
Look into properties and dunder methods. It could make the class a lot better! :) Also I am not a fan of the cache. It makes the increment method useless after one use and thus the timer class useless for other possible usecases.

P.S.: double underscores in front of a method/variable causes name mangling which I don't see how it would be useful here. If you want to mark something private, use a single underscore.

sigmaSd:
@PythonicChemist Thanks a lot for the review and valuable info :)
I think what I'm missing now is how to use property in a way that simplify the class, I'll look into it more later, and I also would appreciate any advice or pointer you have ^^

sigmaSd:
5 - generalize Timer class:
Now its a "timer" class with start stop and duration methods, and the + operator is overloaded depending on what the timer is used for

lutris/util/timer.py
```diff
@@ -2,20 +2,30 @@
 
 
 class Timer:
+    """Simple Timer class to time code.
+    You can overload the + operator according to your need by using a new timer_type.
+    """
+
+    def __init__(self, timer_type=None):
+        self.timer_type = timer_type
```
PythonicChemist: This is what subtyping or composition is used for.

lutris/util/timer.py
```diff
@@ -2,20 +2,30 @@
...
+        self.finsihed = False
 
     def start_t(self):
-        self.start = datetime.datetime.now()
+        self._start = datetime.datetime.now()
```
PythonicChemist: _start and _end may be referenced before assignment. does start_t expand to start_timer? so Timer.start_timer?

lutris/util/timer.py
```diff
@@ -2,20 +2,30 @@
 
 
 class Timer:
+    """Simple Timer class to time code.
+    You can overload the + operator according to your need by using a new timer_type.
+    """
+
+    def __init__(self, timer_type=None):
+        self.timer_type = timer_type
+        self.finsihed = False
```
PythonicChemist: spelling

lutris/util/timer.py
```diff
@@ -2,20 +2,30 @@
...
 
     def start_t(self):
-        self.start = datetime.datetime.now()
+        self._start = datetime.datetime.now()
 
     def end_t(self):
-        self.end = datetime.datetime.now()
+        self._end = datetime.datetime.now()
+        self.finsihed = True
```
PythonicChemist: since you are not setting self.finished in start_t, timing with an instance would only work once.

lutris/util/timer.py
```diff
@@ -2,20 +2,30 @@
...
 
     def duration(self):
```
PythonicChemist: with property a user could access the duration as timer.duration and the class would be doing the work behind the scenes to convert it to seconds/hours/minutes/days or whatever is specified in self.format. Additionally when self._end hasn't been set, return current duration.

lutris/util/timer.py
```diff
@@ -2,20 +2,30 @@
...
-        return (self.end - self.start).seconds / 3600
+        return (self._end - self._start).seconds / 3600
 
-    def increment(self, saved_dur):
-        if saved_dur == '':
-            return "%.1f hrs" % self.duration()
+    def __add__(self, saved_dur):
```
PythonicChemist: \+ is used for addition or concatenation in Python. This method does neither really. While it does addition, it also converts the number to a string and does formatting. Also \_\_add\_\_ shouldn't return None rather raise an exception.

\_\_add\_\_ should probably accept either a number type or timer instance.

Having said that when I proposed the usage of dunder methods I had \_\_str\_\_ in mind for formatting purposes.

PythonicChemist:
Overall I think a major inconvenience is that the playtime is stored as a string. It would be more convenient to have it stored as something useful and then only format to a string for the view. I don't think we need increment or \+ overloading for the class at all.

Keep up the good work :+1:

sigmaSd:
Thanks a lot for taking the time to review it :) allright back to WIP ^^

strycore:

sigmaSd:
rebased

PR #1336, opened 2018-11-12
[WIP-Controversial] Ntfs check

### LINKED ISSUE #1313: Disallow to install/run games on NTFS partition
Since using Wine prefixed located on NTFS partitions very often causes problems, Lutris should check if either
- the game directory (when installing the game)
- the game’s executable (that can be chosen manually in the config menu) 

are located on an NTFS partition, and if they are, display an error popup that explains the issue and prompts user to select a different location.

sigmaSd:
fix #1313

### DIFF
--- a/lutris/gui/installerwindow.py
+++ b/lutris/gui/installerwindow.py
@@ -12,7 +12,7 @@ from lutris.installer import interpreter
 from lutris.installer.errors import ScriptingError
 from lutris.game import Game
 from lutris.gui.config_dialogs import AddGameDialog
-from lutris.gui.dialogs import NoInstallerDialog, DirectoryDialog, ErrorDialog
+from lutris.gui.dialogs import NoInstallerDialog, DirectoryDialog, InstallerSourceDialog, ErrorDialog
 from lutris.gui.widgets.download_progress import DownloadProgressBox
 from lutris.gui.widgets.common import FileChooserEntry
 from lutris.gui.logwindow import LogTextView


JanGalek:
Why dont install for ntfs ? It works for me, This will disable it :-1:

tannisroot:
Well I am glad it works for you, but for others it usually introduces a lot of problems.

JanGalek:
> Well I am glad it works for you, but for others it usually introduces a lot of problems.

yeah, so and some option to disable this check should be ok.

sigmaSd:
@JanGalek you make a valid point, Its a controversial subject I guess, the problem is making it dismissible will make it useless since most people will just dismiss warnings without thinking, I'll just let the maintainers decide what they want

JanGalek:
@sigmaSd Its only warning, I see you call ErrorDialog :), so if this will not disable install to ntfs, for me it will be ok ;)

sigmaSd:
It actually disable it with an early return ^^ But It doesn't matter its really simple to change,  probably this issue needs to be handled differently since right now its too restrictive,  I'll just keep the pr open to maybe get other opinions or ideas

lutris/util/system.py
```diff
@@ -455,3 +455,18 @@ def is_running(process):
...
+    """Returns True if the given path is located on a NTFS partition"""
+
+    out = None
+
+    df_cmd = find_executable('df')
+    if df_cmd and path:
+        out = subprocess.check_output(
+            [df_cmd, "-T", path]).decode("UTF-8")
+
+    if out and "fuseblk" in out:
```
RobertZenz: Correct me, but that just means that the path is mounted over the FUSE driver and does not tell you anything about the type of the filesystem.

sigmaSd: I did a bit of research on this and experimentation and found that df command reports ntfs as fuseblk, and that's why I use it, I'm not sure if it's 100% valid but I don't know if it's not either^^

RobertZenz: `fuseblk` is simply the indicator that it is mounted through the FUSE driver, which could be NTFS or some other filesystem. So it is not limited to NTFS. Something like [this might be more reliable](https://unix.stackexchange.com/a/402236/4106).

sigmaSd: that's what fuseblk stands for, fuseblk returned from df or stat command which both share the same code base http://git.savannah.gnu.org/gitweb/?p=coreutils.git;a=blob_plain;f=src/stat.c will report ntfs as fuseblk at least on recent kernels, after more digging it seems these gnu tools used to report it as ntfs, so I'll probably add or ntfs in the check

sigmaSd: even if stat.c has a ntfs check git.savannah.gnu.org/gitweb/?p=coreutils.git;a=blob;f=src/stat.c#l415, apparently on recent kernels this path is not hit, I got curious now wonder if someone can explain it ^^

sigmaSd: This `lsblk -no MOUNTPOINT,fstype` actually is the first tool I used that reports ntfs, I guess I'll just change to that, guess linux tools can do what gnu tools cant ^^

strycore:
Please make it show a warning message like the one displayed when installing over an existing path, do not show an error dialog and do not prevent the install from going on.

PR #1338, opened 2018-11-12
improve execute script fn

sigmaSd:


lutris/gui/gameviews.py
```diff
@@ -562,6 +562,12 @@ def popup(self, event, game_row=None, game=None):
             self.add_menuitems(runner_entries)
         self.show_all()
 
+        def manual_script_not_set():
+            game = Game(game_id)
+            if not game.runner.system_config.get("manual_command", None):
+                return True
+            return False
```
BO41: you could write
```python
return not game.runner.system_config.get("manual_command", None)
```

PR #1345, opened 2018-11-13
Improve playtime code and generalize Timer class

sigmaSd:


lutris/game.py
```diff
@@ -502,12 +503,16 @@ def stop(self):
...
+            self.playtime = self.timer.duration / 3600 + self.playtime
 
     def on_game_quit(self):
         """Restore some settings and cleanup after game quit."""
 
-        self.timer.end_t()
-        self.playtime = self.timer.increment(self.playtime)
+        if not self.timer.finished:
+            self.timer.end()
+            self.playtime = self.timer.duration / 3600 + self.playtime
```
PythonicChemist: ```self.playtime += self.timer.duration / 3600```

conversion should be done within the Timer class depending on what format has been set when instantiating.

lutris/util/timer.py
```diff
@@ -2,20 +2,29 @@
...
+    def __init__(self):
+        self._start = None
+        self._end = None
+        self.finished = False
 
-    def end_t(self):
-        self.end = datetime.datetime.now()
+    def start(self):
+        self._start = datetime.datetime.now()
+        self.finished = False
```
PythonicChemist: You have to reset self._end to None otherwise accessing duration after a second call to Timer.start() will behave unexpectedly.

lutris/game.py
sigmaSd: Ok, thought its not necessary since its in seconds but I guess you'll right I'll add that

lutris/util/timer.py
sigmaSd: nice catch ^^

lutris/game.py
sigmaSd: I used a simplistic conversation method, is that good enough?

lutris/util/timer.py
```diff
@@ -2,20 +2,41 @@
...
-        if saved_dur == '':
-            return "%.1f hrs" % self.duration()
+        else:
+            result = (self._end - self._start).seconds
 
-        saved_dur = float(saved_dur.split()[0])
-        new_dur = saved_dur + self.duration()
-        return "%.1f hrs" % new_dur
+        # Retrun time in the specified format
+        if self.format == 'seconds':
```
PythonicChemist: put the conversion in a private helper method. Also this should work as a single if-elif construct with a single return, if I am not mistaken.

lutris/util/timer.py
```diff
@@ -2,20 +2,41 @@
 
 
 class Timer:
+    """Simple Timer class to time code"""
 
-    def start_t(self):
-        self.start = datetime.datetime.now()
+    def __init__(self, format):
```
PythonicChemist: ```def __init__(self, format='hours'):```

Using a sane default is useful in this case, I think.

lutris/util/timer.py
sigmaSd: I can make it do with one return but i just like the expressiveness with current style, is it ok to keep it like this or this is just not pythonic ?

PR #1360, opened 2018-11-20
Small clean ups for util/display.py

PythonicChemist:
Just some small clean ups for some functions in display.py.


lutris/util/display.py
```diff
@@ -172,9 +164,9 @@ def change_resolution(resolution):
 def restore_gamma():
     """Restores gamma to a normal level."""
     xgamma_path = system.find_executable("xgamma")
-    if xgamma_path:
+    try:
         subprocess.Popen([xgamma_path, "-gamma", "1.0"])
-    else:
+    except:
```
strycore: don't add blank exception handlers, revert to the if / else, this was intended.

PR #1363, opened 2018-11-21
Include new subpackages of lutris.util in setup.py

openglfreak:


strycore:
Thanks! I always forget about that!

Issue #1370, opened 2018-11-25
DXVK versions not updating (git)


clapbr:
![screenshot_20181124_233328](https://user-images.githubusercontent.com/6379485/48974484-5b980f80-f042-11e8-8d26-963cca6f1a70.png)

Tried to delete dxvk_versions.json manually but it doesnt get recreated.

tannisroot:
I can't reproduce this.
Please provide more information.

clapbr:
Really nothing much besides default everything. Tried and reproduced in a clean user, video https://www.youtube.com/watch?v=aiz05PMA1ws&feature=youtu.be - sorry for potato quality, I sent 1080p but youtube encoder butchered it.

clapbr:
Ok, figured it out. get/init_dxvk_versions only triggers if I install a game on steam wine and then change it's configuration. This behavior could change to also trigger when you open Wine and Wine Steam configs without any game.

strycore:
Makes sense, that should give me enough info to start working on a fix.

PR #1373, opened 2018-11-25
Update dxvk on game addition + change to decorator method

sigmaSd:
fix https://github.com/lutris/lutris/issues/1370


sigmaSd:
Just discovered another method to do this without decorators
```python 
init_dxvk_versions(first = [true]):
    if first and first.pop():
        ...
```
So many way to do this xD

strycore:
very nice implementation!

sigmaSd:
Thanks !

PR #1376, opened 2018-11-26
Fix call to sync_with_lutris()

RobLoach:
Was running into this error...

```
service.sync_with_lutris()
TypeError: sync_with_lutris() missing 1 required positional argument: 'games'
```


strycore:
This is not a bug, this is unfinished code

strycore:
But Strider, this is not a bug report, this is a PR.

PR #1420, opened 2018-12-18
Fix misplaced parenthesis.

Xenega:


strycore:
indeed, the sidebar in next is not resizable.

PR #1421, opened 2018-12-18
make LutrisThread not a thread

AaronOpfer:
`LutrisThread` doesn't actually do anything threaded because the body of `run` method doesn't block. Make it no longer threaded. Also move `set_child_subreaper` to application initialization so that we don't have to keep calling it over and over again on every game launch for no reason, and because I originally put it there because I wasn't sure if it was a per-thread setting (which is now irrelevant).

Also, I noticed that `attach_thread` is dead code so I removed it.


strycore:
Looks cool! I'll have to think of some naming stuff since LutrisThread in obviously no longer appropriate

Issue #1012, opened 2018-08-09
Re-enabling of desktop effects takes too long


tannisroot:
After trying new "Disable Desktop Effects" option I have experienced quite a big nuance with it - it takes a lot of time for it to re-enable the compositor, and sometimes it actually never happens and I have to do that manually.
I assume it's because the option waits for the prefix to completely shutdown (as I can manually force it re-enable compositor by pressing the stop button), which can take a while or never even happen for whatever reasons. 
So maybe it should re-enable the compositor after only monitored processes exit instead to make it do that faster?
Or do whatever "Restore Gamma..." (that got fixed in #1011 ) does as it restores default gamma immediately after one of my games that changes gamma exits.

strycore:
I don't know what to say...

I don't use either of those options and the fact that xgamma is called **after** restoring the compositor when the game quits makes the issue even more difficult to work on.

You might have to dive in the code a bit and add some log statements if you want to provide a ticket I can solve.

strycore:
post launch actions will now be executed as soon as the game quits

PR #1423, opened 2018-12-19
speed up detection of wine process exit

AaronOpfer:
This changeset installs a `SIGCHLD` signal handler which reaps child processes, and on successful child process reaping, triggers an early execution of `watch_children` to evaluate the new process state. This also removes the cycle count requirement for marking a process as exited since we now have more reliable ways to determine the state of the application. It also adds more background wine processes to the exclude list by default. Combined, these changes mean much more rapid detection of an exited wine game.

This should significantly help #1012.

### DIFF
--- a/docs/installers.rst
+++ b/docs/installers.rst
@@ -1022,8 +1022,6 @@ Sysoptions
 
 ``disable_runtime`` (example: ``true``)
 
-``monitor_max_cycles`` adjusts the number of cycles the monitor will wait until declaring a game as exited (example: 10)
-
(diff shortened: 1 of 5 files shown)


strycore:
Does your patch allow for a period of time where no process is running?

strycore:
It looks really nice and I like the idea of being able to eliminate the max cycle option. I'll merge it and see if it passes tests with Warframe or Battle.net games

strycore:
yep, seems to be working fine! :ok_hand:

PR #1424, opened 2018-12-19
speed up lutris/util/wine/registry.py

### LINKED ISSUE #1418: Lutris takes extremely long time to parse very long string keys in registry
If user.reg contains very long (we a talking megabytes long here) string key it takes a very long time to start anything in this wineprefix with lutris. With ~15MB key lutris was taking ~30 minutes. It happens before starting wine, in "preparing to start" phase.

P.S. If you ask how does this happens - Battletech ny HBS writes some analytics into string key in registry. And it only adds to it, never cleans.

PythonicChemist:
Fixes #1418.
- ordered conditionals by frequency avoids calling startswith() numerous times
- sped up access of the last subkey in WineRegistryKey.add_to_last()
- bundled the additional values to be added to the last subkey

```
Before:
1785064 function calls (1785054 primitive calls) in 401.188 seconds
After:
609004 function calls (608994 primitive calls) in 0.139 seconds
```

I've compared the provided user.reg file with the output of this PR and couldn't find any deviations.

### DIFF
--- a/lutris/util/wine/registry.py
+++ b/lutris/util/wine/registry.py
@@ -109,32 +109,29 @@ class WineRegistry:
         registry_lines = self.get_raw_registry(reg_filename)
         current_key = None
         add_next_to_value = False
+        additional_values = []
         for line in registry_lines:
-            line = line.rstrip("\n")  # Remove trailing newlines
-
-            if line.startswith(self.version_header):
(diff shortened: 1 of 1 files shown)


strycore:
Didn't pass unit tests, reverted

PR #1447, opened 2018-12-22
[WIP] Collections feature

Riesi:
This pull request adds collections in the sidebar, which helps sorting games. 
This is by no means ready to merge and will probably take some iterations to make it fit the quality of Lutris. There is probably some overhead to be cut by saving the collections in a smarter way. (I dont really know whats possible in SQL)
The UI for adding the collections to a game should be changed to something like a combo-box or another better fitting element, but first get the data management correct.

I dont have much experience with Python, but I think we can make this feature work in a way that is merge-able.

### DIFF
--- a/lutris/game.py
+++ b/lutris/game.py
@@ -37,6 +37,18 @@ def watch_lutris_errors(function):
     return wrapper
 
 
+def join_coll(collection):
+    if collection is None:
+        return None
+    return "§".join(collection)
+
+
+def split_coll(collection):
+    if collection is None:
+        return ""
+    return collection.split("§")
+
+
 class Game(GObject.Object):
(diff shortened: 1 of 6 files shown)


Riesi:
Screenshot of the feature

PythonicChemist:
Having the user use syntax in a text field is not a good UX. 

Maybe we should take inspiration from steam's "Set Category..." feature. I really like this feature and might be beneficial to rename collections to categories, because that is the term most users will be most familiar with.

Good work so far and am looking forward to this feature! 👍 
 
P.S.: The new code in the update method in sidebar.py maybe is better put into a private method and try to find better names for the variables than something with temp in it.

PR #1452, opened 2018-12-25
Category feature

Riesi:
Ok I feel a bit stupid for messing up  and having to do a new pull request. 
This code should be much better in regards to quality and efficiency.

- SQL tablse for categories and m2m relationship
- REFERENCES in a very basic way
- UNIQUE constraint; only single entry constraint for now, because else stuff broke
- sidebar entries for categories
- entry in the right click menu
- dedicated window for adding and selecting categories
- excessive, leading and tailing whitespaces get removed
- categories get deleted, if they have no entry in the m2m table

screenshot

### DIFF
--- a/lutris/game_actions.py
+++ b/lutris/game_actions.py
@@ -8,6 +8,7 @@ from lutris.gui import dialogs
 from lutris.gui.widgets.utils import open_uri
 from lutris.gui.config.add_game import AddGameDialog
 from lutris.gui.config.edit_game import EditGameConfigDialog
+from lutris.gui.config.edit_game_categories import EditGameCategoriesDialog
 from lutris.gui.installerwindow import InstallerWindow
 from lutris.gui.dialogs.uninstall_game import UninstallGameDialog
 from lutris.gui.dialogs.log import LogWindow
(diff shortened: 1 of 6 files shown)


tannisroot:
I think "WIP" part should be dropped since you are not actively improving the PR.

Riesi:
I think WIP still fits, since I dont think this is completely ready.
But more feedback would be great for that, because I don't know what you guys like and what not.

tannisroot:
Yeah but I feel like WIP kind of implies it is not ready even for a review.

Riesi:
Ahh ok that also makes sense! Rebased it onto latest next.

tannisroot:
I think it should be rebased against master now, btw.

strycore:
Pushing back to 0.5.3 because there is a lot to review. I'll merge this in a separate branch and to the review from there. You can then submit other PRs against that branch once I've created it.

strycore:
This is now merged in the `categories` (https://github.com/lutris/lutris/tree/categories) branch and PRs related to this feature can be submitted against that branch.

PR #1455, opened 2018-12-27
Fix wrong order of tuple for assigning soft- and hardlimits value

Riesi:
Here is the documentation of the method:
https://docs.python.org/3/library/resource.html#resource.getrlimit

> Returns a tuple (soft, hard) with the current soft and hard limits of resource. Raises ValueError if an invalid resource is specified, or error if the underlying system call fails unexpectedly.

### DIFF
--- a/lutris/util/system.py
+++ b/lutris/util/system.py
@@ -116,7 +116,7 @@ class LinuxSystem:
 
         self.populate_libraries()
         self.populate_sound_fonts()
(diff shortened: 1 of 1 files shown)


PythonicChemist:
For more reference the commit 3b569bcc2d4da403ebf21f97811be4c3e3de9c72 also shows that it was just a mistake during a refactor. LGTM 👍

PR #1464, opened 2019-01-02
Added additional dependencies for Fedora

BunnyApocalypse:
I decided to make this in the Fedora check, as I do not know if the package names are consistent across all RPM based distributions. I can move them outside if you'd like, but I can't be sure that it'll work.


strycore:
can you send thing to next?

strycore:
it's good actually

PR #1476, opened 2019-01-03
PICO-8 runner!

daniel-j:
This is a runner for the [PICO-8 Fantasy Console](https://www.lexaloffle.com/pico-8.php). You can choose game by it's id, url or .p8.png file. The runner also has a web-based runtime which uses the free web build of PICO-8 and is downloaded from their website. This runner depends on the `web` runner, but only if you don't own the game. If you do own the game, go configure it and select your `pico8` or `pico8_dyn` executable, and then restart Lutris to be detected. The web based runner does not include the editor, and can only play .p8.png cartridges. Enjoy!


daniel-j:
Runner config:
screenshot from 2019-01-03 02-05-47
Running Celeste with the web engine:
screenshot from 2019-01-03 02-06-45
Games can be just a simple id! :D
screenshot from 2019-01-03 02-07-41
When you buy PICO-8 you can access the editor:
screenshot from 2019-01-03 02-08-53

daniel-j:
TODO: Replace the web runner usage with a Python gio WebKit2 implementation instead.

daniel-j:
Removed the web runner dependency, uses a brand new python and Gtk WebKit2-based webview windows instead! This might become the new web runner in the future for non-flash content.

PR #1490, opened 2019-01-04
[next] Fix sorting

daniel-j:
Sorting some fields from the dropdown menu didn't work, which this fixes.

### DIFF
--- a/lutris/gui/gameviews.py
+++ b/lutris/gui/gameviews.py
@@ -48,8 +48,9 @@ sortings = {
     "year": COL_YEAR,
     "runner": COL_RUNNER_HUMAN_NAME,
     "platform": COL_PLATFORM,
-    "lastplayed": COL_LASTPLAYED,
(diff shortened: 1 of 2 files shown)


lutris/gui/gameviews.py
```diff
@@ -408,9 +409,9 @@ def __init__(self, store):
         self.set_column(default_text_cell, "Year", COL_YEAR, 60)
         self.set_column(default_text_cell, "Runner", COL_RUNNER_HUMAN_NAME, 120)
         self.set_column(default_text_cell, "Platform", COL_PLATFORM, 120)
-        self.set_column(default_text_cell, "Last played", COL_LASTPLAYED_TEXT, 120)
+        self.set_column(default_text_cell, "Last Played", COL_LASTPLAYED_TEXT, 120)
```
TingPing: This seems very wrong, you sort times by timestamp not text.

Issue #1097, opened 2018-09-07
[Feature] Auto-sandbox Firejail


asddsaz:
**Feature**: Automatically isolate games using [Firejail](https://firejail.wordpress.com/documentation-2/basic-usage/) for added security.

**Why**?: Many games are proprietary this would substantially improve privacy.

Hiradur:
Sandboxing games with firejail isn't very difficult. I've been doing this for quite a while now. As far as I can tell the performance impact is also negligible but I didn't do any benchmarking.
For each game I create a specific firejail profile. Usually only the whitelisted directories need to be adjusted for each game, but I additionally sometimes block network access where I don't need it (see comments below).

Firejail needs to be invoked as a prefix to the game command. For example, under Rightclick on game icon->system settings->command prefix add: `firejail --profile=~/.config/firejail/lutris-KOTOR2.profile`.

The Firejail profile for the game looks like this:
```
# Firejail profile for KOTOR2 running on Wine (Lutris)

# Adjust this
noblacklist ~/Games/star-wars-knights-of-the-old-republic-ii-the-sith-

include ~/.config/firejail/openal.inc
include ~/.config/firejail/lutris-common.inc
include /etc/firejail/disable-common.inc
include /etc/firejail/disable-devel.inc
include /etc/firejail/disable-passwdmgr.inc
include /etc/firejail/disable-programs.inc

# Adjust these
mkdir ~/Games/star-wars-knights-of-the-old-republic-ii-the-sith- 
whitelist ~/Games/star-wars-knights-of-the-old-republic-ii-the-sith-

caps.drop all
ipc-namespace

# Adjust this
net none

#nodbus
nodvd
nogroups
nonewprivs
noroot
notv
#nou2f

# Add network protocols if you need networking
protocol unix

# Seccomp is disabled for performance because it enables security mitigations for hyper threading I think
#seccomp

#shell none
tracelog

machine-id

disable-mnt
#private-bin bash,dpkg,ls,wine,wine-stable,wineserver,wineserver-stable,sh,ntlm_auth
#private-etc drirc,group,hostname,localtime,nsswitch.conf,passwd,resolv.conf
#private-dev
private-tmp

noexec /tmp
```

In the firejail profile above I include two custom files which are:

lutris-common.inc:
```
noblacklist ~/.local/share/lutris/runners

mkdir ~/.local/share/lutris/runners
whitelist ~/.local/share/lutris/runners
read-only ~/.local/share/lutris/runners/
```

[...]

PR #1507, opened 2019-01-05
use wrapper processes to manage running games

AaronOpfer:
This is attempting to address multiple issues involving running multiple games or multiple instances of the same game from the same `WINEPREFIX` at the same time. These issues include:

 * When a second application starts and the first one closes, lutris doesn't recognize that the first application has closed.
 * When a second application closes, the first still-running application is killed.
 * When one application is stopped, all applications using that prefix are stopped. This is because behind the scenes, stop is using `wineserver -k`.

All of these issues are addressed in the below PR.

Instead of Lutris directly setting `PR_SET_CHILD_SUBREAPER` and having lots of children running around whose original parents are unknown and unknowable, we add an intermediate daemon (here `child_wrapper.py`, invoked as `python3 -m lutris.child_wrapper`) who sets `PR_SET_CHILD_SUBREAPER` and watches a game's process and its child processes on Lutris' behalf. It basically has two responsibilities:

 * collect all of a game's processes, no matter how tragic their lineage is, while ignoring other processes.
 * when `SIGINT` or `SIGTERM` are received, propagate those signals *only* to those directly relevant processes. Do NOT send them to `wineserver`.
 * Exit when all of the game processes exit, giving Lutris a simple indication that the game has exited.

Extra care had to be taken to avoid breaking the "run in terminal" feature and it actually works better now (it doesn't spawn an extra shell anymore and the terminal closes timely when the game closes).

Wine applications no longer invoke `wineserver -k`. Users can invoke `wineserver -k` by right-clicking a game and clicking "Kill all wine processes". This context menu is getting a little crowded now, though.

I believe there may be significant amounts of dead code laying around the process monitoring system and the wine runner now as I tore out a lot of code based on the old paradigm. It may need a once-over to find unused code and remove it.

I changed the wine runner to have a prelaunch step where we launch the `wineserver` as well, so that none of the `child_wrapper` instances will have the `wineserver` inside their perview. This is technically unnecessary because of the exclude process system working just fine to accomplish the same thing, I figured the more we can isolate the process trees the better off we'll be. However, processes like `plugplay.exe` and `winedevice.exe` are still created inside the `child_wrapper` process tree, so I'm not sure if my efforts have done much here.

### DIFF
--- a/lutris/child_wrapper.py
+++ b/lutris/child_wrapper.py
@@ -1,0 +1,77 @@
+# coding: utf-8
+import os
+import time
+import ctypes
+import sys
+import subprocess
+import signal
+from ctypes.util import find_library
+from lutris.util.log import logger
+from lutris.util.monitor import ProcessMonitor
+
+PR_SET_CHILD_SUBREAPER = 36
+def set_child_subreaper():
+    """Sets the current process to a subreaper.
+
+    A subreaper fulfills the role of init(1) for its descendant
+    processes.  When a process becomes orphaned (i.e., its
+    immediate parent terminates) then that process will be
+    reparented to the nearest still living ancestor subreaper.
+    Subsequently, calls to getppid() in the orphaned process will
+    now return the PID of the subreaper process, and when the
+    orphan terminates, it is the subreaper process that will
+    receive a SIGCHLD signal and will be able to wait(2) on the
+    process to discover its termination status.
+
+    The setting of this bit is not inherited by children created
+    by fork(2) and clone(2).  The setting is preserved across
+    execve(2).
+
+    Establishing a subreaper process is useful in session
+    management frameworks where a hierarchical group of processes
+    is managed by a subreaper process that needs to be informed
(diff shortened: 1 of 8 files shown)


AaronOpfer:
I discovered a regression in battle.net launcher where lutris kills the games after my changeset. I'll investigate.

AaronOpfer:
OK, looks like splitting out the `wineserver` into prelaunch is broken in some way, I suspect something to do with environment variables. It's not very important to the patch so just don't bother.

strycore:
I've moved some stuff around recently (mostly renamed lutris.thread to lutris.command) so there are going to be some adjustments needed on this PR

AaronOpfer:
PR updated with a new rebased & squashed version of the patch.

strycore:
I've reviewed everything and it's really neat, great work!

AaronOpfer:
Thanks! I think this may make #1097 easier to implement as well since process monitoring is now delegated to the subprocess.

PR #1580, opened 2019-01-12
Update get_game_by_id() to check for game.id

RobLoach:
There were cases where running_games objects didn't have `id` properties, running into python errors. Checking to ensure the property exists first keeps it from breaking.

### DIFF
--- a/lutris/gui/application.py
+++ b/lutris/gui/application.py
@@ -374,7 +374,7 @@ class Application(Gtk.Application):
 
     def get_game_by_id(self, game_id):
         for game in self.running_games:
-            if game.id == game_id:
+            if hasattr(game, 'id') and game.id == game_id:
                 return game
 
     @staticmethod


AaronOpfer:
Hmm... feels like the right fix is to give all `running_game` objects IDs instead.

RobLoach:
Yaaaassssss, where do we do that?

strycore:
@RobLoach  Don't add checks like that as it creates potential issues later on. There's a very good article on the subject: https://medium.com/@vcarl/overly-defensive-programming-e7a1b3d234c2

AaronOpfer:
@strycore Am I correct in assuming `ServiceGame` and `Game` instances are meant to have the same interface? If they are, `ServiceGame` calls it a `game_id` while `Game` calls it `id`.

In the future to prevent bugs like these we might want to make an `ABCGame` class that defines the interface (using Python [abstract base classes](https://docs.python.org/3/library/abc.html)).

strycore:
No ServiceGame and Game classes are not related in any kind.

strycore:
the game_id attribute of ServiceGame is here to store a Game.id reference.

PR #1605, opened 2019-01-15
make lutris.util.downloader use monotonic time

AaronOpfer:
`time.time` can go backwards if the system clock time is changed. This makes it inappropriate for measuring the length of time between two events. Replace with `time.monotonic` inside `lutris.util.downloader`.

### DIFF
--- a/lutris/util/downloader.py
+++ b/lutris/util/downloader.py
@@ -6,6 +6,10 @@ from lutris.util import http, jobs
 from lutris.util.log import logger
 
 
+# `time.time` can skip ahead or even go backwards if the current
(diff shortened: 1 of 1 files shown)


AaronOpfer:
This includes some changes I made to fix some of CodeFactor's complaints, which is a mistake on my part. I intended to make them in another PR. The average download speed calculation has some funny things going on and I changed its behavior slightly in the process. Not sure what the author originally intended.

AaronOpfer:
@strycore how is it now?

strycore:
Perfect!

PR #1609, opened 2019-01-15
Add wine SampleCount option

TRPB:
A trivial change to expose WINE's registry option SampleCount in Lutris' GUI. This allows forcing MSAA in games. It's useful for older games that don't have an in-game option for anti-aliasing. 

I'm fairly sure this will also need adding in the code that parses installers so you can set the SampleCount in the installer, if it's required.

### DIFF
--- a/lutris/runners/wine.py
+++ b/lutris/runners/wine.py
@@ -102,6 +102,7 @@ class wine(Runner):
         "MouseWarpOverride": r"%s/DirectInput" % reg_prefix,
         "OffscreenRenderingMode": r"%s/Direct3D" % reg_prefix,
         "StrictDrawOrdering": r"%s/Direct3D" % reg_prefix,
+        "SampleCount": r"%s/Direct3D" % reg_prefix,
         "Desktop": "MANAGED",
         "WineDesktop": "MANAGED",
         "ShowCrashDialog": "MANAGED",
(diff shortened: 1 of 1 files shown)


tannisroot:
This has no effect on DXVK, this only affects Wined3d.

lutris/runners/wine.py
```diff
@@ -647,7 +667,11 @@ def set_regedit_keys(self):
                         value = None
                     managed_keys[key](value)
                     continue
-                prefix_manager.set_registry_key(path, key, value)
+                # Convert numeric strings to integers so they are saved as dword
+                if value.isdigit():
+                    value = int(value)
+
+                    prefix_manager.set_registry_key(path, key, value)
```
strycore: is this supposed to be indented like that?

AaronOpfer:
I don't know if it's possible with the way configuration pages are created right now, but I'd like to see the SampleCount option disabled/grayed out if DXVK is enabled so that users don't get confused.

If we can't do that, I'd like to see it mentioned in the configuration option's help text instead.

lutris/runners/wine.py
```diff
@@ -647,6 +667,10 @@ def set_regedit_keys(self):
                         value = None
                     managed_keys[key](value)
                     continue
+                # Convert numeric strings to integers so they are saved as dword
+                if value.isdigit():
```
strycore: This seems like an unsafe change to make and might cause bugs.

strycore:
If it's an advanced option, users are expected to know the effects and we can just add a note in the label to point out that this wont work on DXVK games (as long as the option doesn't crash any games).

Going to merge that in dev because I'm not sure about the isdigit line and it may need more testing.

Issue #1140, opened 2018-10-01
Lutris - Unable to get runners


Leopard1907:
So , i have this strange problem. Lutris doesn't let me to manage Wine versions etc.

Gives that error.

![lutris](https://user-images.githubusercontent.com/24878978/46320821-d0257380-c5e8-11e8-9f72-8a642d3a682b.png)

Output of `lutris -d`:

`ERROR    2018-10-02 02:09:31,774 [dxvk]:<urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:645)>
WARNING  2018-10-02 02:09:31,870 [application]:This version of Gtk doesn't support set_option_context_summary
ERROR    2018-10-02 02:09:32,202 [http]:Unable to connect to server (https://lutris.net/api/runtime): <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:645)>
DEBUG    2018-10-02 02:09:32,206 [api]:Fetching game library
DEBUG    2018-10-02 02:09:32,399 [steam]:Watching Steam folder /home/asus/.local/share/Steam/steamapps
DEBUG    2018-10-02 02:09:32,400 [steam]:Watching Steam folder /Steam Oyunları/steamapps
ERROR    2018-10-02 02:09:40,479 [http]:Unable to connect to server (https://lutris.net/api/runners/wine): <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:645)>
ERROR    2018-10-02 02:09:43,854 [http]:Unable to connect to server (https://lutris.net/api/runners/wine): <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:645)>
ERROR    2018-10-02 02:10:57,897 [http]:Unable to connect to server (https://lutris.net/api/runners/wine): <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:645)>
`

Though i can ping to Lutris.net just fine:

![lutris2](https://user-images.githubusercontent.com/24878978/46320893-25fa1b80-c5e9-11e8-9fb2-04139237eb40.png)

Distro: Linux Mint 18.3

AccountOneOff:
Duplicate #1134

Leopard1907:
I think that has to do with libssl. My version is 1.0.2 something.

[...]

PR #1613, opened 2019-01-16
 Fix SSL: CERTIFICATE_VERIFY_FAILED Web, Runtime, Resources Modules

kelebek333:
Fix SSL: CERTIFICATE_VERIFY_FAILED Web, Runtime, Resources Modules

https://github.com/lutris/lutris/issues/1140

### DIFF
--- a/lutris/runners/web.py
+++ b/lutris/runners/web.py
@@ -1,6 +1,7 @@
 import os
 import string
 import shlex
+import locale
 from urllib.parse import urlparse
 
 from lutris.runners.runner import Runner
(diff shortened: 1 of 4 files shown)


strycore:
It would be very nice if that does indeed solve this problem but there got to be a more elegant way to implement this, I can't just have setlocale calls dropped in random modules of the application.

kelebek333:
I think the problem is related to the version of opensll. The problem was solved with any fix, when I installed version 1.1.1 from PPA. 

If you want, you can close this commit.

strycore:
Oh, that's good news, I'll point people with that issue to this direction. Btw, I don't think you meant to push your translation to this PR.

kelebek333:
Thank you very much for your interest.

I send new commit for "update tr.po".

PR #1655, opened 2019-01-23
Fix install runners button

gasinvein:
Button probably was broken by 35b7230


RobLoach:
I believe this is already fixed in `master`.

PR #1693, opened 2019-02-01
checks can crash the program when an unmigrated game needs to be provided a platform

roei12:
if an un-migrated game needs to be provided a platform, lutris checks will crash the program when trying to save the game's data.  
this happens because `playtime` couldn't be initialized with un-migrated data and thus isn't an attribute in the Game object.  

so this PR should fix 3 issues:  
- `Game` object doesn't always have `playtime` attribute  
- `fill_missing_platforms` is called before migration, and tries to save un-migrated game data   
- `fill_missing_platforms` is saving data even when it fails to get platform

Log: https://pastebin.com/HBZN7R5c

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -3,6 +3,7 @@ Section: games
 Priority: optional
 Build-Depends: debhelper,
  python3,
+ python3-yaml,
  python3-setuptools,
  python3-requests,
  python3-pil,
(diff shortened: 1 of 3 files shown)


strycore:
This was manually merged, closing

PR #1704, opened 2019-02-02
Update version in appdata.xml - & other suggested change

BunnyApocalypse:
There should probably be some automated process that bumps version numbers in this file.

Also, the changed names of this appdata & the desktop file to match Flatpak(?) naming requirements was something that made me, and probably a few other packagers need to change things to find these newly named files. Perhaps we could have both the Flatpak and normally named files so that this doesn't cause problems?

### DIFF
--- a/share/appdata/net.lutris.Lutris.appdata.xml
+++ b/share/appdata/net.lutris.Lutris.appdata.xml
@@ -26,6 +26,6 @@
   <project_license>GPL-3.0</project_license>
   <metadata_license>CC0-1.0</metadata_license>
   <releases>
-    <release version="0.4.23" date="2018-11-07"/>
+    <release version="0.5.0" date="2018-02-02"/>
   </releases>
 </application>


TingPing:
> Perhaps we could have both the Flatpak and normally named files so that this doesn't cause problems?

That is a terrible idea. Apps should only have one id so you avoid things like app stores showing two duplicates with different reviews and whatnot. Packages will already have done the rename the 5 seconds of pain will pass.

share/appdata/net.lutris.Lutris.appdata.xml
```diff
@@ -26,6 +26,6 @@
   <project_license>GPL-3.0</project_license>
   <metadata_license>CC0-1.0</metadata_license>
   <releases>
-    <release version="0.4.23" date="2018-11-07"/>
+    <release version="0.5.0" date="2018-02-02"/>
```
TingPing: Would be nice to get the changelog in here too.

PR #1726, opened 2019-02-05
lutris.spec fix

andrewschott:
Fixed a few errors regarding fedora in the spec file.  Issue with .desktop file registration and the 'gnome-desktop3' runtime dependency.

PR #1727, opened 2019-02-05
Fix typo in lutris.spec

AsciiWolf:


AsciiWolf:
Oops, I see that there is already fix for this in #1726.

PR #1728, opened 2019-02-05
Fix invalid get_row_by_id() access in list-view

Riesi:
This error occurs in list-view when a game is updated for example when configuring and then saving.


Riesi:
Seems like #1723 is the exact same thing.

PR #1752, opened 2019-02-12
add PrimusVK as a bumblebee option (adds Vulkan functionality)

nephitejnf:
Implements PrimusVK as another option for Vulkan support for those that have it installed on their system with bumblebee. This is mostly because it is a viable option now for those that take advantage of bumblebee. The project itself can be found here. https://github.com/felixdoerre/primus_vk


strycore:
Cannot test it but looks safe enough of a change to merge

Issue #1758, opened 2019-02-13
Do not set DRI_PRIME to 0 if hybrid graphics are not detected


tannisroot:
Currently, Lutris will set DRI_PRIME to 0 if it can't detect dedicated GPU, which is pretty bad as a bug in the detection of hybrid graphics (like https://github.com/lutris/lutris/issues/1732) will make it impossible to set DRI_PRIME envar to 1 manually through envar menu (Lutris will override manually set value and set it to 0 anyway) and workaround such bug in Lutris, thus making it impossible to use dedicated GPU for games installed through Lutris.
What Lutris should do instead is just not set DRI_PRIME to anything if it can't detect hybrid graphics.

sonicpp:
Agree. Do you have any example of bug in detection of hybrid graphics? (bug #1732 is already fixed).

tannisroot:
Well, not at the moment, but in case a similar bug appears, it would be nice if we had a workaround option available.

ririen:
I confirm as another person that there is such an error. In earlier versions, `DRI_PRIME` worked correctly. I've tried to use Lutris 0.5 lately and the game does not see/work the variable `DRI_PRIME=1` at all.

Issue #1749, opened 2019-02-11
Move "Use PRIME" out of advanced options


tannisroot:
This is not really an advanced feature and many people will look for it to use dedicated graphics in their games instead of relying on integrated graphics.

strycore:
It is an advanced feature. Can we default to enabling prime by default instead?

tannisroot:
Yeah but I don't get why something like DLL overrides is not an advanced feature while such commonly used and needed feature as PRIME is.

sonicpp:
There should be no side effects when setting DRI_PRIME=1 environment even for GPUs not supporting this feature or for single GPU setup as far as I know. But it may confuse players with single GPU - it will show PRIME enabled with no option to disable it...

tannisroot:
PRIME is currently greyed out if you don't have a second GPU, so it's not an issue,

sonicpp:
I know, but my point was that if we enable it by default, it will be shown as enabled (even though you have single GPU system).

tannisroot:
Despite "Enable Feral Gamemode" being enabled by default when it's present on the system, it's disabled when it's not present.

sonicpp:
Yes, that is what I meant. In that case I vote for enabling it by default too :)

strycore:
I still don't understand why this is an advanced feature, you should need to set this one, in the system options and be done with it.

tannisroot:
> you should need to set this once

Well, this can be applied to pretty much any option in Lutris.
From **my perspective**, advanced options are those options that are not essential and meant to not be tampered with (as that may cause issues) if you are just a regular user and don't really know what you're doing, which is why they are hidden by default. 
I don't consider PRIME option something that can mess up your games (if you don't have Hybrid graphics you won't even be able to enable it) and it's **quite essential** as enabling the use of dedicated GPU can bring massive performance boost. So in a way, it's in the same category as Feral's Gamemode.

strycore:
so it's just a matter of having it enabled by default.

tannisroot:
Well yeah, that would be nice too. But PRIME option is really something the user will look for after they install Lutris for the first so it would be nice to show the user that it's already enabled, as they might try and do it manually through envar menu.

strycore:
Prime should now be enabled by default.

PR #1780, opened 2019-02-18
Prime fixes

sonicpp:
Small fixes for DRI_PRIME environment variable.
First commit 9d2ffd8 resolves issue #1758, while commit 49f7439 enables DRI_PRIME by default, as per discussion from #1749


strycore:
Very cool!

PR #1786, opened 2019-02-19
give every installer cache dir a unique name

### LINKED ISSUE #1773: Don't remove installation files after cancelling installation if another installer is open
How to reproduce:
1. Use an installer that downloads files (like https://lutris.net/games/the-sims/)
2. Start the installation, continue until installer downloads installation files (which happens after selecting installation directory)
3. After that, without cancelling existing installation, use the same installer and repeat step 2
4. Cancel the first installation and close the window.

As a result, Lutris removes installation files in `~./cache/lutris/installer/`, and the other installer doesn't execute correctly and errors out due to missing files.

In the log, I see:
```Removing folder /home/alex/.cache/lutris/installer/the-sims```
So perhaps, the issue could be resolved by giving each installer instance a unique name for the cache folder. So instead of just `the-sims`, it could be `the-sims-654656654`.

Medath:
Fixes #1773 
This patch gives every cache directory of every installer its own unique name by appending the start time of the installer in milliseconds to the directory name. I have done some basic testing with two different games and it seems to work just fine.

### DIFF
--- a/lutris/installer/installer_file.py
+++ b/lutris/installer/installer_file.py
@@ -9,7 +9,8 @@ from lutris.util import system
 
 class InstallerFile:
     """Representation of a file in the `files` sections of an installer"""
-    def __init__(self, game_slug, file_id, file_meta):
+    def __init__(self, game_slug, file_id, file_meta, launch_time):
+        self.launch_time = launch_time
         self.game_slug = game_slug
         self.id = file_id  # pylint: disable=invalid-name
         self.dest_file = None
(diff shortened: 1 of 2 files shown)


strycore:
I don't agree with this design. Some changes to how caching is handled are currently being made and PR just make the re-architecturing process a lot more difficult. If you want to send a workaround, send one that makes it impossible to open 2 InstallerWindows for the same installer_slug (if that happens, it presents the first window)

PR #1845, opened 2019-03-09
Add 5px margin to all sides of search entry

Feichtmeier:
The search entry has currently no margin, making it look baked to the sides. This adds 5px margin to all sides.

Before:
before_lutris_search_entry

After:
after_lutris_search_entry

### DIFF
--- a/share/lutris/ui/lutris-window.ui
+++ b/share/lutris/ui/lutris-window.ui
@@ -614,6 +614,10 @@
                             <property name="can_focus">True</property>
                             <property name="halign">center</property>
                             <property name="valign">center</property>
+                            <property name="margin_left">5</property>
+                            <property name="margin_right">5</property>
+                            <property name="margin_top">5</property>
+                            <property name="margin_bottom">5</property>
                             <property name="hexpand">True</property>
                             <property name="width_chars">50</property>
                             <property name="placeholder_text" translatable="yes">Filter the list of games</property>


share/lutris/ui/lutris-window.ui
```diff
@@ -614,6 +614,10 @@
                             <property name="can_focus">True</property>
                             <property name="halign">center</property>
                             <property name="valign">center</property>
+                            <property name="margin_left">5</property>
```
TingPing: I believe you only need one property: `<property name="margin">5</property>`

Feichtmeier: Glade does not provide a way to change all the margin.
However when I edit the file in a text editor with your proposed change, Glade produces this preview, which misses a margin-top, and also can't display the change it it's "margin" config panel on the right side:

image

TingPing: Well glade isn't very useful most of the time and you can just edit the file directly.

But its not a big deal eitherway.

Feichtmeier: I disagree ^.^ it's very helpful for designing :)

TingPing: I mean, in theory, but it doesn't actually understand the format its reading/writing often.

strycore: It is a big deal to me, if Glade were usable, I would actually write .ui files instead of going for Python code, which is current preferred method. Really not a fan of editing XML files manually.

strycore:
I don't notice any difference while QAing this PR. I think it's broken.

PR #1862, opened 2019-03-16
Ask for confirmation when changing the runner for a game

Askannz:
Hi,

It's my first pull request to Lutris so my apologies for anything done wrong.

I noticed that when changing the runner for game, Lutris just erases any configuration written in the Game, Runner and System tabs. I believe this behavior is a little dangerous, even if nothing is actually applied until the user presses "Save". After all, one of the first things a new Lutris user does is often opening the configuration menus and play with anything that can be tweaked. The other day, an user requested help in the Lutris Discord because he had accidentally zero-ed out his configuration for Overwatch by changing the runner.

This pull request simply adds a confirmation message when the user changes the current runner in the "Game info" tab of a game. If the user answers "No", the drop-down menu is reverted to the previously selected runner.

I am publishing the request as a draft for now because it creates two problems : 

 - first, the confirmation pop-up will also appear when the user manually adds a new game and selects a runner, even if the rest of the configuration is already blank.
- second, the change seems to break the current test suite, because it asks for an additional user input that the tests have not been written for.
 
The second problem can be fixed by adapting the tests, but the first one requires more work (we need to create a special case for new games somehow).

It's a small feature, so if you think it is not worth the effort, I'll just drop the pull request.

### DIFF
--- a/lutris/gui/config/common.py
+++ b/lutris/gui/config/common.py
@@ -9,7 +9,7 @@ from lutris import settings
 from lutris.cache import get_cache_path, save_cache_path
 from lutris.gui.widgets.common import VBox, SlugEntry, NumberEntry, Label, FileChooserEntry
 from lutris.gui.config.boxes import GameBox, RunnerBox, SystemBox
-from lutris.gui.dialogs import ErrorDialog
+from lutris.gui.dialogs import ErrorDialog, QuestionDialog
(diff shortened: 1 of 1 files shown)


strycore:
This looks pretty good so far.
Maybe we should be handling the configuration in a better way so that the config isn't lost on changes but until then, this is a good solution.

Issue #1869, opened 2019-03-18
Option to use xdg-open instead of Gtk.show_uri to browse folders


sparr:
I don't have Gnome installed or any gnome-specific management packages, so my gio mime settings are unmanaged, leaving a random `inode/directory` handler as the "default" for the "Browse Files" functionality for games in Lutris.

I would like to see an option to use the `xdg-open` script instead, which has its own logic for resolving handlers across different desktop environments and otherwise.

strycore:
Open a bug against show_uri in Gtk since it's not doing the right thing.

sparr:
What "right thing" would you expect it to do in this situation? The freedesktop standards say that if a DE hasn't installed any default defaults that the default should be chosen effectively at random (actually the first on the list, and the list is randomized when the cache is regenerated). show_uri seems to be behaving exactly as the spec requires.

TingPing:
> show_uri seems to be behaving exactly as the spec requires.

Indeed, no bug here IMO.

Just run `gio mime inode/directory foo.desktop` if you want to change it.

sparr:
My motivation here is to not need to manage mime associations in two different places. Very few non-Gnome-specific applications gnome/gtk/gio-only systems like `show_uri`; most of them honor settings from other desktop environments by using xdg tools.

TingPing:
Gio is not gnome-specific in any way. GLib is a cross-platform project and the freedesktop spec is a cross-desktop standard.

sparr:
While gio is not gnome-specific, most users have no idea what gio is, and won't know how to configure it without having tools installed from Gnome (or another DE).

strycore:
I've never had to "configure" Gio since 1996.

PR #1870, opened 2019-03-18
Add option to use xdg-open instead of Gtk.show_uri for browsing files

sparr:
Implementation of #1869
Even if it's not welcome for merge, at least this makes it visible to other people having the same problem I did, with gtk opening a random program (as the spec indicates it should when no default has been configured with gtk tools).

### DIFF
--- a/lutris/game_actions.py
+++ b/lutris/game_actions.py
@@ -1,6 +1,7 @@
 """Handle game specific actions"""
 import os
 import signal
+import subprocess
 from gi.repository import Gio
 from lutris.command import MonitoredCommand
 from lutris.game import Game
@@ -217,7 +218,10 @@ class GameActions:
         if not path:
             dialogs.NoticeDialog("This game has no installation directory")
         elif path_exists(path):
-            open_uri("file://%s" % path)
+            if self.game.runner.system_config.get("use_xdg_utils"):
+                subprocess.run(["xdg-open", path])
+            else:
+                open_uri("file://%s" % path)
         else:
             dialogs.NoticeDialog("Can't open %s \nThe folder doesn't exist." % path)
 
--- a/lutris/sysoptions.py
+++ b/lutris/sysoptions.py
@@ -353,6 +353,14 @@ system_options = [  # pylint: disable=invalid-name
         "advanced": True,
         "help": "Open Xephyr in fullscreen (at the desktop resolution)",
     },
+    {
+        "option": "use_xdg_utils",
+        "type": "bool",
+        "label": "Use xdg utils",
+        "default": False,
+        "advanced": True,
+        "help": "Use xdg-open instead of Gtk.show_uri for browsing, and other xdg utils",
+    },
 ]
 
 


strycore:
Do not add another option, use feature detection if you want to sent that as a PR

TingPing:
>  use feature detection if you want to sent that as a PR

What is there to detect? The freedesktop spec applies to all desktops. `xdg-open` just happens to not follow the spec thus behave differently. Its easy to argue that its wrong.

strycore:
I don't know. I just don't want to add option to work around bugs. That's it.

sparr:
@TingPing xdg-open is part of xdg-utils which is published by the freedesktop group. What spec do you think xdg-open is not following?

sparr:
@strycore I'm not sure how to accomplish this with feature detection. I can't think of a reliable way to detect a case where a user would certainly want this.

TingPing:
Read the source, xdg-open is just hacks to maybe do something. The fact it has the xdg name is just an ironic joke or something.

Lutris is currently correctly implementing the standard. It makes no sense to do anything other than the standard. If your desktop does not implement the standard then it is a bug of your desktop.

sparr:
Hacks? Ironic joke?

https://www.freedesktop.org/wiki/Software/xdg-utils/

The entire purpose of the package is to make it smoother for applications to integrate with various desktop environments, rather than calling lower level libraries that might not be DE-aware.

PS: The features used by KDE to disambiguate lack of defaults are actually mentioned in the spec, but no other DE has adopted them: https://specifications.freedesktop.org/desktop-entry-spec/latest/apb.html (`InitialPreference`, in particular)

TingPing:
The spec is a well defined standard:

- Read foo config
- Fallback to bar config
- Execute desktop file in config found

`xdg-open` is a hack:

- Try kde tools from 1998
- Fallback to gnome tools from 2001
- Fallback to $BROWSER env var
- Fallback to gnome tools from 2013
- Fallback to xfce tools
- Fallback to gnome tools from 2017
- Maybe it works *shrug*?

sparr:
It is not a hack. It's a thoughtfully considered branching decision tree for many different environments the freedesktop devs (the same people who wrote the spec you keep referring to) have encountered in use.

If you are running Gnome and have gio then it will use gio. But if you aren't running Gnome or otherwise haven't configured gio then gio is not the best way to answer the question "what program should open this file/directory/path?".

Your description of trying things in a fallback order is entirely off the mark; there is no situation in which tools will be attempted from that many different environments and eras.

strycore:
Yadda, yadda, I'm just here to make sure that we don't get another checkbox in the system options.

TingPing:
> Your description of trying things in a fallback order is entirely off the mark;

I did exaggerate cross-desktop fallbacks but:

>  there is no situation in which tools will be attempted from that many ... eras.

That is not true:

`gnome-open` -> `gvfs-open` -> `gio open` Are all historical GNOME tools. The current iteration being the final one because it implements the standard we all should.

And I belive `$BROWSER` is always used.

sparr:
@strycore if you have any suggestions about feature detection here I would be glad to implement it. that effort is probably less than maintaining a fork indefinitely.

sparr:
@TingPing The KDE implementation of the freedesktop standard is a superset of the "standard we all should [implement]". Why should they downgrade and give up functionality that predates the spec in question?

strycore:
No, I don't have any suggestion. I just don't want any options added.

TingPing:
I believe the superset that is relevant here is not actually required. KDE can install a desktop-specific default file already that is equal to setting a default browser via another method. If that is mistaken let me know and if I am wrong lets improve the standard (I will contribute such changes to GLib).

sparr:
@strycore how about detecting if no default handler is set (and thus gio will do something unpredictable), and xdg-open is available?

@TingPing yeah, everyone on every side of this issue "can" fix it, but nobody wants to, and they all (including you) think that pointing out that someone else "can" fix it absolves them of responsibility.

sparr:
KDE won't fix it (by writing additional config that KDE doesn't need) because it all works fine in KDE thanks to their extension to the spec.

Gnome won't fix it (by adding support for KDE's extensions to gio) because it's not in the spec.

FreeDesktop won't fix it (by extending the spec) because not enough DE representatives want this functionality.

Arch won't fix it (by shipping a defaults file) because they don't want to force config on users and there's not a good way to manage such a config without knowing what default handlers will be installed.

You won't fix it (by offering an option or using xdg-open by default) because you have fixated on a particular spec and implementation of that spec at the expense of anyone using systems that work differently than that spec.

Everyone passes the buck to someone else.

PS: I reported this same problem to Sublime Text and they might actually fix it on their end for what that's worth, so kudos to them!

TingPing:
I like how you ignored my comment. Please tell me what feature the standard doesn't cover. How is `InitialPreference` not handled by a desktop specific config. And how I offered to improve GLib if it is not.

sparr:
The standard doesn't cover programs specifying their own appropriateness-as-default-handler relative to other programs that can handle the same mime types.

When I have not set a default handler for inode/directory and I install kid3 then launching kid3 is better than nothing (which we might argue about, but I hope we won't since it's what the spec dictates).

When I then install konqueror, the package maintainers for konqueror and kid3 can have agreed in advance that konqueror is more appropriate as an unconfigured default handler for that type, without either package needing to specifically refer to the other by name (or the dozens of other such programs).

sparr:
re: improving glib: https://gitlab.gnome.org/GNOME/glib/issues/1721

Like I said, it's just a big buck-passing circle.

TingPing:
> re: improving glib: https://gitlab.gnome.org/GNOME/glib/issues/1721
> 
> Like I said, it's just a big buck-passing circle.

We will move this conversation there. I'll be busy in the near future sadly but when I have some time I'll reach out to the KDE/desktop-file-spec maintainer and try to see what can be done.

strycore:
Implementing such an option sends the wrong message. It would suggest that it's ok to not follow standards, that it's ok to make application developers write workarounds for potential issues.
I don't care if that should be fixed on Gnome's side or on KDE's side, the only thing I am 200% sure of is that it's not Lutris' responsibility to handle such an issue.

This is a hard NO on my part.

strycore:
Plus I don't get paid enough for this nerd shit.

PR #1884, opened 2019-03-20
Add file_glob command, aliases in input_menu

sparr:
Per discussion on Discord, this PR adds a new installer directive `file_glob` which allows searching the filesystem for files and putting them in an alias to be referenced later. The immediate use case is to locate Vulkan json files that have slightly different naming schemes on different distros.

For testing purposes I also added alias substitution to input_menu keys and values, and that feels like a useful feature to keep. I can take it out, if necessary.


strycore:
I don't want a file_glob directive, what the fuck is this shit?!

Implement globbing on the src arguments of functions that have them, ffs.

Issue #578, opened 2017-06-06
Close Lutris after CLI rungame/rungameid


RobLoach:
When using `lutris lutris:rungame/%`, the Lutris window stays open after the game has quit. Would be great to:

1. Not have Lutris pop up at all
2. Quit Lutris after it was run

Originally reported in https://github.com/RobLoach/lutris-kodi-addon/issues/14 by @solbero . Another solution might be a `--no-gui` option or something.

## Testing

1. Open up your terminal
2. Run `lutris lutris:rungame/tuxracer`
3. Lutris window pops up and runs the game
4. When you quit the game, Lutris stays open.

solbero:
Implementing a feature which is similar in behaviour to Steam’s `steam steam://rungameid/####` would be a boon for people using Lutris on a HTPC. When launching games with the above command, the Steam GUI is never shown on screen.

As @RobLoach points out, as of now the Lutris GUI is left open after the user exits a game. For HTPC usage this is not ideal if the user is using a gamepad for navigation and gaming. As implemented, the user has to use a mouse and keyboard or simulate a mouse with the gamepad to close the GUI or assign ALT+F4 to a gamepad button.

lefty420:
I would also love a no gui mode. Any plans to include such a feature?

strycore:
@lefty420 this specific issue will be fixed in next release, other than that, there are no immediate plans to add other command line only features.

lefty420:
Thanks a lot :) That would make a massive difference when loading from other front ends. Well I can keep the keyboard in the draw at least ;)

arustler:
Awesome! Can’t wait for next release.

sorsasampo:
This would be very nice. Closely related, I think Lutris should immediately exit if the given game does not exist, instead of showing the UI and a traceback in console.

tannisroot:
This has regressed.
I am not sure if it still relevant as the workflow in 0.5.x has changed, but I will add it to 0.5.1 milestone so that we don't forget about it.

seirl:
@tannisroot Could you clarify what you mean by "new workflow"? As you mentioned, it regressed and my scripts now launch a lutris window instead of the game directly. Should I launch my games in a different fashion in 0.5.x?

tannisroot:
@seirl no, I've clarified this with strider and it's definitely a bug. Sorry for any inconvenience this might be causing.

Vedrillan:
[...]

PR #1885, opened 2019-03-20
GH-578 Hide GUI window on rungame command

Vedrillan:
Prevent Lutris window to open when running a game from command line.
The opening window seems to be a regression introduced by commit dabf7ef
which was a revert of commit 7ca5f6c9. Adding back chunks of commit
7ca5f6c9 prevent the Lutris window from opening when running game from
command line.

Also removing a duplicate entry from the .gitignore file and adding
JetBrains IDE projet directory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,8 +9,8 @@ build
 .pydevproject
 .settings
 .ropeproject
+.idea
 tags
-.ropeproject
 *.pyc
 *.pyo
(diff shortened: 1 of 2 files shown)


strycore:
Looks simple enough that even if there was something wrong with this PR, it would be an easy fix.
Requires QAing on my end but this looks good and ready to merge.

PR #1904, opened 2019-03-22
Change default icon folder to 128x128

### LINKED ISSUE #1903: Higher resolution icons are still saved in 32x32 directory
Despite Lutris bumping its icon size to 128x128 (currently it's still 64x64 as bump to 128x128 hasn't been deployed yet), it still saves the icons into `icons/hicolor/32x32`.
It should probably not do that and save it in proper 128x128 location instead.

tannisroot:
Fixes #1903
Btw, I noticed that even without my changes, I always see `gtk-update-icon-cache: The generated cache was invalid.` pop up in the log. Is that normal?


strycore:
> Btw, I noticed that even without my changes, I always see `gtk-update-icon-cache: The generated cache was invalid.` pop up in the log. Is that normal?

I've seen that as well, probably not related to lutris

PR #1922, opened 2019-03-24
Improve "Please support Lutris!" text

theDrake:
Remove extra space before "like," remove -ing from "cover" and "fund," and break final sentence into two sentences.


strycore:
Thanks!

PR #1928, opened 2019-03-24
Add detection of custom proton builds in compatibilitytools.d folder,…

GloriousEggroll:
… as documented here: https://github.com/valvesoftware/proton#install-proton-locally


GloriousEggroll:
P.S. hi strider, I'm still alive ;p

Riesi:
Your search for folders containing the name `Proton` is case sensitive, which resulted in `proton_tkg_4.4.r7.gd963a250` not being found. Maybe you can change it to a case insensitive search.

strycore:
I won't test this has I don't have a use case but looks good enough to merge!

Issue #1936, opened 2019-03-25
Error in Lutris 0.5.1.1


eNuKeD:
Yesterday I was playing Overwatch without problems and today with the new update, I have a problem when starting the game:

```
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/lutris/exceptions.py", line 26, in wrapper
    return function(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/lutris/game.py", line 299, in configure_game
    gameplay_info = self.runner.play()
  File "/usr/lib/python3/dist-packages/lutris/runners/wine.py", line 832, in play
    launch_info = {"env": self.get_env(os_env=False)}
  File "/usr/lib/python3/dist-packages/lutris/runners/wine.py", line 718, in get_env
    env = super(wine, self).get_env(False)
  File "/usr/lib/python3/dist-packages/lutris/runners/runner.py", line 149, in get_env
    runtime_env = self.get_runtime_env()
  File "/usr/lib/python3/dist-packages/lutris/runners/wine.py", line 744, in get_runtime_env
    if entry in wine_path:
TypeError: 'in <string>' requires string as left operand, not NoneType
```

The play button, stay stuck on "Launching".
This happend with all games from wine runner.

![gGmiPI3](https://user-images.githubusercontent.com/5289911/54913999-4cdc3f00-4ef4-11e9-9ffd-6d6c61dd512c.png)

EzraT:
Can confirm, this also happens to me on Linux Mint 19.1.

eNuKeD:
I also use Linux Mint 19.1

kuchenohnegraeten:
The same happens to me on Ubuntu 18.04 with wine games / wine launchers (Epic Launcher, Battle.Net, Anno 1404 (GOG)), hanging with "Launching...". Here the error output from the command-line-started lutris:
`Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/lutris/exceptions.py", line 26, in wrapper
    return function(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/lutris/game.py", line 299, in configure_game
    gameplay_info = self.runner.play()
  File "/usr/lib/python3/dist-packages/lutris/runners/wine.py", line 832, in play
    launch_info = {"env": self.get_env(os_env=False)}
  File "/usr/lib/python3/dist-packages/lutris/runners/wine.py", line 718, in get_env
    env = super(wine, self).get_env(False)
[...]

PR #1937, opened 2019-03-25
None check when looking for PROTON_PATH

Lrss:
Broke wine runner according to #1936


Lrss:
The error is happening if get_proton_custom() in lutris/util/wine/wine.py can't find any custom versions of proton in the various steamapps folders.

strycore:
Please look at https://github.com/lutris/lutris/commit/fa8907f54ae6c0a14764a07c3366ab0fc6422a2c for the final implementation.

AaronOpfer:
I don't think a `None` check here is wise. We shouldn't have non-strings inside `PROTON_PATH` and I'd like to see the problem fixed there.

PR #1947, opened 2019-03-26
GH-578 Hide GUI window on rungame command

Vedrillan:
Instead of showing the window unconditionally in the do_activate method,
move the window.present() into the install action, the full behavior
should then be:
* not show the window for rungame/rungameid action, unless the provided
game is not installed
* show the window in any other cases


strycore:
This is still not the right way to solve this issue.

Issue #1395, opened 2018-12-07
Add search to logwindow.py


AccountOneOff:
Instead of scrolling endlessly I want to be able to find errors fast with searching. Like ctrl+f in a browser.

strycore:
you that's not shipping in the next release, right?

AccountOneOff:
No, no I didn't. What is?

strycore:
That's just a small test program I coded to get the GOG API working. Eventually, all the features in that file will be moved to the import games window (which *will* have search!)

AccountOneOff:
I'm talking about https://github.com/lutris/lutris/blob/master/lutris/gui/logwindow.py which is called logdialog.py now I guess.

strycore:
I totally misread that as gogwindow (because the request also made sense for it)

strycore:
The PR implements the feature very nicely, it would be nice to add a search icon in the title bar because not everyone will know about the ctrl + f shortcut but the search itself is good now!

PR #1959, opened 2019-03-28
Implemented search feature in LogWindow (#1395)

a0kami:
Hi,
I may be new around here but I wanted to give it a shot anyway.

I wasn't sure about the coding conventions, so I used exclusively spaces and followed snake_case syntax to stay consistent with the code in place, I just didn't set up the lint yet. :disappointed: 

But here's a proposition of an implementation of the search feature for the `LogWindow` class in order to adress issue #1395 .
It massively relies on the Gtk implementation to do the actual search, using the various signals of `Gtk.SearchEntry` but also the `Gtk.TextIter` `forward_search` and `backward_search` functions while using `Gtk.TextMark` because the buffer is constantly mutated and we can't rely on `Gtk.TextIter` alone.

The event part stays in the view/controller part (`LogWindow(Gtk.ApplicationWindow)`) and the search/algorithmic part stays in the model part (`LogTextView(Gtk.TextView)`).

There is just a slightly front-end correction to add so highlighted text near the edges of the window stay visible and maybe better support the already existing autoscroll function but aside from it should be 100% functional.

**How-to**:

- **Ctrl + F** on log window brings the `SearchEntry`
- **Escape** while there's the `SearchEntry` hides it
- Typing in the `SearchEntry` searches for the first occurrence
- **Ctrl + G** (default) and **Return** hops to next occurrence (while in `SearchEntry`)
- **Ctrl + Shift + G** (default) and **Shift + Return** hops to previous occurrence (while in `SearchEntry`)

Finally, I wasn't sure if I should have added a `Gtk.SearchBar` and/or `Gtk.Button`'s but there's definitely room for improvement.

### DIFF
--- a/lutris/gui/dialogs/log.py
+++ b/lutris/gui/dialogs/log.py
@@ -1,5 +1,5 @@
 """Window to show game logs"""
-from gi.repository import Gtk
+from gi.repository import Gtk, Gdk
 from lutris.gui.widgets.log_text_view import LogTextView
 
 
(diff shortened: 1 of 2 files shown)


strycore:
Looks good at first sight, so I think this is good for a merge! We still have time to include some improvement before the next release.

a0kami:
I missed one test (using len to check empty string) and had a typo though.
Sorry about that. :worried: 

But thanks a lot!
If we could all further discuss on the issue page about what people would like ?

strycore:
it's all good, the code is really clean and this the kind of PR that makes me really happy!
I suggested adding an icon on the title bar to toggle the search but ctrl+f is widespread enough for this to be enough at the moment.

PR #1968, opened 2019-03-30
Newly designed library search mechanism based on ldconfig.

aTosser:
This approach will solve the runtime not being able to find where libraries are located by using the ld.so.cache. This way we can rely on the system to know where it's own libraries are located instead of needing hard coded paths.

### DIFF
--- a/lutris/util/linux.py
+++ b/lutris/util/linux.py
@@ -273,7 +273,6 @@ class LinuxSystem:
                 # On non amd64 setups, only the first element is relevant
                 lib_paths = [lib_paths[0]]
             #if all([os.path.exists(path) for path in lib_paths]):
-            print(lib_paths)
             yield lib_paths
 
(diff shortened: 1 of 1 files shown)


aTosser:
The test failure noted here have nothing to do with the changes that I made to the codebase.

strycore:
Yeah, I'm really not sure about such a risky change and going 100% with ldconfig. This might introduce some kind of major issue down the line. That said, we're far enough from the next release that we can get this well tested by the time of the release. 

The PR itself looks good and looks like it's doing the right thing. Going to merge it so we can catch any issue and refactor some of the code.

aTosser:
Let me know if there is anything you would like me to modify for style for refactoring.

strycore:
the style is ok but we have a lot nesting going on there, I think it's time to decompose some of the stuff into smaller units

PR #1982, opened 2019-04-02
ldconfig correction

aTosser:
Now including all lib folders detected on system as well as keeping ordering for duplicates for distros that care.


AccountOneOff:
Tested and working.

PR #1989, opened 2019-04-04
iteritems() does not exist in python3 use items() instead

Riesi:


strycore:
Thanks, but I already commited the fix

PR #2017, opened 2019-04-09
BlastEm got libretro support

telanus:
https://www.retrodev.com/repos/blastem/rev/8500bd987dda


RobLoach:
We're working on getting the new upstream core on buildbot. In the mean time, this applies nicely. Thanks!

PR #2041, opened 2019-04-15
add options to install from remote script (download to tmp)

yucefsourani:
I hope this option is added because it helps a lot
ex:
lutris -r https://raw.githubusercontent.com/yucefsourani/my_lutris_scripts/master/scratch333.yml

### DIFF
--- a/lutris/gui/application.py
+++ b/lutris/gui/application.py
@@ -22,6 +22,8 @@ import signal
 import sys
 import gettext
 from gettext import gettext as _
+import requests
+import tempfile
 
 import gi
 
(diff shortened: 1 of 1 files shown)


strycore:
@yucefsourani I like the feature but can you please change the code so that it doesn't introduce a new option? Lutris should be able to determine automatically when using the -i flag that the file is either local or remote.

lutris/gui/application.py
```diff
@@ -113,6 +115,10 @@ def add_arguments(self):
             "install", ord("i"), GLib.OptionFlags.NONE, GLib.OptionArg.STRING,
             _("Install a game from a yml file"), None,
         )
+        self.add_main_option(
```
strycore: This shouldn't be needed, you can adapt the "install" option to detect http links

strycore:
Totally missed the updated PR, sorry for that! Going to merge this.

PR #2048, opened 2019-04-16
lutris/command: use pool to notify on exit

szekelyszilv:
This pull request aims to improve the exit detection of MonitoredCommands.

The patch changes it to use to use a ThreadPoolExecutor, where a task is inserted to wait for the process to finish, notifying the main thread when wait() returns. This replaces listening for the SIGCHLD signal and dispatching the exit notification from there.

The motivation for this pull request is that I have experienced an issue with installing games. After either the installer or winetricks exited the installation did not proceed to the next step, just waited indefinitely. This patch let me finish the installation process.

Platform where the above issue was noticed:
- Lutris 0.5.2.1
- Python 3.7.3
- GTK 3.28.4
- ArchLinux (latest updates as of 2019-04-16)

I tried testing it on Ubuntu as well, but the original issue didn't surface there. The modifications still let me finish the installation process.


strycore:
@AaronOpfer can you please take a look at this PR and see if it looks ok from your side?

AaronOpfer:
I am uncomfortable with making this change without reliable reproduction steps of the OP's problem. I think adding threads to an application usually causes more problems than it solves and I don't see how threads can solve this particular problem better than asynchronous signals can.

If the OP or someone else can provide:
  * a unittest test case,
 *  test lutris game or windows EXE,
 *  or a .py file that imports `lutris.command` and shows the bad behavior

 I'll be happy to reconsider.

szekelyszilv:
Thank you for the feedback!

I will try to create a testcase or find an alternative solution without threads. In case the problem only manifests with a certain combination of dependency versions, what would be the most convenient for you, how should I provide a test environment?

AaronOpfer:
First of all, thank you for your interest in improving Lutris for everyone.

> I will try to create a testcase or find an alternative solution without threads.

Sorry, but an alternate solution doesn't appeal to me until we can demonstrate that something is wrong via a test case or reproduction steps. I appreciate you're having an issue with lutris that this fixes; but as far as we know other people don't have this problem, and making changes to this code could break lutris for many more people who currently have lutris working for them. Until we know exactly what problem, I am not interested in changing the code. I can't speak for @strycore but I can imagine him feeling the same way.

> In case the problem only manifests with a certain combination of dependency versions, what would be the most convenient for you, how should I provide a test environment?

If the dependencies are Python libraries, you can list them out using `pip freeze`. That, combined with `python --version` might be enough for us to create a virtual environment that matches yours closely. If that doesn't work, maybe we can muck around with an Arch VM that mirrors your setup closely, although I'm unfamiliar with arch and am not sure how fast I can accomplish something like that. :slightly_smiling_face: 

If you can provide this and some basic [reproduction steps](https://secure.phabricator.com/book/phabcontrib/article/reproduction_steps/), that should be enough to look into the problem and consider your pull request.

szekelyszilv:
I have looked into the issue more, and it seems to be cause by the WebKitGTK login window. The latest stable version of WebKitGTK installs signal handlers when it initializes, including SIGCHLD, which override the ones set by Lutris.

# Reproduction steps
- I started with a fresh install of Ubuntu 18.04
- Compile and install WebKit2GTK 2.24.1 (if not available in the repos)
- Start Lutris where GOG is not connected
- Start to install any Wine GOG game
- GOG login window pops up (WebKitGTK gets initialized)
- Cancel and select an .exe file
- Lutris starts installing and hangs after the first task, as signals are no longer delivered

If I modify the GOG service not to open a WebKitGTK dialog then the installation proceeds as normal, as the signal handlers are not replaced.

I have not yet looked into when this change was made, or why, or whether it is possible to switch it off.

Reinstalling the signal handler when starting a new MonitoredCommand also seems to solve the issue.

## WebKitGTK build steps used
```
wget https://webkitgtk.org/releases/webkitgtk-2.24.1.tar.xz
tar -xf webkitgtk-2.24.1.tar.xz
cd webkitgtk-2.24.1
sudo apt install -y build-essential
sudo Tools/gtk/install-dependencies
sudo apt install -y libopenjp2-7-dev libwoff-dev
cd ..
cmake -Hwebkitgtk-2.24.1 -Bbuild \
-DPORT=GTK \
-DCMAKE_BUILD_TYPE=Release \
-DCMAKE_INSTALL_PREFIX=~/.local \
-DCMAKE_SKIP_RPATH=ON \
-DENABLE_GTKDOC=ON \
-DENABLE_PLUGIN_PROCESS_GTK2=OFF \
-DLIBEXEC_INSTALL_DIR=~/.local/lib \
-DLIB_INSTALL_DIR=~/.local/lib
cd build
make -j4
make install
```

It failed at a point for me, I had to run `make JavaScriptCore-4-gir` and then make again.

AaronOpfer:
Thank you @szekelyszilv, this makes the entire issue very clear now.

Since messing with signals is essentially modifying global state, I think it's apparent that the `SIGCHLD` method is inadequate and needs to be replaced.

But, instead of a threadpool, can you modify your PR to use [`Glib.child_watch_add`](https://developer.gnome.org/pygobject/stable/glib-functions.html#function-glib--child-watch-add)? It looks like this would allow us to get that asynchronous callback we're looking for without spawning a thread. The fewer threads in lutris the better IMO.

Thanks again for hunting down the details around this issue!

AaronOpfer:
LGTM :+1:

szekelyszilv:
Thank you for pointing me towards `child_watch_add`! It does indeed look better this way. :)

AaronOpfer:
@strycore this is good to merge IMO unless there is something you still want to see from it?

Issue #1948, opened 2019-03-27
Refresh desktop shortcut and app shortcut items


AccountOneOff:
When pressing "Create desktop/app shortcut" it won't change to "Delete desktop/app shortcut" until clicking off the game entry.

You can see here it didn't change when I clicked Create app shortcut:

https://streamable.com/0wj1y

dino8890:
Hi,

I took a shot at this and created a decorator for the action callback which toggles the visibility between create and delete shortcut buttons. Do you think this is a proper way to go? I can submit a pull request if you want to review the changes.

Dino

strycore:
@dino8890 I'd like to take a look at it but I'm not sure this is the way I want to go with things.

Since we handle the action in multiple place (contextual menu, sidebar) I think it's better to go with a solution using GObject signals. We've increased the usage of signals in 0.5.x and it has solved a lot of problems we previously had.

dino8890:
I rewrote it to use signals and you're right, it's better this way. I don't know why I didn't do it like this the first time.

Here's the relevant [pull request](https://github.com/lutris/lutris/pull/2056).

PR #2056, opened 2019-04-17
Refresh functionality when clicking on create/remove shortcut button

dino8890:
I took at stab at the issue #1948 to familiarise myself with lutris code base(first time working on lutris) as this seemed relatively simple thing to do.

This solution achieves the original purpose, but creates one problem, clicking the same option in the context menu doesn't update the buttons on the panel.

@strycore I'm willing to devote some time on this and rewrite it using GObject signals, it will probably also solve the context menu problem.

Pardon my code style and other errors, as I'm creating this PR in order to get feedback.

### DIFF
--- a/lutris/gui/views/game_panel.py
+++ b/lutris/gui/views/game_panel.py
@@ -149,13 +149,6 @@ class GamePanel(GenericPanel):
                 buttons[name] = button
         return buttons
 
-    def on_shortcut_edited(self, widget, action_id):
-        self.buttons[action_id].hide()
-        if 'rm' == action_id[0:2]:
-            self.buttons[action_id[3:]].show()
-        else:
-            self.buttons['rm-' + action_id].show()
-
     def place_buttons(self, base_height):
         play_x_offset = 87
         icon_offset = 6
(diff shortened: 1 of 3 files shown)


dino8890:
In menu.py on line 34
```
        action.connect("activate", lambda a: self.emit('shortcut-edited', name))
```
Do you think it's better to use lambda here or to create a method and then call emit from that method? 

Another question, should we create different handler for different shortcuts(menu vs desktop), different actions(create vs remove), both or leave it as it is? They would essentially do the same thing, I am just not sure which is the cleanest.

strycore:
I'd rather have a method than a lambda, those are rather hackish and having a named function is more explicit. 

If there's a way to pass parameters to the callback (menu / desktop, create / remove) then it should be possible to have just a single function to handle all cases.

lutris/gui/views/game_panel.py
```diff
@@ -201,6 +194,13 @@ def place_buttons(self, base_height):
 
             self.put(button, position[0], position[1])
 
+    def on_shortcut_edited(self, widget, action_id):
```
dino8890: I just moved it further to down to group it with other on_<something> methods.

dino8890:
I'm not sure I understand your last statement. Currently there is a single method which handles all cases in GamePanel
```
    def on_shortcut_edited(self, widget, action_id):
        self.buttons[action_id].hide()
        if 'rm' == action_id[0:2]:
            self.buttons[action_id[3:]].show()
        else:
            self.buttons['rm-' + action_id].show()
```
Is this what you meant?

strycore:
I don't think the method I'm thinking about exists currently. the on_shortcut_edited method looks suitable for that.

dino8890:
If you're happy with the result then that's it, if you're not let me know so I can make additional changes.

strycore:
looks good!

PR #2072, opened 2019-04-21
Rename zh-tw.po to zh_TW.po

laichiaheng:
It's zh_TW, instead of zh-tw


laichiaheng:
My Lutris client is still in English, please merge it.

PR #2076, opened 2019-04-22
Expose environment variables to scripts

TiZ-EX1:
Hi there, hopefully this is a pretty minor change that won't have much friction. The pre-launch and post-exit scripts are very useful, but I would like to be able to use environment vars defined per-game > per-runner > global to influence the things that the scripts do. That way I can use existing GUI configuration for my specific use cases.

I did add another change as well: the name of the game being run is exported into the environment. That way, scripts can do intelligent things like adding the game's name to screenshot or recording filenames.


strycore:
That variable name is gonna get changed really fast but otherwise, I do like the idea!

Issue #1625, opened 2019-01-18
add gallium nine standalone for option


Meklahi:
[https://github.com/dhewg/nine](url)

Gallium Nine Standalone, as the name implies, is a standalone version of the WINE parts of Gallium Nine.
This decouples Gallium Nine from the WINE tree, so that it can be used with any WINE version. There is no need for any WINE patches. A stable, development, or staging WINE release is sufficient.
WINE can be updated independent of Gallium Nine Standalone. Users can mix releases of both projects to their liking. Switching between staging and non-staging does not require a rebuild.

romulasry:
This would be nice to have. A toggle button would be cool.

strycore:
the problem is getting a gallium9 build, once that is sorted out, it's easy to add the option in lutris.

dhewg:
Oh nice, just stumbled upon this issue!

The project moved: https://github.com/iXit/wine-nine-standalone
Since v0.2 we also provide binary releases: https://github.com/iXit/wine-nine-standalone/releases

Those can be installed by the bundled script, see https://github.com/iXit/wine-nine-standalone#usage
Or you can do it by hand, like winetricks: https://github.com/Winetricks/winetricks/commit/2a4ec13d86132fbe1c69afa75bd0981f45f15af0

Does that work for you guys?

dhewg:
Is there anything that needs to be improved for the standalone version to get this into lutris?

sonicpp:
@strycore Do you want to provide custom build or rely on gallium-nine provided by distribution package (like Lutris already does with Feral gamemode and FPS limiter for example)?

I can start working on this if you are busy with other features...

strycore:
@sonicpp yeah, I have very little experience with Gallium nine so that would be very helpful. We build wine with this script https://github.com/lutris/buildbot/blob/master/runners/wine/build.sh but this requires a specific setup (it's explained in the readme of the buildbot repo)

I'll push that to 0.5.3 to give more time but if we get progress made earlier this might be ready for 0.5.2 (or maybe there's no work to be done on the client side at all and a custom build will be enough)

sonicpp:
Hello,

So I have finally time to work on this.

At the moment I rely on gallium-nine provided by distribution, but later we can provide custom build as part of Lutris.

So far I investigated following (I already coded simple implementation and tested it on my localhost for 32 bit prefix):
[...]

PR #2092, opened 2019-04-26
RFC: Gallium Nine

sonicpp:
This pull request adds Gallium Nine option to Wine runner, as requested in #1625 

So far this option rely on Gallium Nine standalone provided by system package (like Feral gamemode), but we may change it to pack it with Lutris or let it download (like DXVK) from Gallium Nine standalone Github project in the future.

Gallium Nine can be enabled if:
* MESA is built with gallium nine state tracker support (_d3dadapter9_ library is found)
* Gallium Nine standalone is installed on the system (_d3d9-nine.dll_ and _ninewinecfg.exe_ binaries are found)

First condition is checked every time the configuration window is opened. I tried to add info about MESA support into `LIBRARIES` in `util.linux.SYSTEM_COMPONENTS` and check Mesa for Gallium support by simple calling `system.LINUX_SYSTEM.is_feature_supported("GALLIUM_NINE")`, but _d3dadapter9_ is not detected by `ldconfig -p`. If we want to do it like this, we need to manually check system library paths (for example by _linux.iter_lib_folders()_).

Gallium Nine standalone provides _ninewinecfg.exe_, which is used for enabling and disabling Gallium Nine itself for active Wine prefix. But we can do it manually, as described in #1625

There is one problem - when you enable Gallium Nine on GPU which does not support it, game will not start. For example my setup - dedicated AMD GPU with MESA driver supporting Gallium Nine - everything works fine unless I switch GPU to integrated Intel. Probably the only way how to detect if chosen GPU supports Gallium Nine or not is by calling some executable which will try to initialize D3D9 device by calling _Direct3DCreate9Ex_. This detection is not part of this pull request - we will need to create custom exe file (or kindly ask GN standalone guys to add command line option to _ninewinecfg.exe_ for this?)

### DIFF
--- a/lutris/runners/wine.py
+++ b/lutris/runners/wine.py
@@ -739,7 +739,10 @@ class wine(Runner):
         self.sandbox(prefix_manager)
         self.set_regedit_keys()
         self.setup_x360ce(self.runner_config.get("x360ce-path"))
-        self.setup_nine(self.runner_config.get("gallium_nine"))
+        try:
+            self.setup_nine(self.runner_config.get("gallium_nine"))
+        except nine.NineUnavailable as e:
+            raise GameConfigError("Unable to configure GalliumNine: " + str(e))
         try:
             dxvk_version = self.runner_config.get("dxvk_version")
             self.toggle_dxvk(
--- a/lutris/runners/wine.py
+++ b/lutris/runners/wine.py
@@ -17,6 +17,7 @@ from lutris.util.graphics.vkquery import is_vulkan_supported
 from lutris.util.wine.prefix import WinePrefixManager
 from lutris.util.wine.x360ce import X360ce
 from lutris.util.wine import dxvk
+from lutris.util.wine import nine
 from lutris.util.wine.wine import (
     POL_PATH,
     WINE_DIR,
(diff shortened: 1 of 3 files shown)


lutris/runners/wine.py
```diff
@@ -725,6 +739,7 @@ def prelaunch(self):
         self.sandbox(prefix_manager)
         self.set_regedit_keys()
         self.setup_x360ce(self.runner_config.get("x360ce-path"))
+        self.setup_nine(self.runner_config.get("gallium_nine"))
```
strycore: put this in a try / except with an exception handler please

sonicpp: Exception handler added

soredake:
Any progress on this?

sonicpp:
I am waiting for feedback from @strycore

sonicpp:
I see there is a conflict now (not unexpected - I wrote this PR several months ago). If anyone could review this PR, I can fix conflict and/or update PR if necessary. Until then I will keep it untouched, since there is no interest in this PR from the Lutris devs ATM...

strycore:
Going to merge this manually since the conflicts shouldn't be too bad and we'll have this feature out in Lutris 0.5.4

soredake:
@sonicpp hi, tried lutris with this pr merged, gallium nine option is grayed out, i am using mesa-git from tkg.

sonicpp:
> @sonicpp hi, tried lutris with this pr merged, gallium nine option is grayed out, i am using mesa-git from tkg.

Hi, thanks for report. I will test it tomorrow

sonicpp:
@soredake: After quick checking - yes, there is an issue in detecting Nine. I created a bug #2332
Thanks

Issue #701, opened 2017-12-01
Distribute Lutris in a Flatpak


LeandroStanger:

RobLoach:
Was Gnome Software going to have a Flatpak browser or something? Think something in Lutris would work like.......

1. Lutris Flatpak Runner  requires pre-installation of Flatpak
2. Lutris game installer defines Flatpak remote and flatpak ref URL
3. Instlaling would make the Runner trigger Flatpak remote add and game install

We'd also have to have the Flatpak runner handle uninstall too.

strycore:
@RobLoach that ticket is for distributing Lutris itself in a flatpak, we'll also support installing Flatpak games but that will come later.

RobLoach:
Ah, cool!

TingPing:
As I've mentioned in the past you can currently only choose one of those and IMO the games are more valuable of the two.

Saroufim:
Lutris as a Flatpak cannot detect Steam (Flatpak as well). If there is a choice to be made between providing Lutris as a Flatpak app or having Lutris as a Flatpak games manager, which of those two will the development team choose? Or will this be sorted out somehow by having both?

TingPing:
**I** think having a non-flatpak Lutris launch Flatpak runners is the far more valuable first step that allows the custom lutris runtime and all its problems to be removed. Now I have not necessarily looked into the complexities of this but it seems doable.

strycore:
@LeandroStanger I don't know, I don't work on Flatpak related features.

Currently, the focus is on better moderation tools and support for GOG, at least as far as I'm concerned.

RobLoach:
Had an attempt at a [Lutris Flatpak](https://github.com/robloach/lutris-flatpak). While it compiles, still run into runtime errors during execution. Would appreciate a look!

Saroufim:
@TingPing has a sort of working flatpak package on his repository. It runs but it can't find its own runtimes.

AsciiWolf:
Any news?

RobLoach:
No... @LeandroStanger and I have a pretty good proof of concept to run [Lutris in Flatpak](https://github.com/robloach/lutris-flatpak). No work has been put into having a Flatpak Runner in Lutris.

LeandroStanger:
https://github.com/flathub/flathub/pull/797

LeandroStanger:
https://github.com/flathub/flathub/pull/919

LeandroStanger:
[...]

PR #2106, opened 2019-05-04
Flatpak support

gasinvein:
This is a set of patches needed to run Lutris inside Flatpak sandbox.
I'm not familiar with Lutris internals, so feel free to correct me.

This should solve following problems (specific to running in flatpak):
- [x] Steam binary can't be launched directly, so use `steam://` URLs where possible and let xdg handle it.
- [x] Some warnings are irrelevant, e.g. the ones related to dependencies 

For flathub/flathub#926 and #701. Closes RobLoach/net.lutris.Lutris#13.


gasinvein:
Are there any plans for merging this, or should I change something?

strycore:
Should be good to merge. I havent tested it yet but the client isn't affected by that PR so it's safe.

PR #2108, opened 2019-05-05
Snap package, a step foward

MrTheSoulz:
This still fails to actualy install any game and refuses to run in strict mode but at least it opens :)
i belive the not installing games part is beacuse i probably missed a lutris-runtime dependency.

to complile localy use the commands:
* snapcraft try
* snap try time

then to debug run with:
* lutris.lutris

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1,6 +1,14 @@
 name: lutris
-version: v0.5.1.3
-summary: Lutris
+version: git
+confinement: devmode #chaning to strict causes errors trying to read nvidia version file, need to figure it out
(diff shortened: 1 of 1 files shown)


snapcraft.yaml
```diff
@@ -1,59 +1,126 @@
 name: lutris
-version: v0.5.1.3
-summary: Lutris
+version: git
+confinement: devmode #chaning to strict causes errors trying to read nvidia version file, need to figure it out
```
strycore: it's fine, leave it in devmode until the snap is stable

PR #2116, opened 2019-05-07
Kill Processes on Install Cancellation

### LINKED ISSUE #2084: Lutris doesn't completely kill Wine processes in the prefix after cancellation of the installation
It's not uncommon for me to get "creating prefix" messages just after cancellation of the game installation and see active processes that were launched during installation process in System Monitor (which are what triggering prefix re-creation after Lutris deletes it).
And same happens if the installation is not cancelled (which might caused issues with WINEESYNC option).
Perhaps Lutris should use `wineserver -k9` for this.

matthewkovacs:
Attempt to fix #2084


lutris/gui/installerwindow.py
```diff
@@ -481,6 +482,7 @@ def cancel_installation(self, widget=None):
             }
         )
         if confirm_cancel_dialog.result != Gtk.ResponseType.YES:
+            Popen(['wineserver', '-k9'], shell=True)
```
strycore: don't use shell=True on subprocess calls

matthewkovacs:
I'm assuming the location of the wineserver executable as `/usr/bin/wineserver` to address security concerns here.  That path might not always be correct.  If I attempt to run `locate wineserver` first I would just run into the same problem of guessing the full path of the `locate` executable.

I'm not sure what the best practice is here.

tannisroot:
@matthewkovacs This PR assumes that Wine is installed on the system but it might not be. Why not call `wineserver` that is shipped with Lutris builds?

tannisroot:
This doesn't even work...

Issue #826, opened 2018-04-14
Discord integration


fosspill:
Since it seems to be possible to add "sub titles" to running games in Discord now I was thinking it would be interesting to know if it would be possible to add this for games ran through lutris.

As in, it would show you as playing "lutris", but if someone clicked you they would see "World of Warcraft" as the sub title. Reason for this is of course that Discord + wine games doesn't exactly go hand in hand as of now.

Picture of the mentioned feature:
https://runelite.net/img/features/discord.png

strycore:
I'll have to double check the Discord API but I think there's a way to set the "Now playing" status with HTTP requests. This would be preferable to just having "User is playing Lutris".

strycore:
@RobLoach Our Discord channel is hosted by the Linux Gamers Group: https://discord.gg/GC8vA6C 

Also, I've had a look at Discord.py a few months back, it's neat but I wouldn't want to add this as a mandatory dependency, we should maybe think about adding a plug-in system (or just make the feature present ootb but only enabled when requirements are available)

mattjraives:
I've been looking in to this myself for a couple of days; I think that [pypresence](https://github.com/qwertyquerty/pypresence) is probably a better solution than [discord.py](https://github.com/Rapptz/discord.py).  Discord.py, from what I can tell, is more about writing bots, and while you *can* use a "self-bot" (i.e., run a bot on your user account) using discord.py, self-bots are [officially against the Discord TOS and can result in a ban](https://support.discordapp.com/hc/en-us/articles/115002192352-Automated-user-accounts-self-bots-).

pypresence, on the other hand, is a wrapper specifically for discord rich presence, and so shouldn't have these issues.

Feichtmeier:
@strycore 

when launching world of warcraft the battle.net app is started
discord shows "is playing battle.net login" :+1: :+1:  :+1: 
after I start World of Warcraft discord shows "is playing lutris"

Is there anything else I need to set up to make this work with WoW?
Thank in advance

PR #2125, opened 2019-05-10
Implementation of Discord Rich Presence, along with some typo/flake8 …

isugimpy:
…fixes

Got the wild urge to write this for myself, figured it might be useful to the community.  Turns out there is/was an open issue for a feature request on it a while back  https://github.com/lutris/lutris/issues/826

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ before_install:
   - "/sbin/start-stop-daemon --start --quiet --pidfile /tmp/custom_xvfb_99.pid --make-pidfile --background --exec /usr/bin/Xvfb -- :99 -ac -screen 0 1280x1024x16"
   - sleep 3 # Give xvfb some time to start
 install:
-  - pip install --upgrade pip
+  - pip install --upgrade pip pypresence
   - sudo apt-get -qq update
   - sudo apt-get install -y python3-yaml python3-gi python3-pil gir1.2-gtk-3.0 psmisc gir1.2-glib-2.0 libgirepository1.0-dev gir1.2-gnomedesktop-3.0 gir1.2-webkit2-4.0 gir1.2-notify-0.7
 script: nosetests
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ before_install:
   - "/sbin/start-stop-daemon --start --quiet --pidfile /tmp/custom_xvfb_99.pid --make-pidfile --background --exec /usr/bin/Xvfb -- :99 -ac -screen 0 1280x1024x16"
   - sleep 3 # Give xvfb some time to start
 install:
-  - pip install --upgrade pip pypresence
+  - pip install --upgrade pip pypresence~=3.3.2
   - sudo apt-get -qq update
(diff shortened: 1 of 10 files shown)


isugimpy:
Apologies for the messiness here.  I'm closing this back out for now.  Apparently pypresence isn't 3.5 compatible.  I'm going to see what I can do about this.

strycore:
no need to close it :) you can say when it's ready for review, we'll leave it here until then!

isugimpy:
@strycore Thanks!  Reopened.  I got the pypresence maintainers to release a newer version of the package, it was actually unintentional that the previous release wasn't 3.5 compatible.  This code has been tested pretty extensively locally over the last couple days and seems to work very well.  Not certain that it meets all the design considerations of the project, but I think it's ready for review.

For clarification of intent:

- Allows users to set their presence on Discord automatically
- Uses a default app ID (which I've created, feel free to replace it) that presents a game name of "a game through Lutris"
- Uses the Lutris game name by default
- Provides a secondary detail which indicates the runner the game is running under
- Allows you to override per-game the app ID, game name, and visibility of the runner
- Allows you to disable showing the runner and the Discord update entirely

Example attached below:
Screenshot_20190509_214613

PythonicChemist:
Perhaps extracting the Discord Rich Presence feature out of the Game class would be good since the Game class is already pretty big and the feature would add another 100+ lines to it.

Awesome feature and I am sure many people on Discord are looking forward to it. :+1:

isugimpy:
On the one hand, I agree with you because it increases file complexity significantly as is, and it'd make it significantly more testable. On the other, it's a very integration-heavy feature, and it seems to me that it's a behavior of the game class itself. The game is notifying Discord that it's running and no longer running. I also don't see a situation where anything else in the codebase would consume this code, which reduces the value of extracting it into its own class. I'll do what's best for the project and desired by the maintainers, at the end of the day, but I just don't know if the time investment of that provides real gain.

telanus:
Just a question, how will this work with systems that already feature discord's DRP? IE MAME, PPSSPP & Retroarch?

isugimpy:
You just disable it on those. Things like that are one of the big reasons I offered a toggle for it. :)

isugimpy:
@strycore I think the latest two commits address your review.  Added the dependency to extras_require.  The config options get hidden if the dependency isn't present on the system, and it just logs that the feature isn't enabled.

strycore:
Nicely done! Is the default client id something I need to change?

isugimpy:
@strycore You don't necessarily NEED to, I have no intention of changing any settings on it. But if you want to, absolutely feel free. It might be wise to have the ID owned by a member of the team, in case any additional features get added on the Discord side which could be taken advantage of. I'm hoping this is the first of many contributions by me to the project, though, so I have no problem with holding onto the ID. Basically, do what the team feels makes sense. :)

strycore:
Thanks for clarifying. As you said, this would be better to have as an asset owned by the team. I'll create an app before shipping this feature.

PR #2131, opened 2019-05-12
Add background

LeandroStanger:
The Background is used in the App Center of Endless OS
An example image
Captura de tela de 2019-05-11 20-02-07


strycore:
The link for the image is still not correct

TingPing:
I suggest reverting this. Its low quality, not widely used, and not clearly licensed.

strycore:
yeah and the image shouldn't be hosted on github anyway

PR #2137, opened 2019-05-16
Add System Info (Closes #1782)

### LINKED ISSUE #1782: Add 'System Information" to GUI
It's nice to have information about graphics drivers printed in terminal, but it's not ideal and I would prefer if it could be accessible through GUI too.
We could implement it in a way Steam does (a window that gets brought up from "Help" menu) , but leave only useful information like graphics devices, driver type, version, Desktop Environment, kernel version, distro (version of it), session type (Wayland/Xorg) and maybe some other stuff too.
image

Citiroller:
This is a initial commit to try and add the system information to lutris as proposed in #1782 
It's not complete as of yet.
When the Preferences are opened on a system level (not on a game level) a new tab appears which contains the system information.
I also changed the `get_mem_info()` function to read from `/proc/meminfo` as the keys were localized in the old function. The old function was deleted. Another possibility to have the keys always in english would be  to export the system locale before running the `free` command with `export LANG=en_US.UTF-8 && free`.

Missing features / issues:
* Only displays information about the first CPU socket, could be problematic with server boards which have two or more CPU sockets.
* Needs testing on different hardware configurations, especially with amdgpu drivers.

Bildschirmfoto vom 2019-05-22 13-47-29

As this is my first commit to this project I'm happy about feedback.

### DIFF
--- a/lutris/util/linux.py
+++ b/lutris/util/linux.py
@@ -171,20 +171,6 @@ class LinuxSystem:
             drive for drive in json.loads(output)["blockdevices"]
             if drive["fstype"] != "squashfs"
         ]
-
-    @staticmethod
-    def get_ram_info_old():
-        """Return RAM information"""
-        try:
-            output = subprocess.check_output(["free"]).decode().split("\n")
-        except subprocess.CalledProcessError as ex:
-            logger.error("Failed to get RAM information: %s", ex)
-            return None
-        columns = output[0].split()
-        meminfo = {}
-        for parts in [line.split() for line in output[1:] if line]:
-            meminfo[parts[0].strip(":").lower()] = dict(zip(columns, parts[1:]))
-        return meminfo
     
     @staticmethod
     def get_ram_info():
@@ -443,17 +429,14 @@ def gather_system_info_str():
     system_info_readable["Memory"] = ram_dict
     #Add graphics information
     graphics_dict = {}
-    if drivers.is_nvidia():
-        graphics_dict["Vendor"] = "nVidia"
-        logger.info("nVidia info not yet implemnted")
-    elif LINUX_SYSTEM.glxinfo:
+    if LINUX_SYSTEM.glxinfo:
         graphics_dict["Vendor"] = system_info["glxinfo"]["opengl_vendor"]
         graphics_dict["OpenGL Renderer"] = system_info["glxinfo"]["opengl_renderer"]
         graphics_dict["OpenGL Version"] = system_info["glxinfo"]["opengl_version"]
         graphics_dict["OpenGL Core"] = system_info["glxinfo"]["opengl_core_profile_version"]
         graphics_dict["OpenGL ES"] = system_info["glxinfo"]["opengl_es_profile_version"]
     else:
-        graphics_dict["Vendor"] = "Unable to detect driver"
+        graphics_dict["Vendor"] = "Unable to obtain glxinfo"
     #check Vulkan support
     if vkquery.is_vulkan_supported():
         graphics_dict["Vulkan"] = "Supported"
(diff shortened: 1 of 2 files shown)


strycore:
That looks pretty cool so far

jaymehoward:
Couple things I'd love to see in this:  Proper column splits with alignment from top to bottom on the page, and a button to get it in a fixed-width font text box, so it can easily be selected and copied for pasting into issues with the insert code option.

isugimpy:
Couple things I'd love to see in this:  Proper column formatting from top to bottom for visibility, and button that displays a fixed-width font text box that could be copied and pasted into code tags on issues.

Citiroller:
I'll try using the integrated python formatting to format it like two columns.
Regarding the copy function: I use a text box which is non editable, so it's possible to mark the text and copy it by using ctrl+c, I can add a button wich says copy to put the text into the system copy buffer if that is what you propose.

isugimpy:
Ah, I missed the fact that it was already a text box.  If that's the case, there's no need for a button!

Citiroller:
I just added some graphics information for gpus using mesa. I will add Nvidia graphics info soon. I need to clone my repo to my other machine to test it. I'm not sure if this will work with the amdgpu drivers though as there doesn't seem to be an implementation yet to get the information.

lutris/util/linux.py
```diff
@@ -390,5 +394,66 @@ def gather_system_info():
...
+    print(system_info["gpus"])
+    print("GLX:")
+    print(system_info["glxinfo"])
+    system_info_readable = {}
+    #Add system information
+    system_dict = {}
+    system_dict["OS"] = ' '.join(system_info["dist"])
+    system_dict["Arch"] = system_info["arch"]
+    system_dict["Kernel"] = system_info["kernel"]
+    system_dict["Desktop"] = system_info["env"]["XDG_CURRENT_DESKTOP"]
```
isugimpy: You may want to make this line and the one below it call .get() instead of just doing the subscript lookup.  If somebody's not running a full DE and their WM doesn't present that variable to the env, it'll raise an exception.

Citiroller: So something like
```suggestion
    system_dict["Desktop"] = system_info["env"].get("XDG_CURRENT_DESKTOP", "No DE found")
```

isugimpy: Seems reasonable to me.  Not a maintainer, so I don't want to make a decision on the verbiage, but functionally that's what I was proposing.

Citiroller: Good, I'll wait on more info from the maintainers. Maybe they want me to expose more information or remove some lines. Thanks for your comment!

LiamDawe:
If I might have one suggestion: The ability to click to copy to clipboard? Could end up being useful for people to paste their sys info.

strycore:
Yeah, the text should totally be selectable

Citiroller:
In this implementation the text is displayed in a non editable text box, so the text is selectable and you can copy it by pressing ctrl+c or right click and select copy
Bildschirmfoto vom 2019-05-26 20-09-41
But yeah, I can add a button, for convenience.

LiamDawe:
Great, top stuff!

Citiroller:
I've added a copy button at the end of the textview. See the attached screenshot.
Bildschirmfoto vom 2019-05-26 22-26-08
In the last commit I've also added the changes suggested by @isugimpy .
Is there any more information the system info should expose like missing libs? And I hope the way I've implemented it is fine with the maintainers.

Medath:
This is a great feature, but for some reason the tabs seem to be a bit messed up on my system
SCR_20190623_232451

Citiroller:
The thing is, when a non monospace font is used I can't calculate the number of tabs I need for each line to fit perfectly. One way could be to use a monospace font. Or use two text boxes? But I think then you can't select the whole text.
Does anyone know a lib or some function in gtk to do this?

Issue #2046, opened 2019-04-16
Game listed last if it starts with a lowercase


AccountOneOff:
```
cxf@PC:~/.lutris$ git bisect bad
284ee18c2f506aee524926c1b089112568c56e4b is the first bad commit
commit 284ee18c2f506aee524926c1b089112568c56e4b
Author: Mathieu Comandon <strycore@gmail.com>
Date:   Mon Mar 18 19:48:21 2019 -0700

    Apply sort_func to modelsort on creation (Closes: #1805)

:040000 040000 4f22545a5217a6111c1e20c6533cebe49aa535fb e6af73dfe5f4932c1e6b60e5e7a104fb394101c9 M	lutris
```

![Screenshot from 2019-04-15 22-15-36](https://user-images.githubusercontent.com/15083452/56184013-7d803780-5fcd-11e9-9850-e18304e143df.png)
![Screenshot from 2019-04-15 22-33-40](https://user-images.githubusercontent.com/15083452/56184408-c1277100-5fce-11e9-9de8-c131dbf6556d.png)

isugimpy:
This looks like a really trivial fix, I'll take a crack at it tonight.

PR #2138, opened 2019-05-17
Fix sorting (Closes: #2046)

isugimpy:
I maybe overdid it on the test coverage here, and the try_lower() method is kinda an ugly hack, but it's way better than sticking a try/except around every attempt to lower things.  Also, this makes sure that if we somehow ever try to compare two rows that don't have a name/runner (okay, yeah, this should never happen, but stranger things have happened), we still return a value rather than raising an exception and just barfing.

### DIFF
--- a/lutris/gui/views/store.py
+++ b/lutris/gui/views/store.py
@@ -30,6 +30,14 @@ from . import (
 )
 
 
+def try_lower(value):
+    try:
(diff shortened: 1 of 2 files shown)


strycore:
Well I was going to say "Do, or do not, there is no try" but then I saw you added some unit tests so it's all good!

Issue #2081, opened 2019-04-24
Lutris gets stuck on launching a game when quotation is incomplete


tannisroot:
This happened to me after I forgot to end a quote in `Arguments` field:
```
 File "/usr/lib/python3.7/site-packages/lutris/exceptions.py", line 26, in wrapper
    return function(*args, **kwargs)
  File "/usr/lib/python3.7/site-packages/lutris/game.py", line 299, in configure_game
    gameplay_info = self.runner.play()
  File "/usr/lib/python3.7/site-packages/lutris/runners/wine.py", line 897, in play
    for arg in shlex.split(arguments):
  File "/usr/lib/python3.7/shlex.py", line 305, in split
    return list(lex)
  File "/usr/lib/python3.7/shlex.py", line 295, in __next__
    token = self.get_token()
  File "/usr/lib/python3.7/shlex.py", line 105, in get_token
    raw = self.read_token()
  File "/usr/lib/python3.7/shlex.py", line 187, in read_token
    raise ValueError("No closing quotation")
ValueError: No closing quotation
```

hwang381:
This one seems quite beginner-friendly. May I take a look? :)

PR #2139, opened 2019-05-17
Config validation, base implementation.  Closes: #2081

isugimpy:
Only enabled for the args field of the wine runner.  Works by doing magic imports and enumerating fields, and if there is a validator for the field, we make sure it does not raise an exception.  If it does, refuse to save.  This does lean more heavily toward using custom validators than core functions, but should be acceptable, I think.

### DIFF
--- a/lutris/runners/wine.py
+++ b/lutris/runners/wine.py
@@ -566,9 +566,9 @@ class wine(Runner):
                 # Update the version in the config
                 if version == self.runner_config.get("version"):
                     self.runner_config["version"] = default_version
-                    # TODO: runner_config is a dict so we have to instanciate a
+                    # TODO: runner_config is a dict so we have to instanciate a #  pylint: disable=fixme
                     # LutrisConfig object to save it.
-                    # XXX: The version key could be either in the game specific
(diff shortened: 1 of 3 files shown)


isugimpy:
This should be ready to review.  Of note, I did put a couple pylint ignores in because CodeFactor was dinging me on them despite them not being in my actual changes.  I believe the comments do reflect changes that are needed, but they're really out of scope for what I'm doing here.

@tannisroot, particularly interested in your thoughts on this one, because it should fix your issue in #2081

tannisroot:
This is pretty cool but I am going to tag this for 0.5.4 as the issue this PR resolves is very minor.

lutris/runners/wine.py
```diff
@@ -565,9 +566,9 @@ def get_executable(self, version=None, fallback=True):
                 # Update the version in the config
                 if version == self.runner_config.get("version"):
                     self.runner_config["version"] = default_version
-                    # TODO: runner_config is a dict so we have to instanciate a
+                    # TODO: runner_config is a dict so we have to instanciate a #  pylint: disable=fixme
```
strycore: Remove those

PR #2152, opened 2019-05-19
Rip out the game_options injection, no longer needed.  Put the discor…

isugimpy:
…d_options block into system options for full inheritance.

Moved the options to system options at the request of @tannisroot and @telanus.  Also set the default to disable the feature, so it's explicitly opt-in.  This allows users to change it globally, at the runner level, and at the game level.  Opting in at the global level turns it on for all games, you can then explicitly opt out for a specific runner (say, LibRetro), and still opt-in for an individual game under that runner.  @strycore, what are your thoughts on this?

### DIFF
--- a/lutris/game.py
+++ b/lutris/game.py
@@ -162,11 +162,11 @@ class Game(GObject.Object):
         )
         self.runner = self._get_runner()
         if self.discord_presence.available:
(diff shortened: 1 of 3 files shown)


tannisroot:
I like the changes. However, I think the options should be grey-ed out instead of simply removing them from the menu when `pypresence` is not available. Also, I think language in the descriptions of new options should be changed, specifically replace `send` with `pass` and `notification` with `status`, but I could do it in a separate PR myself :D

PR #2221, opened 2019-07-01
Begin switch to meson build system

DevilishSpirits:
I just discovered Lutris... very good ! In your README, you was saying that the system-wide installation is not clean. I've learned *meson* and made a `meson.build` file, to use it :
```sh
meson build && cd build && sudo meson install
```
Lutris *python* files are now installed in `$DESTDIR/lib/lutris`.
I also edited some code about path searching (see the *commit* message for all changes), Lutris should be usable from any installation root (*/tmp/lutris* work for me).

With some additions, *translations* and *automated tests* should work.

This patch stay very minimalist and doesn't break old behaviors (running from source and *AUR makepkg* for `lutris-git` stays working).

Are you interested ?

### DIFF
--- a/bin/lutris
+++ b/bin/lutris
@@ -17,10 +17,12 @@ import locale
 from os.path import realpath, dirname, normpath
 
 LAUNCH_PATH = dirname(realpath(__file__))
-if not LAUNCH_PATH.startswith("/usr"):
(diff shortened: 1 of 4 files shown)


strycore:
That looks simple enough and we were considering switching to Meson in the future (Well @TingPing mostly). 
Thanks a lot for providing some initial Meson support

Issue #2210, opened 2019-06-22
side panels consume too much space


foresto:
I have limited screen space, and like to be able to see other things on my screen while using Lutris. Unfortunately, the side panels are rather wide, so I have to choose between accepting them and devoting a huge amount of screen space to the Lutris window, or hiding them and being unable to see what game is currently running.

Some ideas that might help with this problem:
* Make the panels shrink to fit their contents. Right now my left panel's contents need less than a third of the panel's width, so that panel is mostly empty space.
* Make it possible to show/hide each panel independently. I don't care about the left-hand panel. I wish I could hide it without hiding the right-hand one.
* Automatically hid the left-hand panel when only one item in it is in use. All my Lutris games are Wine games. I don't need a panel devoted to selecting other runners that none of my games use.
* Show running game status in a status bar. This way I could hide both panels, reclaim all the wasted screen space, and still see a little text at the bottom of the window indicating whether a game is running.

Faalagorn:
I have the same problem and I'd like to see especially dot number 2 being addressed, as I'd also love to hide just one panel.

tannisroot:
Panels can now be hidden.

Issue #1916, opened 2019-03-23
[v0.5.1] lutris not opening, always redirecting to donate webpage.


telans:
`@Desktop:~$ lutris`
`GDBus.Error:org.freedesktop.DBus.Error.NoReply: Message recipient disconnected from message bus without replying
`

v0.5.0.1 works without problem.

telans:
Fixed by running lutris --submit-issue despite not submitting. Lutris now opens correctly.

PR #2239, opened 2019-07-19
Toggle each side panel independently

Xenega:
Allows to show/hide each panel independently. Use two buttons and two hotkeys (F9 and F10).

Solve partially #2210.
Solve bug found in #1916


strycore:
F10 is a non standard for that type of action, but eh, whatever.

PR #2271, opened 2019-08-13
Add liberapay and the donate page to the sponsor button

rogersachan:
This pull request adds liberapay and the donation page to the sponsor button. allowing people to find liberapay and paypal donation methods faster.


strycore:
I cherry picked the commit to avoid having to do a rebase, can be closed.

PR #2275, opened 2019-08-14
Add auto-hide-launcher option

danielfvm:
This pull request adds an auto-hide-launcher option. If enabled, it hides lutris when a game is running and appears again when the game is closed.


strycore:
Thanks, that's neat!

Issue #2292, opened 2019-08-27
Wine sandbox on non english system


G4E:
Hello,
By default, the option "Create a sandbox for wine folders" is ticked when we create a new prefix, by creating a new game with a wine runner.
It doesn't seems to work on my system, as I have found some data in my ~/Documents folder created by some app in a wine prefix.
If I go in the wine's prefix/drive_c/users/tom
I can see that indeed, wine has created some symlinks to my home folders :
```
ls -l
total 88
drwxr-xr-x 3 tom tom 4096 14 août  22:19  AppData/
drwxr-xr-x 3 tom tom 4096 14 août  22:19 'Application Data'/
lrwxrwxrwx 1 tom tom   17 14 août  22:19  Bureau -> /home/tom/Desktop/
drwxr-xr-x 2 tom tom 4096 14 août  22:19  Contacts/
drwxr-xr-x 2 tom tom 4096 14 août  22:19  Cookies/
drwxr-xr-x 2 tom tom 4096 18 août  22:30  Desktop/
drwxr-xr-x 2 tom tom 4096 14 août  22:19  Downloads/
drwxr-xr-x 2 tom tom 4096 14 août  22:19  Favorites/
drwxr-xr-x 2 tom tom 4096 14 août  22:19  Links/
drwxr-xr-x 5 tom tom 4096 14 août  22:19 'Local Settings'/
lrwxrwxrwx 1 tom tom   17 14 août  22:19 'Ma musique' -> /home/tom/Musique/
lrwxrwxrwx 1 tom tom   19 14 août  22:19 'Mes documents' -> /home/tom/Documents/
lrwxrwxrwx 1 tom tom   16 14 août  22:19 'Mes images' -> /home/tom/Images/
lrwxrwxrwx 1 tom tom   17 14 août  22:19 'Mes vidéos' -> /home/tom/Vidéos/
drwxr-xr-x 2 tom tom 4096 18 août  22:30 'My Documents'/
drwxr-xr-x 2 tom tom 4096 18 août  22:30 'My Music'/
drwxr-xr-x 2 tom tom 4096 18 août  22:30 'My Pictures'/
drwxr-xr-x 2 tom tom 4096 18 août  22:30 'My Videos'/
drwxr-xr-x 2 tom tom 4096 14 août  22:19  NetHood/
drwxr-xr-x 2 tom tom 4096 14 août  22:19  PrintHood/
drwxr-xr-x 2 tom tom 4096 14 août  22:19  Recent/
[...]

PR #2300, opened 2019-09-01
Wine sandbox from registry

G4E:
Here is a pull request to avoid internationalization problems when we try to remove wine's symlinks to the user's folders (#2292)
Be careful, there might be encoding issues at the moment.


G4E:
Latest commit works as-is for me. 
Previously, non ascii characters were encoded like "Mes Vid\xe9os" (with \xe9 encoded on 4 characters). My modification in registry.py solve this, but has the side effect of also removing **\\** hard escape to simply **\**. Which might break get_unix_path and potentially other functions ?

strycore:
really nice, thanks!

PR #2301, opened 2019-09-01
Add german language to DESKTOP_FOLDERS

Medath:
~Entirely untested~
Seems to work for me


G4E:
Could you try my pull request #2300 ? It should solve it as well for all languages.

PR #2322, opened 2019-09-08
Fetching user's folder by using wine's registery

G4E:
Basically the same as pull #2300, but with more work on the unicode decoding part.
Tested and working with locales:
-  Russian (ru_RU)
-  French (fr_FR)
-  Spanish (es_ES)

@Medath Could you test in German ?
If some peoples have non English systems, please give it a try, unicode handling is tough !


gasinvein:
Waitasec. Does this mean that Lutris' "wine sandbox" feature didn't work with non-English locales?

Medath:
Yes

Issue #2385, opened 2019-10-06
Some lutris runners cannot be installed


CartoonFan:
**Describe the bug**

The Higan, melonDS, Redream, or TIC-80 runners fail silently with no popups and a "urllib.error.HTTPError: HTTP Error 404: Not Found" error. These four appear to use https://lutris.net/api/runners/ instead of https://lutris.net/runners, where they are not present. Additionally, lutris' list of libretro cores does not agree with the list on https://lutris.net/runners: TIC-80 and Redream are the only ones on both lists, with higan only showing up in lutris and melonDS only appearing on the website.

**Expected behavior**

All runners available in lutris should be installable.

**Current behavior**

Four runners are not able to be installed.

**Steps to reproduce**

- Open Lutris and navigate to the "Manage Runners" section.
- Try to install any of the higan, melonDS, Redream, or TIC-80 runners.
-  Optionally try to install either of the higan libretro cores  through the lutris app.

**Lutris debugging output**

```
INFO     2019-10-06 10:32:16,039 [application.do_command_line:225]:Running Lutris 0.5.3
INFO     2019-10-06 10:32:16,039 [startup.check_driver:54]:Using NVIDIA drivers 435.24.02 for x86_64
INFO     2019-10-06 10:32:16,039 [startup.check_driver:58]:GPU: GeForce GTX 960
INFO     2019-10-06 10:32:16,039 [startup.check_driver:75]:GPU: 8086:0402 1043:8534 using i915 drivers
INFO     2019-10-06 10:32:16,223 [startup.check_vulkan:129]:Vulkan is supported
INFO     2019-10-06 10:32:16,225 [dxvk.get_dxvk_versions:21]:Updating DXVK versions
INFO     2019-10-06 10:32:16,622 [dxvk.get_dxvk_versions:21]:Updating D9VK versions
DEBUG    2019-10-06 10:32:16,791 [http.get:65]:GET https://lutris.net/api/runtime
INFO     2019-10-06 10:32:16,798 [lutriswindow.toggle_connection:580]:Connected to lutris.net as CartoonFan
DEBUG    2019-10-06 10:32:16,802 [http.get:65]:GET https://lutris.net/api/games/library/CartoonFan
DEBUG    2019-10-06 10:32:16,818 [store.get_missing_media:194]:Requesting missing icons from API for 113 games
DEBUG    2019-10-06 10:32:16,818 [http.get:65]:GET https://lutris.net/api/games
[...]

PR #2396, opened 2019-10-12
Fix issue #2385

skriptmonkey:
Fixes issue #2385 
Higan, MelonDS, Redream, and TIC-80 are no longer available to download and install via https://lutris.net/runners. Also, their functionality looks to already exist within the Libretro runner.


telanus:
Higan Yes, It can be replaced with Libretro's bsnes core. Byuu actively help to keep the core inline with the stand-alone emu.

melonDS: The Current Libretro Core is quite a bit out of sync with the stand-alone emu. 

Redream: The Libretro core is ages behind the stand-alnoe emu, and will never be able to get on par with what the stand-alone. and after the drama the dev had with libretro, it went close source. I will try and get permission of the dev to host stable releases, but doubt it'll happen.

 TIC-80: The Libretro core isn't as usable as the stand-alone. The way Libretro implement their retropad isn't conducive to  TIC-80's usability

soredake:
Redream can be replaced with flycast https://github.com/libretro/flycast/

skriptmonkey:
@telanus Understood, but these runners were already unavailable via the website/server. All I'm doing is syncing the client to the servers by removing the entries from showing up in Manage Runners. Otherwise users will expect the stand-alone functionality and try to install, only to be greeted by a 404.

@soredake Thank you. Flycast is one of the cores included in the Libretro runner that is currently present in Lutris. :)

telanus:
> I think it would be better to instead just comment out those runners until we add actually build them and start distributing.
For redream, one need to use official files, as the emu is closed-source

strycore:
I agree that those runner should not have been displayed. Please submit build scripts in the buildbot repo so we can distributing those emulators and have runners for them.

PR #2382, opened 2019-10-05
Added the option to restore the window position (fixes #2354)

G4E:
I implemented here the option to save the window position between launches. There is a new option in the lutris preferences. Fixes #2354 :)
image

The logic might seems convoluted (storing the inverse of the checkbox) but with this way everything is still coherent when the option is not in the config file.

PR #2404, opened 2019-10-15
Fixes #2354 by saving the window's position in the config file

### LINKED ISSUE #2354: Lutris window always starts in the middle of the screen
**Describe the bug**

Lutris does not remember its window position. Instead it always starts in the middle of the screen. (This is using KWin.)

**Expected behavior**

If I set the window to be in a particular position, next time I start it, it should be in that position.

**Current behavior**

The window spawns in the middle of the screen, no matter what.

**Steps to reproduce**

* Start Lutris.
* Move window to non-middle position.
* Quit Lutris.
* Start Lutris.

**Lutris debugging output**

```
$ lutris -d
INFO     2019-09-17 00:13:49,893 [application.do_command_line:225]:Running Lutris 0.5.3
INFO     2019-09-17 00:13:49,894 [startup.check_driver:60]:Using X.Org
INFO     2019-09-17 00:13:49,894 [startup.check_driver:65]:Running Mesa driver 19.1.6 on Radeon RX 580 Series (POLARIS10, DRM 3.32.0, 5.2.14-arch2-1-ARCH, LLVM 8.0.1) (0x67df)
INFO     2019-09-17 00:13:49,894 [startup.check_driver:75]:GPU: 1002:67DF 1682:C580 using amdgpu drivers
INFO     2019-09-17 00:13:49,934 [startup.check_vulkan:129]:Vulkan is supported
INFO     2019-09-17 00:13:49,935 [dxvk.get_dxvk_versions:21]:Updating DXVK versions
DEBUG    2019-09-17 00:13:50,095 [store.get_missing_media:198]:Requesting missing icons from API for world-of-warcraft-classic
DEBUG    2019-09-17 00:13:50,097 [http.get:68]:GET https://lutris.net/api/runtime
DEBUG    2019-09-17 00:13:50,108 [http.get:68]:GET https://lutris.net/api/games
INFO     2019-09-17 00:13:50,201 [dxvk.get_dxvk_versions:21]:Updating D9VK versions
DEBUG    2019-09-17 00:13:50,688 [api.get_game_api_page:132]:Loaded 1 games from page 1
INFO     2019-09-17 00:13:57,684 [application.do_shutdown:446]:Shutting down Lutris
```

**System information**

[lutris-issue-report.json.txt](https://github.com/lutris/lutris/files/3619225/lutris-issue-report.json.txt)

(JSON files cannot be attached for some reason.)

G4E:
The same work as pr #2382 but without the option to disable this behavior.
Fix #2354
[WIP] Don't merge as-is.

### DIFF
--- a/lutris/gui/lutriswindow.py
+++ b/lutris/gui/lutriswindow.py
@@ -73,6 +73,7 @@ class LutrisWindow(Gtk.ApplicationWindow):
         super().__init__(
             default_width=width,
             default_height=height,
+            window_position=Gtk.WindowPosition.NONE,
             icon_name="lutris",
             application=application,
             **kwargs
@@ -108,6 +109,14 @@ class LutrisWindow(Gtk.ApplicationWindow):
         self.connect("delete-event", self.on_window_delete)
         if self.maximized:
             self.maximize()
+        else:
+            #If we can't retrieve the window coordinates from the config file, the window will stay at the center of the screen
+            try:
+                x,y = [int(i) for i in settings.read_setting("window_position").split(',')]
+                self.move(x,y)
+            except ValueError:
+                pass
+
         self.init_template()
         self._init_actions()
         self._bind_zoom_adjustment()
@@ -592,6 +601,8 @@ class LutrisWindow(Gtk.ApplicationWindow):
             self.window_size = widget.get_size()
 
     def on_window_delete(self, *_args):
+
+
         if self.application.running_games.get_n_items():
             dlg = dialogs.QuestionDialog(
                 {
@@ -602,6 +613,9 @@ class LutrisWindow(Gtk.ApplicationWindow):
             )
             if dlg.result != Gtk.ResponseType.YES:
                 return True
+
+        self.set_gravity(Gdk.Gravity.STATIC)
+        settings.write_setting("window_position","{},{}".format(self.get_position().root_x,self.get_position().root_y))
 
     @GtkTemplate.Callback
     def on_destroy(self, *_args):


G4E:
As stated in the pygtk documentation : https://developer.gnome.org/pygtk/stable/class-gtkwindow.html#method-gtkwindow--get-position
```
The get_position() method is not 100% reliable because the X Window System does not specify a way to obtain the geometry of the decorations placed on a window by the window manager. Thus PyGTK is using a "best guess" that works with most window managers. Moreover, nearly all window managers are historically broken with respect to their handling of window gravity. So moving a window to its current position as returned by the get_position() method tends to result in moving the window slightly. Window managers are slowly getting better over time.

If a window has gravity gtk.gdk.GRAVITY_STATIC the window manager frame is not relevant, and thus the get_position() method will always produce accurate results. However you can't use static gravity to do things like place a window in a corner of the screen, because static gravity ignores the window manager decorations. If you are saving and restoring your application's window positions, you should know that it's impossible for applications to do this without getting it somewhat wrong because applications do not have sufficient knowledge of window manager state. The correct mechanism is to support the session management protocol (see the "GnomeClient" object in the GNOME libraries for example) and allow the window manager to save your window sizes and positions.
```
The proposed behavior in the pr isn't correctly working, as the window is slowly moving between restarts.

strycore:
A more detailed explanation of how to achieve this is published here: https://wiki.gnome.org/HowDoI/SaveWindowState

It's in C but since we're using GObject, the principles are the same.

G4E:
After reading carefully, this shows a standardish way to save the window size and the window state (maximized or not), but not the window position. 
Lutris already implements those two parameters in its lutris.conf file. 
I can do the job to store the state of the window like that if you wish.
But it won't solve the present problem : to store the *window position*.

strycore:
Because it's not Lutris' job to save the position, we should not have to do this. The above documentation states: "The position of the window is best left to the window manager."

I have removed the forced centering of the window in the Glade file.

lutris/gui/lutriswindow.py
```diff
@@ -73,6 +73,7 @@ def __init__(self, application, **kwargs):
         super().__init__(
             default_width=width,
             default_height=height,
+            window_position=Gtk.WindowPosition.NONE,
```
G4E: Great ! You may want to merge just this line then, to be clear (i don't know what's the default is)

tannisroot:
Since this is already fixed in master, it should probably be closed.

PR #2423, opened 2019-10-24
Disconnect game hook after game has stopped

manuelVo:
Everytime time application.launch a hook is added to the stop event of that game. This hook never gets removed. However sometimes a game instance gets reused, causing another hook being added to the stop event so that the event handler is called multiple times.

This PR removes the hook from the game instance once the game has quit ensuring that the quit hook is always called exactly once.


strycore:
Thanks!

PR #2476, opened 2019-11-21
make package Debian compliant

stephanlachnit:
This series of patches tries to make Lutris compliant for a upload in the Debian archive. An official Debian upload would make Lutris easier accessible for more users. Please let me know if you're interested in this, then I can start to explain several changes I've made and ask for a sponsor for the upload.

Any help is welcome, see `debian/TODO`.

Signed-off-by: Stephan Lachnit <stephanlachnit@protonmail.com>


stephanlachnit:
Since I'm not sure if I'll find a satisfying solution for the changelogs, it may also be a good idea to give the Debian package an own branch.
The current patches set their target to bionic, which theoretically allows a backport to Debian stable, but would drop xenial. I don't think that's a problem since gaming on a 3 year old distro may not be the best idea anyway, but I can change it back to xenial if wished.
I also wanted to note that I'll squash all patches together in the end.

stephanlachnit:
So I have a new idea which is I think is easier to maintain: all Debian specific files will be marked with .debian in the end, so right now the changelog, the format and some lintian-overrides. Before the Debian package is build, all .debian files replace their non-.debian partner files. This could be done either in this repository or in a repository on salsa, I actually think it would be easier on salsa.

strycore:
That is really nice and much needed so thank you!

I too would prefer an integration with Salsa rather than having to maintain a separate branch here.
The client's own release cycle will be vastly different from Debian's so I'd rather not have things mixed up in the same repo.

stephanlachnit:
Will upload the modified files for Debian here: https://salsa.debian.org/stephanlachnit-guest/lutris
But right now the upload is not a priority since I need to find a sponsor first and will wait for version 5.4 for the first package, in case this PR gets merged in time.

I'm almost done, the only thing left now is the copyright file (I would remove the README in the directory with the svgs). So if you want you can start to review, I (hopefully) won't change much from now on (and I'll probably need some more time until I can finish the copyright file).

stephanlachnit:
Because I saw that you are preparing 5.4 but it's not released yet, I hurried and finished my work. The PR can be merged as is, but the work isn't finished completely, that's why I left the `TODO` file in the `debian` dir, so everyone knows what should be done to improve the package.
The copyright file is pretty sure incomplete, meaning the old `README.rst` in the svgs directory was. Some files don't exist any more, others have been added without notice. But this is kinda out of my hand.
Second I don't know if `lutris-wrapper` has a proper cli, if it has than there should be a man page, else there should be a lintian override.
Also the copyright file should be reviewed from someone on the Debian project, that's why I didn't delete the old copyright notice for the svgs. It could be, that the package will be rejected because some of the svgs (like the Playstation logos) aren't _really_ free. A possible fix would be that Lutris just downloads the icons once it's started the first time, but that's a topic for another PR.

debian/lintian-overrides
```diff
@@ -0,0 +1,5 @@
+# not a game but a gaming platform
+lutris: package-section-games-but-has-usr-bin
+
+# graphical frontend without cli
+lutris: binary-without-manpage usr/games/lutris
```
stephanlachnit: That's actually not correct and shouldn't be overridden.

stephanlachnit:
Force-push was just a little fixup for `debian/lintian-overrides` and `debian/copyright`, no significant changes.

stephanlachnit:
I checked the git commits and added following copyright notes:
`gog-symbolic.svg`: romlok, CC0-1.0
`ms-dos-symbolic.svg` & `necpc-98-symbolic.svg`: matoro, CC0-1.0
`windows-symbolic.svg`: you, CC0-1.0

stephanlachnit:
Since 0.5.4 is now released I uploaded my version on salsa. While doing this I noticed, that the symbolic link `CHANGELOG` to `debian/changelog` doesn't work nicely when building the package as a non-native package since debhelper thinks this is the upstream changelog. While that would be right, I have to change the changelog to actually build the package, so I removed it.

I also added `libwine-development` to the recommendations.

Issue #2344, opened 2019-09-13
Move Items from Gnome's App Menu into the already existing dedicated Menu


ZahnPastor:
See https://gitlab.gnome.org/GNOME/Initiatives/wikis/App-Menu-Retirement

Lutris is still using the App Menu in Gnome to display Settings 
![grafik](https://user-images.githubusercontent.com/21054889/64883266-420e9580-d65f-11e9-9304-377152e74ab1.png)

Gnome wants to retire this App Menu, so Lutris should move all necessary items into the already existing dedicated menu button 
![grafik](https://user-images.githubusercontent.com/21054889/64883341-6e2a1680-d65f-11e9-983c-48fe508dea07.png)

(dont know if this has already been brought up in another issue, i searched for app menu and didnt find any results)

strycore:
Fixed in master

Issue #2154, opened 2019-05-20
Suggestion for new menus since app menu retirement in Gnome 3.32.


pedrofleck:
Since Gnome 3.32, app menu is [deprecated](https://gitlab.gnome.org/GNOME/gnome-shell/issues/624) and a legacy menu like [this](https://i.imgur.com/k1uzKky.png) in Lutris is automatically added to CSD bar. My suggestion is based on Nautilus appearance: move the old app menu (screenshot above) to the current "hamburger" menu on the left and include the options "Lutris forums", "Discord" and "Support Lutris!" there. And the current options shown in the hamburger menu can be moved to a new dropdown menu next to "toggle view" like [this](https://i.imgur.com/J15pyXc.png). In my opinion, it would help consistency throughout the Gnome apps (I know Lutris isn't affiliated to Gnome, but anyway) since I'm always opening the wrong menus searching for the option I want.

strycore:
Fixed in master

PR #2508, opened 2019-12-02
Remove usage of application menu

Appadeia:
The application menu is a design pattern abandoned by GNOME, and makes Lutris stick out on the GNOME desktop for having one. This PR moves the app menu's content into the primary menu of the window. Besides being more GNOME-ish, removing the application menu also reduces the number of hamburger menus a user has to look through in order to find the item they want and makes the headerbar appear less cluttered.


gikari:
I believe this fixes #2344 and #2154

PR #2523, opened 2019-12-15
Epic Games Store runner/service implementation

### LINKED ISSUE #2521: [WIP] Epic Games Store Runner
I'm not sure about the etiquette of asking questions in closed issues (#1794) but it has not received any attention, so I'm opening this one.

I started working on this because i was curious how Lutris works in general. Right now it's a big hack but it works for me. 

If there is interest in the community, I'd love to make this "production ready" together. It took some guesswork how things work in lutris so i definitely need some feedback on my [fork](https://github.com/buxel/lutris/tree/epic_games_store) and comments on [the changes i made](https://github.com/lutris/lutris/compare/master...buxel:epic_games_store).

> **Works**
> * Installing EGS as runner
> * Importer for installed games from EGS to lutris
> * Launching games
>   * Directly (DRM free)
>   * Trough EGS
> 
> **Issues**
> * still (ab)using steamId in some places
> * save/load game settings takes a while because wine registry is initialized each time (probably easy to fix)
> * lutris does not see game exiting, thinks its still running when launched trough EGS
> * most likely some dead code from winesteam
> 
> **Not working**
> * quit EGS after game exits
> * un/-install games (not sure if EGS supports this at all)

buxel:
I started integrating EGS for my personal use. Some parts are very hacky but they worked for me. I'd appreciate feedback and hope we can clean things up together to get this integrated.

Close #2521 in favor of this draft PR.

**Working**

- [x] Installing EGS as runner
- [x] Importer for installed games from EGS to lutris
- [x] Launching games
    - [x] Directly (DRM free)
    - [x] Trough EGS

**Open Issues **
- [ ] still (ab)using steamId in some places
- [ ] quit EGS after game exits 
- [ ] un/-install games (not sure if EGS supports this at all via arguments)
    - [ ] not sure how to handle install scripts
- [ ] Defaut directory for installing games does not work properly ([workaround](https://github.com/lutris/lutris/pull/2523#issuecomment-616096208))
- [ ] lutris does not see game exiting, thinks its still running when launched trough EGS (**help needed**)


buxel:
I had a quick check why travis fails but couldn't put my finger on it.
The CodeFactor issues are just TODOs i already mentioned in the initial PR.

I'm off to vacation now so i'll not code anything until next year. 
Till then: Happy new year to you! :tada:

buxel:
I'm terribly sorry - this was a rebase attempt gone horribly wrong :sob: . If there is no way of salvaging this (any idea?), i will create a clean fork/pr.

buxel:
I don't have too much free time to work on this but i'd greatly appreciate some feedback on what still needs to be done so this could get merged.

Especially the unticked TODOs give me trouble, because i don't know how to approach them.

buxel:
Updated to latest master - it has been a while. 

The initial setup (downloading, winetricks, install) takes some time and it might seem as if it crashed. Also, Lutris needs to be restarted for the installed to show up as installed. 
Does anybody have an idea why that could be? 

Apart from that, everything works smoothly for me.

strycore:
I want to wrap up the 0.5.6 release pretty soon and don't want to delay it with a new service but  can we plan this for inclusion in 0.5.7?

buxel:
Definitely. So far it is in a "works on my machine" state and I'd be more comfortable with having it reviewed before.

Steinhagen:
Small thing I've seen cloning your pull request:
`2020-04-12 14:53:52,349: Unable to find icon '/home/username/Projects/epicstore/lutris/share/lutris/media/runner_icons/wineegs.png'`
 You should probably add an icon for the store as well.

buxel:
@Steinhagen thanks for the heads up! I added one from wikipedia, even tough it lacks a fancy glass of wine. :wink:

Steinhagen:
Looks good. However, you are still missing a hicolor icon for EGS: 
> lutris/share/lutris/icons/hicolor/symbolic/apps/

buxel:
I'm using photopea as an editor but for some reason i cannot export it as svg. It's always missing colors/layers. A PR for a hicolor icon would be welcome :wink:

Steinhagen:
@buxel I created a pull request on your fork with the Epic Games Store symbolic icon.

buxel:
Thank you! :grin:

Steinhagen:
I played with it for a bit, and there is a thing that is broken (in my opinion). The default installation folder for this runner is being ignored. At the present time, all games installed using the EGS runner are being installed in the runner directory (/home/username/.local/share/lutris/runners/wineegs). I'm not sure that's a good ideea.
Also, it would be good to have a demo install script to show this runner working before merging.

buxel:
@Steinhagen you are right, this does not work properly or as you might expect. The problem is that EGS is still not very well documented (or good software in general) so the current way to use it is:
1) Start the runner
2) Install a game and select the path (can also be outside prefix)
3) Trigger importing from EGS to Lutris. 

There are launch arguments (see [here](https://github.com/buxel/lutris/blob/epic_games_store/lutris/runners/wineegs.py#L380)) which allow starting a game, but i haven't found any for installing.

Steinhagen:
I gave it a try and it worked (with some quirks).
It's not exactly intuitive, but if you know what you have to do, it will work.
Question now is how will it work with install scripts?

Steinhagen:
Small issues that should be fixed before integrating this:
1. Looks like the import library is not being updated after you remove a game.
If you enable Sync all games at startup, it will automatically add it as an entry even if the game is no longer installed.

Steinhagen:
2. The default game installation
This issue is a bit annoying. A solution would be to modify the Epic Games Store default installation based on the location you provide to the runner. In this case, let's say:
a) In Lutris EGS default location you have: '/home/username/Games'
b) When you apply this change, your python script should convert this unix location in a winepath (https://wiki.winehq.org/Winepath). Example:
```
# winepath -w /home/username/Games/
Z:\home\username\Games\
```
c) Using the winepath, modify the default installation folder for EGS inside the runner directory. Example of it's location:
`# head -5 /home/username/.local/share/lutris/runners/wineegs/prefix64/drive_c/users/username/Local Settings/Application Data/EpicGamesLauncher/Saved/Config/Windows/GameUserSettings.ini

[Launcher]
LastActiveDateTime=2020-04-19T00:00:00.000Z
LastActiveVersion=10.15.5
AutoEnabledStartOnBootVer=10.15.3
DefaultAppInstallLocation=Z:\home\crapitea\Game
`

Is this a great solution? Not really. Will it work? Probably.

Steinhagen:
3. This is more of a question, but does EGS in it's current state support individual game prefixes? From what I can tell, it doesn't (and I guess it should, considering how finicky some games are).

buxel:
Thank you for you in depth look at this.

1) I couldn't reproduce this issue. what i did was:
    - launch the EGS runner and install a game
    - enable sync, restart lutris
    - the game is properly imported
    - remove the game from lutris
    - launch EGS, uninstall the game
    - restart lutris -> the game is not synchronized anymore

    If you meant that the game is not automatically removed from lutris after uninstalling it from EGS, you are right. I followed the winesteam runner implementation and i think it is that done  way intentionally. @strycore any comment on this?

2) I mostly agree with you on this. The underlying problem is, in my opinion, that we cannot control EGS enough to allow scripted installations. We might need to take a **closer look at EGS if it supports more arguments besides launching a game** to solve this.
I wouldn't want to meddle with EGS configurations just to set the default directory there. If a user installs a game manually in EGS, there should be no interference from lutris. The sync feature should take care of it appearing in lutris then.

3) You caught me there :wink: I haven't paid much attention to this, since i usually have one prefix for all my games and so far it worked. You can set a custom prefix path in `Game Options -> Prefix` but it throws an error. I'll look into this.

Steinhagen:
> We might need to take a closer look at EGS if it supports more arguments besides launching a game to solve this.
I kinda already looked at this before suggesting the previous option.
Unfortunately, I haven't found any supported arguments other than those needed to launch an application (and to directly access your games library). From the looks of things, they are not supporting cmdline arguments.
For the present time, until such an API will be available, I suggest at least implementing the default installation path using config manipulation.

buxel:
I checked the EGS [debug log](https://gist.github.com/buxel/04253cc795b70aa9a065867a9484614b) (can be enabled in settings) after installing a game. I couldn't find anything useable, tough. 
Another approach might be scanning the binaries for printable strings and see if we can find at least the arguments for launching games. Going from there we might find others, too - if they exist at all.

Steinhagen:
During testing, I've seen an unusual behaviour. My Epic application created 2 Data Manifests for the same application (probably because I've installed/uninstalled the app twice). Because of this, there are 2 different files for the same game in: `~/.local/share/lutris/runners/wineegs/prefix64/drive_c/ProgramData/Epic/EpicGamesLauncher/Data/Manifests/`
Those items have identical fields except for "InstallationGuid" and the "InstallSessionId". Do you have any ideea on how to get around this?
A solution would be to just read each file and get all unique ones, but I'm not sure if it's the correct one.

A reminder for other contribuitors, we are using those files to determine what games are installed on your system so a duplicate file will result in a duplicate game in the Syncer class.

Steinhagen:
A way to trigger such a scenario would be:
1. Install a game using EGS Launcher. This will install the game in: _/home/username/Games/egsgame_. This will create a item manifest for this game in:
_~/.local/share/lutris/runners/wineegs/prefix64/drive_c/ProgramData/Epic/EpicGamesLauncher/Data/Manifests/_
2. Close the EGS Launcher.
3. Remove the game manually (not using the EGS Launcher):
_rm -rf /home/username/Games/egsgame_
4. Open the EGS Launcher. It will realize that the game is no longer there and it will how it as 'Uninstalled' in the GUI. However, when doing that, it will not also remove/clean the manifest it previously created for that game.
5. If you are installing the game again, the EGS will recreate a new manifest for the game.

Doing this, you will have 2 different manifests and, of course, 2 different entries inside the EGS Syncer.

buxel:
I can confirm this issue. Possible approaches:
1) Stop fiddling with folders :wink: 
2) Filter out duplicates -> might give more trouble if the 2nd install was in a different directory
3) Check, if the install directory exists

I tought 3) would be easy enough because there is already a function being called, which checks if a game is installed and should be added to the import list [here](https://github.com/buxel/lutris/blob/epic_games_store/lutris/util/egs/appmanifest.py#L47)

Something along the lines of:
```python
def is_installed(self):
        # Check if the application is installed according to manifest
        complete = not self.appmanifest_data.get("bIsIncompleteInstall") # true, if INcomplete   
        
        # check, if the game directory exists. 
        # EGS might not clean up manifest files if a game directory is manually deleted.
        # this could create duplicate manifests when installing the game again
        game_path = get_unix_path(<<prefix_path>>, self.installdir)
        dir_exists = os.path.exists(game_path)
        return complete and dir_exists
```
The problem with this is, that at this stage, we do not know the prefix_path, because we do not have a lutris game at hands yet. Am i missing something?

buxel:
As discussed with @Steinhagen on Discord, this is [my take](https://github.com/lutris/lutris/pull/2523/commits/5090f55c1e4be179c32ab8e45b7c5aa6931ea019)  avoiding duplicates when syncing:
- we only assume the default wine prefix for EGS games for now
- the installation path of the games in manifests are checked within the prefix and skipped if they don't exist
- if two manifests point to the same install dir, we assume they are the same and ignore one (using a python set)

buxel:
This PR is on hold in favor of integrating legendary as runner. 

Progress can be tracked here: https://github.com/buxel/lutris/commits/legendary-runner

DestroyerXyz:
Any update on this?

strycore:
We now have a working integration with EGS, this is no longer needed.

PR #2535, opened 2019-12-22
Move view configuration items into separate popover

Appadeia:
The zoom and sorting options are now in a separate popover triggered by a menu button next to the view configuration button.

image


strycore:
That's really cool, thanks!

PR #2536, opened 2019-12-22
Preferences Refresh

pontaoski:
- libhandy widgets are now used for preferences windows
- editable grids are put in their own popover
- some icons are changed
- reset button is removed due to technical issues (plus they're mostly pointless from a UX perspective considering the vast majority of options can be reset either via blanking a text field, clicking the option labelled default, or triggering a switch)
- clicking anywhere in a row will trigger a switch
- preferences windows now have search and are adaptive

image
image
image
image

### DIFF
--- a/lutris/gui/application.py
+++ b/lutris/gui/application.py
@@ -29,6 +29,7 @@ import gi
 gi.require_version("Gdk", "3.0")
 gi.require_version("Gtk", "3.0")
 gi.require_version("GnomeDesktop", "3.0")
+gi.require_version("Handy", "0.0")
 
 from gi.repository import Gio, GLib, Gtk
 from lutris import pga
--- a/lutris/gui/config/add_game.py
+++ b/lutris/gui/config/add_game.py
@@ -24,12 +24,8 @@ class AddGameDialog(Dialog, GameDialogCommon):
             runner_slug=self.runner_name,
             level="game",
         )
-        self.build_notebook()
         self.build_tabs("game")
         self.build_action_area(self.on_save)
         self.name_entry.grab_focus()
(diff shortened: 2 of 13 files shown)


Appadeia:
Closing since Lutris supports distros too old for libhandy.

tannisroot:
I would keep this opened, we might be able to drop support for Bionic and Stretch next year.

strycore:
> I would keep this opened, we might be able to drop support for Bionic and Stretch next year.

We don't have a plan to deprecate Bionic and Stretch at the moment. Marking this as out of scope until we change our shipping targets.

YihaoPeng:
Personal opinion only: I like the reset button and use it often, please do not remove it.

### Why is the reset button important?

1. Consider the DXVK version in the wine runner. If the user does not select any version, it will always update to the latest version automatically. But if the user selects a specific version, it will be locked in that version and will no longer be automatically updated.

2. In addition, it also allows us to easily delete all environment variables or DLL Overrides (sometimes very many). It is painful to manually delete one by one.

3. Finally, the appearance of the reset button allows users to easily see which options have been modified by themselves or the installer. They can also make it easier to ensure that an option has been set as a default value.

### What happens when the reset button is removed?

1. Once the user selects a DXVK version, there is no chance to reset it back to the automatic update state. Or we must add an additional "auto update" entry to the drop-down box.

2. The user must delete environment variables or DLL overrides one by one, it's hard.

3. Users can no longer easily find the options they have modified. The user must pull down each drop-down box to confirm that he has set the option to the default value.

strycore:
We are not getting rid of the reset button. To prove it, I will send this PR to hell.

PR #2540, opened 2019-12-26
Fixed one lintian error when building deb

sharkwouter:
This fixes the package-section-games-but-has-usr-bin error which was
shown when running debuild.

With this change, there will no longer be any binaries in the /usr/bin
directory in the deb package.

There are a couple more errors when building the deb package. I'll try to look into those too.

### DIFF
--- a/debian/lintian-overrides
+++ b/debian/lintian-overrides
@@ -1,2 +1,0 @@
-# not a game but a gaming platform
-lutris: package-section-games-but-has-usr-bin
--- a/debian/lutris.links
+++ b/debian/lutris.links
@@ -1,2 +1,0 @@
-/usr/bin/lutris /usr/games/lutris
-/usr/bin/lutris-wrapper /usr/games/lutris-wrapper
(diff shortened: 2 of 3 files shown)


stephanlachnit:
@strycore are you sure this should be merged? I think it should be reverted since Lutris is not a game.
Reading through the lintian tag and policy manual ([link](https://lintian.debian.org/tags/package-section-games-but-has-usr-bin.html)), this warning is mainly for actual games so they can set certain security features for example for online games. However as already noted in the comment for the override, Lutris is not a game and therefore shouldn't be in this location, it might even cause problems.

sharkwouter:
I don't think Debian makes a destinction between games and game related tools for this one. Steam and Gnome Games also have their binaries installed in ``/usr/games``

stephanlachnit:
> I don't think Debian makes a destinction between games and game related tools for this one. Steam and Gnome Games also have their binaries installed in `/usr/games`

Indeed, didn't thought steam would be in there. I found some overrides which add for example mapping tools is `/usr/bin`, but if steam is in `/usr/games` then Lutris probably should as well.

PR #2543, opened 2019-12-27
Added dpkg build files to .gitignore file

sharkwouter:
When building the debian package these files are generated, but you wouldn't want to commit them to git.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,15 @@ pga.db
 tests/coverage/*
 /dist
 /lutris.egg-info
+
+#dpkg build files
+.pybuild/
+debian/.debhelper/
+debian/debhelper-build-stamp
+debian/files
+debian/licence-test
+debian/lutris.postinst.debhelper
+debian/lutris.prerm.debhelper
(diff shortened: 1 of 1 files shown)


stephanlachnit:
Wasn't really necessary though. I added a `clean` file, meaning running `dh_clean` deletes all build files.

sharkwouter:
Ah, I see. I was building with debuild instead of using the Makefile.

stephanlachnit:
> Ah, I see. I was building with debuild instead of using the Makefile.

No I'm talking about `debuild`. The [`debian/clean`](https://github.com/lutris/lutris/blob/master/debian/clean) file will be read by `debuild`. The only thing is that `debuild` only cleans _before_ the build, not after it. Just run `dh_clean` after `debuild` and you're good to go. I'm actually not sure if the makefile even cleans after the build, it might actually should be updated now that I take a look at it.

strycore:
I agree with @stephanlachnit, this is not necessary and I'd rather have build files not ignored by git and have them cleaned up by debclean.

PR #2547, opened 2019-12-27
lutris-wrapper: setproctitle to game name

AaronOpfer:
Makes it slightly easier to identify what game a particular `lutris-wrapper` instance is responsible for.

image


AaronOpfer:
It looks like adding `setproctitle` to the Python dependencies has to be done in a few different places. @strycore do you have a comprehensive list of every place I should add it? Or we can drop this idea as its only a nice-to-have.

strycore:
If it's just a nice to have then it doesn't need to be a hard dependency and can be provided in the "Recommends" section of the package description (or equivalent section).
Otherwise, dependencies have to be added to: `debian/control`, `lutris.spec`, `setup.py` and to the `INSTALL` documentation as well.

PR #2569, opened 2020-01-03
Remove unused attribute 'exit_main_loop'

Swalkyn:
While going through lutris code trying to fix some issue, I discovered this unused attribute in `game.py`. I think it can be safely removed since doing a project-wide search yielded only these two occurrences, meaning it is never set to true.
This is one of my first contribution to FOSS. I made sure to read and follow the contribution guidelines, even though introduced changes are small. However, please feel free to refuse the PR and shortly explain why, so that the next one I make is better.


strycore:
Always helpful to have some fresh eyes looking at the source code :)

Issue #2553, opened 2019-12-28
Debian package


sharkwouter:
I was wondering if it would be possible to include Lutris into Debian, but I found a ton of work on this has already been done. @stephanlachnit updated the Debian package big time in #2476 and the ITP bug in Debian [was recently updated](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=754129#44) as well. There is still a [TODO file](https://github.com/lutris/lutris/blob/master/debian/TODO), but I couldn't find a bug report.

The TODO file currently contains the following items:
- review copyright file and remove the copyright.old for the svgs
- create man-page for lutris
- create lintian-override or man-page for lutris-wrapper

I would like to help out on this if possible, is any help needed? I think the first item has probably been resolved with #2542. Is there any documentation on what lutris wrapper does?

sharkwouter:
Just to add some information, I am not a Debian contributor. I just have some experience making deb packages and I've read the new maintainers guide.

stephanlachnit:
As you said I already tried to get the package into Debian. I filled an [RFS](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=945588), but it looks like no DD or DM is currently interested in sponsoring (or even reviewing) the package. I'll still "maintain" it though, with the source on [salsa](https://salsa.debian.org/stephanlachnit-guest/lutris) and the dsc on [mentors](https://mentors.debian.net/package/lutris).

You could look into the man pages, that would be the the easiest thing to do here in upstream, but actually it's not that important for an official upload.

The major issue is still the copyright, that's why I leaft it in the TODO file. I checked everything as good as I could but without a review from the [debian-legal](https://lists.debian.org/debian-legal/) mailing list I don't see high chances for an official upload (copyright file could be fine but lutris still not DFSG compliant). Tbh I haven't asked on the mailing for a review though, you could do that if you want.

stephanlachnit:
So I have some good and some bad news considering the an official Debian package. @JohnScott623 did a [review](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=945588#14), and the svgs are not DFSG-compliant. However, here's the good news: that doesn't block an upload, Debian could just ship the package without the svgs ([explanation](https://wiki.debian.org/DebianMentorsFaq#What_does_.2BIBw-dfsg.2BIB0_or_.2BIBw-ds.2BIB0_in_the_version_string_mean.3F)). So overall, here's a little to-do list:

More important stuff:
[...]

PR #2576, opened 2020-01-06
More Debian work

stephanlachnit:
As mentioned in #2553, this will make Lutris use the system installtion of `distro.py`, removes the `debian/TODO` file and "for GNU/Linux" from the short description in `debian/control` and adds winetricks to recommends.


stephanlachnit:
Ah I forgot add python3-distro to the dependecies for travis.

stephanlachnit:
I had to change the travis image from xenial to bionic since Ubuntu 16.04 doesn't have `python3-distro`. I think dropping xenial is reasonable anyway, considering it's not just quite old but also almost EOL (only 1 year maintance updates and 20.04 eLTS will be released soon anyway).

strycore:
I made the distro optional since that change is likely to cause issues on other distributions.

PR #2604, opened 2020-01-17
Move lutris wrapper

sharkwouter:
This should be all the work required to move lutris-wrapper to ``/usr/share/lutris/bin/``. This is needed for #2553. Please do test this before merging, since I'm not familiar enough with Lutris to know every scenario in which this script might be used.


sharkwouter:
Does anything need to change in this PR? It's been here for a while, but it hasn't been reviewed yet.

strycore:
I'm a bit afraid of unforeseen consequences on this change but I guess we'll never know if it doesn't get merged.

PR #2620, opened 2020-02-08
Added PAL and Region runner options

fdipilla:
I was unable to append the  PAL mode (-P) and region (-R) to the runner command using the existing options, so I figured it will be nice to have these as of the runner options.

### DIFF
--- a/lutris/runners/dgen.py
+++ b/lutris/runners/dgen.py
@@ -16,7 +16,16 @@ class dgen(Runner):
         }
     ]
     runner_options = [
-        {"option": "fullscreen", "type": "bool", "label": "Fullscreen", "default": True}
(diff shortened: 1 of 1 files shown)


fdipilla:
Sure no problem!
I think I've done it correctly but I'm unsure (first time doing a squash over a PR)

telanus:
Do people still use dgen. I know there was talk to remove it, since Genesis Plus GX and Picodrive (via RetroArch) is way better

fdipilla:
> Yep, you did it right. Can't say anything about the PR since I don't know how dgen works, but looks good else. Is there any reason to do an extra option for PAL if one can also choose it as region?

Now that you mention it, it does sound redundant.
Should I do the change or close the PR because of what @telanus said?

PR #2654, opened 2020-03-13
Updated icon path and typo in readme

egee-irl:
I was working on my own library and noticed that the path listed on the readme for icons is incorrect.

As per [this](https://github.com/lutris/lutris/blob/master/lutris/settings.py#L26) line in `settings.py`, the location Lutris looks for icons is the user's local icon store.


strycore:
Thanks! :+1:

PR #2683, opened 2020-03-24
Check more paths for lutris-wrapper script

### LINKED ISSUE #2681: Commit 9102071 Breaks Gentoo compatibility
Using a relative path breaks Gentoo compatibility.

Due to Gentoo's ability to have multiple versions of python installed simutaneously, they do some symlink resolution magic that determines which python interpreter should be used for a package based on the constraining flags applied at installation.

Because of this lutris script is not located in /usr/bin/lutris. This is a symlink to the resolver that then performs its magic. This causes lutris to fail to be able to call the python-wrapper script.

Recommend non-relative path to lutris-wrapper.

sharkwouter:
This should fix #2681


strycore:
It will break again, eventually. It's only a matter of time. This will do for now...

tannisroot:
This PR causes test failures in Travis and packaging failure on Arch with lutris-git package.

PR #2705, opened 2020-04-02
improve is_dxvk_dll check

Riesi:
Some new wine dlls will break on the old check, because they exceed 256k.


Riesi:
superseded by #2706

Issue #2707, opened 2020-04-02
When sandboxing is disabled symbolic links for non-default user directories are removed


wolfmanx:
**Describe the bug**

When sandboxing is disabled and the user directories are translated into another locale than the default, the symlinks are removed in the section "Security: Remove other symlinks" of method WinePrefixManager.desktop_integration in file util/wine/prefix.py.

The reason is the difference between "desktop_folders", which is used for creation and "DEFAULT_DESKTOP_FOLDERS", which is used for security deletion. E.g.:

```
#    :DBG:    desktop_folders: ]['Desktop', 'Meine Dokumente', 'Eigene Musik', 'Eigene Videos', 'Eigene Bilder'][
#    :DBG:    DEFAULT_DESKTOP_FOLDERS: ]['Desktop', 'My Documents', 'My Music', 'My Videos', 'My Pictures'][
```
Since older versions of wine (lutris-5.2) recreate the symbolic links, the error goes undetected. However, newer versions of wine (lutris-5.4) do no longer touch the user directories, if they exist.

**Expected behavior**

The user directory symlinks should be left alone.

**Current behavior**

The user directory  symlinks are deleted and replaced by real directories.

wolfmanx:
BTW: wine(1) version lutris-5.5.2 also creates a symbolic link for "Downloads", which is removed by the lutris launcher for "Secutrity reasons".
IMHO there should be no removal of symlinks at all, if sandboxing is disabled.

wolfmanx:
This is fixed by PR  #2699.

PR #2713, opened 2020-04-04
Use correct desktop_folders for symlink removal

wolfmanx:
The same set of desktop folders used to create symlinks must be used to remove other symlinks.

This fixes issue  #2707.


wolfmanx:
This is a duplicate of PR  #2699.

PR #2714, opened 2020-04-04
flake8 fixes

connectety:
This PR fixes these flake8 errors:
- unused imports
- incorrect amount of blank lines
- missing whitespace around arithmetic operator

These are all flake8 errors except:
- all lines too long lines with less than 100 characters. 
- imports which are not on top, because the GLib-Interface version has to be defined before importing (E402)
- line breaks before binary operator, because they are somewhat inconsistently linted for compatibility reasons (W503).

Maybe we should add a flake8 test to the PR checks with the following config:
```
[flake8]
ignore = E402, W503
max-line-length = 100
```

### DIFF
--- a/lutris/migrations/d9vk_to_dxvk.py
+++ b/lutris/migrations/d9vk_to_dxvk.py
@@ -1,8 +1,5 @@
-from lutris.pga import PGA_DB, get_games
+from lutris.pga import get_games
 from lutris.game import Game
-from lutris.util import sql
-from lutris.util.log import logger
-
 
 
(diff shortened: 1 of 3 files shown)


strycore:
We currently use CodeFactor for code quality check on Github. I would like to eventually run Pylint checks on the whole project but there are too many warnings at the moment. 
There is a flake8 check running with `make test` and that should be free of any warning. I do have your suggested config in my `~/.config/flake8` file but it would indeed be good to have that documented in the project.

PR #2718, opened 2020-04-05
util.monitor mini-cleanup

connectety:
This PR is a tiny cleanup of lutris.util.monitor.
Is this needed? probably not, feel free to close this PR.

In this PR `iterate_all_processes` is made static, the docstring in `is_game_alive` is defined correctly with 3 quotations marks.
The thing I am most split about is using `return next(ITER, None) is None` to check if an iterator returns something instead of using a for loop. I think this increases readability, but I am not certain about it.


strycore:
Any patch that gets rid of pylint warnings is welcome!

PR #2725, opened 2020-04-08
improve meson support

stephanlachnit:
I improved the meson support a bit. It already does the same like pybuild on my setup, so if you want I can also adjust the `debian/` files to use meson.
Looking at #728 I also prepared i18n, but I disabled it for now. Maybe I'll also look into doing the actual gettext implementation.


stephanlachnit:
@strycore I think I got the `.ui` files working for translations now :)
I had to remove the old translations though, they were just so out of date that no string was correct anymore. Better no translation than a wrong one.
The translations only work when installed with meson, so eventually the default build system should be meson and pybuild removed. I didn't do it in this PR though, let me know when you're interested to switch, I can do the `debian/` stuff.

strycore:
That would be really sweet if we can get i18n working soon!

strycore:
RIP existing translations

Issue #2719, opened 2020-04-05
Steam game overlay broken with lutris shortcut 


buxel:
I'm sorry if this is considered rude, but i also posted the issue to [ValveSoftware/steam-for-linux ](https://github.com/ValveSoftware/steam-for-linux/issues/7012) because it's hard for me to tell what exactly causes the problem.

#### Your system information

* Steam client version (build number or date): Mar 26 2020
* Distribution (e.g. Ubuntu): Manjaro
* Opted into Steam client beta?: [Yes/No] No
* Have you checked for system updates?: [Yes/No] Yes

#### Please describe your issue in as much detail as possible:
I added a non-steam game (lutris shortcut) and it starts as expected. I **expected** the overlay to open when pressing shift-tab. The game is "Totally reliable delivery service", running unity engine `2018.4.14f1` trough wine version `lutris-5.5-2`.
What **did happen**, is that some kind of invisible overlay opened, which prevented any further input. The game/music kept running.

First, i was investigating, how lutris could interfere but after trying many suggestions from reddit/github/forums in vain, i took a closer look at steam itself. 

It looks like it fails to load **both**, 32bit and 64bit, versions of `gameoverlayrenderer.so`. Full log [here](https://gist.github.com/buxel/7b0ffab12c9cf9283f563067e6c9771c).
```
ERROR: ld.so: object '/home/buxel/.local/share/Steam/ubuntu12_32/gameoverlayrenderer.so' from LD_PRELOAD cannot be preloaded (wrong ELF class: ELFCLASS32): ignored.
ERROR: ld.so: object '/home/buxel/.local/share/Steam/ubuntu12_64/gameoverlayrenderer.so' from LD_PRELOAD cannot be preloaded (wrong ELF class: ELFCLASS64): ignored.
```
The logs also tell something about losing focus but I'm not sure what to make of that.

I understand that depending on the game one or the other might fail. **But could having both fail result in the issue I'm having?**. I assume, i accidently created a weird environment for steam, which breaks the overlay but I'm running out of clues where to look. `gameoverlayrenderer.so` is probably just the first lead.

I installed `steam-native` in order to get the gamepad working. The overlay didn't work before that either.

#### Steps for reproducing this issue:
1. Add a game (e.g. 0-AD) to Lutris
2. Create a desktop shortcut
3. Add the shortcut to steam
4. Start the game
5. Open the overlay using shift-tab

buxel:
[...]

Issue #447, opened 2016-11-29
Add lutris games as non-steam games


strycore:
Apparently this project can do it and so should we: https://github.com/sirnuke/steamfootbridge

RobLoach:
How would this work?

testman42:
If I understand correctly, strycore suggests that for example OpenArena, installed trough Lutris, would also appear in your Steam library as non-steam game.
In Steam you can add non-Steam games by pressing "+ ADD A GAME..." in the bottom left corner.

However, I do not see why you would want to (unnecessarily) duplicate the game-launching functionality into multiple game-hub clients.

strycore:
That's exactly it @testman42, I wouldn't used the feature myself but quite a few people are asking for it.

Also, I'll leave this here so I remember to read the source code: https://github.com/scottrice/Ice

mdeguzis:
I believe the core piece of Ice that does this is pysteam, namely within [_shortcut_generator.py](https://github.com/scottrice/pysteam/blob/master/pysteam/_shortcut_generator.py). Check into that as well.

https://github.com/scottrice/pysteam

8BitCerberus:
> However, I do not see why you would want to (unnecessarily) duplicate the game-launching functionality into multiple game-hub clients.

Resurrecting this old thread instead of duplicating a feature request. But one good reason for having the duplicate launching is people who use Steam's in-home streaming either from client to client or using a Steam Link.

Another reason would be for people who use a Steam Controller or use Steam to manage their other/3rd party controller configurations. Launching from Steam gets you access to the overlay so you can manage the configuration. sc-controller can work in many situations, but it's not as fully featured as Steam's own configuration utility.

Fortunately Lutris makes it fairly easy to add individual games as non-Steam games by exporting a .desktop file that we can add to Steam manually. But it would be super convenient to have it directly in Lutris both on a per-game basis, and as a one shot "add all to Steam" option.

RobLoach:
Automatically add shortcuts to all non-Steam Lutris games in Steam... That'd be cool.

Here's how to do it manually:
https://github.com/lutris/lutris/wiki#steam

buxel:
For the poeple coming here from google: some time ago i created [a script](https://gist.github.com/buxel/fa710969bbc650447d924cf8f7d15204) for personal use to do this. Feel free to take & adopt it. 
- can be run multiple times without creating duplicates (using tag/category as identifier)
- uses 'lutris lutris:rungame/{slug}' as target
[...]

PR #2729, opened 2020-04-09
Add Lutris games as steam shortcuts

buxel:
This PR implements functionality to synchronize Lutris games to steam by adding them as a non-steam shortcut as described in #447. 

It is implemented in `lutris.util.steam.shortcut` without being wired up to any UI yet. 

So far, it
- can be run multiple times without creating duplicates (using a tag as identifier)
- uses 'lutris lutris:rungameid/{id}' as target
- adds icons & banners (pixelated, tough)
- leaves user created shortcuts alone

The main use case is for using steam remote play with games managed by lutris.

Open questions
- [ ] Where/when to trigger this functionality
- [ ] Sometimes, the steam overlay won't work when a game is launched via lutris, see #2719 
- [ ] Steam "tags" are not visible in the UI anymore, as they have been replaced by "collections". If tags get deprecated, we should create a "Lutris" collection for imported games. I haven't found the place where those collections are stored.
- [ ] Parts of the  [vdf](https://pypi.org/project/vdf/) package has simply been copied to `lutris/vendor`. I'm not sure if adding a git submodule or a different approach is more suitable


connectety:
> Parts of the vdf package has simply been copied to lutris/vendor. I'm not sure if adding a git submodule or a different approach is more suitable
Can't the vdf package just be added to the requirements?
I think you can just add it to the install_requires in setup.py:

buxel:
As mentioned [here](https://github.com/lutris/lutris/issues/447#issuecomment-609890718), this wouldn't be suitable for Lutris. 
I'm not too happy with this solution, as well (missing tests, updates) but I'll just add the original license file to the directory for now. Any suggestion/contribution concerning this is welcome!

stephanlachnit:
I'm not a big fan of git submodules, they make distributing the source for distro way harder (they aren't included in the source tarballs created by github). If Lutris is going to switch to meson (which was planned IIRC), it would be possible to add this as a [wrap](https://mesonbuild.com/Wrap-dependency-system-manual.html) with a meson patch (for example like [this](https://github.com/stephanlachnit/vulkanheaders-wrap)). This has the advantage that distros can either pack vdf themselves or just install it with the wrap file.

buxel:
This looks like a promising approach. So instead of a submodule/copying code we could just add a wrap file, which would get picked up at build time?

I guess it's time to brush up my local dev environment...

stephanlachnit:
Yep. First you need to check in the meson code (use #2725) wether the vdf module loads, and if not run `subproject('vdf')`.
However, vdf does not have meson as a build system, so it needs to be patched, i.e. a `meson.build` needs to be added.
It would also be possible to add meson support on vdf upstream, maybe they accept it.

strycore:
I will not accept git submodules to be merged in lutris' source code

buxel:
For now, i just took the file. I agree that submodules wouldn't be a good solution. 
Once/if meson is used in Lutris, we can reconsider this approach. Maybe we can get mesos support upstream in the vdf package so using it will be pretty straight forward.

strycore:
No. We do not use git submodules. Ever.

buxel:
I think we all agree here. I was talking about mesos, the solution @stephanlachnit suggested, not submodules.

strycore:
Oh right. The Meson stuff is still alien to me but it seems to be something we should be using in the future.

stephanlachnit:
I guess copying the file is easier for now. I'll look if I can get vdf packed for Debian, it will miss 20.04 LTS so it will be stuck here for at least another two years.

strycore:
I'm glad for the library that it's getting into the Debian repos but that doesn't mean this lib will be available on other distros.
I do not want this vdf stuff added as a Lutris dependency unless it's in the vendor folder. I will personally rewrite the code into lutris if I have to.

buxel:
@strycore i don't think anybody has to rewrite the code here. The single code file has been copied and  @stephanlachnit added the lincense to te proper place.

What I'm really wondering is when this functionality should be triggered and where to add an option for that.

Lucki:
> * I haven't found the place where those collections are stored.

They're stored inside the local storage of the embedded chromium web browser.
Have fun: https://github.com/tkashkin/GameHub/pull/368/files

liberavia:
We should make sure, that shortcuts.vdf exists and therefore know, if steam is fully installed and working
https://github.com/buxel/lutris/blob/0bac943fe1d98f37bd5aeb94c67210b1a59aa5ea/lutris/util/steam/shortcut.py#L14

buxel:
I have not been active on this for quite a while. Times have changed and all my free time is occupied now so i don't have time to update this.If there is anybody willing to pick this up, go for it! If i can help by giving permissions on my repo, let me know (not sure if this is the right way to do it). 
-------- Ursprüngliche Nachricht --------Von: liberavia ***@***.***> Datum: 08.03.22  17:24  (GMT+01:00) An: lutris/lutris ***@***.***> Cc: buxel ***@***.***>, Author ***@***.***> Betreff: Re: [lutris/lutris] Add Lutris games as steam shortcuts (#2729) 
We should make sure, that shortcuts.vdf exists and therefore know, if steam is fully installed and working
https://github.com/buxel/lutris/blob/0bac943fe1d98f37bd5aeb94c67210b1a59aa5ea/lutris/util/steam/shortcut.py#L14

liberavia:
I had a recent view on the code, which is a good starting point IMHO, so I would like to take over. Instead of adding all games to steam, I'd rather have an option in context and installer to add the non-steam-game.

However, as a single dad of three kids, I fully undrestand your point @buxel . Maybe I will have the one or other question while ongoing on this

strycore:
I'd like to get this merged in before the 0.5.10 release because this is quite an important feature for the Steam deck integration.
The only conflict in this PR is the copyright file so I might just go ahead and merge it so we can have a base to build upon.

buxel:
great :) 
@liberavia it is nice to know somebody picks up where i left off, thank you!

liberavia:
ok, I'll start from master then

PR #2742, opened 2020-04-13
Update to current year again

testman42:
number changed since last update


testman42:
Oh, there is one 2019 in [settings.py](https://github.com/lutris/lutris/blob/master/lutris/settings.py#L9) as well. That should be updated too.

PR #2766, opened 2020-04-17
Feature: Add game manually when installer exists

AngryUbuntuNerd:
When installing a game from the Lutris DB, if there is an installer already, I have no option to configure it manually instead, similar to the way it works when there is no installer at all. That way I loose the bonus of using data from the Lutris DB to prefill some fields (cover, release year, etc.)

This change fixes it by adding a button at the bottom of the Choose Installer dialog.

Screenshot from 2020-04-17 23-11-55
Screenshot from 2020-04-17 23-12-00


tannisroot:
I am against this. IMO people are supposed to use Manual installation to test game configurations, various options and winetricks workarounds and then write installers that automate all of this, not as a primary way of installing games... Such feature will give users an idea that our installer db is some side gig and that Lutris is just another GUI for Wine that you're supposed to interact with by following some awful 5 year old guide on Reddit (which what PoL has become) to get anything working and get rid of an incentive for them to look into installer creation themselves.

AngryUbuntuNerd:
Let me give you my specific usecase, which I believe other people might have as well.

I have a bunch of DOS games on the hard drive for which I dont have the original install media (disc, CD) anymore. Pretty much all of them exist in the Lutris DB, so I want to add them manually but keep the nice images and metadata.

If you want to encourage writing installers, I could change the PR to add the "Write installer" button as well, similar to the way it works when there is no installer.

strycore:
There is a legit use for this. Sometimes I have a game already installed from a old system and I just want to add it to Lutris and not go through the installer again.

PR #2769, opened 2020-04-18
Updated the Mame Runer

telanus:
Added support to differentiate between the systems it can run, the Game & Watch, Plug & Play TV games (Jakks Pacific type)  & Tiger handhelds

### DIFF
--- a/lutris/runners/mame.py
+++ b/lutris/runners/mame.py
@@ -58,9 +58,7 @@ class mame(Runner):
         selected_platform = self.game_config.get("platform")
         if selected_platform:
             return self.platforms[int(selected_platform)]
-        else:
-            return ("Arcade")
-        return ""
+        return ("Arcade")
 
     def prelaunch(self):
         if not system.path_exists(os.path.join(self.config_dir, "mame.ini")):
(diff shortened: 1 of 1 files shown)


lutris/runners/mame.py
```diff
@@ -41,6 +54,14 @@ def config_dir(self):
     def working_dir(self):
         return self.config_dir
 
+    def get_platform(self):
+        selected_platform = self.game_config.get("platform")
+        if selected_platform:
+            return self.platforms[int(selected_platform)]
+        else:
+            return ("Arcade")
+        return ""
```
Steinhagen, line 63: This return will never be called. There is already a return for both if and else.

telanus: I found that if I don't include this, current games will loose their platform, if nothing gets manually changed. Also if you do not choose a platform at the time you create the entry, you will end up without a platform, if I remove the "else:". With it in the platform gets auto-assigned to arcade. I tested a few times this morning, trying to get it to auto-convert games already in lutris

telanus: Think I fixed it correctly :D

strycore:
We can get started with that but that doesn't really cover the amount of platforms MAME supports. I'll look into the MESS runner and try to reuse some code if useful.

PR #2777, opened 2020-04-20
(Libretro) Use the info files to build the list of cores

AlexanderRavenheart:


telanus:
One thing I ran into. When you do a "clean" install, removing all the lutris directories in .config .local/shared & .cache Lutris won't start as it can't find the info.zip file

```
  File "./bin/lutris", line 40, in <module>
    from lutris.gui.application import Application
  File "~/src/lutris/lutris/gui/application.py", line 54, in <module>
    from .lutriswindow import LutrisWindow
  File "~/src/lutris/lutris/gui/lutriswindow.py", line 28, in <module>
    from lutris.gui.widgets.sidebar import SidebarListBox
  File "~/src/lutris/lutris/gui/widgets/sidebar.py", line 6, in <module>
    from lutris import platforms
  File "~/src/lutris/lutris/platforms.py", line 17, in <module>
    _init_platforms()
  File "~/src/lutris/lutris/platforms.py", line 12, in _init_platforms
    runner = runners.import_runner(runner_name)()
  File "~/src/lutris/lutris/runners/__init__.py", line 87, in import_runner
    runner_module = get_runner_module(runner_name)
  File "~/src/lutris/lutris/runners/__init__.py", line 81, in get_runner_module
    "lutris.runners.%s" % runner_name, globals(), locals(), [runner_name], 0
  File "~/src/lutris/lutris/runners/libretro.py", line 29, in <module>
    open(get_default_config_path('info.zip'), 'wb').write(req.content)
FileNotFoundError: [Errno 2] No such file or directory: '~/.local/share/lutris/runners/retroarch/info.zip'
```

AlexanderRavenheart:
On a clean install none of the runners are present so the retroarch folder will be missing completely, right? I will update accordingly.

telanus:
> On a clean install none of the runners are present so the retroarch folder will be missing completely, right? I will update accordingly.

Yes, nothing will be installed

AlexanderRavenheart:
Added check for runner dir.

strycore:
That is an awesome PR

matoro:
@AlexanderRavenheart I pulled this from git today and found that it does cause one issue - the naming scheme of some libretro platforms is inconsistent with what Lutris has for other runners.  Namely, Lutris uses "Company Platform" whereas libretro uses just "Platform".  So whereas the platform filters previously looked like:
```
Nintendo Game Boy/Game Boy Color
Nintendo Gamecube
Nintendo NES
Nintendo SNES
Sega Dreamcast
Sony Playstation
Sony Playstation 2
```
It now looks like:
```
Game Boy/Game Boy Color
Nintendo Entertainment System
Nintendo Gamecube
PlayStation
Sega Dreamcast
Sony PlayStation 2
Super Nintendo Entertainment System
```
which is both a lot uglier and causes the icons for those platform filters to no longer be found.  I was wondering if you and @strycore had seen this and possibly had an idea for a solution...

AlexanderRavenheart:
In regards to naming, Nintendo NES (Nintendo Entertainment System), see the double Nintendo, doesn't make sense, thus Libretro naming makes more sense.

matoro:
Well sure, but if that's the case then shouldn't we get rid of the manufacturer name in the other runners?  It just doesn't make sense to have `PlayStation` next to `Sony PlayStation 2`.  Alternatively - we could use the `database` field from the info files instead and add `.replace(" - ", "")` - though this still would end up with platform names like `Nintendo NES / Famicom`.

For the icons, the mapping is from platform name -> icon.  For example - if somebody has a game where they are using the PPSSPP libretro core and another where they are using standalone PPSSPP, they should be grouped under the same platform, and they are not right now.

AlexanderRavenheart:
I would agree to drop the Manufacturer from the other runners. After all, it is a platforms filter, not a manufacturer + platform filter. This would also mean we could have an additional Manufacturer filter added to the sidebar to show only Nintendo, Sony, etc.

matoro:
2020-05-03-130746_245x835_scrot
Okay, so after patching all the other runners to use platform names only, here's what mine looks like.  One thing I did find is that even the libretro info files are not consistent...the `systemname` for DC is still Sega Dreamcast.

Also, this would require a database update, otherwise it would only apply to already-installed games.  I did it by manually opening the configure box for each game and re-selecting the runner/core, but that's not feasible for those with larger collections of course...how would you like to handle that?

AlexanderRavenheart:
It wouldn't require a database update unless the runner/core has changed. If the core name is the same as before then no change is required. Display name used for the combo box and the platform name used for the filter don't impact the selected core/runner.

PR #2789, opened 2020-04-23
Better installation guide

XakepSDK:
Current installation guide may break some distros. Warning about that it's not recommended way to install lutris written too late, some people may execute apt command.


tannisroot:
Superseded by https://github.com/lutris/lutris/pull/2790

PR #2790, opened 2020-04-23
Improve Installation guide

tannisroot:
The changes most notably emphasize that you should install Lutris from a PPA or distros's repository, improve the steps that describe how to install it manually (we now suggest to use virtualenv to avoid breaking the system), but there are also some minor fixes and additions (like Wine being an optional dependency) here and there.

### DIFF
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -5,8 +5,10 @@ Requirements
 ------------
 
 Lutris should work on any up to date Linux system. It is based on Python and
-Gtk but will run on any desktop environment. The following dependencies should
-be installed:
+Gtk but will run on any desktop environment.
+If you installed Lutris from our PPA or some other repository, it should already
+come with all of its essential dependencies. However, if you need to install
+Lutris manually, it requires the following components:
 
     * Python > 3.4
     * PyGObject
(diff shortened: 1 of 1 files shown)


strycore:
Since when do we support virtualenvs? Installing Lutris in a virtualenv has always been frowned upon.
PyGobject is a pain to get working inside a virtualenv or at least it was last time I checked.
Also, you really gotta try hard to break your system by running Lutris from the source folder.

tannisroot:
> Since when do we support virtualenvs? Installing Lutris in a virtualenv has always been frowned upon.

Not sure why that is but it seems to work pretty well ¯\_(ツ)_/¯
> PyGobject is a pain to get working inside a virtualenv or at least it was last time I checked.

Yes, that's why in the guide you create virtualenv that uses system libraries (and there's a note that you need to install dependencies system-wide). With them Lutris works just fine in virtualenv.
> Also, you really gotta try hard to break your system by running Lutris from the source folder.

I had a very nasty conversation with a user on VKx server and using our setup.py broke their installation of Ubuntu 20.04, it's the sole reason I made this PR.

strycore:
If people are dumb enough to use the setup.py script while we explicitly discourage its use, there's not much we can do. 
Leaving this script in the hands of incompetent people will likely have some disastrous effect on their system. There's a reason why we discourage its use.
Anyway, I was just worried that virtualenv would cause issues but if they work then I'm fine with this recommendation.

PR #2845, opened 2020-05-07
Give more informative error message for unset USER

FlyingWombat:
If USER is unset, this error in lutris/util/wine/prefix.py:98:
    `TypeError: expected str, bytes or os.PathLike object, not NoneType`
appears when attempting to install or launch a Wine game.
This changes that to manually raise a ValueError with appropriate message.

An alternative to this could be using the `subprocess` module to run `whoami` if USER is unset. I chose sticking with the error message because it is less complicated than running a subprocess, and USER should probably be set in the user's env anyway.

Note: the only other place this env var is used is lutris/installer/interpreter.py:745  
I don't know if that, or other uses of env vars, need null checks.

### DIFF
--- a/lutris/util/wine/prefix.py
+++ b/lutris/util/wine/prefix.py
@@ -95,6 +95,8 @@ class WinePrefixManager:
         # pylint: disable=too-many-branches
         # TODO: reduce complexity (18)
         user = os.getenv("USER")
+        if not user:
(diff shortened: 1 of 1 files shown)


FlyingWombat:
> Instead of crashing lutris, you can make a default for the username like "lutrisuser"

Oh, ya that's a better idea.
Done

PR #2860, opened 2020-05-11
Fix for bare PICO-8 invocation.

### LINKED ISSUE #2859: Bare PICO-8 invocation throws error
Directly invoking the PICO-8 runner doesn't actually run PICO-8.

The expected behavior is that PICO-8 should launch itself either in its default mode or "Splore" mode depending on the runner's configuration.

The current behavior is that invoking the runner directly doesn't appear to do anything. Behind the scenes it throws a TypeError exception.

To make this happen, just click the little "play" icon next to the PICO-8 runner entry in the sidebar.

```
> lutris -d
Gtk-Message: 00:16:38.297: Failed to load module "appmenu-gtk-module"
INFO     2020-05-11 00:16:38,618 [application.do_command_line:245]:Running Lutris 0.5.6
INFO     2020-05-11 00:16:38,619 [startup.check_driver:54]:Using NVIDIA drivers 440.82 for x86_64
INFO     2020-05-11 00:16:38,619 [startup.check_driver:60]:GPU: GeForce GTX 1070 with Max-Q Design
INFO     2020-05-11 00:16:38,619 [startup.check_driver:77]:GPU: 10DE:1BA1 1558:95E1 using nvidia drivers
INFO     2020-05-11 00:16:38,619 [startup.check_driver:77]:GPU: 8086:3E9B 1558:95E1 using i915 drivers
INFO     2020-05-11 00:16:38,665 [startup.check_vulkan:132]:Vulkan is supported
INFO     2020-05-11 00:16:38,666 [dxvk.get_dxvk_versions:22]:Updating DXVK versions
DEBUG    2020-05-11 00:16:38,829 [http.get:65]:GET https://lutris.net/api/runtime
DEBUG    2020-05-11 00:16:38,831 [store.get_missing_media:208]:Requesting missing icons from API for 30 games
DEBUG    2020-05-11 00:16:38,832 [http.get:65]:GET https://lutris.net/api/games
DEBUG    2020-05-11 00:16:39,457 [api.get_game_api_page:132]:Loaded 16 games from page 1
INFO     2020-05-11 00:16:42,194 [runner.use_runtime:243]:Runtime disabled by system configuration
  File "/usr/lib/python3/dist-packages/lutris/runners/runner.py", line 233, in run
    self.prelaunch()
  File "/usr/lib/python3/dist-packages/lutris/runners/pico8.py", line 173, in prelaunch
    if not os.path.exists(self.game_config.get("main_file")) and (
  File "/usr/lib/python3.8/genericpath.py", line 19, in exists
    os.stat(path)
TypeError: stat: path should be string, bytes, os.PathLike or integer, not NoneType
INFO     2020-05-11 00:16:44,462 [application.do_shutdown:491]:Shutting down Lutris
```

The fix is to check for an empty main_file prior to the sensitive point in code and (if PICO-8 is also installed) just return True anyway. I have fixed and manually tested this as well and will submit a pull request including this fix.

Feneric:
Closes #2859

This is a fairly trivial change. It enables the existing PICO-8 runner "play" button to invoke PICO-8 directly without throwing an error.

Oddly I wasn't able to successfully run `make dev` or `make sc` on my machine (I'm running Pop!_OS 20.04) as the former was unable to build wheels for pygobject and the latter threw a ton of `Unable to import 'gi.repository'` sorts of errors. I'll have to dig into both of these to see what's happening, but in the meanwhile this particular fix passes `pep8`, `pylint`, and `flake8` when they are run manually.

### DIFF
--- a/lutris/runners/pico8.py
+++ b/lutris/runners/pico8.py
@@ -154,6 +154,8 @@ class pico8(Runner):
         return system.path_exists(os.path.join(settings.RUNNER_DIR, "pico8/web/player.html"))
 
     def prelaunch(self):
+        if not self.game_config.get("main_file") and self.is_installed():
+            return True
         if os.path.exists(os.path.join(settings.RUNNER_DIR, "pico8/cartridges", "tmp.p8.png")):
             os.remove(os.path.join(settings.RUNNER_DIR, "pico8/cartridges", "tmp.p8.png"))
 


strycore:
If you want to run Pico8 by itself, the way to do it is to set the "runnable_alone" attribute on the class to True.

Feneric:
Oh interesting. Thanks for that info. Something a little deeper is going on as at first blush it looks like it ought to be setting that attribute and yet the play button definitely doesn't work. I'll dig in.

Feneric:
OK, so digging in further, I can confirm that `runnable_alone` is being properly set but the exception I reported earlier is what's preventing it from working, so the fix I provided to shield the test from the `None` does appear to be the correct approach.

strycore:
That's weird that you would have to modify the play method in order to get the runner to launch by itself.
The patch itself is harmless so I can go ahead and merge it, it can always be improved later.

Feneric:
Yeah, in going through a few runners it looks like while many run fine there may be others that have different (but possibly related) problems that prevent them from running even when they are `runnable_alone`. My guess is that maybe not a lot of people use it and it doesn't get tested as much. I only just discovered Lutris when I upgraded Pop!_OS and noticed it prominently listed under games. I wasn't sure how runners were supposed to work (did they make use of pre-existing software, did they do new installs, etc.) and so trying the bare case seemed like a pretty good way to start. Thanks for pulling in the change. I'll keep looking a bit and see what I find.

PR #2869, opened 2020-05-12
Support Valve's futex() extensions as a possible synchronization method

lazerl0rd:
As detailed by Valve [1], their futex() extensions provide an optimal
improvement over the current eventfd-based synchronization implementation.

As well as causing file descriptor exhastion in event-hungry applications,
esync also causes extraneous spinning in the kernel. The futex() extensions
hope to solve this by providing functionality within the kernel for a thread
to wait on multiple futexes at a time.

1: https://lkml.org/lkml/2019/7/30/1399

### DIFF
--- a/lutris/runners/wine.py
+++ b/lutris/runners/wine.py
@@ -239,19 +239,27 @@ class wine(Runner):
                 "option":
                 "esync",
                 "label":
-                "Enable Esync",
+                "Use custom synchronization",
                 "type":
-                "extended_bool",
+                "choice",
+                "choices": [
+                    ("Esync", "esync"),
+                    ("Fsync", "fsync"),
+                    ("None", "none"),
+                ],
                 "callback":
                 esync_limit_callback,
                 "callback_on":
-                True,
-                "active":
-                True,
+                "esync",
+                "default":
+                "none",
                 "help": (
-                    "Enable eventfd-based synchronization (esync). "
-                    "This will increase performance in applications "
-                    "that take advantage of multi-core processors."
+                    "Use custom synchronization to improve performance in multi-threaded "
+                    "applications\n"
+                    "<b>Esync</b>: Use eventfd-based synchronization (esync).\n"
+                    "<b>Fsync</b>: Use Valve's extensions to the futex() syscall "
+                    "(requires a custom kernel witht the fsync patchset).\n"
+                    "<b>None</b>: Use the default synchronization mechanism"
                 ),
             },
             {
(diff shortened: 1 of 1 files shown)


lazerl0rd:
The force-pushes should take care of all the linting issues.

lazerl0rd:
An example of the comparison between esync and fsync can be seen with this graph by Valve:

image

lazerl0rd:
@strycore I may be wrong but Valve's futex() extensions are meant to replace esync. I believe WINE automatically enables futex() extensions if using a supported kernel, but esync may override this.

If so, we could just do a check if extended futex() is supported under the esync loop and off esync. I'd need your conformation on this, however.

I think I'd need some more clarity on the matter TBH, if we are to add a secondary option as opposed to my proposal.

GloriousEggroll:
fsync is not enabled by default. Proton uses WINEFSYNC to enable it. If both WINEESYNC and WINEFSYNC are enabled, and the kernel supports fsync, fsync runs instead of esync and takes priority.

That being said, fsync should have its own toggle, because fsync may not work in some games while esync does, thus presenting a case where esync needs to be enabled but fsync disabled, even if the system is fsync capable.

lazerl0rd:
@GloriousEggroll thanks for the on about Proton and WINEFSYNC. However, I’m not sure about your statement that “fsync might not work where esync does work”. How does that work?

strycore:
that's not even the problem. All our installers use a value of esync: true, it expects a boolean.
changing that to a string is not possible.
I agree with @GloriousEggroll , both options should have their own toggle

lazerl0rd:
> that's not even the problem. All our installers use a value of esync: true, it expects a boolean.
> changing that to a string is not possible.

I understand that, but to have two options I need to know if I should disable esync when “fsync” is set on or not. Also, do we have a way to “grey out” the esync option if this is necessary?

strycore:
no, don't bother with greying out anything, that is more complicated to implement.

GloriousEggroll:
> @GloriousEggroll thanks for the on about Proton and WINEFSYNC. However, I’m confused about your statement that “fsync might not work where esync does work”. How does that work?

Wine builds can have both esync and fsync patched in.

A game may break with fsync and not esync, OR may break with esync and not fsync.

In which case you would need individual toggles for both instances, such as WINEESYNC=1 WINEFSYNC=0 or vice versa.

soredake:
cool!

strycore:
I merged PR #2893 which implements the fsync toggle so I'm going to close this PR.

PR #2875, opened 2020-05-16
Snap progress

diddledan:
I moved the `snapcraft.yaml` into `snap/` so that if any snap-specific files need to be added then they can be contained there inside `snap/local/` similarly to how the `debian/` directory holds all deb-specifics.

With these changes, I can successfully start lutris, install at least one game through steam and launch that game, login to gog.com, and login to lutris.net. Despite these successes it is still very brittle and not ready for prime-time as a snap yet.

Signed-off-by: Daniel Llewellyn <daniel@bowlhat.net>


strycore:
Thanks!

PR #2879, opened 2020-05-16
Enforce runner play method verification before runtime

Steinhagen:
This check makes sure that all derived runners will define the play method before execution. This is extremely helpful since the end goal would be to make runners independent of Lutris.

The metaclass approach was chosen instead of __init_subclass__ to keep compatibility with the current minimum python version.

### DIFF
--- a/lutris/runners/runner.py
+++ b/lutris/runners/runner.py
@@ -19,7 +19,7 @@ from lutris.util.log import logger
 
 class RunnerMeta(type):
     def __new__(mcs, name, bases, body):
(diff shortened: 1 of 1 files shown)


AlexanderRavenheart:
This check would make any runner that doesn't have play implemented to raise a TypeError during it's initialization, right? And aren't all runners initialized at Lutris startup, effectively making Lutris not run at all if a runner doesn't have play implemented?

strycore:
I do not believe in the value of this patch but it looks harmless enough to be merged in.

PR #2888, opened 2020-05-18
Improve compositor handling.

openglfreak:


openglfreak:
Made after 0 in the morning, please forgive me if I wrote nonsense somewhere.

lutris/util/display.py
```diff
@@ -130,37 +130,121 @@ def get_display_manager():
 USE_DRI_PRIME = len(_get_graphics_adapters()) > 1
 
 
-def get_compositor_commands():
-    """Nominated for the worst function in lutris"""
+# Pretend this is an enum.
+class DE:
```
AlexanderRavenheart, line 134: Why pretend? Just make it a proper enum: https://docs.python.org/3.6/library/enum.html#creating-an-enum

lutris/util/display.py
```diff
@@ -130,37 +130,121 @@ def get_display_manager():
...
+    """Enum of desktop environments."""
+
+    PLASMA = object()
+    MATE = object()
+    XFCE = object()
+    DEEPIN = object()
+    UNKNOWN = object()
+
+
+def get_desktop_environment():
```
AlexanderRavenheart, line 145: I don't recall where but I think we already have a function that determines what the current desktop environment is. Maybe @strycore knows where it is.

lutris/util/display.py
openglfreak: I did not use enum.Enum because afaict nothing else in lutris uses the enum module, but if you say it's okay I'll make a commit that converts this class to an Enum.

lutris/util/display.py
openglfreak: `grep -iF DESKTOP_SESSION -r .` did not yield any other place where DESKTOP_SESSION is read, so I don't think there's any other place.

lutris/util/display.py
```diff
@@ -130,37 +131,120 @@ def get_display_manager():
 USE_DRI_PRIME = len(_get_graphics_adapters()) > 1
 
 
-def get_compositor_commands():
-    """Nominated for the worst function in lutris"""
+class DE(enum.Enum):
```
strycore, line 134: DE is not a valid class name
Also, you don't seem to be using a code editor that tells you that DE is not a valid class name.

lutris/util/display.py
AlexanderRavenheart: False alarm, it was this function that you are rewriting.

lutris/util/display.py
AlexanderRavenheart: Rename to DesktopEnvironment.

openglfreak: In my defense, yes I am using a code editor that doesn't show `DE` as an invalid class name (Visual Studio Code), but only because the underlying linter (pylint) does not report the class name as invalid.
I also ran flake8 on the file and got no errors, and I ran `make test` which threw no errors.

I will change the class name now, give me a few seconds.

strycore: That is weird, Pylint should totally report this as an invalid name.

lutris/util/display.py
```diff
@@ -130,37 +131,120 @@ def get_display_manager():
...
+    if desktop_environment is DesktopEnvironment.PLASMA:
+        stop_compositor = ("qdbus", "org.kde.KWin", "/Compositor", "org.kde.kwin.Compositing.suspend")
+        start_compositor = ("qdbus", "org.kde.KWin", "/Compositor", "org.kde.kwin.Compositing.resume")
+    elif desktop_environment is DesktopEnvironment.MATE:
+        stop_compositor = ("gsettings", "set org.mate.Marco.general", "compositing-manager", "false")
+        start_compositor = ("gsettings", "set org.mate.Marco.general", "compositing-manager", "true")
+    elif desktop_environment is DesktopEnvironment.XFCE:
+        stop_compositor = ("xfconf-query", "--channel=xfwm4", "--property=/general/use_compositing", "--set=false")
+        start_compositor = ("xfconf-query", "--channel=xfwm4", "--property=/general/use_compositing", "--set=true")
+    elif desktop_environment is DesktopEnvironment.PLASMA:
```
xordspar0, line 214: This must be a copy/paste mistake.

```suggestion
    elif desktop_environment is DesktopEnvironment.DEEPIN:
```

openglfreak: Yup, thanks for catching that!

xordspar0:
I think one side effect of this is that compositor enable/disable commands will no longer be logged. Previously the commands used lutris.util.system.execute, which debug logs commands. I'm not sure if this is intended or important.

PR #2893, opened 2020-05-23
Implement fsync toggle

openglfreak:
Mostly a copy&paste of the esync option.

Supersedes #2869

### DIFF
--- a/lutris/runners/wine.py
+++ b/lutris/runners/wine.py
@@ -1018,7 +1018,7 @@ class wine(Runner):
         else:
             wine_prefix.desktop_integration(restore=True)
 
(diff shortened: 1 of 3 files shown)


openglfreak:
Why are some merged commits listed???
I'll try rebasing onto master

openglfreak:
That worked.

robzombie91:
Will this option be disabled like gamemode if fsync is not detected in the current wine build or will it just throw an error if its not detected during startup?

openglfreak:
@robzombie91 It will be disabled if fsync is not detected in the Wine version.

PR #2903, opened 2020-05-26
add imports needed for wine installers to function properly

Jan200101:
wine installers that use set_regedit or set_regedit_file (and possibly others too) fail because of the missing imports


AlexanderRavenheart:
I already opened a PR for this 3 hours ago: https://github.com/lutris/lutris/pull/2902

Jan200101:
woops, my bad

PR #2896, opened 2020-05-24
Update Croatian language and README.md

muzena:

PR #2947, opened 2020-06-17
Full i18n support: includes almost all translatable content in all interfaces

YihaoPeng:
* mark almost all translatable content in python codes with `_()` or `gettext.ngettext` (for plurality).
* `meson` can extract translatable content in code to the `.po` files.
* `meson` can build and deploy translation files (`*.mo`) to the correct directory.
* `lutris` can load `.mo` translation files from the current installation location, not just `/usr/share/locale`.
* Translation is effective for all windows.

A complete simplified Chinese (zh_CN) translation likes this:

截图录屏_20200617105040


stephanlachnit:
I'm against the autogenerated files in `po/`. This allows for example to remove an old broken translation, without deleting it for future translators.
Also I think you should separate the `po` changes from the Chinese translation in different PRs for better diffs.

po/meson.build
```diff
@@ -1,6 +1,12 @@
...
 
+run_target(
+  'build',
+  command: './autogen.sh'
+)
+
 i18n.gettext(
   'lutris',
   install_dir: localedir,
+  args: '--from-code=UTF-8'
```
stephanlachnit, line 11: This isn't necessary

.gitignore
```diff
@@ -21,6 +21,9 @@ builddir
 
 # i18n files
 po/lutris.pot
+po/POTFILES
+po/LINGUAS
+po/*.mo
```
stephanlachnit, line 26: `.mo` files won't be generated in the `po` dir.

bin/lutris
```diff
@@ -27,16 +27,26 @@ if os.path.isdir(os.path.join(LAUNCH_PATH, "../lutris")):
...
     sys.stderr.write("Unsupported locale setting. Fix your locales\n")
 
+try:
+    # Application install prefix
+    # Example:
+    #   /usr/bin/lutris/../..         -> /usr
+    #   $HOME/.local/bin/lutris/../.. -> $HOME/.local
+    prefixPath = os.path.realpath(os.path.join(__file__, os.path.pardir, os.path.pardir))
+
+    localDir = "%s/share/locale" % prefixPath
```
stephanlachnit, line 42: I think this is pretty sketchy. It might work in all default configurations, but I'm not a fan of it. The proper way would be to configure a file via Meson, and take the localedir from there. For example like here: https://github.com/cvfosammmm/Setzer/blob/master/meson.build
Since we don't want to break the current build system just now, I think it's best to make that file "optional", i. e. design it that if it's not configured the localedir won't be set.
Also, I think `localedir` is sounds better than `localDir`, since that is what Meson uses as built-in option.

po/POTFILES
```diff
@@ -1,8 +0,0 @@
-share/lutris/ui/about-dialog.ui
-share/lutris/ui/dialog-lutris-login.ui
-share/lutris/ui/dialog-pga-sources.ui
-share/lutris/ui/dialog-uninstall-game.ui
-share/lutris/ui/runner-entry.ui
-share/lutris/ui/runner-remove-all-versions-dialog.ui
-share/lutris/ui/runner-remove-confirm-dialog.ui
-share/lutris/ui/runners-dialog.ui
```
stephanlachnit, line 8: Please keep POTFILES. `autogen.sh` is more than sketchy, not every `.py` file is translatable. Also, one might want to add i18n support for the metainfo or .desktop file.

po/LINGUAS
```diff
@@ -1 +0,0 @@
-fr
```
stephanlachnit, line 1: As already mentioned, please restore the LINGUAS files. It's fine to remove all current translations in there since they are useless anyway.

po/README.md
```diff
@@ -1,14 +1,15 @@
 # i18n
 
-Translations are not implemented yet, see github issue #728. Please read the notes before opening a PR.
+i18n build only works with the meson build system. See github issue #728 for more details.
```
stephanlachnit, line 3: Honestly, if this will be merged, I will open a PR to default to Meson (change is fine I just wanted to mention it).

po/meson.build
YihaoPeng: > This isn't necessary

This **is** necessary because there are some Chinese full-width ellipsis in the python code (`…`, `\u2026`). I was surprised by this but it does exist:

image

Without this you will get some errors:

```
xgettext: Non-ASCII string at /parent/work/winegame/lutris/./lutris/gui/dialogs/runner_install.py:222.
          Please specify the source encoding through --from-code or through a comment
          as specified in http://www.python.org/peps/pep-0263.html.
```

.gitignore
YihaoPeng: > `.mo` files won't be generated in the `po` dir.

If you use `poedit` or other tools to edit the `.po` file, the corresponding `.mo` will be generated in this directory by default. Although I can set the software to stop the generating, but I just want to give myself a convenience. Others may also get help.

po/POTFILES
YihaoPeng: Have added `POTFILES` back and deleted `autogen.sh`.

po/LINGUAS
YihaoPeng: Have added LINGUAS back.

po/README.md
YihaoPeng: I rolled back the README changes

bin/lutris
YihaoPeng: > I'm against the autogenerated files in `po/`. This allows for example to remove an old broken translation, without deleting it for future translators.
> Also I think you should separate the `po` changes from the Chinese translation in different PRs for better diffs.

Chinese translation file removed

YihaoPeng: > I think this is pretty sketchy. It might work in all default configurations, but I'm not a fan of it. The proper way would be to configure a file via Meson, and take the localedir from there. For example like here: https://github.com/cvfosammmm/Setzer/blob/master/meson.build
> Since we don't want to break the current build system just now, I think it's best to make that file "optional", i. e. design it that if it's not configured the localedir won't be set.
> Also, I think `localedir` is sounds better than `localDir`, since that is what Meson uses as built-in option.

I done the change. Now it will find locale dir from the optional configuration file generated by meson.

89ddec3c138ffc508145d17bdb32faee749470b3

po/README.md
YihaoPeng: I rolled back `README.md` too when I rolled back `autogen.sh` changes. You can change it yourself later. 😃

.gitignore
stephanlachnit: Hm that's weird. But I guess it doesn't hurt.

po/meson.build
stephanlachnit: That's super strange. Never had that problem with non-ascii characters (like ä, å, etc), but I guess Chinese is a bit different from that.

po/README.md
```diff
@@ -35,8 +34,8 @@ and commit your changes.
 ## Notes
 
 - Only commit changes for the translation file you actually edited.
-- Ignore or delete the first four lines (copyright notice) in the `$LANG.po` files.
```
stephanlachnit: Honestly I think that line was quite nice, like the rest of the readme changes. Just the line below should be changed.

bin/lutris
```diff
@@ -27,16 +27,35 @@ if os.path.isdir(os.path.join(LAUNCH_PATH, "../lutris")):
...
+    sys.stderr.write(
+        "Unable to load optional_settings, locale dir is not set, "
+        "translations may not work.\n"
+    )
+
+try:
+    locale.textdomain("lutris")
+    gettext.textdomain("lutris")
+except:
+    sys.stderr.write("Couldn't set gettext domain, translations won't work.\n")
```
stephanlachnit, line 57: I guess this can be merged into one try and except:
```python3
try:
    # optional_settings does not exist if you don't use the meson build system
    from lutris import optional_settings

    try:
        locale.bindtextdomain("lutris", optional_settings.LOCALE_DIR)
        gettext.bindtextdomain("lutris", optional_settings.LOCALE_DIR)
        locale.textdomain("lutris")
        gettext.textdomain("lutris")
    except:
        sys.stderr.write(
            "Couldn't bind gettext domain, translations won't work.\n"
            "LOCALE_DIR: %s\n" % optional_settings.LOCALE_DIR
        )
except ModuleNotFoundError:
    sys.stdout.write(
        "Unable to load locale dir, translations won't work.\n"
    )
```
There is no real need to print out to stderr, since for now Meson isn't the default.

stephanlachnit:
Pretty cool with the Meson stuff! Once this is merged, Lutris should definitely switch to Meson @strycore, I can do the PR and the `debian/` stuff.

bin/lutris
YihaoPeng: > I guess this can be merged into one try and except:
> 
> ```python
> try:
>     # optional_settings does not exist if you don't use the meson build system
>     from lutris import optional_settings
> 
>     try:
>         locale.bindtextdomain("lutris", optional_settings.LOCALE_DIR)
>         gettext.bindtextdomain("lutris", optional_settings.LOCALE_DIR)
>         locale.textdomain("lutris")
>         gettext.textdomain("lutris")
>     except:
>         sys.stderr.write(
>             "Couldn't bind gettext domain, translations won't work.\n"
>             "LOCALE_DIR: %s\n" % optional_settings.LOCALE_DIR
>         )
> except ModuleNotFoundError:
>     sys.stdout.write(
>         "Unable to load locale dir, translations won't work.\n"
>     )
> ```
> 
> There is no real need to print out to stderr, since for now Meson isn't the default.

Done the change. But it should be an `ImportError`:
```
  File "/home/hu60/.local/bin/lutris", line 37, in <module>
    from lutris import optional_settings
ImportError: cannot import name 'optional_settings' from 'lutris' (/home/hu60/.local/lib/python3.7/site-packages/lutris/__init__.py)
```

po/README.md
YihaoPeng: > Honestly I think that line was quite nice, like the rest of the readme changes. Just the line below should be changed.

added it back.

YihaoPeng:
> just a style improvement: put the prefix above all other dirs (after `# Set folders`).

done.

muzena:
@YihaoPeng 
Is localization now working?
I translated about 40% of hr.po in #2896 commit.
I tried to test it, cloning locally on PC update hr.po with latest lutris.pot., add hr to LINGUAS and build lutris.deb package.
I've install lutris.deb but localization dosen't work, Lutris GUI is not translated.

PR #2962, opened 2020-06-23
Added Dutch translation

Vistaus:
Now it should be all good to go :)

### DIFF
--- a/po/POTFILES
+++ b/po/POTFILES
@@ -2,6 +2,7 @@ share/lutris/ui/about-dialog.ui
 share/lutris/ui/dialog-lutris-login.ui
 share/lutris/ui/dialog-pga-sources.ui
 share/lutris/ui/dialog-uninstall-game.ui
+share/lutris/ui/lutris-window.ui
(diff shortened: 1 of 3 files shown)


po/POTFILES
```diff
@@ -2,6 +2,7 @@ share/lutris/ui/about-dialog.ui
 share/lutris/ui/dialog-lutris-login.ui
 share/lutris/ui/dialog-pga-sources.ui
 share/lutris/ui/dialog-uninstall-game.ui
+share/lutris/ui/lutris-window.ui
```
stephanlachnit, line 5: Could be that this will conflict with #2961.

Vistaus: But in my previous PR, @YihaoPeng asked me to include this to make the missing strings available, so I assume it'll be okay then.

strycore:
Thanks a lot!

po/POTFILES
YihaoPeng: > Could be that this will conflict with #2961.

Two identical changes will not cause conflicts, git will handle these intelligently.

PR #2982, opened 2020-07-01
Fix copy paste issue in auth webviews

### LINKED ISSUE #2132: Cannot paste GoG login credentials into login helper window
Using `lutris-0.5.2.1` I cannot paste my GoG login credentials into the GoG login helper window. This is quite annoying as I have a password with 40+ characters and typing such a password is quite error-prone.
Neither Linux middle mouse button works nor CTRL+V but typing from keyboard works.

malefice:
I have tried debugging with `Gtk.Clipboard`'s `owner-change` signal, and copying text from other apps, including KeePassXC, triggers this signal. Calling `Gtk.Clipboard.wait_for_text()` from inside the signal handler does output the copied text in other parts of Lutris except in these auth webviews where I get `None` if I copy from KeePassXC.

I also tried replacing the `WebConnectDialog`'s content with a simple `Gtk.Entry`, and I can copy paste fine from KeePassXC. There is really something that interferes with the GTK clipboard if a `WebKit2.WebView` is inside a `Gtk.Dialog`, and if the containing dialog was displayed with `Gtk.Dialog.run()`. Using `Gtk.Dialog.set_modal(True)` and `Gtk.Dialog.show()` instead works fine, even copy pasting from KeePassXC.

Fixes #2132

### DIFF
--- a/lutris/services/gog.py
+++ b/lutris/services/gog.py
@@ -279,7 +279,8 @@ def connect(parent=None):
     """Connect to GOG"""
     logger.debug("Connecting to GOG")
     dialog = WebConnectDialog(SERVICE, parent)
-    dialog.run()
+    dialog.set_modal(True)
+    dialog.show()
 
 
 def disconnect():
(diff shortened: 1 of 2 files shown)


strycore:
Good research and patch made as small as possible, really nice!

Issue #2832, opened 2020-05-03
[Icons] Mixed naming schemes for platform icons


AlexanderRavenheart:
There are currently 3 types of platform names defined in Lutris runners:
- Manufacturer + Abbreviated system name (Ex: Nintendo NES, Nintendo SNES)
- Manufacturer + System name (Ex: Nintendo GameCube, Sony PlayStation, Nintendo Game Boy Advance)
- System  name (Ex: Nintendo Entertainment System, GameCube, Wii, Game Boy, Game Boy Color)

Some runners use only one of the first two types while others (Ex: mednafen) use both.
The third type is used by Libetro cores. 

We should decide and conform to just one of the three types of platform names.
Personally I vote for using only the full system/platform name (third type) as the icons are platform specific and should not contain manufacturer information in their naming.

AlexanderRavenheart:
@matoro Opened this issue to continue the discussion that was opened in the PR.

matoro:
@AlexanderRavenheart I am sure this is going to need some review from you all.  But hopefully https://github.com/lutris/lutris/pull/2990 is a good starting point for making the naming scheme more flexible and getting the platform name -> icon mapping right.  What do you think?

AlexanderRavenheart:
@matoro Hey, I just checked out the PR you linked and I think adding options is the wrong way to go. We need a standard, to align all runners to it and reduce the maintainability effort when cores are added/updated and not keep lists upon lists of hard coded names in the code.

Can you explain which of the 3 variants mentioned in the description of this issue did you make your PR to support as default?

lukasfink1:
The problem with the libretro runner is that it extracts the platform name from the chosen libretro core.
To me it seems as though the only complete fix would be to add an option to manually specify the platform.
It could of course try to translate the libretro platform names to the names we choose. The problem with this approach is that some cores support multiple platforms (e.g. mesen-s supports SNES, GB, GBC).

strycore:
To avoid licensing issues for the inclusion of Lutris in Debian, I've removed all icons from the project and moved them to an icon pack that gets downloaded with the lutris runtime.
It makes little sense to distribute icons in a .tar.xz like I'm doing. It would be much more flexible to have icons provided through our REST API. This would allow us much more flexibility in platform naming. We can easily make aliases like we do for games.

matoro:
[...]

PR #2990, opened 2020-07-04
Add config option to use long platform names

matoro:
This PR includes a few options to support the auto-configuration of
platform name filters...first, it makes the update of platform names in
the database unconditional for all games.  I only have about 50 titles
in my library and I did not see a significant performance impact from
this, but testing would be appreciated from somebody with a larger
library.

Second, it adds a system config option to use long platform names
(Manufacturer - System) style instead of the official platform names.  A
hardcoded list of platform names is included for the most common
systems.  This option is off by default.

Thirdly there is a small change in gui/widgets/sidebar.py that I had to
make, but I'm not sure the full consequences of.  It constructs the
unfiltered list of platforms only out of the active platform names.  The
only possible negative effect I can think of is that when adding a game
that is the first on a new platform, it won't show up in the filter list
until the next application restart.

Lastly I renamed some icons and made the old names into symlinks in
order to support multiple naming conventions for the same platform.  I'm
using the short version (no manufacturer name) as the "canonical" style.


matoro:
This is hopefully a good solution for https://github.com/lutris/lutris/issues/2832 .  Over time it might be expanded to include all alternate naming schemes for various platforms.

strycore:
Remove the option and format the code and it should be good to go.

matoro:
@strycore By remove the option, do you mean that long names should be default and unchangeable?  Because that's contrary to the decision @AlexanderRavenheart made here:  https://github.com/lutris/lutris/pull/2777#issuecomment-623141860

strycore:
We can have options for trivial stuff like that otherwise Lutris is going to become more and more bloated.
Choose a default and stick with it.

AlexanderRavenheart:
I'm not a fan of adding flexibility and options to naming conventions of runners and cores. We should choose one of the 3 variants mentioned in the linked issue and align all runners/cores to conform to that. The decision should take into account the maintainability effort when new runners/cores are added or existing ones are updated and preferably avoid having to keep hard coded lists in the code.

matoro:
This has the default to use short names, i.e. whatever the runner gives.  But the runners are wildly inconsistent right now, both the Lutris ones and even within libretro itself e.g. some .info files have long-style names, some short-style names, some runners for the same platform use different conventions, e.g. Genesis vs Mega Drive or even abbreviations that other cores don't use.  If you go with that the platform filter will continue to be a mess.  The way I see it this is us providing a "canonical" platform name, or at least the option to use one.  And it's not like there's a lot of them, this wouldn't need to be updated every time a new runner is added, only when it's a new runner for a previously-un-emulated platform.

strycore:
please update against

matoro:
@strycore : rebased

strycore:
thanks!

Issue #3005, opened 2020-07-11
PPA missing package for Ubuntu 18.04 (Bionic)


rockerbacon:
The PPA only contains packages for 20.04 (Focal) and 19.10 (Eoan).

Because of that, the install instructions on Lutris.net won't work for Ubuntu 18.04 and derivatives (Elementary, which has its icon in the download page, still uses 18.04). People using this version of the OS must specify either Focal or Eoan when adding the PPA or the lutris package will never be available for installation.

strycore:
I think some of the recent updates are preventing the package to build. At least on OBS. On Launchpad, I usually copy over the package binaries to other releases and had no issue doing that so far.

Now the situation seems to be very messed up on Launchpad. We have an amd64 package for focal, an i386 package for eoan and no package at all for bionic.

My guess is that with the introduction of meson and ninja to build i18n files, the binary packages need to be rebuilt for each release. I've tried passing the release name to dput like `dput ppa:lutris-team/lutris/bionic build/lutris_0.5.7*_source.changes` but that got rejected.

We may have to adapt our build system to generate packages with different filenames. We may also have to build each package with a different release name in debian/changelog. 

I don't really know how to deal with this so anyone is welcome to experiment with their own PPA. One of the issues with bionic is that the version of debhelper should be downgraded.

@stephanlachnit you probably know better how to handle this. Do you think it's easy to get back the package on bionic?

stephanlachnit:
Hm that's weird. What is the exact reject message? I've never worked with PPA's, but if the reject reasons are the same as on Debian it should be possible to figure it out.
Theoretically, the build system shouldn't really matter for the installation since it doesn't really change the `.deb` itself.

If the builds really require building them on bionic or eoan, one could generate such a package with pbuilder without installing the distros.

muzena:
@rockerbacon 
Try from this repo:https://launchpad.net/~trebelnik-stefina/+archive/ubuntu/games?field.series_filter=bionic

@strycore 
If you want, you can add me to Launchpad Lutris team, I can make auto build PPA eg. stable PPA and git build PPA for testing.
Like this: [Lutris git recipe](https://code.launchpad.net/~trebelnik-stefina/+recipe/lutris-daily)
                [Example of PPA](https://launchpad.net/~devilutionx)

strycore:
@muzena thanks, I can sure use some help with Launchpad!

I added you to lutris-team in LP so, if you want to give it a shot, fell free.

[...]

PR #3006, opened 2020-07-11
change to arch: all

stephanlachnit:
Related to #3005.
@strycore to mark it as `arch: all` no arch specific recommends are allowed (don't really get it but nothing one can do about it).
This means all the 32bit gcc libs have to be removed - obviously one should only do this if they are really required, but I'm not sure why they are in the recommends field in the first place.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -37,8 +37,6 @@ Recommends: python3-evdev,
             gvfs-backends,
             libwine-development,
             winetricks,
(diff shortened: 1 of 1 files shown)


strycore:
That's cool, hope our package build goes back to normal after!

I think the 32bit libs should be recommended by the client and not by the package.

PR #3021, opened 2020-07-14
Dist: Fix manpage installation dir

akien-mga:
Seems like `meson` was already configured to install it in the right location but not `setup.py` (which is what I'm using for now, will poke at meson next).


akien-mga:
BTW the `meson` installation instructions are not documented yet in https://github.com/lutris/lutris/blob/master/INSTALL.rst

Does the Meson buildsystem superseded the setuptools one, or should both be used?

strycore:
that's a good question for @stephanlachnit

stephanlachnit:
I didn't remove setuptools since travis still uses it and I don't have the time to look into the CI stuff. Also some packages still use setuptools for 0.5.7, for example Arch Linux (/cc @FFY00).
But eventually it seems reasonable to drop setuptools and fully switch to Meson due to the translations.

FFY00:
I can change that if needed.

Issue #3020, opened 2020-07-14
HTTPError download error


telanus:
**Describe the bug**

Testing This [commit](https://github.com/lutris/lutris/commit/60b1976145f6a098a361e9d335d94e10661e5b25) I got the following on the terminal:

```
INFO     2020-07-14 08:03:11,615 [runner.get_runner_version:324]:Getting runner information for melonds
DEBUG    2020-07-14 08:03:11,615 [http.get:71]:GET https://lutris.net/api/runners/melonds
Traceback (most recent call last):
  File "~/src/lutris/lutris/util/http.py", line 77, in get
    request = urllib.request.urlopen(req, timeout=self.timeout)
  File "/usr/lib/python3.6/urllib/request.py", line 223, in urlopen
    return opener.open(url, data, timeout)
  File "/usr/lib/python3.6/urllib/request.py", line 532, in open
    response = meth(req, response)
  File "/usr/lib/python3.6/urllib/request.py", line 642, in http_response
    'http', request, response, code, msg, hdrs)
  File "/usr/lib/python3.6/urllib/request.py", line 570, in error
    return self._call_chain(*args)
  File "/usr/lib/python3.6/urllib/request.py", line 504, in _call_chain
    result = func(*args)
  File "/usr/lib/python3.6/urllib/request.py", line 650, in http_error_default
    raise HTTPError(req.full_url, code, msg, hdrs, fp)
urllib.error.HTTPError: HTTP Error 404: Not Found

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "~/src/lutris/lutris/gui/dialogs/runners.py", line 142, in on_install_clicked
    runner.install(downloader=simple_downloader)
  File "~/src/lutris/lutris/runners/runner.py", line 365, in install
    runner = self.get_runner_version(version)
  File "~/src/lutris/lutris/runners/runner.py", line 327, in get_runner_version
    runner_info = request.get().json
  File "~/src/lutris/lutris/util/http.py", line 81, in get
    raise HTTPError("Request to %s failed: %s" % (self.url, error))
lutris.util.http.HTTPError: Request to https://lutris.net/api/runners/melonds failed: HTTP Error 404: Not Found
```

Was testing the download function to see if it work for inclusion in a few other runners

strycore:
we do not have a melonds build

strycore:
I added a download_url field to the JSON runner. I'll have to detect that field and override the API's URL during the install.
We won't have builds for custom runners so this is necessary

PR #3022, opened 2020-07-14
Added a tic-80, MicroM8 & Redream Runner

telanus:
Can be merged once https://github.com/lutris/lutris/issues/3020 gets fully resolved.

The Redream runner was previously in lutris, but was removed due to not being available on the Lutris download server, with the new "download_url" option this can go back, as the removal reason is no longer valid. I tried jsonfied the redream runner, but it opened way more corner cases than what it's worth, so I'm leaving it as is

### DIFF
--- a/lutris/runners/__init__.py
+++ b/lutris/runners/__init__.py
@@ -30,6 +30,7 @@ __all__ = [
     # Sony
     "pcsx2",
     "rpcs3",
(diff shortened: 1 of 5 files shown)


AlexanderRavenheart:
Please fix the issues reported by the CI.

PR #3024, opened 2020-07-15
Fix ppa building

muzena:
@strycore 
I moved debian folder in packaging folder because I can not use in ppa recipes "nest-part packaging" command like in this [Devilution example](https://code.launchpad.net/~devilutionx/+recipe/devilutionx-git)
Can you add in https://launchpad.net/lutris for Driver: lutris-team?


stephanlachnit:
Can you elaborate a bit more why this is needed? I worked before and changing the folder makes it impossible to build the package without moving folders.

muzena:
@stephanlachnit 
When in master branch is dir in /debian path and not /packaging/debian, It is impossible to use [nest-packaging](https://help.launchpad.net/Packaging/SourceBuilds/Recipes), for easier testing, when error occurred you can test in a few clicks, without that it's hard to test all releases of Ubuntu, to much time to install in eg. Virtualbox and test.
You can not nest debian dir from 'some_your_branch' over master' debian dir.

stephanlachnit:
Simply run `debuild` to test if it builds, or if you want to build on a specific distro use `pdebuild` via [pbuilder](https://wiki.ubuntu.com/PbuilderHowto).
You can have multiple `base.tgz` files for every distro (focal, groovy, etc), and write a simple script that changes the `/var/cache/pbuilder/base.tgz` file depending on the distro you want.
Thanks to Meson building like this should only take a minute, and you instantly get the resulting `.deb` (hint: there is a config option to change the output folder).

Also I don't get why you want to nest in there? Is it just to nest the debian folder from a testing branch into master via launchpad? That's such a specific thing which shouldn't happen often anyway but makes building from source much more annoying.

strycore:
I would like to have something in place that ensures consistent and automated builds in Launchpad.

If possible, I would like to setup Travis to be able to build a deb and push it to Launchpad.

@muzena do you have some kind of automation setup for your PPAs? is it possible to use Travis here?

stephanlachnit:
I can send in a PR to have travis build the `.deb`, and since travis uses boinic it should work on any newer systems as well.

I don't know it's possible to setup several jobs with travis. I know that one can run jobs on different distro with CircleCI, but I think it's not possible to publish the result to Launchpad without releasing the private signing key.

muzena:
@stephanlachnit 
I made this change because I had problem with debhelper-compat failed to build, since this now work (I don' know why) and I see that caused the problem for you I will revert back those changes. I think it should now work.

> since this now work (I don' know why)

You can test building locally on pc and it all work and when you upload to ppa build failing, there seem to be some lack and flaws in launchpad build system.
Eg. like this ridiculous problem https://github.com/diasurgical/devilutionX/pull/547

I will take look in pbuilder manual when I get time.
Thanks.

stephanlachnit:
Yeah sadly the Launchpad documentation is the best one...

PR #3026, opened 2020-07-16
Fix debian building

muzena:
@strycore 
Ubuntu dosen't contain **debhelper-compat** package it has **debhelper** and separate **compat** file in debian folder.

 Can you add in https://launchpad.net/lutris for Driver: lutris-team, to be able to edit [git repositories](https://code.launchpad.net/lutris)?


stephanlachnit:
> Ubuntu dosen't contain **debhelper-compat** package it has **debhelper** and separate **compat** file in debian folder.

This is wrong. There is no `debhelper-compat` package at all, not on Debian either. It is a virtual package, provided by `debhelper`.
Using a seperate file to specify the debhlper compatibility is deprecated, please revert this.

stephanlachnit:
Also this changes the `arch` which has nothing to do with this at all.

muzena:
@stephanlachnit 
Strange, had a problem with debhelper-compat, It gave me error in build log that debhelper-compat don't exist instead use debhelper and compat file with version (I tried to build a few times).
Now I tried building with  debhelper-compat, and It work, I don't know why.
I will revert this,I hope there won't be this debhelper-compat error again

PR #3029, opened 2020-07-17
Install/Remove/Uninstall from Command line without GUI

### LINKED ISSUE #2907: Unattened Install
lutris --reinstall gamename

is a good command to create scripts to bundle lutris with games. But it shows asking paths, wine screen etc..
I think it would be better something silent install command to install a game without any prompt to user. such as

lutris --silentinstall gamename

tarTG:
Fixes #2907

Currently it works like
**Silent Install**
lutris --install (game slug / url) --bin_path path_to_binary1; path_to_binary2 [--install_path path_to_install_folder]

If install folder is not given, the default one is used.

**Silent Remove**
lutris --remove id/game_slug

**Silent Uninstall**
lutris --uninstall id/game_slug

Missing Runners are installed without asking
Disks are searched in the standard search paths

I don't know how to handle input menus for this case. So the program prints an error and stops.
Any suggestions?

### DIFF
--- a/lutris/gui/application.py
+++ b/lutris/gui/application.py
@@ -46,6 +46,7 @@ from lutris.util.log import logger
 from lutris.util.steam.appmanifest import AppManifest, get_appmanifests
 from lutris.util.steam.config import get_steamapps_paths
 from lutris.util.wine.dxvk import init_dxvk_versions, wait_for_dxvk_init
+from lutris.installer import silient_install
 
 from .lutriswindow import LutrisWindow
 
@@ -124,6 +125,22 @@ class Application(Gtk.Application):
             None,
         )
         self.add_main_option(
+            "bin_path",
+            0,
+            GLib.OptionFlags.NONE,
+            GLib.OptionArg.STRING,
+            _("Binary path for a silent install"),
+            None,
+        )
+        self.add_main_option(
+            "install_path",
+            0,
+            GLib.OptionFlags.NONE,
+            GLib.OptionArg.STRING,
+            _("Install path for a silent install. If not declared, default path will be used"),
+            None,
+        )
+        self.add_main_option(
             "exec",
             ord("e"),
             GLib.OptionFlags.NONE,
@@ -192,7 +209,7 @@ class Application(Gtk.Application):
     def do_startup(self):  # pylint: disable=arguments-differ
         Gtk.Application.do_startup(self)
         signal.signal(signal.SIGINT, signal.SIG_DFL)
-
+        
         action = Gio.SimpleAction.new("quit")
         action.connect("activate", lambda *x: self.quit())
         self.add_action(action)
@@ -367,6 +384,10 @@ class Application(Gtk.Application):
                 self._print(command_line, _("No such file: %s") % installer_file)
                 return 1
 
+            # Do we have an silent install?
+            if options.contains("bin_path"):
+                action = "silent_install"
+
         db_game = None
         if game_slug:
             if action == "rungameid":
@@ -388,11 +409,43 @@ class Application(Gtk.Application):
                     or pga.get_game_by_field(game_slug, "installer_slug")
                 )
 
+<<<<<<< HEAD
         if action == "write-script":
             if not db_game or not db_game["id"]:
                 logger.warning("No game provided to generate the script")
                 return 1
             self.generate_script(db_game, options.lookup_value("output-script").get_string())
+=======
+        # check if path is provided with install
+        if (options.contains("bin_path") or options.contains("install_path")) and not options.contains("install"):
+            self._print(command_line, _("No installer provided!"))
+            return 1
+
+        if action == "silent_install":
+            self.window = LutrisWindow(application=self)
+            screen = self.window.props.screen
+            Gtk.StyleContext.add_provider_for_screen(screen, self.css_provider, Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION)
+
+            if options.contains("install_path"):
+                silient_install.SilenInstall(
+                    game_slug=game_slug,
+                    installer_file=installer_file,
+                    revision=revision,
+                    binary_path=options.lookup_value("bin_path").get_string().split("; "),
+                    install_path=options.lookup_value("install_path").get_string(),
+                    cmd_print=self._print,
+                    commandline=command_line
+                )
+            else:
+                silient_install.SilenInstall(
+                    game_slug=game_slug,
+                    installer_file=installer_file,
+                    revision=revision,
+                    binary_path=options.lookup_value("bin_path").get_string().split("; "),
+                    cmd_print=self._print,
+                    commandline=command_line
+                )
+>>>>>>> a294b699... Added option to install from command line
             return 0
 
         # Graphical commands
--- a/lutris/gui/application.py
+++ b/lutris/gui/application.py
@@ -46,7 +46,7 @@ from lutris.util.log import logger
 from lutris.util.steam.appmanifest import AppManifest, get_appmanifests
 from lutris.util.steam.config import get_steamapps_paths
 from lutris.util.wine.dxvk import init_dxvk_versions, wait_for_dxvk_init
-from lutris.installer import silient_install
+from lutris.installer import unattended_install
 
 from .lutriswindow import LutrisWindow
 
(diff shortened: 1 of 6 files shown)


lutris/installer/silient_install.py
```diff
@@ -0,0 +1,219 @@
...
+    def prepare_install(self, script):
+
+        install_script = script
+        if not install_script:
+            raise ValueError("Could not find script %s" % install_script)
+        try:
+            self.interpreter = interpreter.ScriptInterpreter(install_script, self)
+        except MissingGameDependency as ex:
+            # call recursive dependencies
+            SilenInstall(
```
AlexanderRavenheart, line 91: Missing t from Silent

lutris/installer/silient_install.py
```diff
@@ -0,0 +1,219 @@
...
+# Lutris Modules
+from lutris import settings
+from lutris.installer import interpreter
+from lutris.installer.errors import MissingGameDependency, ScriptingError
+from lutris.util import jobs, system
+from lutris.util.log import logger
+from lutris.util.downloader import Downloader
+
+
+class SilenInstall:
```
AlexanderRavenheart, line 16: Missing t from Silent

lutris/installer/silient_install.py
```diff
@@ -0,0 +1,219 @@
...
+        except ScriptingError as ex:
+            self._print(self.commandline, ex.__str__)
+            logger.error(ex.__str__)
+            return
+
+    def set_status(self, text):
+        self._print(self.commandline, text)
+
+    # required by interpreter
+    def clean_widgets(self):
```
AlexanderRavenheart, line 118: Required? This SilentInstall class does not extend or inherit any class, so where is this requirement coming from?

lutris/installer/silient_install.py
```diff
@@ -0,0 +1,219 @@
...
+        pass
+
+    # required by interpreter
+    def add_spinner(self):
```
AlexanderRavenheart, line 122: Required? This SilentInstall class does not extend or inherit any class, so where is this requirement coming from?

lutris/installer/silient_install.py
```diff
@@ -0,0 +1,219 @@
...
+        pass
+
+    # required by interpreter
+    def set_cancle_butten_sensitive(self):
```
AlexanderRavenheart, line 126: Required? This SilentInstall class does not extend or inherit any class, so where is this requirement coming from?

lutris/installer/silient_install.py
```diff
@@ -0,0 +1,219 @@
...
+        pass
+
+    # required by interpreter
+    def continue_button_hide(self):
```
AlexanderRavenheart, line 130: Required? This SilentInstall class does not extend or inherit any class, so where is this requirement coming from?

lutris/installer/silient_install.py
```diff
@@ -0,0 +1,219 @@
...
+        pass
+
+    def attach_logger(self, command):
+        pass
```
AlexanderRavenheart, line 134: Why does this function even exist?

AlexanderRavenheart:
If the actions are silent then there shouldn't be any input menus or other user interaction required as that kinda defeats the purpose of being silent, right?

tarTG:
I invoke the install process, like the installwindow. 
But the process and the installwindow are not decoupled and invoke each other.

The GUI functions (addSpinner etc.) are required, because of the direct GUI accesses. 
Without this, i have to refactor a lot of function within the original interpreter, the installerwindow and the commands files.

lutris/installer/interpreter.py
```diff
@@ -492,7 +492,7 @@ def get_runner_class(self, runner_name):
         try:
             runner = import_runner(runner_name)
         except InvalidRunner:
-            GLib.idle_add(self.parent.cancel_button.set_sensitive, True)
+            self.parent.set_cancle_butten_sensitive()
```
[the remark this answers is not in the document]
tarTG: I did not. 
I just moved it into the function of the installer window, so the silent installer has no need for having a cancel button but just an empty function

lutris/installer/commands.py
```diff
@@ -216,7 +216,7 @@ def insert_disc(self, data):
               "<i>%s</i>") % requires
         )
         if self.runner == "wine":
-            GLib.idle_add(self.parent.eject_button.show)
+            self.parent.eject_button_show()
```
strycore, line 219: why did you remove that,

strycore:
Input menu values have to be passed from the command line. The unattended installer must check that all values have been provided before launching the install and quit if one is not provided/invalid.

Also, I don't think the remove command line flag should be added.

lutris/installer/commands.py
tarTG: I moved it into the installer window.
https://github.com/lutris/lutris/pull/3029/files#diff-09064015f83e9cec3fca291ce9ae5954R538-R539

Otherwise the silent installer needs an eject_button with all the GUI dependencies.
This way I only need to provide an empty function.

lutris/installer/silient_install.py
tarTG: The [Script Interpreter](https://github.com/lutris/lutris/blob/155dc4e61ac9e65bec0deb24e8fe27faa200caf4/lutris/installer/interpreter.py#L88) requires a parent class. At the moment this is only the installer window and via this function, the script interpreter make changes on the Window.

Either there had to be a lot of changes in the script interpreter (like raising signals instead of calling the installer window as parent) or the GUI functions have to be available, so the program does not crash.
This is also the reason why there are some changes to GUI accesses like
`continue_button.hide()` 
into functions
`continue_button_hide()`
Otherwise there would be a lot of GUI dependencies in the silent installer.

tarTG:
Deleted remove command line option.
Added -input_menu option:option... for menu choices.
Check all arguments before installing.
Multiple arguments now divided by **:** because **;** terminates command line statements.
Fixed errors and typos.

The only thing which bothers me, is that installers who don't need any files are installed with 
`./lutris --install installer --bin_path (random_argument) `
The argument never gets evaluated but it is really ugly.

kirtr:
I'd love to see this feature added to Lutris. Thank you for the patches, @tarTG! It would greatly help me to automate some of the game installation for the gaming computers I maintain for our kids and their friends. Any chance of signing the Contributor License Agreement to finish this off?

lutris/gui/application.py
```diff
@@ -395,6 +433,42 @@ def do_command_line(self, command_line):  # noqa: C901  # pylint: disable=argume
...
+            if options.contains("install_path"):
+                install_path = options.lookup_value("install_path").get_string()
+            if options.contains("input_menu"):
+                input_menu = options.lookup_value("input_menu").get_string().split(":")
+
+            unattended_install.UnattendedInstall(
+                game_slug=game_slug,
+                installer_file=installer_file,
+                revision=revision,
+                binary_path=options.lookup_value("bin_path").get_string().split(":"),
```
yoyonel, line 455: you're check (above) for `install_path` and `input_menu` but for `bin_path` no if statement ... is it safe ?

lutris/gui/application.py
```diff
@@ -395,6 +433,42 @@ def do_command_line(self, command_line):  # noqa: C901  # pylint: disable=argume
...
+                install_path=install_path,
+                install_options=input_menu,
+                cmd_print=self._print,
+                commandline=command_line,
+                quit_callback=self.do_shutdown
+            )
+            return 0
+
+        if action == "uninstall":
+            if not db_game or not db_game["id"]:
```
yoyonel, line 465: field/key `id` exist necessarly ?

lutris/installer/unattended_install.py
```diff
@@ -0,0 +1,253 @@
...
+        for item in ["description", "notes"]:
+            self.script[item] = self.script.get(item) or ""
+        for item in ["name", "runner", "version"]:
+            if item not in self.script:
+                self.on_install_error(_("Invalid script: %s" % self.script))
+                return -1
+
+        # Check if given files match the number of needed files in the skript. Also test if the files available
+        if "files" in self.script["script"]:
+            req_files = list(filter(lambda file: "N/A" in file[next(iter(file))], self.script["script"]["files"]))
```
yoyonel, line 87: list-comprehension is (a bit) more readable:
```python
req_files = [file for self.script["script"]["files"] if "N/A" in file[next(iter(file))]]
```

PR #3030, opened 2020-07-17
Vkd3d changes

tannisroot:
This will:
-  Disable the Vkd3d option (and also the d3d9 option), even for those people who still have it enabled in the configuration.
- Bump DXVK version
- Add override for d3d12.dll
- Rename DXVK option to DXVK/VKD3D as the opiton now handles VKD3D as well.


AlexanderRavenheart:
Please take care of the issues reported by the CI. Add a space after # on the lines reported there.

PR #3035, opened 2020-07-20
Simplify copyright of Lutris

strycore:
Over time, the project has gathered a lot of different files from different sources.
With Debian being who they are, this is creating friction in getting the project into the official repositories.

This pull request simplifies the ownership of files in the project. Any file not owned by myself has been removed.

We can circumvent abusive copyright rules by having the client download resources such as icons at runtime.


strycore:
This unblocks #2553

strycore:
It's not viable to manage icons for all platforms and runners from the client itself anyway. Especially now that we have json runners

strycore:
Icons are now downloaded as part of the lutris runtime

stephanlachnit:
Cool, thanks!

PR #3038, opened 2020-07-20
Fix missing parentheses in default value of gamemode option.

openglfreak:
Feral GameMode support in Lutris seemed to be always enabled since some time ago. This probably fixes that.


openglfreak:
Specifically commit 6dc584f78b2ccee4839ebbe3c51a1aee2bb28858 seems to have broken it.

PR #3054, opened 2020-07-27
Gtktemplate bugfixed

huo-ju:
Fix the Gtk Template bug before 50a7ad27447dee59d0908985c839ec300ac600fc

### DIFF
--- a/lutris/gui/lutriswindow.py
+++ b/lutris/gui/lutriswindow.py
@@ -660,7 +660,7 @@ class LutrisWindow(Gtk.ApplicationWindow):  # pylint: disable=too-many-public-me
     def on_preferences_activate(self, *_args):
         """Callback when preferences is activated."""
         self.application.show_window(SystemConfigDialog)
-        
+
     def on_manage_runners(self, *args):
         self.application.show_window(RunnersDialog, transient_for=self)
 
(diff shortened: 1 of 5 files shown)


strycore:
Thanks but I have already reverted those changes and pushed them to master.

strycore:
Wait, I think I'm missing something. Does this fix something else? Is the master branch working ok for you now?

virushuo:
I updated the code on Friday and dialog boxes were not pop up, then I realized that Gtk.Template had some differences with gi_composites. So I fixed it, then I found you revert to the old version and also Fix color icon loading issues. I cherry-pick the d3c1a6414d16537bb12ddc4b12ae7117baf50979 and c44794258706840a6455e5b1dc70f1d30c112ded  , and merged with my forked branch. If you still want to use Gtk.Template you can use my commit. I also tested is on the Ubuntu 18.04 and it works.

strycore:
ah ok, yes that will be useful in the future! I noticed the shim was behaving differently than the current implementation so I just reverted the shim. Since we still support 18.04, that's just simpler. But if we can conditionally load it if not available and use the regular implementation by default that's even better. There's not a huge diff left so it will be easy to test.

strycore:
I'm going to close this since I would like to explore solutions to get rid of Gtk.Template (and the shim) entirely.

PR #3061, opened 2020-08-01
Fix error message in prepare_game_files

gabor-udvari:
The prepare_game_files function has separate blocks for GOG and
Humble Bundle services, but the error messages print GOG for
both of them.


strycore:
With the merging of the file-picker branch, this part of the code has changed quite a bit. I'll check if that still needs to be applied.

strycore:
That has been fixed in another branch.

PR #3097, opened 2020-08-19
Russian translate updated

RDL-28:


tannisroot:
This is not a good translation at all...

PR #3132, opened 2020-09-03
Tab order fix

veksha:
Cool feature for keyboard lovers.
Now you can finally use Tab key to change focus (even left or right keys work when focus is on side panels).

Bonus: games list is selecting by default on startup so you can just quickly select your game ~~and hit enter to play.~~


strycore:
This needs to be rebased against the current master.

veksha:
I think i made it.
I noticed that new Lutris is not launching games by pressing ENTER or SPACE anymore. is it just me?

strycore:
@veksha no it's not just you, I have to put back this handler.
Thanks for the update!

Issue #3145, opened 2020-09-15
Duplicated Steam games


mjdiliscia:
When importing Steam games, all of them are shown twice. If I choose to sync all games at startup, all games are imported twice as well. This is happening on two PCs, one with ArchLinux and the other with Manjaro.

I think I've isolated the problem, as when I run `lutris --list-steam-folders` its output is:

```
Unable to load locale dir, translations won't work.
2020-09-14 20:59:03,910: Getting full game list from MAME...
2020-09-14 20:59:03,911: MAME XML generation launched in the background, not returning anything this time
2020-09-14 20:59:03,918: MAME isn't installed, can't retrieve systems list.
2020-09-14 20:59:03,980: Running Lutris 0.5.7.1
2020-09-14 20:59:03,981: Using NVIDIA drivers 450.66 for x86_64
2020-09-14 20:59:03,981: GPU: GeForce GTX 950
2020-09-14 20:59:03,981: GPU: 10DE:1402 1458:36C3 using nvidia drivers
2020-09-14 20:59:04,074: Vulkan is supported
/home/lorddrako/.local/share/Steam/steamapps
/home/lorddrako/.steam/steam/steamapps
2020-09-14 20:59:04,268: Shutting down Lutris
2020-09-14 20:59:04,269: MAME XML written
```

Looking through the code, it seems lutris doesn't notice there are symlinks involved in one of the two paths (I understand Steam sometimes does this as a backward compatibility measure).
I tried to add a `device` and `inode` check on every path, and filter them with that, and seems to work fine. If you want, I can make a pull request with my fix. Or maybe there is a prettier solution.

Steps to reproduce:
- Open lutris.
- Go to Import games.
- Select Steam icon.
- Watch games list.

I thinks there is nothing useful in `lutris -d`, but here it goes anyway:

```
Unable to load locale dir, translations won't work.
2020-09-14 21:16:35,965: Getting full game list from MAME...
2020-09-14 21:16:35,965: MAME XML generation launched in the background, not returning anything this time
2020-09-14 21:16:35,966: MAME isn't installed, can't retrieve systems list.
INFO     2020-09-14 21:16:36,035 [application.do_command_line:275]:Running Lutris 0.5.7.1
INFO     2020-09-14 21:16:36,035 [startup.check_driver:54]:Using NVIDIA drivers 450.66 for x86_64
INFO     2020-09-14 21:16:36,036 [startup.check_driver:58]:GPU: GeForce GTX 950
[...]

PR #3150, opened 2020-09-18
Filtered steamapps dirs with the same device.inode

mjdiliscia:
Fixes a problem where Steam creates an additional folder with symlinks to its main one. This confuses lutris, and makes it duplicate all installed Steam games during imports.
(This fixes issue #3145)


AlexanderRavenheart (approved):

strycore:
Thanks, it's something that will be useful for other features as well. I'll take care of moving the code to an appropriate place myself.

PR #3089, opened 2020-08-17
Move services to the sidebar

strycore:
This gets rid of the "import games" dialog and all features associated with syncing service libraries with the local one.
Now all GOG / Humble / Steam games show up in their respective section on the sidebar. With icons matching the service used (This ensures we get an icon for every game).

There has been some major refactoring in how the view is displayed, removing some coupling between the view and the database.
Since I wasn't able to get rid of Gtk critical warnings about sorting the IconView with pixbufs in the store, I delegated the sorting and filtering to the database, which makes things much easier to handle. I had a look at the music player Lollipop and it looks like all sorting is done through the database as well.

There have been some updates to the UI file and it is now possible to use Glade to edit it.

This is not 100% complete. XDG and Steam games still have to be matched correctly. But it is complete enough to go under testing.

All database stuff has been moved to a database package instead of incorrectly referring to db stuff as "PGA".

The Humble Bundle section has a tendency to segfault. It may be related to the icon format used (transparent PNG)

PR #3165, opened 2020-10-01
Configuration UI Improvements

kins-dev:
This improves the config UI by:
- If there are pending changes to the game identifier when you click save, it asks if you want to save those changes rather than throwing them away
- Made the runner class an abstract base class
  - Added a method to find the main file for the runner when possible
- Adds a path type attribute that can be set on the file selector widget
- Added path type to the banner/icon open dialogs
- Controlling the Open dialog starting path with fall backs
  - Use the value of the entry
  - Use the default of the widget
  - Use the last selected path for this path type (i.e. a banners directory)
  - Use the directory for the game executable
  - Use the last selected path (where path type is not set)
  - Use the configured game installation directory
  - Use ~/Games
  - Use ~
  - The path must exist for it to be considered valid
- If the dialog is a file selector, and the default path is pointing to a file, select that file in the open dialog

It isn't perfect yet, but it is a lot better than before when modifying/adding a lot of entries.  (Continuing where you left off is much nicer than jumping back to the user's home directory, especially when adding a bunch of files in the same folder.)


kins-dev:
*sigh*  The failed checks appear to be from existing code (I touched the file so codefactor blamed me)  I'll work on cleaning them up tomorrow.

kins-dev:
So I added a little bit more to this.  Moved most instances of `icon_type` backed by a string to `image_type` backed by an enumeration.

Cleaned up the resulting lint errors.

Also while I hate adding another docstring format (I counted at least 4 distinct different formats already) I found none of them rendered correctly in VSCode, so I added a template (for the [Python Docstring Generator](https://github.com/NilsJPWerner/autoDocstring) extension) so that following this style should be easy for others using VSCode.

Unfortunately, the extension does not store these settings in the folder settings file, just the workspace or system settings and I wasn't going to create a VSCode workspace for everyone right now.  This means you need to set
```json
    "autoDocstring.includeExtendedSummary": true,
    "autoDocstring.customTemplatePath": ".vscode/docstring.template",
```
yourself in your own workspace.

strycore:
This might be hard to integrate once we get this PR merged: https://github.com/lutris/lutris/pull/3089

Please check your PR against the sidebar-service branch.

strycore:
The sidebar-services branch is now merged and as expected it conflicts a lot.

kins-dev:
Yeah, I'm going to split the two sets.  Going to modify this PR to just be the Directory changes.

PR #3182, opened 2020-10-04
Fix isort processing and rearrange setup.py imports

CartoonFan:
I disabled some unsupported isort settings so that isort will correctly process files again. Afterwards, I tested it out on setup.py, and things seem good so far.


.isort.cfg
```diff
@@ -10,14 +10,12 @@ skip=
...
     LOCALFOLDER
-min-similarity-lines=5
+;min-similarity-lines=5
 default_section=THIRDPARTY
-ignore-comments=yes
-ignore-docstrings=yes
-ignore-imports=yes
+;ignore-comments=yes
+;ignore-docstrings=yes
+;ignore-imports=yes
```
AlexanderRavenheart, line 21: Remove these lines entirely. These disabled options are no longer supported on isort 5.x.

CartoonFan: What about min-similarity-lines? Should I delete that too?

AlexanderRavenheart: Yes, support for that was also removed.

CartoonFan: Right, it's done.

AlexanderRavenheart (approved):

Issue #3192, opened 2020-10-10
Test failure


hartmark:
Building the latest code yields a test failure.

```
(nosetests:42864): Gtk-WARNING **: 00:05:12.542: Theme parsing error: gtk.css:4:8: '' is not a valid color name
/home/markus/.cache/yay/lutris-git/src/lutris/lutris/gui/dialogs/__init__.py:10: PyGIWarning: Gtk was imported without specifying a version first. Use gi.require_version('Gtk', '3.0') before import to ensure that the right version gets loaded.
  from gi.repository import GObject, Gtk, WebKit2
.......2020-10-11 00:05:13,202: MAME XML generation launched in the background, not returning anything this time
2020-10-11 00:05:13,203: Getting full game list from MAME...
2020-10-11 00:05:13,208: MAME isn't installed, can't retrieve systems list.
xrandr: Failed to get size of gamma for output screen
xrandr: Failed to get size of gamma for output screen
xrandr: Failed to get size of gamma for output screen
xrandr: Failed to get size of gamma for output screen
2020-10-11 00:05:13,774: Creating new configuration with runner linux
E/home/markus/.cache/yay/lutris-git/src/lutris/lutris/gui/application.py:69: Warning: g_set_application_name() called multiple times
  GLib.set_application_name(_("Lutris"))
2020-10-11 00:05:14,376: Creating new configuration with runner linux
............xrandr: Failed to get size of gamma for output screen
2020-10-11 00:05:15,858: The command "substitute" does not exist.
None
.xrandr: Failed to get size of gamma for output screen
.2020-10-11 00:05:15,866: This installer doesn't have a 'script' section
None
.xrandr: Failed to get size of gamma for output screen
.Exception ignored in: <_io.TextIOWrapper name='<stdout>' mode='w' encoding='utf-8'>
BrokenPipeError: [Errno 32] Broken pipe
.Terminated
.....2020-10-11 00:05:17,873: Migrating basetable field new_field
........................................................
======================================================================
ERROR: test_can_add_game (test_dialogs.TestGameDialog)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/markus/.cache/yay/lutris-git/src/lutris/tests/test_dialogs.py", line 105, in test_can_add_game
    game.remove(from_library=True)
TypeError: remove() got an unexpected keyword argument 'from_library'
-------------------- >> begin captured logging << --------------------
lutris.util.log: WARNING: MAME XML generation launched in the background, not returning anything this time
[...]

Issue #2945, opened 2020-06-16
Allow downloading extras from GOG installers


Kern3l-Panic:
**Describe the bug**

When the Gog Summer Sale started, I decided to pick up Blood: One Unit Whole. because that's the only place I know that sells it (Even thought It's just an Extra of Blood: Fresh Supply)

But when I try to use the Lutris installer for the game, it doesn't register with the 'Extra' that Blood: One Unit Whole is within another game. I only figured this out after 1 hour of waiting for the game to not install and instead just had a message as if it was installing the game.

So on Lutris' Discord server, I asked for some help. And the help was to NOT login to my Gog account when I clicked install on Blood, and instead opt into choosing a local Gog installer, which was that 'Extra' Gog installer from Fresh Supply for Linux. And it worked.

**Expected behavior**

I'd either expect that Lutris would recognize an extra from Gog or that there be an optional choice to choose local Gog installer instead of waiting for nothing.

**Current behavior**

The installer says that Blood is being installed, but nothing happens.

**Steps to reproduce**

1. Go here: https://lutris.net/games/blood-one-unit-whole-blood/

2. Click Install (ONLY IF YOU DON'T HAVE BLOOD IN YOUR GOG ACCOUNT)

3. And wait... because nothing happens

**Lutris debugging output**

INFO     2020-06-14 18:56:54,005 [application.do_command_line:245]:Running Lutris 0.5.6
INFO     2020-06-14 18:56:54,005 [startup.check_driver:55]:Using NVIDIA drivers 440.82 for x86_64
INFO     2020-06-14 18:56:54,005 [startup.check_driver:60]:GPU: GeForce GTX 1070
INFO     2020-06-14 18:56:54,006 [startup.check_driver:79]:GPU: 10DE:1B81 19DA:2445 using nvidia drivers
INFO     2020-06-14 18:56:54,025 [startup.check_vulkan:132]:Vulkan is supported
INFO     2020-06-14 18:56:54,026 [dxvk.get_dxvk_versions:22]:Updating DXVK versions
DEBUG    2020-06-14 18:56:54,189 [http.get:65]:GET https://lutris.net/api/runtime
INFO     2020-06-14 18:56:54,189 [lutriswindow.toggle_connection:642]:Connected to lutris.net as Creature6
DEBUG    2020-06-14 18:56:54,190 [store.get_missing_media:213]:Requesting missing icons from API for yo-frankie, rise-of-the-triad-dark-war
[...]

PR #3193, opened 2020-10-10
Fix failing test #3192

hartmark:


hartmark:
commit from  #2945 fixes the test failure

PR #3200, opened 2020-10-14
Move runner menu to arrow button

smichel17:
From commit message:

> With the updated location for the runner tasks (wine configuration, run exe in prefix, etc), several people did not realize that the wine icon was actually a button that would open this menu.
> 
> This adds a triangle arrow menu next to the runner icon, which people can't possibly miss (…right?). It is disabled if there are no runner tasks, and the icon button is now always disabled.
> 
> It's not the prettiest solution, but I think it's more usable and that this tradeoff is worth it.

As mentioned in Discord, there are some other things that I think would be nice, but are more time/effort to figure out than I want to spend for now (particularly given my lack of familiarity with python):

- Position the text on the right based on the calculated size of the buttons, rather than hard-coding it
- Don't show the menu button when there are no items, as opposed to just disabling it
- Instead of disabling the button with the icon, make clicking it open the menu (of the button next to it)

### DIFF
--- a/lutris/gui/widgets/game_bar.py
+++ b/lutris/gui/widgets/game_bar.py
@@ -62,8 +62,9 @@ class GameBar(Gtk.Fixed):
         y_offset = 42
         if self.game.is_installed:
             self.put(self.get_runner_button(), x_offset, y_offset)
-            self.put(self.get_runner_label(), x_offset + 45, y_offset)
-            x_offset += 135
+            x_offset += 80
+            self.put(self.get_runner_label(), x_offset, y_offset)
+            x_offset += 95
         if self.game.lastplayed:
             self.put(self.get_last_played_label(), x_offset, y_offset)
             x_offset += 95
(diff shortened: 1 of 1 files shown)


lutris/gui/widgets/game_bar.py
```diff
@@ -62,8 +62,9 @@ def update_view(self):
         y_offset = 42
         if self.game.is_installed:
             self.put(self.get_runner_button(), x_offset, y_offset)
-            self.put(self.get_runner_label(), x_offset + 45, y_offset)
-            x_offset += 135
+            x_offset += 80
+            self.put(self.get_runner_label(), x_offset, y_offset)
+            x_offset += 95
```
smichel17, line 67: Style change: incremented the offset in two places. I think this is clearer and probably easier to maintain; happy to revert if you disagree.

lutris/gui/widgets/game_bar.py
```diff
@@ -101,14 +102,25 @@ def get_game_name_label(self):
     def get_runner_button(self):
         icon_name = self.game.runner.name + "-symbolic"
         runner_icon = Gtk.Image.new_from_icon_name(icon_name, Gtk.IconSize.MENU)
-        runner_icon.show()
```
smichel17, line 104: I didn't see any visible negative effect of removing this. Not sure if it's actually unnecessary or I just don't know what to look for.

PR #3204, opened 2020-10-15
update lockfile and generate new requirements from it

pakro:
I'm fairly new to Python and the development environment and had some minor troubles getting Lutris up and running for local development.
For me it looks like the lockfile (e.g python-magic is missing but required to run the application) and the requirements files are outdated.
However I'm not sure if I'm correct but this proposal updates these files.

Sidenote: I now found that `make dev` command in the `CONTRIBUTING.md`, probably would have saved me some minutes but the problem still isn't solved due to that outdated lockfile. Or am I overseeing something?

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "d1fe5d00aa8442f659763e7080c2331c7f51ab9debf4ccf3c3f3b1d7f62c6789"
+            "sha256": "d0f7e8eeb2b6ded83bb0db811e6a8d3c014e7edf067ba4055760bc7e80072239"
         },
         "pipfile-spec": 6,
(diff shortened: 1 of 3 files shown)


AlexanderRavenheart (approved):

strycore:
You shouldn't be using a lockfile for local development, this is mostly used by Travis.

pakro:
I've started to setup a virtual env and tried to start the application from within the IDE. It failed saying that one import couldn't be executed (pyhton-magic). I believe PyCharm is looking for requirements files. So I started looking how to update them. There I realized that some files are outdated. As mentioned before I somehow oversaw the `CONTRIBUTING.md` (did not understand why pipenv_to_requirements was not found as 
package). So I'm actually not sure if I used the lockfile at all but it helped me to better understand how the project is set up.

PR #3245, opened 2020-10-31
Update net.lutris.Lutris.metainfo.xml

LeandroStanger:

PR #3257, opened 2020-11-04
fix copyright information of metainfo

stephanlachnit:
See https://github.com/lutris/lutris/pull/3245


strycore:
I changed the license back to GPL-3

PR #3271, opened 2020-11-11
Add appstream dependency to control file

muzena:
@stephanlachnit 
Is that Ok?
It fails to [build](https://launchpadlibrarian.net/506261860/buildlog_ubuntu-focal-amd64.lutris_0.5.8+git4449-1a44b28c~ubuntu20.04.1_BUILDING.txt.gz) without appstream dependency.


stephanlachnit:
Good catch, ty. Will run it in a clean chroot myself to confirm.

stephanlachnit (approved):

Issue #3277, opened 2020-11-14
metainfo fails on validation after 0.5.8


A6GibKm:
```
$ appstream-util validate net.lutris.Lutris.metainfo.xml 
```

Fails, the output is 

```
net.lutris.Lutris.metainfo.xml: FAILED:
• tag-invalid           : <metadata_license> is not valid [GPL-3.0+]
• attribute-invalid     : <screenshot> width too large [https://lutris.net/static/images/lutris-window.png] maximum is 1600px
• attribute-invalid     : <screenshot> height too large [https://lutris.net/static/images/lutris-window.png] maximum is 900px
• style-invalid         : <image> has vertical padding [https://lutris.net/static/images/lutris-window.png]
• style-invalid         : <image> has horizontal padding [https://lutris.net/static/images/lutris-window.png]
Validation of files failed
```

Somehow 

```
  <metadata_license>GPL-3.0-or-later</metadata_license>
```
is invalid while 

```
  <project_license>GPL-3.0-or-later</project_license>
```

is valid.

A6GibKm:
See https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html.

stephanlachnit:
> ```
> $ appstream-util validate net.lutris.Lutris.metainfo.xml 
> ```
> 
> Fails, the output is 
> 
> ```
> net.lutris.Lutris.metainfo.xml: FAILED:
> • tag-invalid           : <metadata_license> is not valid [GPL-3.0+]
> • attribute-invalid     : <screenshot> width too large [https://lutris.net/static/images/lutris-window.png] maximum is 1600px
> • attribute-invalid     : <screenshot> height too large [https://lutris.net/static/images/lutris-window.png] maximum is 900px
> • style-invalid         : <image> has vertical padding [https://lutris.net/static/images/lutris-window.png]
> • style-invalid         : <image> has horizontal padding [https://lutris.net/static/images/lutris-window.png]
> Validation of files failed
> ```

The image stuff is purely optional, they work fine in KDE Discover and Gnome Software.

> Somehow 
> 
> ```
>   <metadata_license>GPL-3.0-or-later</metadata_license>
> ```
> is invalid while 
> 
> ```
>   <project_license>GPL-3.0-or-later</project_license>
> ```
> 
> is valid.

Run `appstreamcli --pedantic --explain` and you will see why. It's not a big issue (even if e. g. the FSFAP or CC0-1.0 would be more approriate).

[...]

PR #3293, opened 2020-11-17
Reverts metainfo license to CC0-1.0

A6GibKm:
Closes: #3277. 

**cc**: @stephanlachnit  is this Ok?

### DIFF
--- a/debian/copyright
+++ b/debian/copyright
@@ -8,3 +8,9 @@ Copyright: 2009-2020 Mathieu Comandon <strider@strycore.com>
 License: GPL-3.0-or-later
  On Debian systems, the complete text of the General Public License version 3
  can be found in "/usr/share/common-licenses/GPL-3".
+
+Files: share/metainfo/net.lutris.Lutris.metainfo.xml
+Copyright: Lutris Team <strider@lutris.net>
+License: CC0-1.0
+ On Debian systems, the complete text of the CC0-1.0 license
+ can be found in "/usr/share/common-licenses/CC0-1.0".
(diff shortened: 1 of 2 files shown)


debian/copyright
```diff
@@ -8,3 +8,7 @@ Copyright: 2009-2020 Mathieu Comandon <strider@strycore.com>
 License: GPL-3.0-or-later
  On Debian systems, the complete text of the General Public License version 3
  can be found in "/usr/share/common-licenses/GPL-3".
+
+Files: share/metainfo/net.lutris.Lutris.metainfo.xml
+Copyright: Lutris Team <strider@lutris.net>
+License: CC0-1.0
```
stephanlachnit, line 14: Add to line 14
```
License: CC0-1.0
 On Debian systems, the complete text of the CC0-1.0 license
 can be found in "/usr/share/common-licenses/CC0-1.0".
```

stephanlachnit (changes requested):

stephanlachnit:
Btw, the GPL v3 is a perfectly _valid_ license for the metainfo. It just happens that some tools might want to generate a catalog from all apps with metainfo in a certain repository, let's say Flathub or the Debian repository.
It's simply easier to distribute this catalog if every metainfo is licensed under a public domain license, since you don't need to care about copyleft and pedantic attribution.

A6GibKm:
Then I wonder why the tool made by appstream, intended to enforce their own spec fails on this particular license for metainfo. I asked in places, and the response was: it does not make sense to use a code license for an asset.

stephanlachnit:
Explaination from the documentation:

> Omitting the license value can result 			in your data not being incorporated into the distribution metadata (so this is a required tag). 		
> A permissive license ensures your data can be combined with arbitrary other data in one file (this means copyleft licenses like the GPL are not suitable as metadata license).

So yeah it's mainly to simplify distribution. Maybe valid is the wrong word here, the metadata might not be shown in a distribution if it's not permissive, thus the warning. At least for KDE Discover, I was able to find it on a local install.

stephanlachnit (approved):

strycore (approved):

Issue #3285, opened 2020-11-15
Lutris enables "Installed At" column in list view after launch


TotalCaesar659:
**Describe the bug**

Lutris enables "Installed At" column in list view after launch.

**Steps to reproduce**

- Switch to list view.
- Disable all columns (let only "Name" enabled) and close Lutris.
- Re-open Lutris and see "Installed At" column is enabled.

lutris.conf even has saved `installedat_visible = False`.

**System information (Optional)**

Lubuntu 20.04.1, Lutris 0.5.8 (from [Releases](https://github.com/lutris/lutris/releases/tag/0.5.8)).

PR #3300, opened 2020-11-20
Fix for "Installed At" column enabled after launch

TotalCaesar659:
Fixes https://github.com/lutris/lutris/issues/3285.


strycore:
This has been fixed in master already

Issue #3281, opened 2020-11-15
lutris lutris:rungameid/ does not work


soredake:
**Describe the bug**

Using desktop shortcut does work, lutris not starting the game.

**Steps to reproduce**

- Install kubuntu 20.10 + lutris 0.5.8.
- Install game, create a desktop shortcut from lutris.
- Try to launch game using desktop shortcut

**Lutris debugging output (Optional)**

All log:
```
~
❯ lutris -d lutris:rungameid/1219
INFO     2020-11-15 15:53:55,474 [application.do_command_line:319]:Lutris 0.5.8
INFO     2020-11-15 15:53:55,474 [startup.check_driver:69]:Running X.Org Mesa driver 20.2.2 on Radeon RX 560 Series (POLARIS11, DRM 3.39.0, 5.9.8-xanmod1, LLVM 11.0.0) (0x67ff)
INFO     2020-11-15 15:53:55,475 [startup.check_driver:81]:GPU: 1002:67FF 1043:04BC (amdgpu drivers)
```
**System information (Optional)**
[lutris-issue-report.zip](https://github.com/lutris/lutris/files/5542468/lutris-issue-report.zip)

**Screenshots (Optional)**

If your issue is hard to explain with words, feel free to attach screenshots.

pizzadude:
Same issue with lutris git build on Fedora 33 KDE.

roerch:
can confirm this behaviour with the recent update (0.5.8) coming in yesterday for Debian testing. Desktop shortcuts using rungameid not working, however starting from Lutris itself, the games work.

gsantner:
worked in the past, broken with 0.5.8

solbero:
This also affects the [Lutris Kodi Addon](https://github.com/RobLoach/lutris-kodi-addon) since it uses `lutris lutris:rungameid/` to launch games.

jpleau:
I have ran a `git bisect` and have found that a04c640e678e6851015ddaa311ca685f357c425a is the commit that introduced this issue. Investigating a bit more to see if I can find a fix but I have never look at lutris' code before, maybe someone will beat me to it!

tuxtergames:
> Same issue with lutris git build on Fedora 33 KDE.

Could you fixed it? same problem here on F33 KDE, or the solution is create everything manual again?

pizzadude:
> > Same issue with lutris git build on Fedora 33 KDE.
> 
[...]

PR #3320, opened 2020-11-25
(Possibly hacky) adjustment to support file-magic (Possibly Fixes #3292)

### LINKED ISSUE #3292: python-magic dependency conflicts with file-magic (official python bindings for libmagic)
**Describe the bug**

I'm the Lutris packager for Mageia.

Lutris 0.5.8 added a dependency on `python-magic`, which is https://github.com/ahupp/python-magic

`python-magic` provides the `magic.py` module, which conflicts with the official Python bindings for libmagic provided by the `file` package (`python3-file-magic` on Mageia - previously `python3-magic` but I renamed it to make room for the conflicting `python-magic` package).

Long story short, those seem to be two competing implementations for the `magic` module and they can't be installed next to each other. Which means that once packaged in a distro, you can't install packaged software which independently require `file-magic` and `python-magic`.
Downstream issue (`rpmlint` and `lutris` can't be installed together): https://bugs.mageia.org/show_bug.cgi?id=27619

Upstream issue (8 years old, made some progress 2 years ago but still not resolved): https://github.com/ahupp/python-magic/issues/21

That's an unfortunate upstream issue and not directly related to Lutris but it does make packaging trickier. I'm not sure what the best solution would be here, but it might be worth checking if Lutris could use `python-magic` OR `file-magic`, whichever is installed? This would let distros settle on one fork as the main dependency for their packages.

**Steps to reproduce**

- Try to install a package that depends on `file`'s libmagic Python bindings together with Lutris that requires https://github.com/ahupp/python-magic

przerull:
So I had installed lutris using pacman on arch (so it was on 0.5.8), I hadn't realized this at first but I was experiencing issue #3281.

I was getting the "your version of python-magic is too old." message and am pretty sure that python-magic is the file-magic package in pacman.

```
>>> pacman -Si python-magic
Repository      : community
Name            : python-magic
Version         : 5.39-1
Description     : Python bindings to the magic library
Architecture    : any
URL             : https://darwinsys.com/file/
Licenses        : custom
Groups          : None
Provides        : None
Depends On      : python
Optional Deps   : None
Conflicts With  : None
Replaces        : None
Download Size   : 9.46 KiB
Installed Size  : 28.25 KiB
Packager        : Felix Yan <felixonmars@archlinux.org>
Build Date      : Sat 20 Jun 2020 03:55:17 AM EDT
Validated By    : MD5 Sum  SHA-256 Sum  Signature
```

(the URL is what leads me to believe this).

Importing magic in a python shell confirmed that `from_file` was not an attribute of the `magic` module (because I'm pretty sure I had file-magic installed).   I noticed that it had a similar attribute `detect_form_filename`.

Please note that this is my first time interacting with libmagic.

YOLOing a bit here I thought I'd try monkey-patching it and give it a shot.

Much to my pleasure it appeared to solve my issue.  (then I switched back to master and tested again to see that my shortcuts still worked, then read the changelog and saw that the real problem I was experiencing was #3281 - DOH!)

I do believe that this monkey-patch works though.  In separate python shells I used both python-magic's `from_file` and file-magic's `detect_from_filename` and then did a few `string in result` on the results from the respective methods and noticed they behaved similarly.

Granted, this might confuse a reader of this module because `from_file` returns a string whereas `detect_from_filename` returns a `magic.FileMagic` object.  Fortunately, `magic.FileMagic` objects do support the `__contains__` method so lines 48 through 56 end up behaving the same (or at least similarly) in both scenarios.

I didn't see any unit tests for the find_linux_game_executable.  I imagine it would be difficult to write tests against both python-magic and file-magic.

I'd love to hear what y'all think.  I'm really not sure if this should be merged or not, but thought this was worthy of some discussion at least.

Thanks for your time everyone!

### DIFF
--- a/lutris/util/game_finder.py
+++ b/lutris/util/game_finder.py
@@ -13,8 +13,12 @@ except ImportError:
 
 
 if not hasattr(magic, "from_file"):
-    logger.error("Your version of python-magic is too old.")
-    MAGIC_AVAILABLE = False
+    if hasattr(magic, "detect_from_filename"):
+        magic.from_file = magic.detect_from_filename
+        MAGIC_AVAILABLE = True
+    else:
+        logger.error("Your version of python-magic is too old.")
+        MAGIC_AVAILABLE = False
 
 
 def is_excluded_elf(filename):


lutris/util/game_finder.py
```diff
@@ -13,8 +13,12 @@
 
 
 if not hasattr(magic, "from_file"):
-    logger.error("Your version of python-magic is too old.")
-    MAGIC_AVAILABLE = False
+    if hasattr(magic, "detect_from_filename"):
+        magic.from_file = magic.detect_from_filename
+        MATIC_AVAILABLE = True
```
AlexanderRavenheart, line 18: Typo? (MA**T**IC)

AlexanderRavenheart (commented):
Seems okay.

przerull:
Yes it was a typo (facepalm).  Sorry about that.  I made the fix, rebased, then forced push so there's only one commit.  

Not sure if I was supposed to do it that way...  This is my first time actually using git rebase.

strycore (approved):

PR #3330, opened 2020-11-27
lutris: remove proton support from lutris

GloriousEggroll:
Reasons for doing this:

1. As of Proton 5.13, Proton uses a new runtime which runs in its own container.
Due to this, it is no longer viable or recommended to run proton outside of steam.

2. Additionally, neither the old Proton versions nor the new versions properly
generate the wine prefix in the standard way because they do not ship with wineboot.
So if someone removes their prefix then runs a game with proton, it will not properly
correct a new prefix.

3. Lastly, the proton script uses specific environment variables which are not in place
when running proton directly by the wine executable within it, such as the
steam game ID, gstreamer variables, wine username, and others which are important to the
Proton-specific environment.

4. Since lutris does not run proton games using the lutris script, as mentioned it does not use important environment variables. Not specifying the steam game ID can lead to game-specific fixes not working. Proton also requires a custom patch and WINEUSERNAME defined alongside the patch in order to save to any user other than 'steamuser'. And lastly, it needs the gstreamer library envvar defined so that mfplat games can play. I have submitted other PRs for the gstreamer stuff to work in lutris already.

### DIFF
--- a/lutris/runners/steam.py
+++ b/lutris/runners/steam.py
@@ -219,17 +219,6 @@ class steam(Runner):
     def get_steamapps_dirs(self):
         """Return a list of the Steam library main + custom folders."""
         dirs = []
-        # Extra colon-separated compatibility tools dirs environment variable
-        if 'STEAM_EXTRA_COMPAT_TOOLS_PATHS' in os.environ:
-            dirs += os.getenv('STEAM_EXTRA_COMPAT_TOOLS_PATHS').split(':')
-        # Main steamapps dir and compatibilitytools.d dir
-        for data_dir in STEAM_DATA_DIRS:
-            for _dir in ["steamapps", "compatibilitytools.d"]:
-                abs_dir = os.path.join(os.path.expanduser(data_dir), _dir)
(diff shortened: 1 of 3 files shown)


Bananaman:
@strycore Hi, I'm curious... why was this closed without comment?

I see that the latest Lutris still has Proton runtime.

GloriousEggroll is one of the biggest experts on Proton in the world. I would trust his word on this.

strycore:
I trust GE too but Lutris isn't currently able to use Proton in a reliable manner since it has switched to Pressure Vessel.

Once we support Pressure Vessel in Lutris, we'll be able to add support for Proton builds.

Meanwhile, GE also makes some Wine builds that can be used with Lutris: https://github.com/GloriousEggroll/wine-ge-custom/releases

Bananaman:
@strycore Okay I see, thanks for the information! If I understand right, pressure-vessel is Proton's own sub-sandboxing for games. I heard about troubles with this. Gotcha. :)

strycore:
Yes, and were still using something that is very similar to the old Steam runtime, which was used with earlier versions of Proton

PR #3340, opened 2020-11-30
Russian translations and multiple minor spelling fixes.

claike:
I've changed a few original lines here and there and added some that, in my opinion, also need to be translated. I've looked through the contributing guidelines and it says that I better move my commits to another branch, if this is necessary in my case I think I can do that.

Before any merging I'd like to hear your opinions about my changes to the original strings. And I also have a question about a few lines pertaining to the reicast runner:

- In `lutris/runners/reicast.py` `device_id_1/2/3` options have a word Joypad in their labels. As far as a I can google and judge Joypad and Gamepad are essentially the same thing but I can't remember ever hearing the word Joypad in a very long time if ever. Is this particular term specific to Reicast/Dreamcast? If not, I think it would be better if it is changed to Gamepad.


strycore (approved):

strycore:
I'm ok with replacing joypad with gamepad or controller.

claike:
I think it's good enough for now. But now I found an issue I have no idea how to solve because I can't get lutris working in other languages because it doesn't want to respond neither to LC_ALL nor to LANG. There are a few fields like "custom user-agent" in the Web runner that for some reason add the entirety of po file's header with all the meta information therefore adding a lot of unnecessary and irrelevant text. It needs to be solved somehow, it makes it look very weird.

PR #3348, opened 2020-12-05
Add option to disable stats in game bar

TotalCaesar659:
It would be useful for those who don't want to exhibit their game stats.

### DIFF
--- a/lutris/gui/lutriswindow.py
+++ b/lutris/gui/lutriswindow.py
@@ -152,6 +152,11 @@ class LutrisWindow(Gtk.ApplicationWindow):  # pylint: disable=too-many-public-me
                 default=self.side_panel_visible,
                 accel="F9",
             ),
+            "show-game-bar-stats": Action(
+                self.on_game_bar_stats_state_change,
+                type="b",
+                default=self.game_bar_stats_visible,
+            ),
             "show-hidden-games": Action(
                 self.hidden_state_change,
                 type="b",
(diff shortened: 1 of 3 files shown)


strycore:
No.

TotalCaesar659:
Why?

strycore:
Because it's useless, does not add value, adds complexity and you're not going to maintain it, I will have to.

Please re-read and agree with the terms of CONTRIBUTING.md if you wish to send further patches to this project. Otherwise, I'd rather you just keep yourself from contributing.

TotalCaesar659:
> Because it's useless, does not add value, adds complexity and you're not going to maintain it, I will have to.

Game bar that was introduced in 0.5.8 satisfies those categories: appears and disappears, consumes vertical screen space and duplicate context menu. Of course, IMHO (as your one about my PR that was direct consequence of it).

> Please re-read and agree with the terms of CONTRIBUTING.md if you wish to send further patches to this project. Otherwise, I'd rather you just keep yourself from contributing.

I just decided to start from one small PR that would be easy for me. Well, if it's not the right time for it, I'm ready to accept it. But I'm sure that option would be for those people who likes to сustomize GUI.

strycore:
Since you have decided to completely ignore my request of reading and acknowledging the CONTRIBUTING.md file, I'm going to break it down for you.

* Do not create options unless absolutely needed
* Make a design and stick with it
* Each user option is a moving part in the project, increases the risk of future breakages 

This is not "not the time" for your patch. The time for your patch is never. You don't get to make design decisions on my behalf.
If your next patch isn't something related to the 0.5.9, I will block you from contributing to this repository.

Issue #3378, opened 2021-01-23
Use case-folding for alphabetical sorting


solbero:
**Describe the bug**

Games starting with lower case letters will be sorted after games with capital letters when sorting alphabetically in the Lutris interface.

For the large majority of games this is not a problem, since they start with a capital letter. However, there are some games such as [ibb & obb](https://ibbandobb.com/) which are stylised in lower case.

**Steps to reproduce**

1. Add a game entry that starts with a lower case letter
2. Sort the game list alphabetically
3. Games that start with lower case letters will be sorted after all games that start with capital letters

**Screenshots (Optional)**

<img src="https://user-images.githubusercontent.com/7732292/105612081-ca578b80-5db9-11eb-8700-95848c94c9d2.png" width="400">

strycore:
I hope there's a way to apply a `lower()` to the name column for sorting, without having to create another column just for that purpose like we do for playtime.

solbero:
I tried to look at the code and figure out where the sorting is done for the views. Is the sorting done when the entries are are fetched from the database, or further downstream? If it is done later it could be possible to use the Python built-in function `sorted()`with the `key=str.casefold`, e.g. `sorted(unsorted_list, key=str.casefold)`.

Any pointer to where to look in the code would be appreciated, and I’ll see if I can submit possible solution.

strycore:
I think this problem is caused by the fact that I moved a bunch of sorting and filtering from the Gtk view to database queries. 
There is a sort_func in store.py that is only applied on the list view. I haven't spent a lot of time on the listview and mostly refactored the grid view so there might have been some oversight here.
There might be a simple way to fix this at the database level without having to deal with the whole mess in the sort function in the Gtk views.
I kinda want to re-introduce Gtk based sorting but doing so in a much more careful manner than what we once did. There is at least some cleanup to do to either replace the sort_func in store.py with something else or apply it as well on the GridView.

strycore:
Thanks for the fix!

PR #3388, opened 2021-01-26
Use case-folding for alphabetical sorting

solbero:
### About
This PR fixes case-folding for alphabetical sorting in the Lutris GUI. Game titles starting with capital and lower case letters are sorted together. This sorting is applied to the grid and list views, and when using the search function.

### Tests
This PR passes both the style-checks and the test suite.

### Limitations
Unicode characters are not sorted sorted correctly. This is a limitation of `COLLATE NOCASE` which only can collate ASCII characters. To enable Unicode collation the [ICU](https://www.sqlite.org/src/artifact?ci=trunk&filename=ext/icu/README.txt) extension or similar must be enabled.

Also, not installed games are sorted after installed games in the Lutris GUI since they are fetched by a different function and appended to the view.

### DIFF
--- a/lutris/gui/lutriswindow.py
+++ b/lutris/gui/lutriswindow.py
@@ -252,8 +252,8 @@ class LutrisWindow(Gtk.ApplicationWindow):  # pylint: disable=too-many-public-me
 
     @property
     def sort_params(self):
-        _sort_params = [("installed", "DESC")]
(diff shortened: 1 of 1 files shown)


solbero:
This solves Issue https://github.com/lutris/lutris/issues/3378 with some limitations. Unfortunately, I do not have the permission to link this PR to the Issue.

PR #3402, opened 2021-02-02
fix typos and improve some german translations

pakro:
Thank you for the hard work @Sven792! You were faster! This pull request fixes mostly typos and improves some translations. I will check this file from time to time. These are only fixes for the stuff that obviously hit my eye.


liberavia:
Hey. Any idea for an improvement of the translation "Läufer" (Runner). It sounds confusing in my opinion.
Sadly, instead of beeing a native speaker, there aren't really good (and short) words I found so far.
(No superb) suggestions:

- "Betriebsschichten"
- "Ausführungsmodule"
- "Module"
- ...

If we won't find a good translation, I'd suggest to simply change back to "Runner" instead.

pakro:
Yeah I also don't like "Läufer" that much either, because although correct it is wrong in that context (Sport versus Tech). I will think about that.

strycore:
A longer version of Runner is "Program that runs games". It could maybe be replaced with Launcher or Executor?

pakro:
Hi @strycore and @liberavia,

does "Ausführungsmodul" sound like a sane choice for the German translation for the term "runner"? "Ausführungsmodul" is something like "execution module" in English. You need different modules to run different programs. So for me this term makes somehow sense.

Other similar terms are "Ausführungseinheit", "Programmausführungseinheit" but first sounds military to me and the second on is pretty long. The term "Einheit" ist something like "unit" in English but can mean a unit for a some value but also a unit of a squad (thus the military touch). 

If yes I can add this today evening GMT+1.

liberavia:
I confirm that "Ausführungsmodul" describes it best at the moment. It's pretty a shame that the German language does not have a simpler word for the meaning of a "Runner" anyway :grin:

liberavia:
Hey there! A good friend of mine brought me to the best solution. And it's a shame that I did not get this translation first. IMHO the best translation for "Runner" is "Starter"

pakro:
That is indeed a good translation and fits the purpose perfectly! I will change that today or tomorrow!

strycore:
@pakro  No, Travis is trying to download the runtime so it fails, I've written a patch to fix this so if you rebase your branch against master it should stop failing. 
Since this looks ready and it's only a translation file, I'll keep you from having to rebase your branch and just merge.

Issue #3367, opened 2021-01-21
ARM: lutris should identify arm architecture (in 32 or 64 bits) and not install x86 runners


legluondunet:
Hello,
I tested Lutris on ARM architecture:
- Raspberry PI 4 with 8 Go ram
- Manjaro Linux 64 bits

I could installed Lutris and launched it with success.
I tried to install a scummVM game, Lutris began to download and install ScummVM runner but for x86 architecture.
It would be pertinent to avoid x86 runners install and if no arm runner is available, to inform user:
"We can not install your game because no runner is available for your computer architecture".

strycore:
That's a bug because we already have error messages that explicitly say "This runner is not available for your architecture"

pakro:
I've searched for the String you mentioned but couldn't find anything. Do you have a small hint where to search for this error message, maybe I can fix this issue.

strycore:
I went looking for it and maybe it got removed in some recent release. But the code that handles the architecture check is in runner.py in get_runner_version

pakro:
Hi @strycore,

i've searched the code and maybe found the problem. In https://github.com/lutris/lutris/blob/master/lutris/util/linux.py#L191 the program tries to detect the architecture. This is also the place where the warning is logged you mentioned. According to https://stackoverflow.com/a/45125525 the call `platform.machine` can result in many architectures that have the string "64" in them. Namely the possibilites are:

- aarch64_be (arm64)
- aarch64 (arm64)
- armv8b (arm64 compat)
- armv8l (arm64 compat)

One of the last two must be used in the Raspi v4 according to https://www.raspberrypi.org/products/raspberry-pi-4-model-b/specifications/. But it maybe also one of the other two above.

@legluondunet can you please open up a terminal on your Raspi who has this issue and enter the command `uname -m`?
This should print a similar string Python is using in the call `platform.machine`. Maybe we just make the platform detection more precise to solve this issue.

legluondunet:
```
[legluondunet@lgdn-pi4 ~]$ uname -m
aarch64
[legluondunet@lgdn-pi4 ~]$ uname -a
Linux lgdn-pi4 5.4.83-1-MANJARO-ARM #1 SMP PREEMPT Mon Dec 14 08:37:33 CST 2020 aarch64 GNU/Linux
```

pakro:
[...]

PR #3422, opened 2021-02-12
make architecture detection less ambiguous by clearly stating when to…

pakro:
… detect x86 based systems with 32bit and 64bit

solves the issue https://github.com/lutris/lutris/issues/3367


pakro:
honestly I'm not sure if there are any tests that need to be adapted. I hope Travis finds them. But I'm also happy for guidance! :)

strycore:
The travis failure is unrelated to your PR. Your change is straightforward enough that it wouldn't cause any breakage.

PR #3430, opened 2021-02-14
Fix Python 3.9 compatibility

6d6178667269747a:
The following line can cause issues, because xml.etree.ElementTree.Element.getchildren is deprecated and was removed in python 3.9, [see](https://docs.python.org/3.8/library/xml.etree.elementtree.html#xml.etree.ElementTree.Element.getchildren).

The basis of this cabinstall class [here](https://github.com/tonix64/python-installcab/blob/7f7ac03e4359115568eeaf3ed7571bd7522d9083/installcab.py#L174) already fixed it.


strycore:
looks good

PR #3459, opened 2021-03-06
Don't follow symlinks when merging directories

Unrud:
I found this bug when I tried to install Grim Fandango Remastered from GOG. The archive contains dangling symlinks for some files in *.../usr/share/doc* and the installation fails with `FileNotFoundError`.

Without `follow_symlinks=False` it also fails for symlinks to directories because `shutil.copy` can only handle regular files.
A smaller issue caused by following symlinks are redundant copies of libraries (e.g. `libexample.so.1.0.0`, `libexample.so.1 -> libexample.so.1.0.0`, `libexample.so -> libexample.so.1.0.0`).

### DIFF
--- a/lutris/util/system.py
+++ b/lutris/util/system.py
@@ -207,7 +207,7 @@ def merge_folders(source, destination):
             # logger.debug("Copying %s", filename)
             if not os.path.exists(dst_abspath):
                 os.makedirs(dst_abspath)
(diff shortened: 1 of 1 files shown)


pisandelli:
It will be a nice fix. Is there any plan for this PR be merged?

AlexanderRavenheart:
@strycore ☝🏻

Issue #3458, opened 2021-03-06
Banners are not automatically downloaded anymore


tuxrinku:
Since a few months, banners stopped from being automatically added to a new installed game, even tho the "View on Lutris.net" button points correctly to the game page.

JoshStrobl:
Testing on 0.5.9 and issue is still present. After refreshing my Steam library, external games are now detected however the banners for them are never pulled down, resulting in some less-than-ideal results.

![image](https://user-images.githubusercontent.com/156574/136916776-a27964a5-5a84-4ffa-83ec-9d73c54fa363.png)

Interestingly though, if I go to the "Steam" section under Sources and scroll down to the installed games, the banners show up. They just aren't being downloaded.

![image](https://user-images.githubusercontent.com/156574/136916929-a34a3114-fd01-4bc8-8509-2a1a2dc7b6c9.png)

tannisroot:
Why not just convert bigger banners to smaller ones for now if the format differs?

Lrss:
I think this might be a duplicate of this issue #3730, what is the debug messages when trying to fetch banners for those games?

tannisroot:
Oh no it's not, this one is due to a switch to a new banner format for steam games.

danieljohnson2:
I'm seeing this problem quite a bit, but I do not use Steam. It is mostly GOG games for me. I've been experimenting with this.

It appears to me that Lutris installs its banners at a few specific points- after a successful install, when reloading the Lutris library, that sort of thing. This is fiddly and its not surprising it misses things. It also downloads *too many* banners sometimes- like when searching Lutris, it loads every banner for every game that matched the search (that is has not got already).

If you are trying to spare the hamster whose hamster wheel powers your web-server, that is not ideal.

I think it would be better to download banners (and icons) only when they are rendered; this way only visible banners are loaded. This can be done by triggering the download via a CellRenderer. Of course care must be taken not to double-download things due to repeated rendering, or download repeatedly when errors occur. And you want to still do them asynchronously.

That goes beyond a bug-fix though. Does this sort of change interest the Lutris team? I could put together a PR.

strycore:
With our current system, all banners and cover arts should get downloaded and shown:
- Media is downloaded after game installation
- Media is downloaded after adding a game manually
- Missing media will be reattempted every 24h when the client starts

DanMan:
> Missing media will be reattempted every 24h when the client starts

[...]

PR #3460, opened 2021-03-07
Download banners if not on disk

RobinFrcd:
Fix https://github.com/lutris/lutris/issues/3458

- Use the `Downloader` class to download media asynchronously
- Add some typehint


strycore:
Too stale and drifts too far away of future designs. Closing.

PR #3462, opened 2021-03-07
Handle private Steam Web API Key (closes #3461)

### LINKED ISSUE #3461: Add private steam API key to fetch private data
Everybody is able to get an API key from https://steamcommunity.com/dev/apikey

RobinFrcd:
- Get your own key from https://steamcommunity.com/dev/apikey
- Set the key in lutris preferences
- Get games information even if they are private

Fix #3461

### DIFF
--- a/lutris/gui/config/common.py
+++ b/lutris/gui/config/common.py
@@ -100,12 +100,19 @@ class GameDialogCommon:
     def _build_prefs_tab(self):
         prefs_box = VBox()
         settings_options = {
-            "hide_client_on_game_start": _("Minimize client when a game is launched"),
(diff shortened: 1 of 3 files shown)


RobinFrcd:
Hi @strycore, any chance you could have a look at this quick MR ? Would be very cool to have this in the next release!

jcstryker:
I would love to see this feature implemented. I would prefer to keep my steam profile private while syncing with my lutris install.

strycore:
The Steam ID can now be stored in the settings file. This is enough of an edge case that it doesn't need to clutter the UI with additional widgets.

PR #3464, opened 2021-03-08
Added option to disable GStreamer path override

### LINKED ISSUE #3463: GST_PLUGIN_SYSTEM_PATH_1_0 causes errors with Skryim SE
I have a local (non-steam) installation of Skyrim Special Edition. The audio works perfectly when run in the terminal (by calling wine on the EXE), but NPC voices aren't played when run through Lutris. I'm using the system default wine in Lutris.

I found out that the culprit is the environment variable `GST_PLUGIN_SYSTEM_PATH_1_0`. Setting it in the terminal to the value that Lutris sets (`/home/rharish/.local/share/lutris/runners/wine/system/lib64/gstreamer-1.0/:/home/rharish/.local/share/lutris/runners/wine/system/lib/gstreamer-1.0/`) silences NPC voices, and removing it in the [Lutris code](https://github.com/lutris/lutris/blob/master/lutris/runners/wine.py#L878) solves the issue through Lutris.

Please add an option to the wine runner settings in the GUI to disable this environment variable.

Details:
* OS: Arch Linux
* Lutris version: 0.5.8.3
* System Wine version: 6.3 staging
* System GStreamer version: 1.18.3

rharish101:
Fixes #3463.

I've added the option as an advanced option, and the override is enabled by default (the way it was earlier), so it shouldn't cause problems for existing users.

### DIFF
--- a/lutris/runners/wine.py
+++ b/lutris/runners/wine.py
@@ -443,6 +443,14 @@ class wine(Runner):
                 ),
             },
             {
+                "option": "gstreamer",
+                "label": _("Override GStreamer Library Path"),
+                "type": "bool",
+                "default": True,
+                "advanced": True,
+                "help": _('Override the path to the GStreamer library'),
+            },
+            {
                 "option": "UseXVidMode",
                 "label": _("Use XVidMode to switch resolutions"),
                 "type": "bool",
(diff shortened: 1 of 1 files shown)


rharish101:
Closing this in favour of #3473.

Issue #3110, opened 2020-08-24
Missing innoextract gives misleading error message when extracting


gabor-udvari:
**Describe the bug**

If the innoextract executable is missing from the system, an installer using the extract directive with gog format will print out the following error message:

> RuntimeError('specified exe is not a GOG setup file')

This is misleading as the setup file is fine, only innoextract is missing from the system. Installing innoextract makes this error disappear.

**Steps to reproduce**

- Download any Windows installers from GOG
- Write an installer script with an extract directive, eg.:
```
script:
  files:
    - setup: "N/A: Select the Windows installer provided by GOG"                            
  installer:
    - extract:
        format: gog
        file: setup
        dst: $GAMEDIR
```
- Make sure innoextract is not in your $PATH or inside your lutris runtime
- Try to install the game with the installer script, observe the error message

**Lutris debugging output (Optional)**

[lutris-error-innoextract.log](https://github.com/lutris/lutris/files/5120224/lutris-error-innoextract.log)

**Screenshots (Optional)**

![lutris-error-innoextract](https://user-images.githubusercontent.com/1449353/91095118-1b712100-e65c-11ea-9321-28bde2aa67a3.png)

**Possible solutions**

- In `utils/extract.py` the `check_inno_exe` is silently returning False if innoextract is not found on the system, therefore `decompress_gog` is never called which would raise the proper OSError message. Maybe this is as easy as moving the OSError exception from `decompress_gog` to `check_inno_exe`.
- If innoextract is added to the `required_binaries` directive inside the installer, then it will properly alert the user. Maybe this info can be added to the [installers.rst docs.](https://github.com/lutris/lutris/blob/master/docs/installers.rst)
- Some installers are downloading the innoextract binary from the lutris website, eg. [Archimedian Dynasty](https://lutris.net/games/install/7726/view). Maybe this can be done by lutris itself if innoextract is not found on the system.

strycore:
First step is to ship innoextract in the lutris runtime. There is no reason having such tools in an installer. Once we do game caching, users will end up with hundreds of duplicates of innoextract in their cache...
Once innoextract is available from the runtime, we can have proper integration in the client and have proper error messages.

nikitalita:
> First step is to ship innoextract in the lutris runtime. There is no reason having such tools in an installer. Once we do game caching, users will end up with hundreds of duplicates of innoextract in their cache...
> Once innoextract is available from the runtime, we can have proper integration in the client and have proper error messages.

[...]

PR #3497, opened 2021-04-06
Improve innoextract not found error

bexnoss:
Report that innoextract is not found instead of that the target file is not a GOG setup file.

See https://github.com/lutris/lutris/issues/3110


bexnoss:
Also please add innoextract to the runtime. If I understand correctly it just needs to be uploaded, but if there is anything else that needs to be done I'd be happy to help.

bexnoss:
The error message was improved in 5a7a8b80893dfadab1430365e6899b93de026d01 and innoextract was added to the runtime, so this can be closed. Thanks!

PR #3505, opened 2021-04-15
Localize GOG extras

bexnoss:
Instead of localizing by IP this makes it based on the Lutris locale.

### DIFF
--- a/lutris/services/gog.py
+++ b/lutris/services/gog.py
@@ -95,6 +95,16 @@ class GOGService(OnlineService):
         super().__init__()
         self.selected_extras = None
 
+        gog_locales = {
+            "en": "en-US",
+            "de": "de-DE",
+            "fr": "fr-FR",
+            "pl": "pl-PL",
+            "ru": "ru-RU",
+            "zh": "zh-Hans",
+        }
+        self.locale = gog_locales.get(i18n.get_lang(), "en-US")
+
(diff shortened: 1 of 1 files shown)


lutris/services/gog.py
```diff
@@ -95,6 +95,20 @@ def __init__(self):
...
+            "fr": "fr-FR",
+            "pl": "pl-PL",
+            "ru": "ru-RU",
+            "zh": "zh-Hans",
+        }
+        language = i18n.get_lang()
+        if language in gog_locales:
+            self.locale = gog_locales[language]
+        else:
+            self.locale = "en-US"
```
AlexanderRavenheart, line 110:
```suggestion
        self.locale = gog_locales.get(i18.n.get_lang(), "en-US")
```

bexnoss: That's much better, thanks!

lutris/services/gog.py
```diff
@@ -95,6 +95,16 @@ def __init__(self):
         super().__init__()
         self.selected_extras = None
 
+        gog_locales = {
```
AsciiWolf, line 98: Is it possible to add more locales or are these the only ones supported by GOG extras?

bexnoss: These are all locales supported by GOG.

PR #3508, opened 2021-04-16
Fix service id key in generated GOG installers

bexnoss:
GOG generated installers currently use `humbleid` but it should be `gogid`.


bexnoss:
I just noticed https://github.com/lutris/website/blob/36cca8cbf6d2955a2e00cd57097d81348d5f583e/games/models.py#L260-L267

Is `humbleid` intentional in the generated installer? If so why?

strycore:
Probably doesn't matter, we don't use this field. Merging anyway.

PR #3513, opened 2021-04-18
Fix python-magic

bexnoss:
There are two providers for the magic module: the [official file bindings](https://github.com/file/file/tree/master/python) and [ahupp/python-magic](https://github.com/ahupp/python-magic).

Lutris tries to support both by first checking for `magic.from_file` (ahupp) then `magic.detect_from_filename` (official):
- If `magic.detect_from_filename` is found `magic.from_file` is aliased to it, but the APIs are not compatible. The official bindings return a namedtuple while ahupp returns a string. This PR fixes this, tested with official v5.39 and ahupp v4.20.
- If neither is found a message is displayed about python-magic being too old, even when python-magic is not installed. This PR changes this to first check if python-magic is installed and properly inform the user if not.

### DIFF
--- a/lutris/util/game_finder.py
+++ b/lutris/util/game_finder.py
@@ -12,13 +12,15 @@ except ImportError:
     magic = None
 
 
-if not hasattr(magic, "from_file"):
-    if hasattr(magic, "detect_from_filename"):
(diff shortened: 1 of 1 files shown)


strycore:
Looks good, thanks

PR #3518, opened 2021-04-20
Periodically update dxvk_versions.json

j-piecuch:
Currently the `dxvk_versions.json` file isn't automatically updated as new DXVK versions are released.
This PR adds a new optional parameter to the `download_file()` function that allows us to check for a new version if the local one is old enough. We use it to download the newest version of `dxvk_versions.json` every day.


j-piecuch:
This has been fixed in a slightly different way by e608abae, so I'm closing this PR.

Issue #3553, opened 2021-05-29
Lutris dependencies not fully documented


Lewiscowles1986:
Firstly, thanks for making this tool. I've not used it, but I plan to and it's nice that someone took the time to share this with the world.

Unfortunately I hit a problem. Not wanting to add a ppa, I decided to build from source. I would have used a snap, appimage or flatpak if one were provided, but I didn't see it on https://lutris.net/downloads/ or the README.md of this repo. If it does exist, please let me know.

Anyway... I found it prohibitively difficult when considering anyone who is not a coder or sysadmin.

This is one of a series of issues about the experience. I'll make more if it is of interest, but maybe I'm underestimating the target audience.

At the end of my journey, after getting additional system dependencies, I received this message after completing the install of all requirements on ubuntu 18.04 LTS using python 3.9 in a virtualenv, x86_64 using `pip install -r requirements.txt`.

```
~/lutris$ bin/lutris -d
2021-05-29 08:47:10,286: Package 'distro' unavailable. Unable to read Linux distribution
Traceback (most recent call last):
  File "/home/lewis/lutris/bin/lutris", line 52, in <module>
    from lutris.gui.application import Application  # pylint: disable=no-name-in-module
  File "/home/lewis/lutris/lutris/gui/application.py", line 43, in <module>
    from lutris.gui.installerwindow import InstallerWindow
  File "/home/lewis/lutris/lutris/gui/installerwindow.py", line 17, in <module>
    from lutris.installer import interpreter
  File "/home/lewis/lutris/lutris/installer/interpreter.py", line 14, in <module>
    from lutris.installer.installer import LutrisInstaller
  File "/home/lewis/lutris/lutris/installer/installer.py", line 13, in <module>
    from lutris.services import SERVICES
  File "/home/lewis/lutris/lutris/services/__init__.py", line 9, in <module>
    from lutris.services.gog import GOGService
  File "/home/lewis/lutris/lutris/services/gog.py", line 9, in <module>
    from lxml import etree
ModuleNotFoundError: No module named 'lxml'

```

I solved this issue (for me, not for others who might run into it) by using the following command.

```
pip install lxml
```

strycore:
Reminder that while virtualenvs are great, Lutris is not really fitted to this kind of setup due to it's usage of PyGObject. 

I see that lxml is now present in the setup.py so the installation on a very minimal python setup should now work. But there are no guarantees the software will work properly from a venv.

strycore:
[...]

PR #3554, opened 2021-05-29
Ubuntu 18.04 x86_64 install from GitHub

Lewiscowles1986:
This PR is a placeholder for improving the repo, documenting the steps I had to go through using ubuntu 18.04 on x86_64 (amd64)

## Cairo & Python
- if you're interested let me know
## GCC-5
- if you're interested let me know
## Libgirepository
- if you're interested let me know
## LXML

I don't know the lower-bound of lxml, so I've not even attempted to pin the dependency.

I am aware this might need more effort, such as finding that minimum bar.

Let me know if there is interest in documenting this dependency.

Related #3553

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -34,3 +34,4 @@ secretstorage==3.1.2 ; sys_platform == 'linux'
 six==1.15.0
 urllib3==1.25.10
 wrapt==1.12.1
+lxml


strycore:
Hi,

1) What problem are you trying to solve?
2) The requirements.txt file is automatically generated and shouldn't be edited manually
3) Any particular reason to still be using Ubuntu 18.04?

Lewiscowles1986:
1. Making the setup docs more explicit.
2. Whatever is automatically generating it is missing results.
3. I've found it has more compatibility with games. I also use 20.04, it has this same exact issue when installing from GitHub.

strycore:
I'm afraid the way you're installing lutris is wrong.

Lewiscowles1986:
"wrong". Do you mean undesirable? Can you explain it in more words?

strycore:
It's not the supported way, you shouldn't try to sandbox lutris in a virtualenv or other method so python dependencies should come from the operating system packages.

Lewiscowles1986:
So: 

* this software is specifically designed not to work with a virtualenv (which is just a way of keeping a system clean).
* this software should not document all it's application dependencies, because an OS install would?

Honestly I'd assumed the main push-back would be "we depend on a specific version, please document that".
"We don't want people in clean environments using our software." That is new to me. A Virtualenv has access to all system shared libraries, but just overwrites the python interpreter and it's packages. It's not like Docker or a VM entirely. 

Not supporting that could lead to users needing to choose between your package and another if for any reason two packages are incompatible. 

I really don't understand the reasoning here, but respect your decision. Leaving this here as breadcrumbs for anyone else that might want to try this out without complicating their system, is good enough for me.

PR #3580, opened 2021-06-22
Additional fixes for log window LOG_BUFFERS access

fatalis:
Improves my last commit 'Fix log window not showing logs after game closed' - https://github.com/lutris/lutris/commit/6d56b2ef106564ca31ed3e9501a60df997f315a9

It appears `Game` is inconsistently constructed with game_id as either a `str` or `int` throughout the app and I don't know the best way to fix this, so I'm enforcing LOG_BUFFERS to be keyed by str in every occurrence as a fix for now. Also I didn't want to intruduce new bugs by enforcing str or int in the Game constructor.

### DIFF
--- a/lutris/game.py
+++ b/lutris/game.py
@@ -144,7 +144,7 @@ class Game(GObject.Object):
     @property
     def log_buffer(self):
         """Access the log buffer object, creating it if necessary"""
-        _log_buffer = LOG_BUFFERS.get(self.id)
+        _log_buffer = LOG_BUFFERS.get(str(self.id))
         if _log_buffer:
             return _log_buffer
         _log_buffer = Gtk.TextBuffer()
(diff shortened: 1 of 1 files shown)


fatalis:
int:
```
  File "/usr/lib/python3.9/site-packages/lutris/game.py", line 639, in on_game_quit
    self.save(save_config=False)
```

int:
```
  File "/usr/lib/python3.9/site-packages/lutris/gui/widgets/game_bar.py", line 49, in __init__
    self.game = application.get_game_by_id(game_id) or Game(game_id)
```

str:
```
  File "/usr/lib/python3.9/site-packages/lutris/gui/lutriswindow.py", line 811, in on_game_activated
    game = Game(game_id)
```

strycore:
I think the id attribute should be set to a specific type during class instantiation but that might break some things and we are getting close to the release so your fix is safer for now.

PR #3584, opened 2021-06-24
Fix and improve dolphin game cache reader

fatalis:
It was broken for me due to some hardcoded offsets so I implemented the entire struct.


strycore:
Thanks a lot for that patch, the Dolphin importer was still very rough and experimental, this should make it good to release!

PR #3587, opened 2021-06-26
Quote pre/post-launch commands to deal with spaces in path

bsdf:
These scripts fail when there's a space in the path, even if the script was chosen with the Browse button.


strycore:
PR considered abandoned.

PR #3616, opened 2021-08-01
Make GameBar more HiDPI friendly

neworld:
Reworked layout in Boxes to maintain spaces between elements. It is needed for various DPI and also for long texts, which could happen on rare occasions. Since I have very little experience with GTK and python, I do not know how to make spacing pixels indepenend as well. Also, I do not remember how I made larger fonts on my computer. Any suggestions to improve is very welcome.

Previous UI for DPI=160:

After fix:

### DIFF
--- a/lutris/gui/widgets/game_bar.py
+++ b/lutris/gui/widgets/game_bar.py
@@ -10,12 +10,11 @@ from lutris.gui.widgets.utils import get_link_button, get_pixbuf_for_game
 from lutris.util.strings import gtk_safe
 
 
-class GameBar(Gtk.Fixed):
-    play_button_position = (12, 42)
+class GameBar(Gtk.VBox):
 
     def __init__(self, db_game, game_actions, application):
         """Create the game bar with a database row"""
-        super().__init__(visible=True)
+        super().__init__(visible=True, spacing=12)
         GObject.add_emission_hook(Game, "game-start", self.on_game_state_changed)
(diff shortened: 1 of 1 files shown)


strycore:
This will mess up the lo dpi version and the hi dpi should look the same as the lo dpi anyway

If you need to make those adjustments then your hidpi setup is likely broken. I have a hidpi screen and both lodpi and hidpi lutris windows have the same layout.

neworld:
> This will mess up the lo dpi version and the hi dpi should look the same as the lo dpi anyway

I am just wondering, how it will mess up lo dpi?

> I have a hidpi screen and both lodpi and hidpi lutris windows have the same layout.

I guess you are using `GTK_SCALE` or upscaling. It kinda works, but statically positioned layouts have more problems. For example, people could use different fonts or font settings and is nothing wrong with that. If you want to keep it static, then you should fix font sizes and the font itself.

strycore:
Those changes will 

I'd rather have it non static but it's not easy getting a consistent layout that way. I would certainly merge in a patch with a more dynamic layout while keeping elements evenly spaced.

neworld:
> Those changes will

I still do not understand. It will look different, but not messed up. 

> I would certainly merge in a patch with a more dynamic layout while keeping elements evenly spaced.

Do you have something in mind? Maybe make spaces much larger and make sure all texts fit in the expanse of spaces? I could try to do that.

PR #3618, opened 2021-08-02
Add locale option into the game settings

DavidHusicka:
This allows users to choose a locale for their games from a simple drop down menu. 

If a user has changed locale before using the environment variable setting in game configuration, then this environment variable is more important so this won't break games people already set up before this change.

All locales with `utf8` string are marked as recommended so users who do not have an experience dealing with locales won't choose the wrong one.

image

### DIFF
--- a/lutris/sysoptions.py
+++ b/lutris/sysoptions.py
@@ -29,6 +29,24 @@ def get_resolution_choices():
     return resolution_choices
 
 
+def get_locale_choices():
+    """Return list of available locales as label, value tuples
+    suitable for inclusion in drop-downs.
+    """
+    locales = system.get_locale_list()
+
+    # adds "(recommended)" string to utf8 locales
+    locales_humanized = locales.copy()
+    for index, locale in enumerate(locales_humanized):
+        if "utf8" in locale:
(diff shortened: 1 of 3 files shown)


lutris/util/system.py
```diff
@@ -379,6 +379,14 @@ def get_disk_size(path):
     return total_size
 
 
+def get_locale_list():
```
strycore, line 382: we have utility functions to call Popen in utils/system.py

lutris/sysoptions.py
```diff
@@ -29,6 +29,24 @@ def get_resolution_choices():
...
+    """Return list of available locales as label, value tuples
+    suitable for inclusion in drop-downs.
+    """
+    locales = system.get_locale_list()
+
+    # adds "(recommended)" string to utf8 locales
+    locales_humanized = locales.copy()
+    for index, locale in enumerate(locales_humanized):
+        if "utf8" in locale:
+            locales_humanized[index] += " " + _("(recommended)")
```
strycore, line 42: if the utf8 is recommended, why do we give the option of choosing the non recommended one? Shouldn't we just drop them?

PR #3619, opened 2021-08-02
GitHub Action to lint Python code

cclauss:
Because Travis CI seem to be on a (permanent?) vacation.  https://travis-ci.org/github/lutris

Results: https://github.com/cclauss/lutris/actions

`pytest .` throws `E   sqlite3.OperationalError: unable to open database file`

### DIFF
--- a/.github/workflows/lint_python.yml
+++ b/.github/workflows/lint_python.yml
@@ -1,0 +1,21 @@
+name: lint_python
+on: [pull_request, push]
+jobs:
+  lint_python:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


strycore:
Thanks! 
I don't know what's up with Travis, it keeps giving bogus messages so hopefully Github will get it right.

cclauss:
How do we get `pytest .` to succeed?

strycore:
It looks like the SQLite database can't be created. The code tests for 'nosetests' in settings.py to set the PGA_DB  file in /tmp but doesn't check for pytest so that's probably causing the failure.

check for both nosetests and pytest and you should be good (or replace pytest with nose, alternatively)

PR #3662, opened 2021-08-22
Fix order of get_lib_folders

Staudey:
This fixes a longstanding issue on (at least) Solus where Wine launchers (possibly others) wouldn't run. If I'm not misunderstanding things the code did the opposite of what it was supposed to do. If I'm reading that incorrectly this fixes the issue for some weird reason and I'd welcome further input.

Reference to the issue in question:
https://github.com/lutris/agora/issues/21


strycore:
I think this is ok to test at least, I'll merge it and make it part of the beta 2, if no one complains it's all good.

PR #3677, opened 2021-08-28
Remove Python 2 syntax

cclauss:
This pull request was created with [`pyupgrade --py3-only **/*.py`](https://pypi.org/project/pyupgrade.)


AlexanderRavenheart (approved):

strycore:
I see no reason to replace `"%s" % ""` with `"{}".format("")`. One is not a replacement for the other, the original way of doing string interpolation isn't going anywhere.

strycore:
And it is certainly not "Python 2" syntax

PR #3698, opened 2021-09-14
Adding emulators

EduardTheFirst:
Well in this update i added Duckstation 
and a little bit changed the downloader method
The way it works it will check the url and look if its ending with AppImage(so now we can download runners)
i haven't found the downloader for the PlayStation3 if you know where i could find it in the code i would change this one
then we need to activate the file to be executable and run it once
then we need to change the option for custom Executable and lead the path to 
/Applications folder 

reopen Lutris and voialia you have DuckStation in Lutris 

also we can add games to Lutris to activate DuckStation and it would work !

Best PSX Emulator out there!


EduardTheFirst:
i have no clue

EduardTheFirst:
thats it i have no idea how to fix it,
the code it self works without any syntax problems
if anyone want to check it just download the files ...

PR #3714, opened 2021-10-02
Fix invalid syntax error on LMDE 4

msquirogac:
Walrus operator **:=** supported only on python 3.8 an greater, but LMDE which is based on Debian 10 only has python 3.7

- https://docs.python.org/3/whatsnew/3.8.html
- https://linuxmint.com/download_lmde.php

### DIFF
--- a/lutris/runners/steam.py
+++ b/lutris/runners/steam.py
@@ -240,7 +240,8 @@ class steam(Runner):
                 i += 1
 
         # New Custom dirs
-        if library_config := self.get_library_config():
+        library_config = self.get_library_config()
+        if library_config:
             paths = []
             for entry in library_config.values():
                 if "mounted" in entry:


MartinX3:
Hmm, from an objective view the lowest python version which isn't EOL should be the supported minimal version of a public project.

msquirogac:
> LMDE 5 will be based on Debian 11 (python 3.9)
> 
> If you need lutris now you need to find a backport repo and upgrade python.

As far as I know there is no python 3.9 or even python 3.8 for buster-backports and the ETA for LMDE 5 is something between 6 months to 1 year.
Besides that line in the pull request I haven't found any other python 3.8 specific feature that could break compatibility with older versions.

Evernow:
> 
> 
> This is a project management problem. You can't simply require an upgraded version of system python when you have users who installed your application on a lower version of python. Doing so breaks `apt upgrade` and should be considered borderline malicious.

Times change. It's unreasonable to expect the requirements for a piece of software to stay the same forever. Proper solution here is to just add Python 3.8 as a dependency. People who refuse to update can deal with it on their own.

Evernow:
> 
> 
> > Times change. It's unreasonable to expect the requirements for a piece of software to stay the same forever. Proper solution here is to just add Python 3.8 as a dependency. People who refuse to update can deal with it on their own.
> 
> This isn't just a dependency, it's a dependency on the system _for an upgrade_. That's not acceptable, as it causes errors in apt.

Seems like an issue with your distribution refusing to update. Move on, Python 3.8 has been out for two years now, come on

MartinX3:
Yep.
I use arch linux (rolling release), but i dislike dropping support for still supported distributions/libraries.

mihailim:
Indeed. Not to mention that at this point the _official installation instructions_ at https://lutris.net/downloads/ no longer work for multiple still-supported Debian, Ubuntu, Pop!, Elementary distros.

salmedina:
I can confirm that version xxx.9.1 fixed the issue in Ubuntu 18.04.

PR #3786, opened 2021-10-26
Allow value "unset" for SDL_VIDEO_FULLSCREEN_DISPLAY

mwilck:
For a game I have been setting up ("Die Siedler II" aka "Settlers II"),
in-game fullscreen switching with Alt-Enter only works if
SDL_VIDEO_FULLSCREEN_DISPLAY is not present at all in dosbox's environment.
Even if set to the empty string, resolution switchin won't work and the
1024x768 graphics will look tiny on a modern display.

Add the option "unset" which avoids setting SDL_VIDEO_FULLSCREEN_DISPLAY
altogether.

### DIFF
--- a/lutris/runners/runner.py
+++ b/lutris/runners/runner.py
@@ -157,8 +157,9 @@ class Runner:  # pylint: disable=too-many-public-methods
             env["SDL_GAMECONTROLLERCONFIG"] = sdl_gamecontrollerconfig
 
         # Set monitor to use for SDL 1 games
-        if self.system_config.get("sdl_video_fullscreen"):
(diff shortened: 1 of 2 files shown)


strycore:
Don't add an extra option, this is confusing. Use the "off" value to do what you're doing to unset the variable.

mwilck:
That doesn't work, as I tried to explain. "off" sets the variable to the empty string. But I need an environment where this variable isn't set at all.

mwilck:
We could of course change the semantics of "off", but that might cause regressions elsewhere.

strycore:
No it won't. You can use the off value to unset the variable.

PR #3426, opened 2021-02-13
Add datafields to JSON output

solbero:
This PR add the following datafields to the JSON output: `year`, `playtime` and `lastplayed`. I have tried to make the date and time fields human readable, while still making them easy to parse for programs using the output.

Sample output:
```sh
[
  {
    "id": 1,
    "slug": "supertux",
    "name": "SuperTux",
    "runner": "linux",
    "platform": "Linux",
    "year": 2003,
    "playtime": "0:19:14",
    "lastplayed": "2021-02-12 16:24:04",
    "directory": "/home/lutris/Games/supertux"
  }
]
```

Issue #3379, opened 2021-01-23
Expose more data field in JSON terminal output


solbero:
I would like to create a PR to add more data field to the JSON output when running `lutris --list-games --json`. However, I would like to know if such is a PR would be accepted and fits the scope of the project.

The reason why I would like these data field to be output is to to be able to display more game information and more granular categories in [kodi-lutris-addon](https://github.com/RobLoach/lutris-kodi-addon), a Kodi add-on I co-develop with @RobLoach .

As of now the following data fields are output when running `lutris --list-games --json`:
* id
* slug
* name
* runner
* platform
* directory

I would like to expose the following data fields to the JSON output from the Lutris database:
* lastplayed
* year
* playtime
* hidden

The data field `platform` was added to the JSON output in commit https://github.com/lutris/lutris/commit/063232ee1f7a9d0b9e9a9d4a188bb76056dcf4cf by @maxjoehnk. This leads me to believe that adding more data field to the JSON output would be OK.

strycore:
@solbero any effort to make Lutris move away from Gtk and work on other UIs like Kodi is welcome!

solbero:
How would you feel about adding another command flag to dump all the fields from the database? Right now the flags `--list-installed` and `--json` outputs the most most important data fields about the games.

I feel that those two flags should output the same data, only that `--json` formats that that data as JSON. In other words, they should mirror each other. I tried to add more more data fields to the function that generates the JSON, but it did not feel right that adding the `--json` flag suddenly generates a lot more data as compared to just using the `--list-installed` flag. 

A solution could be to add a flag like `--list-games --dump` and `--list-games --json --dump` that would dump all the data fields for all the games in the `pga.db` into the terminal. The flag could be named something other than `--dump` to make it clearer what it does.

strycore:
having a full dump of the sqlite database seems like a nice idea, it would be useful to do imports/exports

strycore:
Patch merged, thanks!

PR #3796, opened 2021-10-31
Update JSON output function to handle empty values

solbero:
This PR is an bugfix for a [PR](https://github.com/lutris/lutris/pull/3426) I got accepted in January 2021.

I have the following changes to the original PR to solve the following problems:
1. Handle keys that can have a value of `None` or `""` from the database. On one of my installs the database started reporting `None` instead of empty strings `""`. This this made Lutris crash when I tried to get the JSON output using the terminal command flag, since my old code expected only empty strings.
2. Replace empty string values from the database with `null` in the JSON output. This makes the JSON output clearer, since all default, empty values are expressed as `null`.
3. Evaluate a `lastplayed` timestamp of 0 from the database as never played instead of an actual timestamp. This stops games that have never been played from reporting their last played date as `1970-01-01 01:00:00` in the JSON output. Do the same with the key `playtime` for consistency.
3. I have also added `hidden` as a date field to the JSON output to enable filtering of installed, hidden games.

Example of a game entry in the JSON output using the this PR:
```json
{
  "id": 120,
  "slug": "valheim",
  "name": "Valheim",
  "runner": "steam",
  "platform": "Linux",
  "year": null,
  "directory": null,
  "hidden": false,
  "playtime": null,
  "lastplayed": null
}

```
Old PR output of the same entry:
```json
{
  "id": 120,
  "slug": "valheim",
  "name": "Valheim",
  "runner": "steam",
  "platform": "Linux",
  "year": null,
  "directory": "",
  "playtime": 0:00:00,
  "lastplayed": 1970-01-01 01:00:00
}
```
In `application.py` both parenthesis and backslashes are used for line continuation. I have used parenthesised in my PR since I felt it was more readable and it conforms with the style recommendations in PEP8.

### DIFF
--- a/lutris/gui/application.py
+++ b/lutris/gui/application.py
@@ -583,11 +583,17 @@ class Application(Gtk.Application):
                 "slug": game["slug"],
                 "name": game["name"],
                 "runner": game["runner"],
-                "platform": game["platform"],
-                "year": game["year"],
-                "playtime": str(timedelta(hours=game["playtime"])),
-                "lastplayed": str(datetime.fromtimestamp(game["lastplayed"])),
-                "directory": game["directory"],
+                "platform": game["platform"] or None,
+                "year": game["year"] or None,
+                "directory": game["directory"] or None,
+                "playtime": (
+                    str(timedelta(hours=game["playtime"]))
+                    if game["playtime"] else None
+                ),
+                "lastplayed": (
+                    str(datetime.fromtimestamp(game["lastplayed"]))
+                    if game["lastplayed"] else None
+                )
             } for game in game_list
         ]
         self._print(command_line, json.dumps(games, indent=2))
--- a/lutris/gui/application.py
+++ b/lutris/gui/application.py
@@ -586,6 +586,7 @@ class Application(Gtk.Application):
                 "platform": game["platform"] or None,
                 "year": game["year"] or None,
                 "directory": game["directory"] or None,
+                "hidden": True if game["hidden"] else False,
                 "playtime": (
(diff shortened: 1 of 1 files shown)


lutris/gui/application.py
```diff
@@ -583,11 +583,18 @@ def print_game_json(self, command_line, game_list):
...
                 "runner": game["runner"],
-                "platform": game["platform"],
-                "year": game["year"],
-                "playtime": str(timedelta(hours=game["playtime"])),
-                "lastplayed": str(datetime.fromtimestamp(game["lastplayed"])),
-                "directory": game["directory"],
+                "platform": game["platform"] or None,
+                "year": game["year"] or None,
+                "directory": game["directory"] or None,
+                "hidden": True if game["hidden"] else False,
```
AlexanderRavenheart, line 589: Isn't game["hidden"] already a boolean value? Why the if?

solbero: The value of `hidden` from the database is either a `1` or a `0` and not a boolean. Line 589 just replaces `1` with `True` and `0` with `False`.

In the [original issue](https://github.com/lutris/lutris/issues/3379) where I asked for more fields in the JSON output, @strycore [asked for the JSON output to be human readable](https://github.com/lutris/lutris/issues/3379#issuecomment-769974455). I think a boolean is a lot more readable than 1s or 0s.

solbero: I could change Line 589  to be `"hidden": bool(game["hidden"])`, but I think using a ternary makes it easier to understand what the line does.

PR #3801, opened 2021-11-02
Initialize the sidebar even if it is hidden

### LINKED ISSUE #3800: Lutris does not initialize the main view when the side panel is hidden
To reproduce this, use the little drop down menu thingy and uncheck "Show Side Panel". Next, close Lutris completely. Re-open it.

It opens to a blank window. I expect to see something here; either the last view I had used, or some default view. If you go back and recheck "Show Side Panel", the view will then appear.

I think this is happening because the side panel will not be realized if hidden, and it is the handler for the realize signal that is selecting the initial view. The signal emitted upon selection is what triggers the view to initialize, but this does not happen.

danieljohnson2:
The game view is initialized in response to the sidebar's selection being
initialized. This does not work if it is hidden on startup- it is never
realized.

This change replaces the relevant on_realize handler with a plain method
that is called directly and unconditionally. This works even if the
sidebar is hidden on startup.

Resolves #3800


strycore (changes requested):
Same as the other PR, please revert changes made to the documentation.

PR #3834, opened 2021-11-18
i18n: Comprehensive update and proofreading of Simplified Chinese translation

YihaoPeng:
* Made the left menu of the preferences dialog translatable.
* Updated the list of translatable files (`po/POTFILES`).
* Fixed existing errors in Simplified Chinese translation.
* The symbols used in Simplified Chinese translation have been unified.
* Completed the untranslated part of Simplified Chinese.
* A note added to `po/README.md`:
   > Note: All the commands below need to be run in the project root directory, not in the `po` directory. Otherwise you may get `Not the project root` error in meson.


YihaoPeng:
As translatable text scattered everywhere is discovered, the pull request may be updated continuously.

It is recommended that the developer of features mark the translatable text with `_()` in advance to provide convenience for translation.

PR #3848, opened 2021-11-21
None safety in print_game_json

mtizim:
Currently, `lutris -ojl` fails if any `playtime` or `lastplayed` is `None`. This prevents that.


AlexanderRavenheart:
There's already a PR for this: https://github.com/lutris/lutris/pull/3796.

PR #3860, opened 2021-11-27
Silence bogus "Could not retrieve game installer" dialog

### LINKED ISSUE #3853: "Could not retrieve game installer" when trying to install a game from EGS source
Whenever I try to install a game through the Epic Games source, I always get a "Could not retrieve game installer" when the installer window pops up. After I close it (which seems to put the installer window to foreground for some reason, very annoying), I can install EGS just fine.

danieljohnson2:
This removes two bogus error reports - one is a dialog, the other a log message.

The offending code treats 'None' coming back from the install() method of a game service as an 'installer not found' error, but this is not correct. This method returns 'None' when it has starting an installation, but it is still ongoing.

This PR also adds a comment on the base-class install() method to explain this behavior.

Resolves #3853

### DIFF
--- a/lutris/services/base.py
+++ b/lutris/services/base.py
@@ -187,7 +187,7 @@ class BaseService(GObject.Object):
         """Install a service game, or starts the installer of the game.
 
         Args:
-            db_name (dict or str): Database fields of the game to add, or (for Lutris service only
+            db_game (dict or str): Database fields of the game to add, or (for Lutris service only
                 the slug of the game.)
 
         Returns:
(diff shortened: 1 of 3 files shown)


lutris/services/base.py
```diff
@@ -184,7 +184,17 @@ def get_installers_from_api(self, appid):
         return service_installers
 
     def install(self, db_game):
-        """Install a service game"""
+        """Install a service game, or starts the installer of the game.
+
+        Args:
+            db_name (dict or str): Database fields of the game to add, or (for Lutris service only
```
AlexanderRavenheart, line 190: db_**g**ame

PR #3864, opened 2021-11-28
Remove games from the running page when they stop

### LINKED ISSUE #3863: "Running" tab is not refreshed when a game stops/starts
If you stop a game that is listed in the running tab, it will not disappear from it, this will only happen if you switch to a different one and back.
Similar thing happens when a game starts through an application/desktop shortcut, the game is running but it's not listed because the view is not refreshed.
Peek 2021-11-28 22-15
Screenshot from 2021-11-28 22-12-54

danieljohnson2:
This hooks up the 'game-stop' emission and updates the page whenever it occurs, but only if the page is the 'Running' page.

It also detects if running page is removed while selected, and selects the first page ('Games') so you don't wind up looking at a misleading message.

I think there's a flaw here; I expect if you have enough running games that you can scroll, your scroll position will be lost. But re-updating the game store seems safer to me than just removing the stopped game, and I think 'so many running games that you scroll' is not likely.

Resolves #3863


danieljohnson2:
This really isn't working as well as I'd hoped. When games exit naturally, it does not work- the game gets re-added immediately when it is updated as part of the 'quit' handling. Disabling this makes newly installed games not appear.

So much for doing this the safe way. I'll need to rethink this fix.

Issue #607, opened 2017-07-02
Allow to bundle multiple executables


basxto:
Like GLX, SDL and server for Xonotic
or map editor and game for MegaGlest
or to directly launch into the map editor of Wesnoth.

It would make sense to hide the alternate versions in the context menu.

strycore:
Yep, that's something I've wanted as well for a while… Approved!

Ryochan7:
I wanted to make an installer script for the GOG release of Civilization 4 Complete only to find out that the single executable limit makes that impossible. The Civ 4 version of Colonization is bundled as a separate executable so switching between normal Civ 4 and Colonization requires extra work to be done to have access to both games. Unfortunately, Colonization is very crashy on Wine so it didn't really matter anyway. Performance was really good otherwise. It was a fun way to celebrate Columbus Day tbh.

basxto:
Now that I think abot it, you could write a script, which asks the the user via zenity.
Starting one version by default if zenity is not found.
But that’s a bit hacky.

hollunder:
Definitely a sorely missing feature. PoL can do it. Well, it doesn't group them unless you name them accordingly, but at least you can have multiple executables.
I stumbled upon this issue just now with Dawn of War, which has three +  graphics configurator. External graphics configurator programs and similar are quite common and right now you have to 'run a exe' and hunt down the file each time.

sigboe:
![untitled](https://user-images.githubusercontent.com/2087930/63679140-e89a1000-c7f0-11e9-9e7d-59b6990335a4.png)

Some examples where the button could be. 
Also maybe it could be launched via a shortcut like `lutris lutris:rungameid/83a` where the `a` letter following the id number would designate the first extra shortcut.

**Usecase 1:**
Add battle.net games like Diablo 3 and make them launch directly without Battle.net, and have a shortcut to open battle.net in the prefix to update and what not.

**Usecase 2:**
Add configuration utilities, some games have configurations in a seperate App

**Usecase 3:**
Open Map Editor or similar additional program

**Usecase 4:**
Open a modded version of the EXE, like skyrim script extender.

**Usecase 5:**
Open a trainer

DenpaBorisu:
for battle net games just add "-launch" as a launching command to avoid the battle.net launcher
![Screenshot_30_230903](https://user-images.githubusercontent.com/22583225/65929170-5a104400-e3d7-11e9-8936-754ae95244fc.png)

basxto:
[...]

PR #3877, opened 2021-12-03
Enable "Install another version"

### LINKED ISSUE #3874: "Install another version" option is missing for GOG games
Games that are installed from the source "GOG" don't have the option "install another version".

1. Start Lutris
2. Install game from GOG
3. Right click the game
4. Try to find the "install another version" option

danieljohnson2:
This enables the "Install another version" command for already installed games.

This also addresses some collateral issues: Installing an already installed game will default to a different directory and name. Also, the install command will re-fetch the game data from the PGA, since it can be stale.

When installing another version of an installed game, Lutris will now detect if the game's name is in use for an installed game and add a numerical suffix (starting with 2) to make it unique.

It also detects if it is defaulting to a target directory path that is in use; if there is a file or a non-empty directory already there, Lutris will add a suffix (starting with '-2') to make that unique also.

Resolves #3874

Consider merging PR #3966 instead of this- it incorporates this PR to reuse its code, but also adds a "Duplicate" command.

### DIFF
--- a/lutris/database/games.py
+++ b/lutris/database/games.py
@@ -112,7 +112,7 @@ def get_service_games(service):
 
 def get_game_by_field(value, field="slug"):
     """Query a game based on a database field"""
-    if field not in ("slug", "installer_slug", "id", "configpath"):
+    if field not in ("slug", "installer_slug", "id", "configpath", "name"):
         raise ValueError("Can't query by field '%s'" % field)
     game_result = sql.db_select(settings.PGA_DB, "games", condition=(field, value))
     if game_result:
--- a/lutris/database/games.py
+++ b/lutris/database/games.py
@@ -211,3 +211,21 @@ def get_used_platforms():
         rows = cursor.execute(query)
         results = rows.fetchall()
     return [result[0] for result in results if result[0]]
+
+
+def get_unusued_game_name(game_name):
+    """Returns the given name, but if this name is already used by an installed
+    game, this adds a number to it to make it unique."""
+    def is_name_in_use(name):
+        """Queries the database to see if a given is in use by an installed
+        game."""
+        existing_game = get_game_by_field(assigned_name, "name")
+        return existing_game and existing_game["installed"]
+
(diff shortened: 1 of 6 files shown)


strycore:
I'm really not sure about this. Services are not supposed to allow duplicates. There is a lot of code that expects a 1 to 1 relation between service games and lutris games. It's never been disallowed to install another version *outside* of the service integration.

It's not that I'm completely opposed to the idea, I'm just pretty sure this will cause issues down the road. No service allows more than 1 version of a game to be installed (Steam, Battle.net, EGS, Origin, ...)

danieljohnson2:
I didn't understand that. But you seemed to endorse having a 'Duplicate' command for games; does not that also have the same   issue?

strycore:
we can have duplicates, as long as they are not tied to a service game. The current service system really isn't designed for multiple games per service game. 

I would really like to keep the service views an exact representation of the user's library, without duplicates. This should be handled in some other way.

danieljohnson2:
Clearly I need to study the connection between Lutris PGA games and service games more.

strycore:
It's not really a study of database schemas, it's more how things are represented for the user. It makes little sense to show multiple entries of a game on a given service.
Maybe we can work out a selector to choose between multiple installs?
This would be related to #607

danieljohnson2:
I did not even know it was possible for one of the 'located' or 'other version' games to appear in a service's game list at all. I thought they all went into the 'Games' section.

strycore:
The "locate game" feature was added so you could add a game to a service without going through the install. But for the reasons explained above, this can only work for Humble and GOG which don't use a client.

I *guess* that we could add this "another version" thing if it didn't create any relation to the service game. 
Note that when I'm reviewing all of this, I also think of how it could be represented in a console like UI.
Also reminder that we can already can install other versions by not using the service.

danieljohnson2:
I'm not maintaining this anymore. Maybe we'll revisit this in future, but for now let's forget it.

PR #3880, opened 2021-12-04
Separate GL shader disk cache - with system option setting

### LINKED ISSUE #3838: Pass Nvidia's _GL_SHADER envars for every game
And when I mean every game, I mean for every runner Lutris has.
Nvidia has this stupid shader cache size limitation for their default folder in .cache/nvidia, so to avoid games losing their cached shader entries, especially ones with very large amounts of shaders like Overwatch, we should set a dedicated shader cache path for every game launched through Lutris. This way, the caches aren't mashed together and pruned by the nvidia driver when the folder size exceeds a certain amount (I don't remember what it is but definitely less than half a gig).
Envars that need to be set:
__GL_SHADER_DISK_CACHE=1
__GL_SHADER_DISK_CACHE_PATH=$GAMEDIR
Setting this will create a GLCache folder in $GAMEDIR.

danieljohnson2:
This is PR #3879, but with a system options to turn it off.

But this has new UI text, and it is not localized. I don't know how to deal with that, so maybe this PR is no good. Or maybe this should not be a setting at all.

Resolves #3838 again


lutris/sysoptions.py
```diff
@@ -227,6 +227,16 @@ def get_vk_icd_choices():
         "advanced": True,
         "help": _("Switch to US keyboard QWERTY layout while game is running"),
     },
+    {
+        "option": "nvidia_separate_glshaderdiskcache",
+        "type": "bool",
+        "default": True,
+        "label": _("Separate GL Shader Disk Caches (NVIDIA)"),
```
tannisroot, line 234: Can this be named "Dedicated NVIDIA Shader Location" instead?

lutris/sysoptions.py
```diff
@@ -227,6 +227,16 @@ def get_vk_icd_choices():
...
+        "advanced": True,
+        "help": _("Use per-game GL shader disk caches for NVidia hardware. "
```
tannisroot, line 236: same request but here it would be better if the tooltip was "Use a dedicated location for a game's shader cache generated by the NVIDIA driver.". that's just imo tho

lutris/sysoptions.py
```diff
@@ -227,6 +227,16 @@ def get_vk_icd_choices():
...
+                  "This allows larger total caches. Turn this off for a single "
```
tannisroot, line 237: "This allows to bypass the default shader cache size limitation in the NVIDIA driver and avoid unnecessary cache pruning for games with many shaders. Disable to use the default (shared) shader cache location."

danieljohnson2: Sure thing. The real problem is naming the option "Koristite namjensko mjesto za predmemoriju shadera igre koju generira NVIDIA upravljački program".

lutris/sysoptions.py
```diff
@@ -227,6 +227,19 @@ def get_vk_icd_choices():
...
+    {
+        "option": "nvidia_separate_glshaderdiskcache",
+        "type": "bool",
+        "default": True,
+        "label": _("Separate NVIDIA Shader Cache Location"),
+        "advanced": True,
+        "help": _("Use a dedicated location for a game's shader cache generated by the NVIDIA driver. "
+                  "This bypasses the default shader cache size limitation in the NVIDIA driver and "
+                  "avoids unnecessary cache pruning for games with many shaders. "
+                  "Disable to use the default (shared) shader cache location. If enabled, this will apply "
```
tannisroot, line 239: "If enabled, this will apply" ?
seems like the sentence ends unfinished

danieljohnson2: The next line continues the sentence with "__GL_SHADER_DISK_CACHE=1" and so on. I just didn't change that line.

I modeled this on the "prime" setting, which explains that it sets environment variables with similar text.

tannisroot: Oh I derped out, i see now.

strycore:
Why does this need to be an option? Is there any need to disable it for any reason?

danieljohnson2:
I figured you might want to save disk space, or maybe NVidia knows best and this change will break something.

But I provided the change both ways, since I'm not convinced these are real issues, and you can explicitly set the environment vars in the settings anyway, and so override this.

strycore:
Ok, I'll go without the option then and close this PR.

PR #3890, opened 2021-12-05
The game box can use a dynamic (box based) layout, and ellipsize the game name

### LINKED ISSUE #3517: Long names cause window expansion
Lutris 0.5.8.3
Very long names cause the window to expand far to the right, even outside of the screenspace, when you set the name in configure and then click on the game icon to activate it (zoom level max, toggle view to four rectangles).

danieljohnson2:
This PR replaces the fixed game-box layout with a dynamic one, using Gtk.Box.

This will be more resilient to things like changing fonts, and is fairly close to the old fixed layout. But it is more dynamic- things can move around as you selecting this or that game.

This also changes the game-name label to use automatic ellipsis, which the box layout supports. This prevents long game names from resizing the Lutris window; the name is instead truncated.

Here's what the Lutris window looks like when enlarged to fit a long name:
Before

And here's what the ellipsized name looks like:
After

Resolves #3517

### DIFF
--- a/lutris/gui/widgets/game_bar.py
+++ b/lutris/gui/widgets/game_bar.py
@@ -10,12 +10,15 @@ from lutris.gui.widgets.utils import get_link_button, get_pixbuf_for_game
 from lutris.util.strings import gtk_safe
 
 
-class GameBar(Gtk.Fixed):
-    play_button_position = (12, 42)
-
+class GameBar(Gtk.Box):
     def __init__(self, db_game, game_actions, application):
         """Create the game bar with a database row"""
-        super().__init__(visible=True)
+        super().__init__(orientation=Gtk.Orientation.VERTICAL, visible=True,
+            margin_top=12,
+            margin_left=12,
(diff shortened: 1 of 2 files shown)


AlexanderRavenheart:
Please include before and after screenshots for such changes.

danieljohnson2:
Hmmmm.

tannisroot:
I don't think the game title under the banner should be longer than one line and go down like that.

danieljohnson2:
It's not my fault! It was like that when I got here!

No, really. That word-wrapping behavior is what the grid does; it's the same in the before-screenshots too. My game-bar changes didn't affect it. If you want to change it, that should be a separate PR.

For what its worth, I like the way that works in the grid. It looks daft for my "RUINER RUINER RUINER" item, but that's extreme. It looks sensible for things like "Serious Sam: The First Encounter". Perhaps it could be limited to 2 or 3 lines or something, but I do not see the need.

PR #3898, opened 2021-12-09
Moved Add Game into hamburger menu

### LINKED ISSUE #3896: Hide "Add game" somewhere
When helping people on our Discord, I've noticed many, many users like to use the "Add Game" menu to install games because it's the most intuitive part of the UI, thinking Lutris works like PoL, and they don't really explore the sources that Lutris provides.
The issue is that when they install games like that, they create broken installations that use default system wine and default $HOME/.wine prefix, which is very bad.
Ideally, the sources UX should be revamped, but for now, the menu can be hidden to force people to explore the 'Sources" section. Hiding it in the hamburger button menu would be ok.
image

DragonOfWar:
Closes #3896 
This removes the "Add Game" button (the "+" on the top left), and moved it into the hamburger menu.
Here's how it look like:
image

### DIFF
--- a/share/lutris/ui/lutris-window.ui
+++ b/share/lutris/ui/lutris-window.ui
@@ -474,7 +474,7 @@
             <property name="can-focus">True</property>
             <property name="receives-default">False</property>
             <property name="action-name">win.add-game</property>
-            <property name="text" translatable="yes">Add Game</property>
+            <property name="text" translatable="yes">Add a game manually</property>
           </object>
           <packing>
             <property name="expand">False</property>
--- a/share/lutris/ui/lutris-window.ui
+++ b/share/lutris/ui/lutris-window.ui
@@ -474,7 +474,7 @@
             <property name="can-focus">True</property>
             <property name="receives-default">False</property>
             <property name="action-name">win.add-game</property>
-            <property name="text" translatable="yes">Add a game manually</property>
+            <property name="text" translatable="yes">Add a Game Manually</property>
           </object>
           <packing>
             <property name="expand">False</property>
--- a/share/lutris/ui/lutris-window.ui
+++ b/share/lutris/ui/lutris-window.ui
@@ -207,35 +207,6 @@
           </object>
         </child>
         <child>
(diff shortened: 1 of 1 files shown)


tannisroot:
Adwaita for reference:
image

share/lutris/ui/lutris-window.ui
```diff
@@ -498,6 +468,31 @@
         <property name="border-width">9</property>
         <property name="orientation">vertical</property>
         <property name="spacing">3</property>
+        <child>
+          <object class="GtkModelButton">
+            <property name="visible">True</property>
+            <property name="can-focus">True</property>
+            <property name="receives-default">False</property>
+            <property name="action-name">win.add-game</property>
+            <property name="text" translatable="yes">Add Game</property>
```
tannisroot, line 477: Can you rename it to "Add a game manually"?

DragonOfWar: image
Done. Is there anything I need to do in translation files to account for this?

danieljohnson2: I think there might be. If I understand this, and maybe I don't, the localized text in the .po files uses the English text as a key. There are translations of "Add game" in there, and I imagine these translations will no longer be used at all.

I'm not sure what ought to be done. You might change "Add Game" in all the .po files to  "Add a game manually", so the old translations will be used. Or maybe these translations should be updated.

By the way, I think it should be "Add a Game Manually", using the same capitalization style as most of the other menu items here. ("Lutris forums" is the only one using sentence-casing  here)

tannisroot: You're right, ideally it should be capitalized, the UI is a bit all over the place regarding that.

DragonOfWar: image
Done.

strycore:
This "Add game" button will become the main point of entry to add any game(s) to Lutris so I'd like it to stay where it is.

tannisroot:
@strycore is that planned for 0.5.10?

strycore:
yes, I would like to have it soon

PR #3919, opened 2021-12-20
Migrate banners to .cache/lutris

### LINKED ISSUE #3906: Add migration for banners cache
Since https://github.com/lutris/lutris/commit/db6c0f4e03395946c6f1dc0abcfe208cfd447056 changed the location for banners, Lutris should add a migration that will move banners that are already stored in the data dir location to the cache dir location.
Because right now, when you start Lutris, all games are missing banners, and Lutris does not seem to redownload them.

danieljohnson2:
This adds a migration that moves each banner from ~/.local/share/lutris/banners to ~/.cache/lutris/banners.

This also adjusts init_lutris() to create the new directory, so we can move stuff in there.

This will try withstand odd conditions like errors moving the files, or if some banners are already there in the cache for some reason.

Resolves #3906

### DIFF
--- a/lutris/migrations/migrate_banners.py
+++ b/lutris/migrations/migrate_banners.py
@@ -2,20 +2,27 @@
 import os
 
 from lutris import settings
+from lutris.util.log import logger
 
 
 def migrate():
     dest_dir = settings.BANNER_PATH
     src_dir = os.path.join(settings.DATA_DIR, "banners")
 
-    # init_lutris() creates the new banners directrory
-    if os.path.isdir(src_dir) and os.path.isdir(dest_dir):
-        for filename in os.listdir(src_dir):
-            try:
-                src_file = os.path.join(src_dir, filename)
-                dest_file = os.path.join(dest_dir, filename)
+    try:
+        # init_lutris() creates the new banners directrory
+        if os.path.isdir(src_dir) and os.path.isdir(dest_dir):
+            for filename in os.listdir(src_dir):
+                try:
+                    src_file = os.path.join(src_dir, filename)
+                    dest_file = os.path.join(dest_dir, filename)
 
-                if not os.path.exists(dest_file):
-                    os.rename(src_file, dest_file)
-            except OSError:
-                pass  # Skip what we can't migrate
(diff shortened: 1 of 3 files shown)


lutris/migrations/migrate_banners.py
```diff
@@ -0,0 +1,21 @@
...
+    if os.path.isdir(src_dir) and os.path.isdir(dest_dir):
+        for filename in os.listdir(src_dir):
+            try:
+                src_file = os.path.join(src_dir, filename)
+                dest_file = os.path.join(dest_dir, filename)
+
+                if not os.path.exists(dest_file):
+                    os.rename(src_file, dest_file)
+            except OSError:
+                pass  # Skip what we can't migrate
```
AlexanderRavenheart, line 21:
```suggestion
    os.makedirs(dest_dir, exist_ok=True)
    # Check if dirs_exist_ok is defined ( Python >= 3.8)
    sig = inspect.signature(shutil.copytree)
    if "dirs_exist_ok" in sig.parameters:
        shutil.copytree(source, destination, symlinks=False, ignore_dangling_symlinks=True, dirs_exist_ok=True)
    else:
        shutil.copytree(source, destination, symlinks=False, ignore_dangling_symlinks=True)
    shutil.rmtree(src_dir, ignore_errors=True)
```

Same code for copying a src folder into a dest folder as provided in merge_folders() of  lutris/util/system.py, with a os.makedirs() call to ensure the folder path is created and shutil.rmtree() to remove the old folder after copying is complete.

danieljohnson2: That's practically a competing PR!

I like mine better. init_lutris() should now be taking care of creating the dest_dir. Mine moves rather than copies, and that should be faster and save (temporary) disk space too. It skips things that can't be moved but it does *not* delete them- it allow manual fixing later if desired. I don't like idea of deleting files as part of migration.

It does leave an empty banners directory behind if everything was moved; I suppose I could delete that as you do.

It occurred to me to just os.rename the entire "banners" directory, but since init_lutris tries to create all the directories everywhere, that would have issues.

PR #3922, opened 2021-12-22
Provide separate NVidia shader caches only for WINE and Lutris runners

### LINKED ISSUE #3920: [feature request] Disable GL_SHADER_DISK_CACHE for older games/runners
Commit https://github.com/lutris/lutris/commit/f575dbb38f9a2ad4fcca3f668cdf2839eab74f4b is great for modern games, but clog up your rom folder if you play older games, IE pre-PSX era games, as these games don't need a GLShader.

Shouldn't we add an option to disable GLShader per runner/game.

@strycore @tannisroot @danieljohnson2

danieljohnson2:
For all other runners, there is a per runner cache kept in .caches/lutris/shaders/<runner name>.

Resolves #3920


lutris/settings.py
```diff
@@ -23,7 +23,8 @@
 GAME_CONFIG_DIR = os.path.join(CONFIG_DIR, "games")
 
 TMP_PATH = os.path.join(CACHE_DIR, "tmp")
-BANNER_PATH = os.path.join(CACHE_DIR, "banners")
+SHADER_CACHE_DIR = os.path.join(CACHE_DIR, "shaders")
+BANNER_PATH = os.path.join(DATA_DIR, "banners")
```
telanus, line 27: Shouldn't this be (DATA_DIR, "banners") be (CACHE_DIR, "banners")

danieljohnson2: Whoops! I botched the merge. Honestly though, GitHub makes that awkward. I can't seem to do these merges before the PR, and the tool for doing it then is not great.

PR #3935, opened 2021-12-25
Add  "Change" button to game directory

### LINKED ISSUE #3914: [Feature request] Select Game folder instead of moving current one
(If I can/should add the feature request label myself or need to change something else please tell me how/what ^^) 

When configuring a game under Game info->Directory you can only choose to move the currently selected 

However sometimes I already have a finished installation or already moved the folder beforehand and just want to update the reference. Clicking move for those cases causes way too many issues (like moving my desktop into my games folder for example), so simply selecting an already existing directory would be much preferred. 
It could also reduce confusion as this is the way most programs/installers handle this kind of thing.

danieljohnson2:
This adds a "Change" button next to "Move" to update the game directory, but not move the files.

This updates the games config file, same as "Move".

Since the button uses the same text as the other "Change" button here, I think no additional localization will be required.

Resolves #3914

### DIFF
--- a/lutris/game.py
+++ b/lutris/game.py
@@ -682,29 +682,39 @@ class Game(GObject.Object):
             return
         export_bash_script(self.runner, gameplay_info, script_path)
 
+    def change_location(self, new_location):
+        """Updates the game and its config file with a new directory, but
+        does not move the game files themselves."""
+        if self.directory != new_location:
+            new_config = ""
+            old_location = self.directory
+
+            self.directory = new_location
+            self.save()
+
+            if not old_location:
+                logger.info("Previous location wasn't set. Cannot update game config file")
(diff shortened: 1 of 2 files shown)


strycore:
This is extremely confusing for the user, no one would have any idea what the change button is for.

I was thinking of implementing something similar to Transmission: 

Screenshot from 2022-01-20 22-05-33

danieljohnson2:
I figured it would be like the identifier field's "Change" button. But I guess the idea is that one "Change" button leads to a further dialog, which leads to yet another dialog?

I think this approach could work well, actually, if you could get those radio buttons into the folder picker dialog. Same number of dialogs then, but the select vs move choice is hidden until you have to choose.

PR #3937, opened 2021-12-25
Make various dialogs properly modal

### LINKED ISSUE #3466: Modal dialog in main window blocks auxiliary window interaction
A modal dialog, e.g. "about Lutris" in the main window inexplicably blocks all interaction in e.g. the download window. Given that the latter can — and usually will —  be on top of and visually block the former, I'm left with no clue as to why I can't interact with the application, and will conclude that its frozen. 

In Windows I seem to recall they "solve" this by flashing the titlebar, to at least indicate the app is alive while still leaving you clueless. MacOS solves it by making the dialogs part of the parent window, and be done with it — the problem doesn't exits there. 

I don't know what gnomes solution is, and I can't guess what they're thinking considering they seem to treat modal dialogs both as independent windows (by giving them a separate titlebar, etc) as well as part of the parent window (by blocking interaction and locking their positions to the parent). 

Anyway, the proper solution from my perspective is to not block interaction in the windows that don't own the modal dialog.

### LINKED ISSUE #3936: Modal dialogs aren't really modal
### Bug description

Lutris has some windows, like game configuration, that seem like they should be modal and others that aren't (like the installer window). They are different- an installer window does not prevent you from accessing the Lutris window again. Just click it and it comes to the front.

But other dialogs do prevent you from bringing the main window to the front. This makes sense- you probably shouldn't be using the main window while reconfiguring its content, and unlike installation the configuration isn't something that you'd do in the background.

The preferences window is the same way.

The thing is, you can still activate the main Lutris window (see media) and interact with it, but it won't come to the front. There also may be some special visual effects for modal dialog (GNOME does this by default- they call it "attached dialogs". It's a bit like OS X this way).

See also issue #3466, which is about the way these half-modal dialogs interact with the about box, which is a proper modal dialog.

### How to Reproduce

Open Lutris. Select any game and use the configure command; a dialog appears. Now, click on the Lutris window hiding behind it. It activates, but it does not come to the front.

The configuration dialog is rather big and really gets in the way of using the main window. This kind of 'inactive, always in front' behavior can work for small windows, but I don't buy it here.

### Expected behavior

Either we should be able to bring Lutris to the front (as with installer windows) or the dialogs should be proper modal dialogs.

### Log output

```shell
INFO     2021-12-25 09:30:32,960 [startup.check_driver:53]:Using NVIDIA drivers 470.86 for x86_64
INFO     2021-12-25 09:30:32,960 [startup.check_driver:57]:GPU: NVIDIA GeForce RTX 3060
INFO     2021-12-25 09:30:32,960 [startup.check_driver:73]:GPU: 10DE:2503 196E:1377 (nvidia drivers)
DEBUG    2021-12-25 09:30:33,114 [lutriswindow.update_store:451]:Showing 30 games
INFO     2021-12-25 09:30:43,716 [application.do_shutdown:631]:Shutting down Lutris
```

### System Information

```shell
[System]
OS:              Pop!_OS 21.10 impish
Arch:            x86_64
Kernel:          5.15.8-76051508-generic
Desktop:         GNOME
Display Server:  x11

[CPU]
Vendor:          AuthenticAMD
Model:           AMD Ryzen 7 5800X 8-Core Processor
Physical cores:  8
Logical cores:   16

[Memory]
RAM:             31.4 GB
Swap:            0.0 GB

[Graphics]
Vendor:          NVIDIA Corporation
OpenGL Renderer: NVIDIA GeForce RTX 3060/PCIe/SSE2
OpenGL Version:  4.6.0 NVIDIA 470.86
OpenGL Core:     4.6.0 NVIDIA 470.86
OpenGL ES:       OpenGL ES 3.2 NVIDIA 470.86
Vulkan:          Supported
```

### Media (optional)

Screenshot from 2021-12-25 09-13-17

### Checklist:

- [X] I'm not asking for support with a game or the wine runner.
- [X] I have followed the above mentioned guides and have all the graphics and wine dependencies installed.
- [X] I have checked for existing issues that describe my problem prior to opening this one.
- [X] I understand that improperly formatted bug reports may be closed without explanation.

danieljohnson2:
The trick to a 'proper' modal dialog is to show it with the run() method (and to set its parent of course).

This changes Lutris to do this for configuration dialogs and for preferences. Not the installer window, which was not modal at all.

There may be more dialog to be done, but this is what I've found so far.

Resolves #3936
Resolves #3466

Of course, it may be that this PR goes the wrong way- #3466 seemed to think so. This PR is the easy thing to do, but maybe we'd rather make all these dialogs fully independent of the Lutris window.

### DIFF
--- a/lutris/game_actions.py
+++ b/lutris/game_actions.py
@@ -167,15 +167,15 @@ class GameActions:
         Params:
             game (Game): Game instance without a database ID, populated with a fields the service can provides
         """
-        AddGameDialog(self.window, game=game)
+        AddGameDialog(self.window, game=game).run()
 
     def on_add_manually(self, _widget, *_args):
         """Callback that presents the Add game dialog"""
-        return AddGameDialog(self.window, game=self.game, runner=self.game.runner_name)
+        AddGameDialog(self.window, game=self.game, runner=self.game.runner_name).run()
 
     def on_edit_game_configuration(self, _widget):
         """Edit game preferences"""
-        EditGameConfigDialog(self.window, self.game)
+        EditGameConfigDialog(self.window, self.game).run()
 
     def on_add_favorite_game(self, _widget):
         """Add to favorite Games list"""
--- a/lutris/game_actions.py
+++ b/lutris/game_actions.py
@@ -237,6 +237,6 @@ class GameActions:
     def on_remove_game(self, *_args):
         """Callback that present the uninstall dialog to the user"""
         if self.game.is_installed:
-            UninstallGameDialog(game_id=self.game.id, parent=self.window)
+            UninstallGameDialog(game_id=self.game.id, parent=self.window).run()
         else:
-            RemoveGameDialog(game_id=self.game.id, parent=self.window)
+            RemoveGameDialog(game_id=self.game.id, parent=self.window).run()
--- a/lutris/game_actions.py
+++ b/lutris/game_actions.py
@@ -208,11 +208,11 @@ class GameActions:
         """Callback to open a game folder in the file browser"""
         path = self.game.get_browse_dir()
         if not path:
-            dialogs.NoticeDialog(_("This game has no installation directory"))
+            dialogs.NoticeDialog(_("This game has no installation directory"), parent=self.window)
         elif path_exists(path):
             open_uri("file://%s" % path)
         else:
-            dialogs.NoticeDialog(_("Can't open %s \nThe folder doesn't exist.") % path)
+            dialogs.NoticeDialog(_("Can't open %s \nThe folder doesn't exist.") % path, parent=self.window)
 
     def on_create_menu_shortcut(self, *_args):
         """Add the selected game to the system's Games menu."""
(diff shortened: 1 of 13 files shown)


strycore:
this is by design.

danieljohnson2:
Could you share what that design is? The current behavior is very odd.

strycore:
Your patch removes several calls to `application.show_window`. This method ensures that only one instance of a given window is created. Your patch removes that.
Not everything in Lutris is a dialog, we also have Gtk.Windows. This prevents opening several times a configuration window for a given game. Or launch several times the same installer.

danieljohnson2:
Thank you.

I figured modality would also achieve this, but maybe I could easily have overlooked a code-path. Perhaps the configure-game window should be like the installer- just an unattached window that does not interfere with the main Lutris window at all.

strycore:
this seems totally safe to do. we don't really need a reference to the main window for those configuration windows. 
any window that is managed by the application shouldn't be bothered by the main client window

danieljohnson2:
I've queued up PR #4004 to do this.

But I found that some configuration windows that seem like they should not be duplicated could be. Application.show_window is not always used. Configuring an existing game was one of these.

I've included a fix for that too.

PR #3955, opened 2021-12-29
DRM-be-gone!

### LINKED ISSUE #3950: [Feature Request] Option to add locally installed games in the Lutris Source view
Currently you can only install games, and it works for installable games, but it's cumbersome at rom based games. 

A better idea would be to have an option to add "locally installed games" to the entry

danieljohnson2:
This removes the DRM-free flag. This merely enables the "Locate installed game" command for every source- it does not enable you to install anything you couldn't before.

I don't think giving only GOG and Humble-Bundle sources the "Locate installed game" command makes a lot of sense; this command does not let you install a DRMed game at all. You would install it via another method, perhaps the Steam client say, and this command would make it easier to connect that game to Lutris.

And anyway, even Steam has a few DRM free games, hasn't it?

Resolves #3950

### DIFF
--- a/lutris/game_actions.py
+++ b/lutris/game_actions.py
@@ -162,7 +162,7 @@ class GameActions:
         self.game.emit("game-install")
 
     def on_locate_installed_game(self, _button, game):
-        """Show the user a dialog to import an existing install to a DRM free service
+        """Show the user a dialog to import an existing install of a game
 
         Params:
             game (Game): Game instance without a database ID, populated with a fields the service can provides
--- a/lutris/gui/widgets/game_bar.py
+++ b/lutris/gui/widgets/game_bar.py
@@ -203,14 +203,13 @@ class GameBar(Gtk.Box):
                 if self.service.local:
                     # Local services don't show an install dialog, they can be launched directly
                     button.set_label(_("Play"))
-                if self.service.drm_free:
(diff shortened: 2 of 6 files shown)


strycore:
I can't even....

danieljohnson2:
Now, now. You could even explain. This is not an unreasonable change.

strycore:
I think there is some confusion with the naming of this attribute. drm_free is supposed to separate services that provide games without the need of an additional launcher. They install games in a directory which can be relocated. This is not the case for Steam games or games from any other launcher.

Removing this drm_free attribute does not solve what the original ticket was asking for, which was to re-import previously installed lutris games.

If you think there is a name better suited for that attribute then I'm all open for it. But right now, the Humble and GOG services work differently from other services, this is why this attribute is present.

danieljohnson2:
I see; the seems related to the idea that each service game is related 1-to-1 with a Lutris PGA game. Just picking some random install via "locate" would be broken in some way, I suppose?

strycore:
the whole purpose of services is to have a way to install and launch games without the need of having install scripts on the website. For every other service, we can detect games that are installed and available. We do this for Steam and EGS. 
in this case 'drm_free' means that Lutris is in charge of managing the game files, where it's not the case for other services.

PR #3966, opened 2021-12-31
Duplicate game command

### LINKED ISSUE #3102: Add game cloning
I would like the ability to have multiple configurations per entry. For example, launching something like `lutris rungameid:18 --config 0` would run the pre-configured set of settings that come from the script, while `lutris rungameid:18 --config 1` would run a user-edited set of settings (for example, that disables DXVK or adds some other environment variable).

An easier (but uglier imo) fix for this would be the ability to clone an entry.

### LINKED ISSUE #3874: "Install another version" option is missing for GOG games

danieljohnson2:
This PR adds a "Duplicate" command to the game actions. This command duplicates the PGA database row and the config file, but *not* the game files. There's a question dialog that explains this and confirms the duplication.

This assigns a new name (and ID) to the duplicate.

This PR is actually built on the same branch as #3877, and includes the "Install another version" command- so that I can reuse the code that assigns unique game names. Hope we don't need to separate them.

Resolves #3102
Resolves #3874


danieljohnson2:
This should be merge-able now, but if you want me to strip out "Install another version" so you can have just "Duplicate game", let me know.

strycore:
there are problems with the "install another version" branch, I don't think it acknowledges that service view should be a one-to-one match with the user's library on said service.
the duplicate command by itself would be nice.

danieljohnson2:
I'll get you a new PR with *just* the duplicate game part.

I may try again with "install another version", since you don't seem entirely against it, but this time based on the duplicate game code instead of the other way around. Of course, this depends on whether I can understand what the bug in it is.

danieljohnson2:
The new PR is #4032; that's just the duplicate-game part of this PR, so I'm closing this one. No choose-your-own-adventure stuff!

PR #3972, opened 2022-01-01
Configuration of the WINE DPI setting

### LINKED ISSUE #3617: Add DPI manager to the Wine Runner
By default, Wine uses DPI of 96, which makes all GUI programs look incredibly small on a 4K screen, even when scaling is enabled in the DE.
While the user may choose to change the DPI themselves in Wine Configuration to remedy that, automating this in Lutris should be relatively easy.
Lutris simply needs to get DPI data from somewhere (at least on X11, this can be done with `xrdb -query | grep dpi`), convert it to hex (192 DPI, 2x scaling on a 4K screen, is `000000c0`, while default, 96 DPI, is `00000060`), and specify it in Wine's registry by setting `"LogPixels"` DWORD variables located in `[Software\\Wine\\Fonts]` and `[Control Panel\\Desktop]` to the according value in the format of `"LogPixels"=dword:000000c0`.
For example:
```
[Control Panel\\Desktop]
"LogPixels"=dword:000000c0
```
```
[Software\\Wine\\Fonts]
"LogPixels"=dword:000000c0
```
Specifying these will set DPI to 192.

danieljohnson2:
This provides configuration options to control WINE's DPI setting.

This works by setting a couple of registry values. If turned off (the default) it sets the DPI to 96 dpi, which produces no scaling. If on, it will auto-detect the scaling used by the primary monitor. You can override this with a second config option.

But it always sets the DPI, so that you can recover to a 'normal' state by turning it off. The bad news- WINE provides access to this setting in 'Wine Configuration', and this code just overwrites that setting.

This seems to work fairly well for integer scaling in GNOME, but gets more dicey with fractional scaling. The Windows GUI stuff works- we provide an integer scaling (x2 = 192 dpi, etc) but GNOME scales the result down so it all balances. But actual game-play seems to bug out quite easily.

I don't think I can fix that; WINE or GNOME will have to.

Resolves #3617

### DIFF
--- a/lutris/runners/wine.py
+++ b/lutris/runners/wine.py
@@ -12,7 +12,7 @@ from lutris.runners.commands.wine import (  # noqa: F401 pylint: disable=unused-
 )
 from lutris.runners.runner import Runner
 from lutris.util import system
-from lutris.util.display import DISPLAY_MANAGER
+from lutris.util.display import DISPLAY_MANAGER, get_default_dpi
 from lutris.util.graphics.vkquery import is_vulkan_supported
 from lutris.util.jobs import thread_safe_call
 from lutris.util.log import logger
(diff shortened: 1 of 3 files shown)


strycore:
except future issue for taking the DPI of the main monitor only but this is fine otherwise.

PR #4002, opened 2022-01-21
Migrate Coverart

telanus:


strycore:
cover art was just introduced, we don't do migration of stuff that has happened between releases.
the cover art folder can be moved manually.

PR #4005, opened 2022-01-23
Correct modal-dialog behavior

### LINKED ISSUE #3466: Modal dialog in main window blocks auxiliary window interaction

danieljohnson2:
Let's try to fix this *again*.

The configuration dialogs are independent of the main window, so a modal dialog that freezes the main window should not freeze them. But it does because GTK grabs all input in the window group.

So, each config dialog needs its own window group. Easy!

But the modal dialogs of the configuration dialog need to know their parent so they can grab input in *that* group, so I have done that too. This means those dialogs now dim the configuration dialog.

That is correct, but my efforts in this direction have been too ambitious in the past, so in this PR I touch only the modal dialogs of the configuration window. That's 1 question dialog, 4 error dialogs, and a directory chooser dialog. I've also added support for a parent in QuestionDialog to do this.

I hope I did not go too far this time, and we can get this bug fixed.

Resolves #3466

### DIFF
--- a/lutris/gui/config/common.py
+++ b/lutris/gui/config/common.py
@@ -52,6 +52,11 @@ class GameDialogCommon(Dialog):
         self.runner_index = None
         self.lutris_config = None
 
+        # These are independent windows, but start centered over
+        # a parent like a dialog. Not modal, not really transient,
+        # and does not share modality with other windows - so it
+        # needs its own window group.
+        Gtk.WindowGroup().add_window(self)
         GLib.idle_add(self.clear_transient_for)
 
     def clear_transient_for(self):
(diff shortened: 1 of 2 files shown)


danieljohnson2:
I seemed to have a failed lint in origin.py, but I did not touch that file. My IDE thinks the imports there are sorted correctly. I don't understand what's wrong here.

Issue #220, opened 2015-11-21
There needs a way to differenciate different versions of a game


strycore:
Even when filtering runners in the sidebar, a game can appear multiple times (with mess or mednafen). I also plan to add a platform filter which would help but what would be nice is overlaying a platform icon on the banner. This requires having the platform aspect implemented and icons for each platforms (consistent and good looking 32x32 icons)

Xodetaetl:
Having the platform information is long overdue. We should have had it from the start, we won't be able to add it backwards automatically to installed games with multiplatform runner.

I'd put the platform icon in the title label. It will be more visible, it won't spoil our beautiful banners and it will surely be easier to code.

Hopefully we can find libre 32x32 icons for most platforms. :)

strycore:
> we won't be able to add it backwards automatically to installed games with multiplatform runner.

Yes we can :) All the multiplatform runners have a 'machine' attribute in the game's config so based on that it's easy to extract the platform from that info.

Xodetaetl:
Oh, that's full of correctness. :P

cheater:
in the lutris launcher, i would like to be able to have a precedence of launchers from most favourite to least favourite, so that only games that are unique to that launcher are shown.

for example, let's say my precedence is like this, from most favourite to least favourite:

Steam, GOG, Amazon, Humble, Epic, Origin, Ubisoft

When I click on GOG on the left, I don't want to see games that I also have on Steam.
When I click on Amazon, I don't want to see games that I also have on either Steam or GOG.
Humble: don't show games I own on Steam, GOG, or Amazon.

etc etc

Have a filter in the hamburger menu that is called "Show only games you prefer to install with this launcher", and make it default on. If it's turned on, you can eg install a game on Amazon if you have it on Steam.

Make the precedence editable of course.

strycore:
@cheater no gonna happen

cheater:
Too bad, how come?

strycore:
Because it's an awful idea.

cheater:
... can you elaborate on this? what makes it so "awful"?

I'd like to be able to browse my games without seeing duplicates over and over.

cheater:
Everything OK over there? Do you normally talk to people like this?

strycore:
On Github, yes.

cheater:
Take it from someone who's been programming for 30 years, Github or not, that's not the right way to talk to people, whether you've written 100% of the code in the repository or are just an onlooker ... just makes you look bad.

[...]

PR #4032, opened 2022-02-05
Add a duplicate game command

### LINKED ISSUE #3102: Add game cloning

danieljohnson2:
This PR adds a "Duplicate" command to the game actions. This command duplicates the PGA database row and the config file, but not the game files. There's a question dialog that explains this and confirms the duplication.

This assigns a new name (and ID) to the duplicate.

Resolves #3102

### DIFF
--- a/lutris/config.py
+++ b/lutris/config.py
@@ -2,6 +2,7 @@
 
 import os
 import time
+from shutil import copyfile
 
 import yaml
 
@@ -26,6 +27,16 @@ def write_game_config(game_slug, config):
         logger.debug("Writing game config to %s", config_filename)
         config_file.write(yaml_config)
     return configpath
+
+
+def duplicate_game_config(game_slug, source_config_id):
+    """Copies an existing configuration file, giving it a new id that this
+    function returns."""
+    new_config_id = make_game_config_id(game_slug)
+    src_path = os.path.join(settings.CONFIG_DIR, "games/%s.yml" % source_config_id)
+    dest_path = os.path.join(settings.CONFIG_DIR, "games/%s.yml" % new_config_id)
+    copyfile(src_path, dest_path)
+    return new_config_id
 
 
 class LutrisConfig:
--- a/lutris/database/games.py
+++ b/lutris/database/games.py
@@ -112,7 +112,7 @@ def get_service_games(service):
 
 def get_game_by_field(value, field="slug"):
     """Query a game based on a database field"""
-    if field not in ("slug", "installer_slug", "id", "configpath"):
+    if field not in ("slug", "installer_slug", "id", "configpath", "name"):
         raise ValueError("Can't query by field '%s'" % field)
     game_result = sql.db_select(settings.PGA_DB, "games", condition=(field, value))
     if game_result:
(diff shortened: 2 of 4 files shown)


danieljohnson2:
Hmm. @strycore has given me a earful about the perils of having >1 PGA game associated with a service game. From my testing, maybe that's happening with this PR.

If you have a GOG game and duplicate it, that works. Go back to the GOG service view and try to start the game from there, and it will start the "first" of the two games- and which is first appears arbitrary. It may be the one with the lowest id, but it may just depend on what the SQLite query optimizer comes up with. Not sure.

Anyway, if we need the user to make an explicit choice here, this will need more work.

strycore:
what if you unset the service and service app id when a game is cloned?
or as I mentioned on another ticket, if you get more than 1 result, you can show a dialog like LaunchConfigDialog

danieljohnson2:
I think the dialog would be much better, and maybe I can fit that in during the long weekend.

But it that an acceptable during a beta? Or is this whole feature just going to wait for 5.11.0 anyway?

Maybe I should just clear the service stuff for now.

strycore:
there's an opportunity to close another issue in the process ( #220 ). Forget about services, we can apply that to the whole UI and show entries by slug instead of by ID. This means there could never be a duplicate in the view and a selection dialog can be shown in every case there is more than 1 result

danieljohnson2:
#220 does not seem obviously related: you'd want to have the dialog regardless of whether you have platform icons in the view. And platform icons don't help with duplicate games either, as they duplicate the platform information- they'd have the same badge on them.

The PR gives the games different *names* (it adds a number on the end) rather than different graphics.

But I guess you'd rather have the dialog in this PR, even though maybe it will have to await the next Lutris release before it is safe to merge.

strycore:
I meant it would be another way of solving #220. Instead of having to draw over game banners, we can just show a single entry then show a selection when launched.

strycore:
don't modify the launch config dialog, this is used at a game level and allows for multiple executables for a game.
instead, either subclass it or write a new similar dialog class for this purpose.

also, the game bar is going to be a problem. didn't think of that one. the game bar works at a game ID level, not slug. I don't really see a solution except ... stacking multiple game bars on top of one another? which seems like a bad solution...

danieljohnson2:
Yes, that's what the current code would do.

When both duplicates remain, though, it will pick a seemingly arbitrary one to use. Perhaps it should use a sort so that it consistently picks the same one every time, in a predictable way.

strycore:
yes, you should remove the service and service ID when cloning so only the initial game is linked to the service.

danieljohnson2:
Well, that's a 'no', but that should work too. So I'll do that.

I really feel dirty with all these forced-pushes, but I can't figure out how to rebase and keep a clean set of commits in the PR without doing that. I hope it is not causing problems for anyone!

strycore:
The feature works great and has already been very useful, thanks!

strycore:
Possible evolution of the feature: show a dialog with a single text entry for the new name and download media for the clone with the new name

danieljohnson2:
You're welcome.

And sure, we could ask for the new name.

But how would the new duplicate ever have different media than the original? The name does not determine what the media will be, does it?

strycore:
The media is queried with the "slug" field, which is generated from the name with the `slugify` function. So changing the name would also generate a new slug that would allow pulling new media.
This shouldn't be specific to the duplicate command but for any time we detect a change in the slug (for example, we don't download any media when the slug gets changed in the preferences)

The specific use case I was using for is with a Need For Speed Anthology I had installed. With duplicates, I could make individual entries for Porsche Unleashed, Hot Pursuit 2, etc.

danieljohnson2:
I see; the current Duplicate command adds a number to the name and does *not* change the slug, since adding a digit to the slug is not likely to match anything. In this way it keeps using the same media, so new download is needed.

The use case I had in mind was 'different configs, same game' not 'different games, same game files', and I think we'd want to support both.

Issue #4031, opened 2022-02-04
Bottom panel switches to a different game when it stops


tannisroot:
If another game is running and you switch to a different one, and the first game is closed by itself (like quitting it manually, closing the game window), not through Lutris, the bottom panel will switch to that stopped game all by itself, without user's interaction:
![Peek 2022-02-04 13-29](https://user-images.githubusercontent.com/10602045/152513690-224ef6c1-b56f-4510-9d0f-897ffe2dd8ea.gif)

It should be noted that so far, only I can reproduce this, for some bizarre reason.

danieljohnson2:
That is very strange, but it does not reproduce for me, either in the master branch or in 0.5.9. 

I notice in your screenie that the selected game does *not* change, only the game bar. But that is strange too, as looking at the code there seems to be no way for the game-bar to switch games except when the selected game changes.

So, two questions:

1. How do you close "Minecraft Dungeons"? F4? Window close box? In game quit command?
2. Does this happen for every game, or only "Minecraft Dungeons"?

tannisroot:
1. I close the game windows itself. Application itself closes and Lutris detects it as closed and switches its stated to stopped/not running.
2. This happens for every game.

tannisroot:
It doesn't really matter how you close it as long as the process of the game self terminates.

danieljohnson2:
I don't understand it, and it does not happen for me.

Are you seeing this on the latest dev stuff (master branch), or in the last release (0.5.9.1)? Maybe post your hardware information and logs?

*Something* must be different between your experience and mine to account for this, but I don't know what.

tannisroot:
I am experiencing this with master and latest release.
System info:
```[System]
OS:              Arch Linux rolling n/a
Arch:            x86_64
Kernel:          5.16.3-241-tkg-pds
Desktop:         GNOME
Display Server:  x11

[CPU]
Vendor:          AuthenticAMD
Model:           AMD Ryzen 7 3800X 8-Core Processor
Physical cores:  8
Logical cores:   16

[Memory]
RAM:             31.3 GB
Swap:            1.9 GB

[Graphics]
Vendor:          NVIDIA Corporation
OpenGL Renderer: NVIDIA GeForce GTX 1070/PCIe/SSE2
OpenGL Version:  4.6.0 NVIDIA 510.47.03
OpenGL Core:     4.6.0 NVIDIA 510.47.03
OpenGL ES:       OpenGL ES 3.2 NVIDIA 510.47.03
[...]

PR #4034, opened 2022-02-05
Add explicit checks that a version is available

danieljohnson2:
While we were failing to repro issue #4031 @tannisroot had a crash while creating a clean new profile. He'd failed to download version data for a DXVK component and Lutris could not cope.

In this weird scenario, a DllManager can be left with *no* version number: the version property is None.

This PR adds some checks and adds a logger warning in this scenario. The component will be unavailable, but Lutris won't try to download it.

This lets Lutris at least start.

I have not included any fancy retry logic for the version numbers. That stuff is treacherous; I'm just trying not to crash here.

I don't really have a solid reproduction for this; I had to test this by munging download_file() to always return None. That's more sabotage than bug.

Still, better to have these checks for robustness, I think.

### DIFF
--- a/lutris/util/wine/dll_manager.py
+++ b/lutris/util/wine/dll_manager.py
@@ -48,8 +48,12 @@ class DLLManager:
 
     @property
     def path(self):
-        """Path to local folder containing DDLs"""
-        return os.path.join(self.base_dir, self.version)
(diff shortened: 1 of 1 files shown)


tannisroot:
You can reproduce the issue with sabotage by blacklisting a domain (github) in pihole

PR #4037, opened 2022-02-06
Unhook the hooks keeping the GameBar alive

danieljohnson2:
While we were failing to repro issue #4031, I discovered that that game update signals were being processed for the non-selected game. That was a dead end; they were being processed by a leaked GameBar object that did not impact the UI.

When you switch between games, the current GameBar is destroyed and a new one takes its place. But the old one doesn't die! It continues, animated by a perverse unlife!

OK, not so much unlife, By perverse emission hooks; this old GameBar is still hooked up to various game related signals and still handles them- forever. It doesn't even crash. But it can never be collected by the GC either; it's a memory leak. It's also wasting time processing these events.

As far as I can tell, it's no worse than that. There are no real symptoms or anything.

Still, we can't old widgets lurching around groping after *braaaaains*. This PR adds code to unhook all the emission hooks for the GameBar when it is destroyed. I've verified that it is now collected promptly.

There are other emission hooks in Lutris, but I think all of those other objects survive for the life of Lutris anyway. They won't survive process termination in any case.

### DIFF
--- a/lutris/gui/widgets/game_bar.py
+++ b/lutris/gui/widgets/game_bar.py
@@ -7,6 +7,7 @@ from lutris import runners, services
 from lutris.database.games import get_game_by_field, get_game_for_service
 from lutris.game import Game
 from lutris.gui.widgets.utils import get_link_button
+from lutris.util.log import logger
 from lutris.util.strings import gtk_safe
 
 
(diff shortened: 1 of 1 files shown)


tannisroot:
could you push something again? For some bizarre reason github didn't run the workflows

danieljohnson2:
'kay. That lint failure is not, I think, due to this PR at all.

danieljohnson2:
Well, I did, and that made GitHub's check pass.

But jeeze, is that supposed to add all those random commits to my PR commit history? What a mess!

PR #4039, opened 2022-02-06
Add support for dgVoodoo.confg

### LINKED ISSUE #3867: Copy dgvoodoo2 config file to the prefix
Lutris needs to copy/symlink the dgVoodoo.conf file from inside the version package into \AppData\Roaming\dgVoodoo\ so that there's no watermark in-game.

danieljohnson2:
This is a config file that goes to a particular spot in the user profile in the WINE prefix, to support the dgvooodoo DLL manager.

This adds support for such files to DllManager. Managers provide the relative path within the profile, the base class does the rest.

The conf files are symlinked, and must be at the root of the component version's directory. If you have files to place in different directories but with the same filename, and they are different, this PR will fall short.

If the .conf file exists, this will rename it (with a ".orig" suffix) so it is not lost.

This PR also removes this symlink when disabled. It will do so even if there is no .orig backup, though I noticed that Lutris does not do this for the DLLs themselves. I don't understand this; it may be a bug.

Resolves #3867

### DIFF
--- a/lutris/util/wine/dgvoodoo2.py
+++ b/lutris/util/wine/dgvoodoo2.py
@@ -9,5 +9,5 @@ class dgvoodoo2Manager(DLLManager):
     base_dir = os.path.join(RUNTIME_DIR, "dgvoodoo2")
     versions_path = os.path.join(base_dir, "dgvoodoo2_versions.json")
     managed_dlls = ("d3dimm", "ddraw", "glide", "glide2x", "glide3x", )
-    managed_user_files = ("AppData/Roaming/dgVoodoo/dgVoodoo.conf", )
+    managed_appdata_files = ("dgVoodoo/dgVoodoo.conf", )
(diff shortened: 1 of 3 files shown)


danieljohnson2:
I can't test this properly since I don't even know what game uses it. @tannisroot said he would test it; I'll post here with the result if he does not.

danieljohnson2:
@tannisroot finds that this code does not always use the right directory. We need to find the correct AppData path, so I'll have to revise this.

tannisroot:
Seems to work OK

QmwJlHuSg9pa:
How can I override this config file, per-prefix? I have resorted to installing dgVoodoo 2 manually in each instance.

danieljohnson2:
I'm afraid that's not supported; what you are doing is what you have to do.

May I ask what you are trying to do here? Maybe we should support it.

PR #4047, opened 2022-02-12
Use mono and gecko from runtime

tannisroot:
To avoid asking user to click a button to download mono and gecko, we can provide them in our runtime and point Wine to the location so that it automatically picks those up as if from cache.
This also adds support for disabling mono/gecko installation through envars instead of overrides, which to me seems cleaner.
Both features require upcoming lutris wine 7.1


tannisroot:
I think there is a better solution to this than having mono in runtime

PR #4055, opened 2022-02-12
Stop button will resort to SIGKILL if SIGTERM fails

### LINKED ISSUE #4046: Lutris leaves behind dead winedevice processes
How to reproduce:
1) Launch a Wine game through Lutris
2) Stop it
Result: all wine processes are killed but winedevice.exe, which persists and can only be terminated via "sigkill".

I am not really sure why this happens, I haven't looked at our process monitoring code but just using wineserver to shut things down shouldn't leave any processes behind.

danieljohnson2:
This PR tries to SIGTERM all of a games processes, but if this does not kill them all, it waits up to 5 seconds and tries again with SIGKILL.

This seems to be an effective way to kill all wine processes, even when they don't want to die- but will still allow processes that will respond to SIGTERM to do so, as long as they do it promptly.

If it must wait for the game to die, it does this on a thread to avoid freezing the UI.

Resolves #4046

### DIFF
--- a/lutris/game.py
+++ b/lutris/game.py
@@ -546,26 +546,44 @@ class Game(GObject.Object):
         """Forces termination of a running game"""
 
         def get_pids():
+            """Finds the PIDs of processes that need killin'!"""
             pids = self.get_game_pids()
             if self.game_thread and self.game_thread.game_process:
                 pids.add(self.game_thread.game_process.pid)
             return pids
 
         def kill(pids, sig):
+            """Sends a signal to a process list, logging errors."""
             for pid in pids:
                 try:
                     os.kill(int(pid), sig)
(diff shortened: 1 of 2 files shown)


tannisroot:
I don't think 1 second is enough, some processes take a while to shut down

danieljohnson2:
Then I'll need to run this on a thread to avoid locking up the UI, I guess.

tannisroot:
I don't think this PR is a good solution to the issue

danieljohnson2:
OK, I've revised it for a 5 second asynchronous wait. It checks at 1/2 second intervals, so it should not be delaying things for a game that does stop promptly.

danieljohnson2:
Superseded by #4078

PR #4064, opened 2022-02-16
Add command line option --force-media-update

danieljohnson2:
Presently the only way to trigger the media update to repair your library's missing media is to edit the source code.

Perhaps we don't want to make this super easy to trigger, but this is going a bit far.

This PR adds a command line option to trigger the update. There's no environment variable; we don't want to encourage people to set this to happen on every Lutris startup, I think.

### DIFF
--- a/lutris/gui/application.py
+++ b/lutris/gui/application.py
@@ -44,7 +44,7 @@ from lutris.gui.installerwindow import InstallerWindow
 from lutris.gui.widgets.status_icon import LutrisStatusIcon
 from lutris.migrations import migrate
 from lutris.startup import init_lutris, run_all_checks, update_runtime
-from lutris.util import datapath, log
+from lutris.util import datapath, log, update_cache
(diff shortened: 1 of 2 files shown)


strycore:
I'm not sure why this is needed at all since medias should not need to be repaired as part of a user action.
Maybe we can consider a --force-updates flag which would ignore the updates.json entirely for both runtimes and media. (Again, not sure why this would be needed.)

But something that is specific to medias, that goes poking into the cache file to remove the update key, is taking it a bit too far as you say.

danieljohnson2:
I think this PR was based on two misapprehensions- that this sync_media feature was supposed to repair *all* media, and that users should be prevented from doing this easily, for some reason. I didn't understand that the 'Reload' button would do this for a service.

So, yeah, there should just be a 'Reload' button on the 'Games' sidebar row to do this, and not this command line switch. I'll close it.

PR #4065, opened 2022-02-16
Add support for system dark mode

A6GibKm:
Adds a StyleManager GObject which reads the user prefernce for dark
mode. StyleManager counts with a GProperty `is_dark` which reflects this,
when the preference changes lutris will change its color scheme via "gtk-application-prefer-dark-mode".

The app has a setting for dark mode, in all cases, the app is set to
dark mode when either the user preference in the system or in the app's
config is set to dark mode.

For the most part, the code is taken from libhandy.


A6GibKm:
In an ideal world, we use libhandy => 1.5 for reading the user preference, but it does not have a stable release yet containing these changes.

A6GibKm:
Is the copyright ok, or should I put my name?

A6GibKm:
I taken the liberty to make the current dark mode switch work at runtime.

strycore:
Looks good, merging so it can be tested in the next beta

PR #4067, opened 2022-02-17
Repair media for all Lutris service games, not just the PGA

danieljohnson2:
This PR just throws the service games into the slug set to be downloaded if missing.

Since sync_media only downloads Lutris media, only the Lutris service is included.

This may be slow if you have no media at all, but hopefully that will be rare.


strycore:
there's no such thing as lutris service games.
is that why the lutris game list is no longer showing?

strycore:
Ok, I removed the weird stuff we were doing while loading the media during a service refresh and now it's doing the right thing.

PR #4074, opened 2022-02-21
Fix incorrect EasyRPG runner executable path

### LINKED ISSUE #3529: After installation of EasyRPG runner executable for the runner has to be set manually
After installation of EasyRPG runner its executable is downloaded to `~/.local/share/lutris/runners/easyrpg/easyrpg-player`, but not used, so runner is not usable from lutris till `Custom executable for runner` in `Runner options` is set accordingly.

Layerex:
Closes #3529

### DIFF
--- a/lutris/runners/easyrpg.py
+++ b/lutris/runners/easyrpg.py
@@ -12,7 +12,7 @@ class easyrpg(Runner):
     platforms = [_("Linux")]
     runnable_alone = True
     entry_point_option = "project_path"
-    runner_executable = "easyrpg-player"
+    runner_executable = "easyrpg/easyrpg-player"
     download_url = "https://easyrpg.org/downloads/player/0.7.0/easyrpg-player-0.7.0-linux.tar.gz"
 
     game_options = [


strycore:
This will make Lutris unable to find the runner since the archive doesn't contain that folder.

There are 2 solutions to this problem:
- we host our own archive for easyrpg, with the correct directory structure (the easiest, and what we usually do)
- we could change the runner extraction process to make sure that they all extract in a directory that matches the runner name

The 2nd solution would be the most sustainable and it would allow using more runners from 3rd parties without having to host them ourselves.

I could see the feature implemented as follows:

- Instead of extracting the runner directly in the runners dir, extract it in a temp directory
- If the temp directory contains another directory matching the exact name of the runner, move it to the runners dir
- If the temp directory contains a directory with a name that doesn't match the runner, rename then move it. (solves fs-uae)
- If the temp directory contains one or more files, move the contents of the temp dir in a folder named after the runner (solves easyrpg)

Layerex:
> This will make Lutris unable to find the runner since the archive doesn't contain that folder.

I am willing to help, but I don't understand the problem.
With my fix everything seems to work just fine, the runner is downloaded and unpacked and a game starts afterwards:

1. Lutris downloads the archive (https://easyrpg.org/downloads/player/0.7.0/easyrpg-player-0.7.0-linux.tar.gz), which contains only the `easyrpg-player` executable 
2. It unpacks the archive to `~/.local/share/lutris/runners/easyrpg`
3. The relative path in `runner_executable` variable (`easyrpg/easyrpg-player`) seems to be resolved relative to `~/.local/share/lutris/runners` directory
4. Lutris finds the `~/.local/share/lutris/runners/easyrpg/easyrpg-player` executable and starts a game with it

All runners seem to be resolved in the same fashion.
I don't understand, where is lutris supposed to fail.

strycore:
ok, I think the problem only affects fs-uae then

PR #4078, opened 2022-02-21
Stop harder, but WINE gets wineserver -k not SIGTERM

### LINKED ISSUE #4046: Lutris leaves behind dead winedevice processes

danieljohnson2:
OK, I think this is what was wanted.

The 'Stop' command will send SIGTERM to all non-WINE games, but WINE games get wineserver -k instead.

In either case, if this leaves any processes alive we'll wait up to 5 seconds for the processes to die. If they do not, they get SIGKILL.

This also removes the "Kill all WINE processes" command, as it would just be 'Stop' without the SIGKILL now. The installer 'winekill' task should still work as before, I believe.

Resolves #4046

### DIFF
--- a/lutris/game.py
+++ b/lutris/game.py
@@ -542,7 +542,7 @@ class Game(GObject.Object):
             np_file.write(self.name)
 
     def force_stop(self):
-        # If SIGTERM fails, wait a few seconds and try SIGKILL on any survivors
+        # If force_stop_game fails, wait a few seconds and try SIGKILL on any survivors
         self.runner.force_stop_game(self)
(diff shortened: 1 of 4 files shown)


strycore:
Looks good, let's have this for testing in the next beta

PR #4103, opened 2022-02-26
Revamped wine runner screen and added subscreen to view apps of given runner

cazzoo:
Changed config runner view to add new button to show active apps for the given installed runners.
I also changed the install action to a dedicated button that is more explicit for the end-users.

Screen now appears as following:
image

Then, when clicking on "View %d apps", the subscreen shows the apps using the runner. This is looking like:
image


cazzoo:
Hi @strycore,
Any chance you get a review/check on this one?
Comments welcome

strycore:
Looks good to merge!

PR #4104, opened 2022-02-26
(i18n) Update pt_BR translation

kassindornelles:
Full Portuguese (Brazil) translation

### DIFF
--- a/po/pt_BR.po
+++ b/po/pt_BR.po
@@ -44,8 +44,8 @@ msgid ""
 msgstr ""
 "Lutris ajuda você a instalar e jogar videogames de todas as épocas e da "
 "maioria dos sistemas de jogos. Ao alavancar e combinar emuladores "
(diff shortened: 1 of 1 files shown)


kassindornelles:
Okay, after some small fixes here and there everything looks correct and i'm pretty happy with it

danieljohnson2:
I think there's still a mistake on line 4347, though I don't speak Portuguese. But I think you have rendered "If blank or 'auto'," as "Se estiver em branco ou 'automático'," and the thing is it actually expects the exact text "auto", not anything longer, here.

Perhaps this string should be localized, but it isn't- and there are other examples like it in the settings where the translation has 'auto'

kassindornelles:
> 

is that line the only one that you saw a problem related to "automatico"?

danieljohnson2:
Yes, that's all I spotted. Though for all I know "automático" means "pickled herring".

PR #4151, opened 2022-03-17
Implementation of steam shortcuts

liberavia:
This is a follow-up of https://github.com/lutris/lutris/pull/2729 for enabling user to add lutris games to steam

### DIFF
--- a/lutris/game_actions.py
+++ b/lutris/game_actions.py
@@ -19,8 +19,8 @@ from lutris.gui.dialogs.log import LogWindow
 from lutris.gui.dialogs.uninstall_game import RemoveGameDialog, UninstallGameDialog
 from lutris.gui.widgets.utils import open_uri
 from lutris.util import xdgshortcuts
-from lutris.util.steam import shortcut as steam_shortcut
 from lutris.util.log import logger
+from lutris.util.steam import shortcut as steam_shortcut
 from lutris.util.strings import gtk_safe
 from lutris.util.system import path_exists
 
--- a/lutris/game_actions.py
+++ b/lutris/game_actions.py
@@ -138,6 +138,7 @@ class GameActions:
             ),
             "steam-shortcut": (
                 self.game.is_installed
+                and steam_shortcut.vdf_file_exists()
                 and not steam_shortcut.shortcut_exists(self.game)
                 and not steam_shortcut.has_steamtype_runner(self.game)
             ),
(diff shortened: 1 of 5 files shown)


liberavia:
Currently WIP

liberavia:
The feature is now working but cannot be merged due there is a big showstopper, which can be seen here:
https://github.com/liberavia/lutris/blob/e5076a8df0d871c6e56329a1fa1c160455de80ef/lutris/util/steam/shortcut.py#L12

My Problem is, that the s**team_user_id** I need is a **9-Digit number** instead of 15-Digit number `get_steam__userid()` delivers

I had some research on vdf file but I wasn't able to determine where this number is coming from. Currently there is only the path to MY shortcuts.vdf, so that current state of the code is only working for me at the moment :(

Anyone an idea where this 9-Digit-Id is coming from?

liberavia:
So I finally got arround the steam id problem in path and it's working without the need of that id

Things are basically working now, so that you can add and delete shortcuts to the steam library via the installer or via contextmenu of the game. Steam games are kept out of that capability (self explaining).

With that, especially steam deck users should receive an easy way to add their games to steamui.

However there are still issues we should keep in mind for further improving:
*  Currently we can't deal with multiple steam users. To be safe here, only the first hit of shortcuts.vdf will be checked and processed
* I did not find a way to determine where to put the coverart from lutris to steam due to I can't find the place where steam is saving and assigning those. Only the icon will be set at the moment.

liberavia:
Okay, at least I got some information on where artwork is stored and how the non-steam-gameid is calculated.
Path, where the artwork is stored is:
`~/.steam/steam/userdata/<9-Digit Steam Userid>/config/grid`
Yeah, again this ID I could not determine...
Files in grid folder are named the following:
<gameid>p.jpg => For modern coverart => Most important
<gameid>_logo.jpg => For details page logo

The gameid seems to be calculated of executable target, whatever that exectly is AND the game's name as we can assume from the following:
https://blog.yo1.dog/calculate-id-for-non-steam-games-js/

liberavia:
Ok this is feature complete now. Please validate and merge

Even I still don't know how to get this 9-Digit Steam UserID, everything is now working, include multiple useraccounts and automatic artwork set!

Also I made sure this shortcut option is only available if a shortcuts.vdf file exists

strycore:
Thanks! I'll test the feature and possibly see if we can find the 9 digit ID

PR #4164, opened 2022-03-22
Add NoneType check for variable folder in util/wine/prefix.py

1reverseengineer:
Screenshot_20220322_202824

Upon opening winetricks for a faulty origin installation, a stack-trace / NoneType exception is thrown out for the variable folder, which calls function get_registry_key.

- Propersal: Adding an if-else statement, on line 65, to check whether the result of function get_registry_key equals NoneType and make use of logger.

### DIFF
--- a/lutris/util/wine/prefix.py
+++ b/lutris/util/wine/prefix.py
@@ -62,8 +62,8 @@ class WinePrefixManager:
             self.hkcu_prefix + "/Software/Microsoft/Windows/CurrentVersion/Explorer/Shell Folders",
             "AppData",
         )
-        if(folder is None):
-            logger.error("Get Registry Key function returned NoneType to variable folder.")
+        if folder is None:
+            logger.warning("Get Registry Key function returned NoneType to variable folder.")
         else:
             # Don't try to resolve the Windows path we get- there's
             # just two options, the Vista-and later option and the


danieljohnson2:
I think this is a sensible defensive check, but stylistically I think it might be more "Pythonic" to say:
```
if  folder:
```
on line 65. The parentheses are not required in Python and I think the error case should trigger for an empty string too, so I would remove 'is None' and just rely on the '"falsiness" of the folder value.

I might also have logged a warning instead of an error for this case, since we're going to go ahead and try to use the directory anyway.

18iteration:
> I think this is a sensible defensive check, but stylistically I think it might be more "Pythonic" to say:
> ```
> if  folder:
> ```
> on line 65. The parentheses are not required in Python and I think the error case should trigger for an empty string too, so I would remove 'is None' and just rely on the '"falsiness" of the folder value.
> 
> I might also have logged a warning instead of an error for this case, since we're going to go ahead and try to use the directory anyway.

Hey thanks for the reply.

Both ways would be sufficient and readable, I personally see the non "Pythonic" way more clearer and readable. Either path is fine, I don't think this is a cause for concern.

Secondly I agree that it is a warning rather than an error, I will edit it later.

PR #4168, opened 2022-03-25
Installer script  YAML schema

mpetuska:
This PR introduces a YAML schema for installer scripts (with an option to easily convert it to JSON schema is needed).

With this, all install scripts now can be linked to the schema and provide on-the-fly autocompletion and validation.


strycore:
Thanks! I feel that should mainly be used for the website, don't you think?
We have some validation rules but validating against a schema would be really nice!

mpetuska:
I don't agree. Ideally, we'd have some command in lutris cli to publish scripts written locally as developing them locally is much more convenient thanks to autoformatting and now even autocompletion via the new schema. Aditionally, any sane dev will want to test their scripts locally before publishing them and as such will need to have it on their local PC anyways.

In short, I found website publishing to be quite botheersome 😞

strycore:
There are some plans to submit install scripts from the desktop client but there's quite a bit of development to get there.
The ideal configuration would be to isolate the schema + our current validation rule into their own library that can be used by both the website and the client (and possibly other tools too)

mpetuska:
I think once we iron out the schema to perfection (my version is still ambiguous in some areas as I wasn't clear on some of your domain entity usage when writing it), it alone should be able to cover all use-cases as pretty much any programming language has some sort of json schema validation library.

strycore:
Some of our validation rules are very specific to Lutris and in some cases, can make database calls. 
That's the current validation rules for our install scripts: https://github.com/lutris/website/blob/master/games/util/installer.py

Some of them can easily be handled with a schema (files_is_an_array,  game_is_a_dict) but others are more complex like uses_a_valid_wine_version. That said it might be possible to reference that data in the schema as well.

PR #4186, opened 2022-04-03
Vulkan loaders: separate amdvlk and amdgpu-pro when possible

AdelKS:
Hello,

As the (fresh) maintainer of the `amdgpu-pro-vulkan` driver on Gentoo, I mirrored something that is also done in the `AUR`'s `PKGBUILD` of the same driver : rename `amd_icd` to `amd_pro_icd` so it can be differentiated from the open source `amdvlk` driver who offers the same name for the icd loader.

This PR is a small tweak in Lutris to recognize when this separation happens then offers both drivers as separate choices.

Note: I relaxed the condition to recognize an icd loaders as `amdvlk` or `amdgpu-pro-vulkan` since other distros can follow suite with a different naming scheme. This should be general enough to capture it properly, and hopefully not too general to confuse it with another driver.

Hopefully this is helpful, otherwise feel free to close this PR.

Adel

### DIFF
--- a/lutris/sysoptions.py
+++ b/lutris/sysoptions.py
@@ -82,6 +82,7 @@ def get_vk_icd_choices():
     amdradv = []
     nvidia = []
     amdvlk = []
+    amdvlkpro = []
     choices = [(_("Auto: WARNING -- No Vulkan Loader detected!"), "")]
     icd_files = defaultdict(list)
     # Add loaders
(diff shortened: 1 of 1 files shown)


MarijnS95 (approved):
I was looking for and made the same change yesterday (on arch, indeed the separation/renaming you describe), this works perfectly :+1:

strycore:
Thanks!

PR #4189, opened 2022-04-03
Misc Meson improvements

A6GibKm:


meson.build
```diff
@@ -16,6 +16,14 @@ mandir = get_option('mandir')
 lutrisdir = join_paths(pylibdir, 'lutris')
 podir = join_paths(meson.source_root(), 'po')
 
+dependency('gio-2.0')
+dependency('glib-2.0')
+dependency('gobject-introspection-1.0')
+dependency('gtk+-3.0')
+dependency('webkit2gtk-4.0')
+dependency('gnome-desktop-3.0', required: false)
+dependency('appindicator-0.1', required: false)
```
Jan200101, line 25: just having dependencies doesn't do anything.
What consumes these?

A6GibKm: Meson obviously, it is meant as a hint that you are missing a dependency. This is specially important with python apps, when you are not going to crash until runtime.

I maintain the flatpak version of the app and I just realized it used libnotify while randomly browsing the code, I had to patch the flatpak and send a patch here too, I would have send those patches earlier if I knew it was used.

Jan200101: yeah but those are the low level dependencies, not the python dependencies.
gtk3 being available does not mean its available in python.

Also since no one consumes these and are only checked against, aren't the dependencies that aren't required obsolete?

Perhaps a better solution would be to generate python code on the fly and check if it can import the required modules

A6GibKm: > aren't the dependencies that aren't required obsolete?

No, libappindicator is used optionally for example.

>  gtk3 being available does not mean its available in python.

if the gobject instrospection dep is satisfied and then having gtk3 means having gtk3 in python.

> Perhaps a better solution would be to generate python code on the fly and check if it can import the required modules

this is already the case, but it is simpler if packagers had a way to determine which dependencies they need, which is what this helps with.

meson.build
```diff
@@ -69,11 +77,33 @@ i18n.merge_file(
...
+appstreamcli = find_program('appstream-util', required: false)
+if appstreamcli.found()
+  test (
+    'Validate metainfo file',
+    appstreamcli,
+    args: ['validate-relax', '--nonet', metainfo_file.full_path() ]
+  )
+endif
+
+meson.add_install_script('utils/meson_post_install.py')
```
eli-schwartz, line 109: If you add a `meson_version: '>=0.57.0'` to the `project()` kwargs, you can use the gnome module's [`post_install()`](https://mesonbuild.com/Gnome-module.html#gnomepost_install) method to do this for you. It correctly handles different names for gtk-update-icon-cache, skips if destdir is set, calculates the right datadir etc...

eli-schwartz (changes requested):

meson.build
A6GibKm: It only works well with >=0.59, given that Lutris stance of supporting distros from other eras I think this is the correct approach.

eli-schwartz: Add a todo comment. :)

Also: it's difficult to know in a bubble what the project policy is, since there is no current minimum version specified. Nevertheless, it uses features not available in Debian Buster or Ubuntu 18.04.

Point taken on Meson 0.59, I overlooked that update-desktop-database is needed but wasn't in the initial implementation alongside update-icon-cache.

A6GibKm: > it uses features not available in Debian Buster or Ubuntu 18.04, dunno if that matters.

What features for example?

eli-schwartz: I set a really low number as a test, and Meson reported the following feature warnings:

```
...

meson.build:29: WARNING: Project targeting '>=0.22.0' but tried to use feature introduced in '0.50.0': install arg in configure_file.
Configuring optional_settings.py using configuration
po/meson.build:6: WARNING: Project targeting '>=0.22.0' but tried to use feature introduced in '0.50.0': install_dir arg in i18n.gettext.

...

WARNING: Project specifies a minimum meson_version '>=0.22.0' but uses features which were added in newer versions:
 * 0.37.0: {'i18n.merge_file'}
 * 0.38.0: {'install_mode arg in install_data'}
 * 0.46.0: {'Python Module'}
 * 0.50.0: {'install arg in configure_file', 'install_dir arg in i18n.gettext'}
```

eli-schwartz: Actually you could drop both 0.50 uses pretty easily.

eli-schwartz: #4192

A6GibKm: Thanks, ill remove my commit setting it to 0.50 then

meson.build
Jan200101: > No, libappindicator is used optionally for example.

Yeah but the actual dependency is not consumed by anything, thus this is used for nothing more than a fancy message and anything that is not required is essentially pointless.

> if the gobject instrospection dep is satisfied then having gtk3 means having gtk3 in python.

right, bad example since the python gobject bindings enable any gobject based library to be used from within python.

Lets take two different examples which are not covered by this:

- libmagic (dynamically loaded [here](https://github.com/lutris/lutris/blob/master/lutris/util/magic.py#L209))
    *this* would fit into this dependency check, but since there are many different names that are checked against (and checking against multiple names would require bumping the required meson version even more) its not feasible to do this
- PyYAML
    technically speaking this depends on libyaml, but we cannot ensure that if libyaml is present pyyaml is also

> this is already the case, but it is simpler if packagers had a way to determine which dependencies they need, which is what this helps with.

How dependencies are structured out, packaged and named differs greatly between distros (and operating systems) so packagers would still need to do some amount of digging around.

But ignoring that, these dependencies generally look for files used for development with those libraries using pkg-config and cmake, but none of which are required to run the python modules.

Taking the Fedora package as an example, this would require adding more dependencies (especially -devel dependencies) that aren't needed to build lutris for effectively no gain

Here is a practical example for gobject

Before this PR
```spec
BuildRequires:  python3-gobject # ropes in the gobject-introspection
```

After this PR
```spec
BuildRequires:  python3-gobject, gobject-introspection-devel
```

A6GibKm: You are right, forgot about that. Removed. Its such a shame thats not possible to declare dependencies in python in any sane way in meson.

eli-schwartz: Find_installation() can check for `modules: []` that must be able to be imported, so there is that. This could help for pyyaml although gobject is a bit more complicated...

PR #4197, opened 2022-04-06
Make steam shortcut support more tolerant of shortcut.vdf variants

danieljohnson2:
This isn't a complete fix, since tolerating errors occurring inside GameActions.get_displayed_entries() is going to be more complicated than this, or so I fear. But this seems to deal with the immediate problem.

This makes Lutris accept 'shortcuts.vdf' files that have the AppName key listed as 'appname', in lower case, or at least to not crash if this is missing entirely.

Resolves #4195
Resolves #4196

Issue #4195, opened 2022-04-05
can't edit games on lutris ~master


WebFreak001:
### Bug description

on current master branch (installed from AUR through lutris-git as regular lutris is currently broken due to #4141) I can start games again, but now the bottom bar is missing, right-clicking games doesn't work and an exception appears when trying to right-click.

### How to Reproduce

1. Install `lutris-git` through AUR
2. Run it
3. Try to edit a game, see that bottom bar is missing and right click doesn't work

### Expected behavior

Bottom bar should be there and right-click should work.

### Log output

```shell
Traceback (most recent call last):
  File "/usr/lib/python3.10/site-packages/lutris/gui/views/base.py", line 52, in popup_contextual_menu
    self.contextual_menu.popup(event, game_actions)
  File "/usr/lib/python3.10/site-packages/lutris/gui/widgets/contextual_menu.py", line 53, in popup
    displayed = game_actions.get_displayed_entries()
  File "/usr/lib/python3.10/site-packages/lutris/game_actions.py", line 141, in get_displayed_entries
    and not steam_shortcut.all_shortcuts_set(self.game)
  File "/usr/lib/python3.10/site-packages/lutris/util/steam/shortcut.py", line 48, in all_shortcuts_set
    shortcut_found = [
  File "/usr/lib/python3.10/site-packages/lutris/util/steam/shortcut.py", line 50, in <listcomp>
    if game.name in s['AppName']
KeyError: 'AppName'
```
```


### System Information

```shell
OS:              Arch Linux rolling n/a
Arch:            x86_64
Kernel:          5.17.1-zen1-1-zen
Desktop:         i3
Display Server:  x11
```

### Checklist:

- [X] I'm not asking for support with a game or the wine runner.
- [X] I have followed the above mentioned guides and have all the graphics and wine dependencies installed.
- [X] I have checked for existing issues that describe my problem prior to opening this one.
- [X] I understand that improperly formatted bug reports may be closed without explanation.

danieljohnson2:
I have a theory about this, but no Steam account to test with. And anyway, this is probably specific to your shortcuts.vdf file.

I read [here](https://github.com/SteamGridDB/steam-rom-manager/issues/2) that somehow Steam writes out the shortcuts with 'AppName' in lower-case. Lutris will fail if that happens.

[...]

PR #4199, opened 2022-04-06
Add exception handling to get_displayed_entries()

danieljohnson2:
This PR protects get_displayed_entries() from exceptions; If they occur they are logged and we resolve to 'False', so the command in question is hidden but other commands can still appear. Propagating the exception here prevents parts of the UI from being built, which is not nice.

The code for this is kinda of ugly; I don't really see a better way that isn't a major rewrite. The dict of actions and flags now has (mostly) lambdas for its values, and then a dict-comprehension is used to resolve each lambda to a bool. That is done with a try-except, and there you go.

I argue that while this *is* nasty, it is better than the status-quo.

This isn't really a fix for #4195, but it is a mitigation.

I already submitted #4197, but that is the actual fix for the underlying  bug. This isn't a choose-your-own adventure PR. You should turn to page 4197 and also page 4199.

### DIFF
--- a/lutris/game_actions.py
+++ b/lutris/game_actions.py
@@ -109,7 +109,7 @@ class GameActions:
 
     def get_displayed_entries(self):
         """Return a dictionary of actions that should be shown for a game"""
-        return {
+        actions = {
             "add": not self.game.is_installed,
             "duplicate": True,
             "install": not self.game.is_installed,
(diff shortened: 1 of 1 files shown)


strycore:
This seems convoluted. Probably because no error handling should be happening here but it should be done in the called functions.

danieljohnson2:
It is convoluted, and I do not like it, but it is fairly general. It handles the error in a place that knows how to respond- in this case by hiding a command.

Still.

Lutris needs some way to handle corrupted shortcuts.vdf. I'll put something together that handles exceptions in the called functions for this instead. It should be doable that way, if there are not too many awkward other callers.

PR #4201, opened 2022-04-07
Remove keyring dependency

A6GibKm:
The file util/keyring.py is not imported anywhere and the keyring
dependency has a big chain of dependencies.

Do note that in the case where one needs to interact with the keyring, doing so via libsecret is preferable since it actually works on a sandbox.


strycore:
At some point I wanted to use Gnome Keyring to store credentials but it never went anywhere.
Now that Flatpak is a thing, this no longer makes sense, thanks for removing it.

PR #4220, opened 2022-04-14
Add a check to make sure the current resolution is usable

### LINKED ISSUE #4218: Getting error : ValueError("invalid literal for int() with base 10: ''") when trying to download any game
### Bug description

Lutris is not allowing me to install any game. It seems like a python dependency but I cannot seem to figure out with dependency it is. Installed lutris through dnf on Fedora 35

Gives error: ValueError("invalid literal for int() with base 10: ''")  when trying to install any game.

### How to Reproduce

1. Go to download any game
2. Click on Install
3. See ValueError("invalid literal for int() with base 10: ''") 

### Expected behavior

Game should install.

### Log output

```shell
INFO 2022-04-13 13:10:17,716 [startup.init_lutris:173]:Starting Lutris 0.5.10WARNING

2022-04-13 13:10:17,731 [libretro.get_libretro_cores:24]:No folder at /home/veri/.local/share/lutris/runners/retroarch/INFO

2022-04-13 13:10:17,748 [startup.check_driver:57]:Using NVIDIA drivers 510.60.02 for x86_64INFO

2022-04-13 13:10:17,748 [startup.check_driver:61]:GPU: NVIDIA GeForce RTX 3060 Laptop GPUINFO

2022-04-13 13:10:17,748 [startup.check_driver:77]:GPU: 10DE:2560 17AA:3801 (nvidia drivers)INFO 2022-04-13 13:10:17,748 [startup.check_driver:77]:GPU: 1002:1638 17AA:380C (amdgpu drivers)INFO

2022-04-13 13:10:17,821 [startup.update_runtime:213]:Startup complete


DEBUG

2022-04-13 13:10:17,879 [lutriswindow.update_store:437]:Showing 0 games

2022-04-13 13:10:49,992: Starting Lutris 0.5.10

2022-04-13 13:10:50,005: Using NVIDIA drivers 510.60.02 for x86_64

2022-04-13 13:10:50,005: GPU: NVIDIA GeForce RTX 3060 Laptop GPU

2022-04-13 13:10:50,005: GPU: 10DE:2560 17AA:3801 (nvidia drivers)

2022-04-13 13:10:50,005: GPU: 1002:1638 17AA:380C (amdgpu drivers)

2022-04-13 13:10:50,052: Startup complete

2022-04-13 13:10:50,054: Fetching installer https://lutris.net/api/installers/final-fantasy-xiv-online-xivlauncher

2022-04-13 13:10:50,054: GET https://lutris.net/api/installers/final-fantasy-xiv-online-xivlauncher

2022-04-13 13:10:50,316: Showing window InstallerWindowfinal-fantasy-xiv-online

2022-04-13 13:10:52,934: Retrieving current resolution

2022-04-13 13:10:52,934: Could not retrieve the current display mode

2022-04-13 13:11:01,331: Checking if service provide extra files

2022-04-13 13:11:01,332: No service to retrieve files from

2022-04-13 13:11:02,592: ⬇ https://kamori.goats.dev/Proxy/Update/Release/Setup.exe

2022-04-13 13:11:03,631: Finished downloading https://kamori.goats.dev/Proxy/Update/Release/Setup.exe

2022-04-13 13:11:04,095: Download completed

2022-04-13 13:11:04,095: launchersetup is available

2022-04-13 13:11:04,095: All files available

2022-04-13 13:11:04,095: All files are available, continuing install

2022-04-13 13:11:04,096: Installer command: {'task': {'arch': 'win64', 'install_gecko': False, 'install_mono': False, 'name': 'create_prefix', 'prefix': '$GAMEDIR'}}

2022-04-13 13:11:04,104: Error while completing task <bound method CommandsMixin.task of <interpreter.ScriptInterpreter object at 0x7fa4d83aa540 (lutris+installer+interpreter+ScriptInterpreter at 0x561dec5156c0)>>: <class 'ValueError'> invalid literal for int() with base 10: '' File "/usr/lib/python3.10/site-packages/lutris/util/jobs.py", line 34, in target result = self.function(*args, **kwargs) File "/usr/lib/python3.10/site-packages/lutris/installer/commands.py", line 432, in task value = self._substitute(data[key]) File "/usr/lib/python3.10/site-packages/lutris/installer/interpreter.py", line 420, in _substitute return system.substitute(template_string, self._get_string_replacements()) File "/usr/lib/python3.10/site-packages/lutris/installer/interpreter.py", line 400, in _get_string_replacements "RESOLUTION_WIDTH_HEX": hex(int(self.current_resolution[0])),

2022-04-13 13:11:04,105: Last install command failed, show error

2022-04-13 13:11:18,402: Cancelling installation of FINAL FANTASY XIV OnlineTraceback (most recent call last): File "/usr/lib/python3.10/site-packages/lutris/gui/installerwindow.py", line 525, in on_destroy if self.confirm_cancel(): File "/usr/lib/python3.10/site-packages/lutris/gui/installerwindow.py", line 571, in confirm_cancel self.interpreter.revert(remove_game_dir=remove_checkbox.get_active()) File "/usr/lib/python3.10/site-packages/lutris/installer/interpreter.py", line 376, in revert self.task({"name": "winekill"}) File "/usr/lib/python3.10/site-packages/lutris/installer/commands.py", line 432, in task value = self._substitute(data[key]) File "/usr/lib/python3.10/site-packages/lutris/installer/interpreter.py", line 420, in _substitute return system.substitute(template_string, self._get_string_replacements()) File "/usr/lib/python3.10/site-packages/lutris/installer/interpreter.py", line 400, in _get_string_replacements "RESOLUTION_WIDTH_HEX": hex(int(self.current_resolution[0])),ValueError: invalid literal for int() with base 10:

''2022-04-13 13:11:18,431: Removed window InstallerWindowfinal-fantasy-xiv-online
```

### System Information

```shell
[System]
OS:              Fedora Linux 35 
Arch:            x86_64
Kernel:          5.16.18-200.fc35.x86_64
Desktop:         GNOME
Display Server:  x11

[CPU]
Vendor:          AuthenticAMD
Model:           AMD Ryzen 7 5800H with Radeon Graphics
Physical cores:  8
Logical cores:   16

[Memory]
RAM:             38.6 GB
Swap:            8.0 GB

[Graphics]
Vendor:          AMD
OpenGL Renderer: AMD RENOIR (DRM 3.44.0, 5.16.18-200.fc35.x86_64, LLVM 13.0.0)
OpenGL Version:  4.6 (Compatibility Profile) Mesa 21.3.8
OpenGL Core:     4.6 (Core Profile) Mesa 21.3.8
OpenGL ES:       OpenGL ES 3.2 Mesa 21.3.8
Vulkan:          Supported
```

### Media (optional)

_No response_

### Checklist:

- [X] I'm not asking for support with a game or the wine runner.
- [X] I have followed the above mentioned guides and have all the graphics and wine dependencies installed.
- [X] I have checked for existing issues that describe my problem prior to opening this one.
- [X] I understand that improperly formatted bug reports may be closed without explanation.

danieljohnson2:
On the LegacyDisplayManager, the resolution is ("" x "") if xrandr provides nothing it can use, which produces various amusing failure modes.

But a crash when building the replacements dict is ugly and hard for users to deal with, so add some validity checks for that. Those replacements are then removed.

This means that installers that use the RESOLUTION replacements will fail, as the replacement won't be done and the variable will be left in.

An alternative I see is to generate blank replacements as Lutris does in other places, but that also won't actually work and may be less obvious. Alternatively again, we could just crash with a nice error at some point- but I suspect most installers don't use these replacements and would therefore actually work if we let them try.

I've also tried to improve the handling of xrandr output. I've adjusted get_resolutions() to no longer require each resolution line to start with two spaces; instead it starts with whitespace, followed by digits, an x, and more digits, then anything else. This should be more resilient to formatting, though if a resolution line does not start with the resolution it won't work. Also, it de-duplicates the resolution list and sorts it. get_unique_resolutions() was not called, so I removed that.

Resolves #4218

### DIFF
--- a/lutris/installer/interpreter.py
+++ b/lutris/installer/interpreter.py
@@ -394,13 +394,19 @@ class ScriptInterpreter(GObject.Object, CommandsMixin):
             "USER": os.getenv("USER"),
             "INPUT": self.user_inputs[-1]["value"] if self.user_inputs else "",
             "VERSION": self.installer.version,
-            "RESOLUTION": "x".join(self.current_resolution),
-            "RESOLUTION_WIDTH": self.current_resolution[0],
-            "RESOLUTION_HEIGHT": self.current_resolution[1],
-            "RESOLUTION_WIDTH_HEX": hex(int(self.current_resolution[0])),
-            "RESOLUTION_HEIGHT_HEX": hex(int(self.current_resolution[1])),
             "WINEBIN": self.get_wine_path(),
         }
+
+        current_resolution = self.current_resolution
+        if current_resolution and len(current_resolution) == 2 and current_resolution[0] and current_resolution[1]:
+            replacements.update({
+                "RESOLUTION": "x".join(self.current_resolution),
+                "RESOLUTION_WIDTH": self.current_resolution[0],
+                "RESOLUTION_HEIGHT": self.current_resolution[1],
+                "RESOLUTION_WIDTH_HEX": hex(int(self.current_resolution[0])),
+                "RESOLUTION_HEIGHT_HEX": hex(int(self.current_resolution[1]))
+            })
+
         replacements.update(self.installer.variables)
         # Add 'INPUT_<id>' replacements for user inputs with an id
         for input_data in self.user_inputs:
--- a/lutris/util/graphics/xrandr.py
+++ b/lutris/util/graphics/xrandr.py
@@ -85,16 +85,10 @@ def get_resolutions():
     """Return the list of supported screen resolutions."""
     resolution_list = []
     for line in _get_vidmodes():
-        if line.startswith("  "):
-            resolution_match = re.match(r".*?(\d+x\d+).*", line)
-            if resolution_match:
-                resolution_list.append(resolution_match.groups()[0])
-    return resolution_list
-
-
-def get_unique_resolutions():
-    """Return available resolutions, without duplicates and ordered with highest resolution first"""
-    return sorted(set(get_resolutions()), key=lambda x: int(x.split("x")[0]), reverse=True)
+        resolution_match = re.match(r"^\s*(\d+x\d+).*", line)
+        if resolution_match:
+            resolution_list.append(resolution_match.groups()[0])
+    return sorted(set(resolution_list), key=lambda x: int(x.split("x")[0]), reverse=True)
 
 
 def change_resolution(resolution):


strycore:
The problem with those conditional replacement is for installer that do use them, having no replacement happen might result in corrupt values, some config file might end up with a resolution value of "x" which might make some game crash.
Another option would be to keep those substitutions in place at all time and have the `current_resolution` always default to something valid, even if xrandr or other methods of getting resolutions fail. Could be something like `(1280, 720)` that is widely supported.

danieljohnson2:
I suppose we could do that, but I think the chief thing is to get the xrandr-based resolution list to work reliably.

But I don't really know why it didn't. My theory that the output format is different for some users is hard to defend, really. But something must be different, surely.

strycore:
If we can't parse xrandr, we could just dump the whole output to a file in ~/.cache/lutris, print an error log telling users to send us that file and then default to some generic resolution (that could possibly be overridden in lutris.conf)

danieljohnson2:
I am more inclined to write the xrandr output to the log in that case; that's what users submit with bug reports.

How about:
1. Sort and dedup the resolutions from xrandr
2. If there are none write the xrandr output to the log and
3. Fall back on just one resolution ["1280x720"]?

strycore:
Writing to the log is an option but the point of writing it to a file is to keep the exact output, with whitespace and all. But that's also possible to reconstruct from the logs.

danieljohnson2:
Does the log not preserve whitespace? I assumed it would do that.

strycore:
It does, but you still have to reformat the output to be able to parse it

danieljohnson2:
OK, I've given this idea a try and I think it's actually not hard to extract the xrandr output from the debug log. It's not broken into a debug message per line; it's all one message. Check out my new PR #4226. I believe that it will at least make this issue more troubleshootable.

strycore:
Closing this in favor of #4226

Issue #4216, opened 2022-04-12
[issue] AttributeError 'NoneType' when calling dll_manager.version.lower()


gbschenkel:
### Bug description

Trying to install Star Citizen, when Lutris try to install DLLs it throw an error. `NoneType' object has no attribute 'lower'`.
I realize is similar to the #3787 and I was using 0.5.9.1, and I upgraded to 0.5.10, but it is not the same.

```bash
ERROR    2022-04-12 00:42:28,265 [jobs.target:36]:Error while completing task <bound method CommandsMixin.task of <interpreter.ScriptInterpreter object at 0x7f5ff2393880 (lutris+installer+interpreter+ScriptInterpreter at 0x2b6ed40)>>: <class 'AttributeError'> 'NoneType' object has no attribute 'lower'
  File "/usr/lib64/python3.9/site-packages/lutris/util/jobs.py", line 34, in target
    result = self.function(*args, **kwargs)
  File "/usr/lib64/python3.9/site-packages/lutris/installer/commands.py", line 436, in task
    command = task(**data)
  File "/usr/lib64/python3.9/site-packages/lutris/runners/commands/wine.py", line 346, in winetricks
    return wineexec(
  File "/usr/lib64/python3.9/site-packages/lutris/runners/commands/wine.py", line 299, in wineexec
    wine.prelaunch()
  File "/usr/lib64/python3.9/site-packages/lutris/runners/wine.py", line 799, in prelaunch
    self.setup_dlls(
  File "/usr/lib64/python3.9/site-packages/lutris/runners/wine.py", line 776, in setup_dlls
    if dll_manager.version.lower() != "manual":
ERROR    2022-04-12 00:42:28,266 [interpreter._iter_commands:302]:Last install command failed, show error
```

### How to Reproduce

1. Try to install Star Citizen
2. After clicking on Install the problem arise

### Expected behavior

Install the game.

### Log output

```shell
INFO     2022-04-12 00:54:54,131 [startup.init_lutris:173]:Starting Lutris 0.5.10
WARNING  2022-04-12 00:54:54,159 [libretro.get_libretro_cores:24]:No folder at /home/schenkel/.local/share/lutris/runners/retroarch/
DEBUG    2022-04-12 00:54:54,167 [xrandr._get_vidmodes:15]:Retrieving video modes from XrandR
[...]

PR #4225, opened 2022-04-16
Fix download timeout measurement

danieljohnson2:
Using .seconds can't return more than a days worth of seconds, so downloads can stop working if you don't get them after a day of inactivity. It'll trigger if you start Lutris in the right time window, but you can miss it and if you're missing components it can lead to crashes.

So, total_seconds() for the win!

Resolves the missing versions.json issue referred to in #4216. Just restart Lutris and it'll notice that it need to download updates, and that should fix things.

### DIFF
--- a/lutris/util/update_cache.py
+++ b/lutris/util/update_cache.py
@@ -42,4 +42,4 @@ def get_last_call(key):
     if not date:
         return 0
     delta = datetime.now() - date
-    return delta.seconds
+    return delta.total_seconds()


strycore:
Thanks for catching that! The timedelta object is being quite sneaky here...

danieljohnson2:
I've written that same bug in .NET. The System.TimeSpan struct does the exact same thing.

strycore:
Worst thing is that I'm pretty sure I've checked if the seconds property would reset when other properties increased but looks like I had only checked for hours and minutes

danieljohnson2:
The docs I'm reading seems to say that it does have hours or minutes as attributes, just as constructor-parameters. The attributes are days, seconds and microseconds, and microseconds and seconds are limited. It's kind of an oddly irregular design.

[Docs](https://docs.python.org/3/library/datetime.html)

Issue #4231, opened 2022-04-20
Winetricks does not open


imLEGR:
Winetricks does not open when you select it from the drop-down menu. But if you run lutris from the terminal using the lutris command, Winetricks opens when you select it from the dropdown menu.

Arch Linux
![image](https://user-images.githubusercontent.com/53097257/164336430-93c7eadb-1bdc-44c2-8fec-7a8182a0a9f2.png)

italoghost:
It seems to be related to the Lutris Runtime. I am experiencing this as well and disabling it [as stated on this older issue](https://github.com/lutris/lutris/issues/713), fixes the problem.

Being more specific, I am having this issue with zenity: 

```
 zenity: symbol lookup error: /usr/lib/libwebkit2gtk-4.0.so.37: undefined symbol: gst_codec_utils_caps_get_mime_codec
``

**Comment by Jeremi360** [comment]
This is from symbol is in libgstreamer try updating/installing it.

**Comment by Jeremi360** [comment]
@italoghost you can run it winetricks as standalone and it works:
```bash
$ WINEPREFIX="path/to/wine/prefix" winetricks
```

italoghost:
Yes! As I said disabling Lutris Runtime fixes the issue. Using it through the terminal with standalone version works as well.

Probably there is another error in Lutris Runtime. Maybe something related to gtk-4.0?

imLEGR:
> Да! Как я уже сказал, отключение Lutris Runtime решает проблему. Использование его через терминал с автономной версией также работает.
> 
> Вероятно, есть еще одна ошибка в Lutris Runtime. Может что-то связанное с gtk-4.0?

Thank you, your answer helped! But I would like it to work all at once and not have to switch back and forth every time.
![image](https://user-images.githubusercontent.com/53097257/164550608-24e5015f-8e63-4d80-bfd4-9e0ff7a060e2.png)

strycore:
> As I said disabling Lutris Runtime fixes the issue. 

No it does not. Having to disable the lutris runtime should never be treated as a solution.

imLEGR:
> No it does not. Having to disable the lutris runtime should never be treated as a solution.

Totally agree. The problem needs to be solved in lutris itself. By the way, it is strange that the Arch Linux repositories still have the old version 0.5.9.1.

italoghost:
> > As I said disabling Lutris Runtime fixes the issue.
> 
> No it does not. Having to disable the lutris runtime should never be treated as a solution.

[...]

PR #4261, opened 2022-05-07
Make System Winetricks System-ier.

danieljohnson2:
The Lutris runtime is interfering with winetricks on Pop!_OS 22.04, but switching to the system winetricks or turning the Lutris runtime off does not help. That's because it doesn't apply fully; the environment is still generated *with* the runtime's stuff in it.

This PR fixes two things. It makes turning off the runtime more effective in WINE by passing the runner around to wineexec(), so the configuration setting there will affect the environment generated better.

It also passes the disabled_runtime flag around in more places to get it to get_env(); this is used when system-winetricks is turned on. It disables the runtime for system-winetricks even if it is not disabled for everything, so that in this case the winetricks can be more system-y.

This is not a complete fix- it allows the existing configuration flags to work better and so help out. But plain default Lutris winetricks should work and still does not.

Still, this helps with #4231.

### DIFF
--- a/lutris/runners/commands/wine.py
+++ b/lutris/runners/commands/wine.py
@@ -6,7 +6,6 @@ import time
 
 from lutris import runtime, settings
 from lutris.command import MonitoredCommand
-from lutris.config import LutrisConfig
 from lutris.runners import import_runner
 from lutris.util import linux, system
 from lutris.util.log import logger
@@ -334,21 +333,21 @@ def winetricks(
     config=None,
     env=None,
     disable_runtime=False,
+    system_winetricks=False,
     runner=None
 ):
     """Execute winetricks."""
-    wine_config = config or LutrisConfig(runner_slug="wine")
     winetricks_path = os.path.join(settings.RUNTIME_DIR, "winetricks/winetricks")
-    if (wine_config.runner_config.get("system_winetricks") or not system.path_exists(winetricks_path)):
+    if system_winetricks or not system.path_exists(winetricks_path):
         winetricks_path = system.find_executable("winetricks")
-        disable_runtime=True
         if not winetricks_path:
             raise RuntimeError("No installation of winetricks found")
     if wine_path:
         winetricks_wine = wine_path
     else:
-        wine = import_runner("wine")
-        winetricks_wine = wine().get_executable()
+        if not runner:
+            runner = import_runner("wine")()
+        winetricks_wine = runner.get_executable()
     if arch not in ("win32", "win64"):
         arch = detect_arch(prefix, winetricks_wine)
     args = app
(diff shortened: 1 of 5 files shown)


lutris/runners/commands/wine.py
```diff
@@ -223,6 +226,7 @@ def wineexec(  # noqa: C901
         blocking (bool): if true, do not run the process in a thread
         config (LutrisConfig): LutrisConfig object for the process context
         watch (list): list of process names to monitor (even when in a ignore list)
+        runner (runner): the wine runner that carries teh configuration to use
```
TrojanerHD, line 229:
```suggestion
        runner (runner): the wine runner that carries the configuration to use
```
Small typo ;)

danieljohnson2: Thank you. Though I'm tempted to leave it in there to see what the localization guys make of it.

strycore:
I don't really like this because disabling the runtime should never be something we recommend. Instead, whatever issues exist with the runtime should be fixed.
Note that in the case of Zenity on 22.04, the problem isn't located in the runtime itself but in additional libraries bundled with wine build (which become part of the runtime, I guess)
Still going to merge this but will probably add a few warnings in the logs to let people know that this is not the recommended way.

danieljohnson2:
Thanks for merging it in spite of your concerns.

It seemed intuitive to me that system-winetricks shouldn't mix Lutris runtime components. I'm not recommending it anywhere, just making it work as I expected it should. Is it that we don't want to use system-winetricks, or that I'm wrong about mixing it with the Lutris-bundled stuff?

I'm not sure how to proceed with Zenity. It seems to me like winetricks can hardly run without Wine, and therefore those bundled components are needed. If there's a way to shield Zenity alone from them, I have no idea what it is.

The only alternative that occurs to me to shipping an old Zenity with Lutris. That seems undesirable also.

strycore:
That way, we could see the output of winetricks like we do in install scripts

danieljohnson2:
We could surely do a higher quality UI than winetricks, but winetricks provides a *lot* of options.

I think we'd wind up sending users off to the system winetricks for anything we did not cover. That would be most of winetricks.

mortenfc:
Silly question perhaps, but ...
Why not have everything Lutris needs to work included in its bundle like a docker container? 
Then the dependency versions could all be bumped together and tested together to verify it works.
That way you won't need to worry about bugs with future releases of each distro etc.

I guess this means to generally not use the system libraries unless absolutely necessary.

strycore:
That's basically what the Flatpak version is. I've considered making Lutris use the Flatpak environment only and not using the runtime in this context

strycore:
> But... perhaps we can know what winetricks to offer by examining the scripts on Lutris.net.

Yes, it's easy to get a detailed list of most used winetricks verbs

danieljohnson2:
On the other hand, apparently there are installers that use zenity (see #4285). Perhaps its worth getting zenity to work regardless of winetricksiness.

PR #4269, opened 2022-05-11
Enable gamescope usage on nvidia >= 515

joebonrichie:
With 515.43.04 and gamescope git, gamescope is working (albeit not greatly) as the nvidia driver now supports `VK_EXT_image_drm_format_modifier`. Therefore, allow usage of gamescope on nvidia if the major driver version is 515 or higher.

AFAIK 515.43.04 is the first release in the 515 series so it should be safe to check for major version only.


strycore:
This is very cool!

Issue #4294, opened 2022-05-21
"Install DLCs" for GOG does not show Linux installers when available


morsvictoria:
Hello,

When using the "Install DLCs" feature for games installed from GOG only the options for WINE are listed even when the developer has provided native Linux installers for the DLC.

An example of this is Pathfinder: Kingmaker which has ".sh" files for the Linux version of the game, but only WINE options are listed.

PR #4297, opened 2022-05-23
#4294 - Linux DLC installers not listed under "Install DLC"

morsvictoria:
Hello,

This is my attempt at resolving #4294.

Modifies get_dlc_installers() to return all installers available for db_game (except mac)
Adds function get_dlc_installers_owned() to return all installers for DLC owned by user
Adds function get_dlc_installers_runner() to return all installers for requested runner (only linux & wine for now)
Adds function get_games_owned() as get_library() did not seem to return owned DLC

This did allow me to successfully install Kingmaker Pathfinder Linux DLC, but it does need more testing.


strycore:
Thanks! I'll do more testing on my end.

PR #4335, opened 2022-06-20
Remove usage of some GTK3 deprecations

melix99:
I've specifically avoided removing a big one: `GtkStatusIcon`. Let me know if you want that to be removed too or not.


lutris/gui/config/runner_box.py
```diff
@@ -130,11 +129,15 @@ def on_remove_clicked(self, widget):
     def on_runner_installed(self, widget):
         """Called after the runnner is installed"""
         self.runner_label_box.set_sensitive(True)
-        self.action_alignment.get_children()[0].destroy()
-        self.action_alignment.add(self.get_action_button())
+        self.action_button.destroy()
+        self.action_button = self.get_action_button()
+        self.action_button.set_valign(Gtk.Align.CENTER)
+        self.pack_start(self.action_button, False, False, 0)
```
A6GibKm, line 135: pack_start is very much deprecated.

melix99: Actually it's not deprecated, but it's still removed in GTK4. So yeah, that has to be removed here.

strycore:
This patch removes widgets that we're currently using. Send a patch that does not change the behavior of the application.

PR #4360, opened 2022-07-02
Capitalize menu items consistently

danieljohnson2:
Some of Lutris's menu items aren't capitalized consistently. "Show Side Panel" versus "Make a donation" for instance.

This PR makes them all 'title cased'. That seems like the norm for menus in GNOME. So, "Add Games", "Lutris Forums" and "Make a Donation"

I updated the msgids in the .po files, but not any translations. I don't know how to capitalize crazy moon-speak!

### DIFF
--- a/po/de.po
+++ b/po/de.po
@@ -166,7 +166,7 @@ msgid "Show Side _Panel"
 msgstr "Zeige seitliches _Bedienfeld"
 
 #: share/lutris/ui/lutris-window.ui:500
-msgid "Add games"
+msgid "Add Games"
 msgstr "Spiele hinzufügen"
 
 #: share/lutris/ui/lutris-window.ui:514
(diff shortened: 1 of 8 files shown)


A6GibKm:
Please don't update translation files.

danieljohnson2:
Why not?

A6GibKm:
It is simply bad practice, you are robbing the translators the opportunity to translate it correctly.

danieljohnson2:
How do you figure?

I changed only the English versions (the msg_ids), and only their capitalizations. Why is that an inconvenience?

A6GibKm:
When someone wants to translate the app they generate a newer pot file, then they check what strings need newer translations based on the diff. If you do the diff for them they can't know the string changed and therefore there won't be a new translation.

danieljohnson2:
So, the idea is that if the capitalization of the English was off, the translations are also, and this diff will trigger a review?

A6GibKm:
Yes, but note that in this particular case since the difference of strings is so small, the older translations will still be displayed to users until it is properly translated (instead of displaying the english version since there is no translation for that msgid). Everyone wins.

danieljohnson2:
Ah, I was not too sure that the msg_ids were case-insensitive. I figured the Unix tradition was for exact string matches, so I ought to match exactly if I wanted this to work.

I will think about this.

strycore:
Do not  change i18n in code patches.

PR #4372, opened 2022-07-11
Add waiting cloudflare check

Dertosh:
Some part of lutris api return 404 on cloudflare check.
This fix wait and get only lutris api answer.

### DIFF
--- a/lutris/api.py
+++ b/lutris/api.py
@@ -6,6 +6,10 @@ import urllib.error
 import urllib.parse
 import urllib.request
 
+import socket
+from collections import OrderedDict
+
+
 import requests
 
 from lutris import settings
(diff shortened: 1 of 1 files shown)


lutris/api.py
```diff
@@ -76,8 +80,18 @@ def get_user_info():
...
+    host = settings.SITE_URL.split("//")[1]
+
+    answers = socket.getaddrinfo(host, 443)
+    (family, type, proto, canonname, (address, port)) = answers[0]
+    headers = OrderedDict({
+        'Host': host
+    })
+    s = requests.Session()
+    s.headers = headers
+    response = s.get(api_url, verify=False, headers=headers)
```
strycore, line 92: Don't think the verify attribute should be set to false.

PR #4390, opened 2022-07-24
Escape special characters when globbing in installers

MinmoTech:
This fixes an issue where paths don't get correctly globbed in the installer if they contain special characters like `[`

### DIFF
--- a/lutris/installer/commands.py
+++ b/lutris/installer/commands.py
@@ -180,7 +180,7 @@ class CommandsMixin:
         if os.path.exists(filespec):
             filenames = [filespec]
         else:
-            filenames = glob.glob(filespec)
+            filenames = glob.glob(glob.escape(filespec))
 
         if not filenames:
             raise ScriptingError(_("%s does not exist") % filespec)


danieljohnson2 (commented):
This does not look right to me- doesn't this mean that if this new code runs, we'll always raise ScriptingError (line 186)?

I think the glob.escape call means that 'glob.glob' can only match the exact filespec. But we just checked for that on line 180, so if we reach this code, it actually won't match and we'll get an empty list, and then an exception right after.

MinmoTech:
@danieljohnson2 You're very right about that, sorry for that.

Since the issue is just with brackets, do you think it's safe to wrap them as described in this [SO post](https://stackoverflow.com/questions/22055500/python-glob-issues-with-directory-with-in-name)?

Failing original:
```bash
$ ls /home/<user>/Games/
'[140425]test.rar'
'[140425]test.zip'
```

```python
import glob

path = "/home/<user>/Games/[140425]test.*"
print(glob.glob(path))
```
result:
```
[]
```

Updated:
```python
import glob

path = "/home/<user>/Games/[140425]test.*"
path = path.translate({ord('['):'[[]', ord(']'):'[]]'})
print(path)
print(glob.glob(path))
```
result:
```
/home/<user>/Games/[[]140425[]]test.*
['/home/<user>/Games/[140425]test.zip', '/home/<user>/Games/[140425]test.rar']
```

danieljohnson2:
That seems like it would work better for your intent.

But that will disable any use of character ranges, and I'm not comfortable with that. Surely this is a rare case- you need to have brackets that are littler, but wildcards like * or ?.

Would it not be better to update the installer .yaml to include the [[] and []] stuff when it is necessary to do so?

MinmoTech:
You are right, I was thinking more about people that write installers who might be confused why their installers don't find the files they specified.
But it's probably better to keep the capability of being able to use character ranges.

Thanks for the help and the insight!

Issue #1984, opened 2019-04-02
Display a thing when the game library is empty


tannisroot:
In 0.4.x we had a welcoming screen with some instructions after launch Lutris for the first time, but in 0.5.x that kinda of disappeared now the user is greeted with the whole lot of nothing.
I think it might be a good idea to bring that back and show a tip about, for example, how you need to click on the search button to start installing games.
As a reference, this is how Gamehub does it:
![image](https://user-images.githubusercontent.com/10602045/55419029-9bd14680-557c-11e9-8fdf-27c30f00def6.png)

AccountOneOff:
Something like Glade's interactive intro would be good.

strycore:
![Screenshot from 2022-02-21 16-08-14](https://user-images.githubusercontent.com/280336/155041661-1674b439-5d0c-4526-bb98-d051a5b660be.png)

There's now a splash.svg image file that can be used to display any info we want

strycore:
Thanks! For the font, you could go with Cantarell since it's the default for Gnome: https://wiki.gnome.org/Projects/CantarellFonts
Maybe the inclusion of Montserrat as well could be nice.

hazelnot:
![splash-dark](https://user-images.githubusercontent.com/10992973/171192086-8c93be38-398c-405d-86fe-c7d350daf844.svg)
![splash-light](https://user-images.githubusercontent.com/10992973/171192093-36ac8ae9-cd8e-441e-9426-9f0aeb3b1fe1.svg)

Sorry for the delay, I legitimately didn't see your reply. How are these for starters?

hazelnot:
That would be ideal, I just didn't know that was a possibility :sweat_smile:

strycore:
Everything is possible!

strycore:
Thanks, I've integrated those into the UI!
The main image is not centered like I'd like it to be but I'm not going to fight Gtk over this.

danieljohnson2:
I'll fight GTK. And win!

The thing that keeps GTK from centering the graphics is the way blank_overlay is set to xscale=.1 and yscale=.1. I do not know why this would set this way; this is in the lutris-window.ui file.

PR #4393 resets the scale factors to 1, and then sets the bits of the splash to appropriate settings.

I used set_center_widget, which appears to require GTK 3.12. Is that okay? A more complex layout could probably be made without it.

PR #4393, opened 2022-07-26
Center the main splash screen; place the side splash thing on the top-left

danieljohnson2:
The main issue was the the alignment object blank_overlay was scaled to .1x.1. I don't know why.

Also, we need to use set_center_widget to center something in an HBox. This, I think, requires GTK 3.12.

I believe this will help get #1984 into shape.


strycore:
Thanks for finding out the issue!

PR #4412, opened 2022-08-07
debian: suggest gamescope

stephanlachnit:
gamescope is not available in Debian (https://tracker.debian.org/pkg/gamescope) and will migrate to Ubuntu in their next release, so let's suggest to users to install it.


strycore:
This better not break anything. We provide support for a lot of distros that don't provide gamescope

stephanlachnit:
Just Suggests, does not break anything.

PR #4415, opened 2022-08-08
Autochange paths on imported games

MrMEEE:
Changed the import feature, so that all paths will automatically be changed to the new import path..


MrMEEE:
Just to clearify, this is to fix the issue where games exported by one user, won't work after importing them with another username..

e.g I'm using this feature to install games on my machine and then export them to my kids machines..

PR #4435, opened 2022-08-18
Add lutris.util.amazon to packages

mackal:


mackal:
Guess I should have checked PRs before, my bad.

PR #4439, opened 2022-08-21
prevent overwriting pre-existing game artwork

crimson-snow:


danieljohnson2 (commented):
I think perhaps this should use system.path_exists(f, exclude_empty=True) instead of path.exists.

I believe Lutris will sometimes leave a zero-byte file as a placeholder if a download fails, so you'd definitely want to replace those. system.path_exists() can check for this.

I'm not certain this actually applies to *these* files, but I suspect so.

crimson-snow:
Good catch. Thanks for the heads up.

PR #4457, opened 2022-08-28
Enable context menu for all games in library

theCalcaholic:
This PR enables the (minimal) context menu for games that aren't installed.

image

For this to work, the following changes were necessary:

- Use a fallback game id of `-1` in lutris/gui/views/base.py#L41 if the game isn't installed. This causes the game_actions to be initialized with an empty "dummy" `Game` object
- Adjust `get_displayed_entries` in `game_actions.py` to only enable the "install", "add installed game" and "view on lutris" entries for non-installed games
- Fix slugs containing underscores (see lutris/game_actions.py#L316). I found that some GOG games had slugs containing underscores instead of dashes. Not sure, if this is the best place to fix this, though - probably should be fixed wherever these slugs are generated.

## Going forward

This is a very simple patch to show the context menu (especially to be able to view games on Lutris before installing them). However, if the Game class was refactored to use the dependency injection and composition patterns, this would make it easy to enable more features for non-installed games (e.g. favoriting) and it would be less likely that things break.
I'm thinking of something like this:

```py
from database.services import ServiceGameCollection
import database.games as games_db

class Game(GObject.Object)
  def __init__(self, service_game=None, local_game=None):
    super().__init__()
    self.service_game = service_game  # contains metadata and all functionality for interacting with installers, lutris.net 
    self.local = local_game  # contains all functionality and configuration for installed games
    ....

game = Game(
  ServiceGameCollection.get_game("lutris", appid),
  games_db.get_game_for_service("lutris", appid))
```

### DIFF
--- a/lutris/game_actions.py
+++ b/lutris/game_actions.py
@@ -313,7 +313,7 @@ class GameActions:
 
     def on_view_game(self, _widget):
         """Callback to open a game on lutris.net"""
-        open_uri("https://lutris.net/games/%s" % self.game.slug)
+        open_uri("https://lutris.net/games/%s" % self.game.slug.replace("_", "-"))
 
     def on_remove_game(self, *_args):
         """Callback that present the uninstall dialog to the user"""
--- a/lutris/game_actions.py
+++ b/lutris/game_actions.py
@@ -111,7 +111,7 @@ class GameActions:
         """Return a dictionary of actions that should be shown for a game"""
         return {
             "add": not self.game.is_installed,
-            "duplicate": True,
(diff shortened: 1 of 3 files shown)


PalmTamino:
I applied this to my own Lutris instance and can confirm that it works like a charm!!

PalmTamino:
@theCalcaholic

theCalcaholic:
@PalmTamino Thanks for the feedback, I'll have a look (I probably just assumed, that uninstall would include remove). Allthough I'm not sure how much interest there is in the PR...

Issue #2970, opened 2020-06-28
can't use lutris due to issues with the ssl


Ds886:
**Describe the bug**

failing to authorize the ssl certificate when trying to connect or get launchers

**Expected behavior**

that It will be able to grab content from the site

**Current behavior**

What is the current behavior?
It is not able to connect to the site

**Lutris debugging output**

```
NFO     2020-06-28 14:15:09,096 [application.do_command_line:245]:Running Lutris 0.5.5
INFO     2020-06-28 14:15:09,096 [startup.check_driver:62]:Using Intel Open Source Technology Center
INFO     2020-06-28 14:15:09,096 [startup.check_driver:67]:Running Mesa driver 19.2.8 on Mesa DRI Intel(R) UHD Graphics 630 (Coffeelake 3x8 GT2)  (0x3e9b)
INFO     2020-06-28 14:15:09,096 [startup.check_driver:79]:GPU: 8086:3E9B 1028:0827 using i915 drivers
INFO     2020-06-28 14:15:09,149 [startup.check_vulkan:132]:Vulkan is supported
INFO     2020-06-28 14:15:09,150 [dxvk.get_dxvk_versions:22]:Updating DXVK versions
DEBUG    2020-06-28 14:15:09,281 [http.get:65]:GET https://lutris.net/api/runtime
ERROR    2020-06-28 14:15:09,314 [dxvk.get_dxvk_versions:31]:<urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1076)>
ERROR    2020-06-28 14:15:09,318 [runtime._iter_remote_runtimes:156]:Failed to get runtimes: Unable to connect to server https://lutris.net/api/runtime: <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1076)>
INFO     2020-06-28 14:15:14,648 [application.do_shutdown:491]:Shutting down Lutris
```
**System information**

Ubuntu MATE 20.04, Vostro 7580

**Screenshots**

strycore:
Your ISP (or someone else) is tampering with your internet.
Do one else has issues with the SSL certificate, it's standard Cloudflare stuff

Do you live in a country where human rights are not respected?

Ds886:
Sorry if the above message was a bit stiff and unwelcoming, had a bit of rough day and it might have reflected in the original request.

[...]

PR #4479, opened 2022-09-04
fix: SSL CERTIFICATE_VERIFY_FAILED issue

rayrapetyan:
I've bumped into the issue described in: https://github.com/lutris/lutris/issues/2970 on the latest Debian (bullseye).
Proposed PR fixes this issue without the need to create `export SSL_CERT_DIR` and updating local certs.


danieljohnson2:
What does this PR *do* exactly? Is this switching certificate files?

rayrapetyan:
Yes, this helps users with a messed-up local certs to use a pure-pythonic independent way to deal with certs.
Because certifi and its certificate list is distributed with the application, it is guaranteed to be present and certificate loading should just work on any system.

theCalcaholic:
@rayrapetyan Is this really desirable? This would break lutris in most coporate environments (which tend to use transparent proxies) if I understand you correctly.
It could be nice to have as a fallback mechanism, though

rayrapetyan:
@theCalcaholic, could you clarify how transparent proxies may affect local certs verification?

ntninja:
Having looked over this again, I believe the underlying issue here is that *lutris/util/http.py* uses the *urllib.request* module (and hence *http.client* and *ssl* from the stdlib in their default configuration), while everything else uses *requests* (and hence *urllib3*’s customized usage of these modules). What @rayrapetyan’s patch does is making the (indirect) *ssl* module usage in *lutris/util/http.py* more similar to that of the remaining code that uses *requests*/*urllib3* instead. It does this by forcing usage of the same certificate bundle also used by *urllib3*, by explicitly querying the path of said bundle from the *certifi* module, constructing an OpenSSL context object using this path and then using said context for all issued requests – in essence duplicating what *requests*/*urllib3* does for all the other code implicitly.

As such, I think it would be better to instead just changing the code in *lutris/util/http.py* to utilize the *requests* module, so that Lutris doesn’t have to deal with these kinds of low-level issues in the future…

@theCalcaholic: At least with system-level installs of Lutris it shouldn’t break anything, as Debian ships its own version of the `certifi` package that just points at */etc/ssl/certs/ca-certificates.crt*. (Distros doing this kind of thing is intended behaviour for the `certifi` package with the ca-certificate list shipped on PyPI being a fallback for environments where using the OS ca-certificates isn’t feasible.)

theCalcaholic:
@ntninja Ah,I see! That's good to know. Would it still break for the flatpak version?

ntninja:
I did some experimentation to find out how to answer your question and I’m afraid that the certificate bundle used in the Flatpak version by `ssl.create_default_context()` (and hence *urllib.request*) differs from the one returned by `certifi.where()`. In fact, it appears that the certificate list returned by `ssl.create_default_context().get_ca_certs()` in the Flatpak environment matches that of the function call outside that environment except for omitting any per-user certificates. The `certifi.where()` function (and hence *requests*/*urllib3*) on the other hand returns the list from PyPI in the Flatpak environment instead, since the Flatpak’d version isn’t patched to delegate to the system trust store. Both inside and outside the Flatpak environment per-user certificates are ignored by anything using `certifi.where()`…

So yes, this patch breaks organizationally installed system certificates for the Flatpak version of Lutris when accessing Lutris services. Any code using the `requests` package is similarly broken, but the breakage may be worked around, as far as I can tell, by setting the `SSL_CERT_FILE` environment variable – which isn’t true for this patch.

theCalcaholic:
I mean I don't know how much of an issue this actually is, if it is at least well documented.

I just wanted to make sure it's considered.

Issue #3555, opened 2021-05-29
[Feature Request] Tagging/Categorising.


Latrolage:
Feature request for tagging/categorizing apps/games in lutris manually (or automatically through lutris.net?).
It could help with sorting the view screen, for example, the user could make custom "views". Which would be like tabs on lutris which each show games with different set of tags (both manually added and tags a game has on lutris.net).

Something else could be folders in lutris. Grouping applications/games in a familiar folder/directory like structure.

strycore:
Implemented in https://github.com/lutris/lutris/pull/4482

Issue #3153, opened 2020-09-22
Feature Request: Add support for game categories


BeatLink:
It would be really cool if games in lutris could be grouped into user created categories like on steam

bcedu:
Hi! I was going to sumbit a new issue but I've seen there is alredy these issue asking for categories. 

Is there any plan to add categories to lutris?

I've also seen that there is a branch (https://github.com/lutris/lutris/pull/1452) that has implemented the categories but it is not integrated with master? Is there any reason?



Thanks!!

feltcat:
It seems that PR #3001, which added support for a "favourites" category and was merged into master, also added backend code to handle arbitrary categories. It seems that the only thing still needed would be UI for managing categories and choosing which category to filter by. Since this issue was added to the 0.5.10 milestone, I'd guess that it's planned to do that, but I can't see if any work has been done on it since then.

strycore:
Implemented in https://github.com/lutris/lutris/pull/4482

PR #4482, opened 2022-09-04
Library game categories

klti:
It's been bugging me for a while now that there is no way to organize larger game collections into categories or tags or something. This has been requested now and then over the years (for example #3153,  #3555 or #4594), and even attempted to implement in PR #1452. That PR lead to a branch "categories", that has no seen any activity since 2019.

This PR caries over some of the UI code from #1452, but uses the now existing category database tables created for favorites to store categories.

Screenshot_20220904_134117


ntninja:
Added it to my Lutris dev version and think its really useful. Some feedback so far:

  * There is no way to rename ~~or remove~~ category labels once they are added
    * Turns out removing them is as simple as removing all games previously added to them
    * (Being able to edit the labels using DB Browser does work, but doesn’t count…)
  * Going through our entire library and categorizing all visual novels I would have really liked some way to add more than one game to some category at once – the intuitive way for me would have been to crtl+click on all games I wish to add in the list and then right-click on the last one to open the category editor, but that of course is impossible since the current listview is single-select only
  * I also think it would be more natural to show the categories list in *Configure* → *Game Info* rather than using its own dialog
  * The *Categories* menu item is displayed above the *Add to favourites* one, but below the *Remove from favourites* one; in effect causing its position in the right-click menu to vary depending on whether the game in question is marked as a favourite or not
  * The edit dialog is some small its a pain to use anything but the alphabetically first three category items
  * It does not appear to be possible to assign categories to uninstalled games

duarm:
Extremely useful feature, basically a tag system. One request I have, not for this PR, but for an extension of this feature, is to automatically tag games based on queried data from igdb, things like first-person, adventure, coop , developer, publisher etc...
I would be nice to query games with specific features on my library

bcedu:
I've made a PR adding some features to this branch: https://github.com/klti/lutris/pull/1

bcedu (approved):

ntninja:
@bcedu: I’ve added a patch to make these and your changes compatible to with latest upstream Lutris: https://github.com/ntninja/lutris/commit/5ab94cb477adeb027596f93a3876a676d33ef0d4

Maybe you want to add that to your branch as well?

As for your actual change: It’s a good improvement over @klti’s work and makes the whole system quite a bit nicer to use in practice! I still think it would be better to have the per-game categories dialog be part of the general game *Configure* dialog and found myself plenty of times looking for them there in recent months – I know it’s not an easy change to pull off though.

Note that I’m not a representative of the Lutris project however – in fact I’m pretty sure the Lutris BDFL actively dislikes me –, so any opinion expressed here is entirely that of myself.

Also, if you want to try and add this feature to Lutris proper (only if you are motivated, not blaming you at all if not), it would probably be best to just open a new PR against this repository instead of relying on @klti (they will still be attributed in GIT history, so its fine).

Reenuip:
This feature is sorely needed. I've been using Lutris as an alternative to Bottles and having everything heaped into a "Games" category is pretty annoying. Also, for things like Skyrim, I have multiple mod tools/managers installed and it's a real mess. Any progress on this feature would be greatly appreciated.

klti:
Awesome work, sorry for the long inactivity. After some "fun" with git this PR should have the up-to-date rebased version of the change, including your changes, in it.

klti:
> It does not appear to be possible to assign categories to uninstalled games

This should work now.

strycore:
Thanks for working on that, this is really awesome!

strycore:
Yeah, it's really nice!

Screenshot from 2023-06-01 19-58-52


Screenshot from 2023-06-01 20-00-54

Once we get better integration with the website API, we can consider things like creating categories from game genres, developer, etc...

danieljohnson2:
And here I thought this was dead. But I suppose *that is not dead which can eternal lie, and with strange versions even death may die*.

I think this here is going to be headline feature for 0.5.14.

It's a bit buggy, but I'm working on it. For future reference: don't call signal handlers (on_game_updated) directly; raise the signal. There are likely to be other controls listening. In this case, the sidebar was not being updated in all cases because of this.

I see @bcedu derived the new dialogs from GameDialogCommon, which has a lot of seemingly irrelevant complexity. Does anyone know why? I'l inclined to refactor the dialogs to not do this, but it may be important for something.

If there's time, I might be able to get in the obvious 'rename category' and 'delete category don't make my uncheck all those damn checkboxes' features. And the context menu desperately needs to be organized and have separators or something. @strycore, any sense how much time I have to work on this?

klti:
> I see @bcedu derived the new dialogs from GameDialogCommon, which has a lot of seemingly irrelevant complexity. Does anyone know why? I'll inclined to refactor the dialogs to not do this, but it may be important for something.

I think this just came about because I borrowed a lot of of initial the frontend code from https://github.com/lutris/lutris/pull/1452 for the first dialog (categories of game), and I think the second one just got duplicated from that.

danieljohnson2:
I'll give them a base class of their own; they do not need very much.

strycore:
> any sense how much time I have to work on this?

Once I finish the EA App integration and reviewed all the open tickets, I'll start bugging you about it.

danieljohnson2:
I'm almost there. The main thing left is the context menu.

Issue #4301, opened 2022-05-27
Multiple accounts for launchers


TomJuri:
A neat feature would be if you could add e.g. multiple epic games launcher accounts, so if you had let's say Shadow of the Tomb Raider on account no.1 and NBA2K21 on account no.2 both would show up in the library. That would make having multiple libraries in the same launcher much easier. Don't know if that's even implementable, but if it is... Why not implement it?

WellBeSerious:
One workaround is to create another OS account, reinstall Lutris/dependencies, and log into the other Epic account on the other OS account.

strycore:
I have Wukong on another Epic account. I didn't have to do any shenanigans.

PR #4487, opened 2022-09-06
Add ability to add more than one account per service

ntninja:
Adds the hidden ability to associate more than one instance of some service accounts (individually whitelisted where it makes sense) and whitelists GoG for this which I could actually test and confirm working.

Caveats / Todos:

  * [ ] Currently all GOG accounts show up with the same label in the UI
     * It should probably include the account user name or email address? But should it do this always or only if there is more than one account of the given type?
  * [ ] There is currently no UI to add further accounts, instead one has to add lines of the form *gog\~\<some-extra-identifier\> = True* to the *[services]* section of the configuration file
     * For testing this is good-enough, I think, and how to actually integrate this capability nicely into the settings screen will need some serious consideration that could perhaps be left for a future date.
  * [ ] As mentioned, this is currently GOG-only since that is all I could test, but with the changes included here any of the other launcherless online-service should *just work* within the limitations outlined above unless I missed something somewhere

Related to #4301 (although it doesn’t do anything for Epic launcher) – maybe there is a better issue somewhere?

### DIFF
--- a/lutris/services/egs.py
+++ b/lutris/services/egs.py
@@ -31,7 +31,6 @@ BOX_ART_SIZE = (200, 267)
 
 
 class DieselGameMedia(ServiceMedia):
-    service = "egs"
     remote_size = (200, 267)
     file_pattern = "%s.jpg"
     file_format = "jpeg"
@@ -116,12 +115,11 @@ class DieselGameBoxLogo(DieselGameMedia):
 
 class EGSGame(ServiceGame):
     """Service game for Epic Games Store"""
-    service = "egs"
 
     @classmethod
-    def new_from_api(cls, egs_game):
+    def new_from_api(cls, service_id, egs_game):
         """Convert an EGS game to a service game"""
-        service_game = cls()
+        service_game = cls(service_id)
         service_game.appid = egs_game["appName"]
         service_game.slug = slugify(egs_game["title"])
         service_game.name = egs_game["title"]
@@ -312,7 +310,7 @@ class EpicGamesStoreService(OnlineService):
             raise AuthTokenExpired from ex
         egs_games = []
         for game in library:
-            egs_game = EGSGame.new_from_api(game)
+            egs_game = EGSGame.new_from_api(self.id, game)
             egs_game.save()
             egs_games.append(egs_game)
         return egs_games
--- a/lutris/services/egs.py
+++ b/lutris/services/egs.py
@@ -132,7 +132,7 @@ class EGSGame(ServiceGame):
 class EpicGamesStoreService(OnlineService):
     """Service class for Epic Games Store"""
 
-    id = "egs"
+    type = "egs"
     name = _("Epic Games Store")
     login_window_width = 500
     login_window_height = 850
@@ -170,8 +170,8 @@ class EpicGamesStoreService(OnlineService):
         'Chrome/84.0.4147.38 Safari/537.36'
     )
 
-    def __init__(self):
-        super().__init__()
+    def __init__(self, id):
+        super().__init__(id)
         self.session = requests.session()
         self.session.headers['User-Agent'] = self.user_agent
         if os.path.exists(self.token_path):
--- a/lutris/services/egs.py
+++ b/lutris/services/egs.py
@@ -149,9 +149,9 @@ class EpicGamesStoreService(OnlineService):
     }
     default_format = "game_banner_small"
     requires_login_page = True
-    cookies_path = os.path.join(settings.CACHE_DIR, ".egs.auth")
-    token_path = os.path.join(settings.CACHE_DIR, ".egs.token")
-    cache_path = os.path.join(settings.CACHE_DIR, "egs-library.json")
+    cache_path_tmpl = "{id}-library.json"
+    cookies_path_tmpl = ".{id}.auth"
+    token_path_tmpl = ".{id}.token"
     login_url = ("https://www.epicgames.com/id/login?redirectUrl="
                  "https%3A//www.epicgames.com/id/api/redirect%3F"
                  "clientId%3D34a02cf8f4414e29b15921876da36f9a%26responseType%3Dcode")
@@ -177,6 +177,10 @@ class EpicGamesStoreService(OnlineService):
                 self.session_data = json.loads(token_file.read())
         else:
             self.session_data = {}
+
+    @property
+    def token_path(self):
+        return os.path.join(settings.CACHE_DIR, self._format_props(self.token_path_tmpl))
 
     @property
     def http_basic_auth(self):
--- a/lutris/gui/application.py
+++ b/lutris/gui/application.py
@@ -52,7 +52,7 @@ from lutris.util.http import HTTPError, Request
 from lutris.util.log import logger
 from lutris.util.steam.appmanifest import AppManifest, get_appmanifests
 from lutris.util.steam.config import get_steamapps_paths
-from lutris.services import get_enabled_services
+from lutris.services import get_service
 from lutris.database.services import ServiceGameCollection
 
 from .lutriswindow import LutrisWindow
@@ -614,7 +614,7 @@ class Application(Gtk.Application):
         if service:
             service_game = ServiceGameCollection.get_game(service, appid)
             if service_game:
-                service = get_enabled_services()[service]()
+                service = get_service(service)
                 service.install(service_game)
                 return 0
 
@@ -698,8 +698,8 @@ class Application(Gtk.Application):
     @watch_errors(error_result=True)
     def on_game_install(self, game):
         """Request installation of a game"""
-        if game.service and game.service != "lutris":
-            service = get_enabled_services()[game.service]()
+        service = get_service(game.service) if game.service else None
+        if service and service.type != "lutris":
             db_game = ServiceGameCollection.get_game(service.id, game.appid)
             if not db_game:
                 logger.error("Can't find %s for %s", game.name, service.name)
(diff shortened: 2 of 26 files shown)


ntninja:
May I humbly request a review of this? Also, feedback on the mentioned caveats would be greatly appreciated…

lutris/services/egs.py
```diff
@@ -169,8 +167,8 @@ class EpicGamesStoreService(OnlineService):
         'Chrome/84.0.4147.38 Safari/537.36'
     )
 
-    def __init__(self):
-        super().__init__()
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
danieljohnson2, line 171: Strange to use *args, **kwargs here. The base class wants just 'id', the service id.

More readable to have the argument list explicit, I say.

danieljohnson2 (commented):
Overall this code is pretty good, but rather uncommented. The id/type/name thing could stand explanation somewhere.

Minor thing- you search search for 'service_name'. You missed a couple and they are actually ids, I think.

Big thing- you're going to hate this, but this PR is in my view both too big and too small.

This is a lot of change at once, and likely to introduce bugs. Yet it is not a complete feature-there's no UI to actually use this.

The 'right thing' here would be to introduce a series of refactoring PRs to prepare Lutris for a change like this. Like adding get_service() and using it everywhere, say. Or introducing the _tmpl mechanism. You'd want to find things that are improvements on their own, ideally, but you should be explaining what these changes are meant to support.

This takes time, but allows for more testing and makes the final PR for the actual feature more manageable. And if @strycore rejects your PRs with "I don't want this feature" you'll have wasted less time building all this.

I guess it's a bit late for that kind of feedback, but if @strycore rejects this as too big, you might try to resubmit in stages like that.

I won't merge this because it is clearly a feature, and @strycore should approve it. But if it were my call, I might accept it *if* you added a UI for managing accounts, and *only* right after a .1 release- way too risky to introduce this before a release.

ntninja:
@danieljohnson2: Thank you for your detailed feedback!

> Overall this code is pretty good, but rather uncommented. The id/type/name thing could stand explanation somewhere.

Agreed! I’ll look into that!

> Yes, you should be able to tell which account is which in the UI, of course.

I can easily add this. I only wanted to know if devs would prefer the account user name to always be visible, or only if more than one account of the given type has been added?

> Minor thing- you search search for 'service_name'. You missed a couple and they are actually ids, I think.

I’ll check all uses of `service_name` again. Thanks!

> Big thing- you're going to hate this, but this PR is in my view both too big and too small. […] The 'right thing' here would be to introduce a series of refactoring PRs to prepare Lutris for a change like this. Like adding get_service() and using it everywhere, say. […]

I always try doing that by making sure each commit is independently reviewable and merge (in order of course). Perhaps these are still too large for @strycore’s tastes each or perhaps you would simply prefer each of these being submitted as independent PRs instead?

> This is a lot of change at once, and likely to introduce bugs. Yet it is not a complete feature- there's no UI to actually use this.

I did not want to start working on an account management UI yet, since it would likely double the size of the patch while almost certainly not ending up the way maintainers want it to look like in any case (the best way to do this is highly subjective IMHO).

If there is serious interest at this point already, I can try and draft up a design document though.

> I won't merge this because it is clearly a feature, and @strycore should approve it. But if it were my call, I might accept it if you added a UI for managing accounts, and only right after a .1 release- way too risky to introduce this before a release.

Very understandable stance!

danieljohnson2:
I don't know how serious the interest is; I wouldn't use this feature myself, but you asked for review.

I think you should not trouble with a design document. Too waterfall. Put the simplest UI you can on it; perhaps something involving extra buttons in the sidebar, or context menus, or something. Just keep it simple.

You want to get this to 'basically usable'; it can evolve from there.

strycore:
I'm ok with this feature existing but I find this patch is over-reaching and modifying too much of the current architecture. It makes things quite harder to understand and I do not think the implementation of this feature require that much change on the side of service modules (it should mostly be a database thing).
The media loader should not be affected by this patch, all media from a service should be shared between accounts.

ntninja:
> I do not think the implementation of this feature require that much change on the side of service modules (it should mostly be a database thing).

I’m sorry I re-read this part of your comment several times and I’m still not sure what you mean with “a database thing”? Could you clarify please?

> The media loader should not be affected by this patch, all media from a service should be shared between accounts.

Thought about this for a bit and I think you are correct. – The service media is per service-type rather than per service-account, so splitting stuff into per-account caches like this really does not make much sense…

ntninja:
Oh, and do you agree with @danieljohnson2 that I should cook up some dummy multi-account management settings UI for this PR already?

And any preference on how and when to distinguish accounts in the sidebar:

> It should probably include the account user name or email address? But should it do this always or only if there is more than one account of the given type?

ntninja:
> > The media loader should not be affected by this patch, all media from a service should be shared between accounts.
> 
> Thought about this for a bit and I think you are correct. – The service media is per service-type rather than per service-account, so splitting stuff into per-account caches like this really does not make much sense… Will revert/drop this part then.

Actually, the those ServiceMedia instance still need to know the service ID, so that they can query the list of games from the database, so I kept that part in and only reverted the part where it would parameterize the cache directory path on said service ID (so images of same game from the same service but from different accounts would now be stored only once).

strycore:
Your patch makes zero changes to the lutris/database package which is precisely where the changes should be.
You're having issues undoing the ServiceMedia stuff because the service files are being modified when they should not be.

danieljohnson2:
I find that a bit unclear, @strycore. By "services files", do you mean the .py files in lutris/services?

If so, that would suggest to me that you'd need to have many 'service-account' objects, each referring to a service- probably a better design. But how does lutris/database come into it? A new table for these accounts maybe?

But that does not seem to fit with how Lutris stores service credentials now. I can't see how you do it without changing the service classes.

ntninja:
> If so, that would suggest to me that you'd need to have many 'service-account' objects, each referring to a service- probably a better design.

I do not believe adding an extra class for “service accounts” would in any way constitute a “better design” as all it would do is needlessly complicate the codebase: The current service classes already manage account-related data (but are only instantiated for a single account ever) and there is no cross-account but per-service state to maintain anywhere… So what’s the point of adding another class for the per-account state, when the current service classes will just wind up devoid of any state as a result?

> But that does not seem to fit with how Lutris stores service credentials now. I can't see how you do it without changing the service classes.

That is correct, hence why I change them. All changes are strictly to make service objects aware of the account they are supposed to represent. Even if you move the management of accounts to the database these changes will *still* be necessary.

> You're having issues undoing the ServiceMedia stuff because the service files are being modified when they should not be.

@strycore: I do not appreciate the smug tone, see above.

ntninja:
Well, any reply would be appreciated at some point. I’ll regularily keep this up-to-date until then, but I won’t rewrite everything or work on the remaining todos until it becomes clear what @strycore actually wants to happen…

danieljohnson2:
You can ignore my views on this PR if you like; it is @strycore you need to satisfy if you are going to get merged.

You might start by appreciating his smug tone.

In this PR here, he's being straight with you, and not insulting or offensive. I don't think he's entirely clear, but he's trying to help you. You can listen, or you can not get merged. Your call.

ntninja:
> You can ignore my views on this PR if you like; it is @strycore you need to satisfy if you are going to get merged.
>
> You might start by appreciating his smug tone.

Let’s let that slide for now then… :roll_eyes: 

Seriously though, I’m trying but it’s *a bit hard*:
You say:

> In this PR here, he's being straight with you, and not insulting or offensive. **I don't think he's entirely clear, but he's trying to help you**. […]

No, no, no. He’s **not clear at all**, not one bit! The last statement I have received was 22 days ago and sounded like:

> Your patch makes zero changes to the lutris/database package which is precisely where the changes should be.

… and it left you, judging by your comment, as confused as it left me. I have tried to explain why the changes are as they are, I have tried to receive further guidance on what it exactly is that @strycore *wants*, but I still don’t have the slightest idea frankly.

Even your idea of adding extra “service account” classes, which I believe to be a bad idea for reasons previously explained, was hardly more than a wild guess on your part unless there was some behind-the-scenes communication going on that I’m not seeing (in which case it would have been very forthcoming to publish the results of such conversion having taken place).

> You can listen, or you can not get merged. Your call.

I will listen if actually talked to… I will not blindly rewrite this thing several times until it just happens to agree with @strycore’s vision.

strycore:
A good start would be to revert the `service.id` to `service.type` change to it original behavior.
That's the change I've found the most off-putting. Once that's done, I will be able to give a better review of the rest.
I'm guessing that the change will make the PR a lot smaller as well, which is also pretty important.

Root-Core:
@ntninja Thank you for your effort, I can absolutely understand that it is frustrating to you. I would love this feature to become reality. @strycore's tone here is a bit harsh first, but if you read it twice it seems mostly informative to me. He just wasn't clear in stating what he wanted you to do, while being clear in stating his opinion on your MR. This obviously was some miscommunication on his side imho.

Are you still working on it or do you want to continue your effort at some point as @strycore clarified a bit on what his vision is? Even if the change from id to type sounds logical to some outsider like me, just from a naming convention standpoint (without any knowledge about the Lutris code base and python in general).

ntninja:
@Root-Core: Thank you for your kind message! I was planning on returning to this PR eventually, but in recent times have mostly been focused on other things in life. I’ll probably get to it eventually, but don’t hold your breath. If you, or anyone else, feels like taking over, by all means please do so!

Regarding the rename from type to id: It is already an independently reviewable commit (https://github.com/lutris/lutris/pull/4487/commits/7facce8c2689217a0061d3a340879481d1a63872) and that could indeed be submitted as a separate PR… Thing is, there is zero point in performing that change alone without the context of the subsequent commits – hence why I grouped these changes together in this PR in the first place. I guess everyone just looked at the whole list of changes and thought “that’s too much” without ever noticing that its made up of several smaller (and IMHO reasonably sized) changes/commits?

@all: Is there any point in doing what @Root-Core suggested and moving each of these commits into a separate PR?

PR #4498, opened 2022-09-11
fix: discord client code issue (Python 3.9: unsupported operand type(s) for |: 't…

rayrapetyan:
…ype' and 'NoneType')

Traceback (most recent call last):
  File "/ara/devel/sandbox/lutris/venv/bin/lutris", line 7, in <module>
    exec(compile(f.read(), __file__, 'exec'))
  File "/ara/devel/sandbox/lutris/bin/lutris", line 52, in <module>
    from lutris.gui.application import Application  # pylint: disable=no-name-in-module
  File "/ara/devel/sandbox/lutris/lutris/gui/application.py", line 39, in <module>
    from lutris.game import Game, export_game, import_game
  File "/ara/devel/sandbox/lutris/lutris/game.py", line 25, in <module>
    from lutris.util import audio, discord, extract, jobs, linux, strings, system, xdgshortcuts
  File "/ara/devel/sandbox/lutris/lutris/util/discord/__init__.py", line 3, in <module>
    from .rpc import client
  File "/ara/devel/sandbox/lutris/lutris/util/discord/rpc.py", line 12, in <module>
    from lutris.util.discord.client import DiscordRichPresenceClient
  File "/ara/devel/sandbox/lutris/lutris/util/discord/client.py", line 6, in <module>
    class DiscordRichPresenceClient(DiscordRichPresenceBase):
  File "/ara/devel/sandbox/lutris/lutris/util/discord/client.py", line 7, in DiscordRichPresenceClient
    rpc: Presence | None  # Presence Object
TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
python-BaseException

PR #4519, opened 2022-09-22
Use Python 3.6+ compatible typing annotation syntax in Discord file

ntninja:
Use Python 3.6+ compatible typing annotation syntax in “lutris/util/discord/client.py”, so that Lutris does not crash on startup on Python 3.9 and below if the `pypresence` package is importable

Like #4498, but actually retains the correct typing information.

### DIFF
--- a/lutris/util/discord/client.py
+++ b/lutris/util/discord/client.py
@@ -1,16 +1,18 @@
+import typing as ty
+
 from pypresence import Presence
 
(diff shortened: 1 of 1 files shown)


SalomonSmeke:
3.6 is deprecated and lutris checks for it in setup.py. If you are happy with 3.7+, maybe this will suit your needs while allowing devs to use future-happy syntax?

https://github.com/lutris/lutris/pull/4521/

PR #4521, opened 2022-09-22
[chore] Add python 3.7 compatiblity with modern 484 syntax.

SalomonSmeke:
I noticed a fair bit of issues regarding python typing compatibility, and some commits removing typing.

Typing is great! You all should be able to type to your heart's content without worrying about compatibility.

This PR does that.

should supersede: https://github.com/lutris/lutris/pull/4498 and https://github.com/lutris/lutris/pull/4519
and allow you to revert: https://github.com/lutris/lutris/commit/cfe45f806d4f3d6ebdf7999cfe04ca527430a535

Note: not python 3.6 compatible.

### DIFF
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -13,3 +13,4 @@ sections =
     FIRSTPARTY,
     LOCALFOLDER
 default_section=THIRDPARTY
+add_imports=from __future__ import annotations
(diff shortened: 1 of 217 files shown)


SalomonSmeke:
Per @strycore 

> I don't want backwards compatibility stuff for 3.7, type hints are tolerated at best but will be removed when they cause the slightest problem

me: My pr message there might not be too clear. 3.7-3.10 all require that import for new-style annotations.

> Then I want neither new style nor old style annotations

PR #4527, opened 2022-09-27
Avoid excessive LutrisWindow updates when deduplicating Steam games

wolfmanx:
While investigation a <span class="title-ref">segmentation fault during Steam game update</span>, I came across an excessive amount of window updates during deduplication.

In `lutris/lutris/gui/lutriswindow.py`
(`/usr/lib/python3/dist-packages/lutris/gui/lutriswindow.py`), in method
*LutrisWindow.update\_store*

``` py
def update_store(self, *_args, **_kwargs):
    self.game_store.store.clear()
```

add a line with debugging output

``` py
def update_store(self, *_args, **_kwargs):
    logger.debug("running update_store")
    self.game_store.store.clear()
```

The log for a Steam update run with the above modification shows, that *LutrisWindow.update\_store* is called 20 times in rapid succession during the interval of 5.151 seconds from 02:31:16,481 to 02:31:21,632, roughly every 257 ms.

``` text
DEBUG    2022-09-27 02:31:16,061 [steam.add_installed_games:170]:271 Steam games detected and installed
DEBUG    2022-09-27 02:31:16,061 [steam.add_installed_games:171]:Games found in: /home/ws/.local/share/Steam/steamapps
WARNING  2022-09-27 02:31:16,120 [steam.add_installed_games:180]:Steam game %s has no AppID
WARNING  2022-09-27 02:31:16,189 [steam.add_installed_games:180]:Steam game %s has no AppID
DEBUG    2022-09-27 02:31:16,265 [steam.add_installed_games:185]:0 Steam games removed
DEBUG    2022-09-27 02:31:16,372 [config.remove:213]:Removed config /home/ws/.config/lutris/games/steam-1017900-1664238673.yml
DEBUG    2022-09-27 02:31:16,481 [lutriswindow.update_store:443]:running update_store
DEBUG    2022-09-27 02:31:16,608 [config.remove:213]:Removed config /home/ws/.config/lutris/games/steam-813780-1664238675.yml
DEBUG    2022-09-27 02:31:16,790 [lutriswindow.update_store:443]:running update_store
DEBUG    2022-09-27 02:31:16,897 [config.remove:213]:Removed config /home/ws/.config/lutris/games/steam-346110-1664238675.yml
DEBUG    2022-09-27 02:31:17,043 [lutriswindow.update_store:443]:running update_store
DEBUG    2022-09-27 02:31:17,208 [config.remove:213]:Removed config /home/ws/.config/lutris/games/steam-25800-1664238673.yml
[...]

Issue #4512, opened 2022-09-17
Lutris crash when actualise steam library


PowaBanga:
hi everybody !
Lutris crash when i try to refresh my steam library : 
```
lutris

(lutris:85669): Gtk-WARNING **: 04:44:21.216: Failed to register client: GDBus.Error:org.gnome.SessionManager.AlreadyRegistered: Unable to register client
2022-09-17 04:44:21,219: Starting Lutris 0.5.11
2022-09-17 04:44:21,387: Using NVIDIA drivers 515.65.01 for x86_64
2022-09-17 04:44:21,387: GPU: NVIDIA GeForce MX150
2022-09-17 04:44:21,387: GPU: 10DE:1D10 103C:84BA (nvidia drivers)
2022-09-17 04:44:21,387: GPU: 8086:5917 103C:84BA (i915 drivers)
MESA-INTEL: warning: Performance support disabled, consider sysctl dev.i915.perf_stream_paranoid=0
2022-09-17 04:44:21,883: Startup complete
2022-09-17 04:44:30,745: Unable to get games from API: HTTP Error 405: Method Not Allowed
2022-09-17 04:44:31,034: Failed to download http://cdn.steamstatic.com/steam/apps/666220/library_600x900.jpg: HTTP Error 404: Not Found
2022-09-17 04:44:31,050: Failed to download http://cdn.steamstatic.com/steam/apps/290930/library_600x900.jpg: HTTP Error 404: Not Found
2022-09-17 04:44:31,051: Failed to download http://cdn.steamstatic.com/steam/apps/258970/library_600x900.jpg: HTTP Error 404: Not Found
2022-09-17 04:44:31,057: Failed to download http://cdn.steamstatic.com/steam/apps/321040/library_600x900.jpg: HTTP Error 404: Not Found
2022-09-17 04:44:31,182: Failed to download http://cdn.steamstatic.com/steam/apps/326960/library_600x900.jpg: HTTP Error 404: Not Found
2022-09-17 04:44:31,260: Failed to download http://cdn.steamstatic.com/steam/apps/238070/library_600x900.jpg: HTTP Error 404: Not Found
2022-09-17 04:44:31,273: Failed to download http://cdn.steamstatic.com/steam/apps/272860/library_600x900.jpg: HTTP Error 404: Not Found
2022-09-17 04:44:31,280: Failed to download http://cdn.steamstatic.com/steam/apps/365300/library_600x900.jpg: HTTP Error 404: Not Found
[...]

Issue #3959, opened 2021-12-30
Cannot update Steam library


aminought:
### Bug description

Cannot update steam library, get small window with text 'appid'.

### How to Reproduce

1. Go to Main page
2. Click on Reload near Steam
3. See window with text 'appid'

### Expected behavior

Steam library should be updated without issues.

### Log output

```shell
2021-12-30 05:30:36,873: Starting Lutris 0.5.9.1
2021-12-30 05:30:36,893: No cores found
2021-12-30 05:30:38,775: Startup complete
INFO     2021-12-30 05:30:38,778 [startup.check_driver:53]:Using NVIDIA drivers 495.44 for x86_64
INFO     2021-12-30 05:30:38,778 [startup.check_driver:57]:GPU: Unknown
INFO     2021-12-30 05:30:38,779 [startup.check_driver:73]:GPU: 10DE:2488 1043:8833 (nvidia drivers)
DEBUG    2021-12-30 05:30:39,007 [lutriswindow.update_store:451]:Showing 132 games
DEBUG    2021-12-30 05:30:41,218 [base.wipe_game_cache:113]:Deleting games from service-games for steam
DEBUG    2021-12-30 05:30:42,732 [http.get:81]:GET https://lutris.net/api/games/service/steam
ERROR    2021-12-30 05:30:44,477 [jobs.target:36]:Error while completing task <bound method BaseService.reload of <steam.SteamService object at 0x7ff134173bc0 (lutris+services+steam+SteamService at 0x559f805ec340)>>: <class 'KeyError'> 'appid'
  File "/usr/lib/python3.9/site-packages/lutris/util/jobs.py", line 34, in target
    result = self.function(*args, **kwargs)
  File "/usr/lib/python3.9/site-packages/lutris/services/base.py", line 85, in reload
    self.add_installed_games()
  File "/usr/lib/python3.9/site-packages/lutris/services/steam.py", line 169, in add_installed_games
    appid = steam_game.config.game_level["game"]["appid"]
INFO     2021-12-30 05:31:27,286 [application.do_shutdown:631]:Shutting down Lutris
```

### System Information

```shell
[System]
OS:              Manjaro Linux 21.2.0 Qonos
Arch:            x86_64
Kernel:          5.15.7-1-MANJARO
Desktop:         GNOME
Display Server:  x11

[CPU]
Vendor:          GenuineIntel
Model:           Intel(R) Core(TM) i7-8700K CPU @ 3.70GHz
Physical cores:  6
[...]

Issue #3731, opened 2021-10-12
New Steam Library/Account Stuff Causes All Installed Steam Games to Show up Twice


gardotd426:
After upgrading to 0.5.9 and re-syncing my Steam account using the left panel, every installed Steam game shows up in Lutris twice:

![Screenshot_20211011_203748](https://user-images.githubusercontent.com/54234607/136871389-16ad21a8-5e7d-42e2-9daa-714cd110380a.png)

Obviously just removing the old config entries from ~/.config/lutris/games/ doesn't do anything, and I can't remove ~/.local/share/lutris/pga.db because that would cause all non-Steam games' entries to be removed (I've learned this the hard way before). 

[lutris-issue-report.json.tar.gz](https://github.com/lutris/lutris/files/7326022/lutris-issue-report.json.tar.gz)

Distro: Arch Linux
Lutris version: 0.5.9

neilunger:
Also having this issue, where refresh adds another copy.

strycore:
Fixed in 0.5.9.1

Brixus3:
Anyone find a fix for this? Or at least how to safely remove the duplicates?

wolfmanx:
I can confirm, that this issue is not fixed. All installed Steam games are duplicated with version 0.5.9.1.

Since it's Christmas and still not fixed, here is a workaround to get rid of the duplicates from failed Steam updates.

There are 2 errors here.

1.  Some game entries in the database do not produce a valid *Game* object:

        ERROR    2021-12-30 10:36:47,614 [steam.add_installed_games:174]:error: missing appid for (game: {'system': {}, 'steam': {}, 'game': {}}) (db: {'id': 122, 'name': 'Europa Universalis III', 'slug': 'europa-universalis-iii', 'installer_slug': 'steam', 'parent_slug': '', 'platform': 'Linux', 'runner': 'steam', 'executable': '', 'directory': '', 'updated': '', 'lastplayed': 1607757128, 'installed': 1, 'installed_at': 1602995509, 'year': 2007, 'steamid': 25800, 'gogid': 0, 'configpath': 'europa-universalis-iii-1602995509', 'has_custom_banner': 0, 'has_custom_icon': 0, 'playtime': 0.0038888888888888888, 'humblestoreid': '', 'hidden': 0, 'service': 'steam', 'service_id': '25800'})
[...]

PR #4528, opened 2022-09-27
Prevent segmentation fault when updating Steam games

wolfmanx:
During a Steam update, GTK fails with memory corruption and segfault (339 Steam games).

This problem was already reported and analyzed in <https://github.com/lutris/lutris/issues/3731#issuecomment-1002958248> (2). It is also reported in <https://github.com/lutris/lutris/issues/3959> and <https://github.com/lutris/lutris/issues/4512>.

In my case this also happens sporadically with approximately 105 GOG games.

In order to be as helpful as possible, some logfiles from different versions are attached to show that the problem does not depend on Linux distribution or Python version.

| Lutris Version       | Python Version | Ubuntu Version |
|----------------------|----------------|----------------|
| lutris-master-0.5.11 | python3.10     | ubuntu-22.04   |
| lutris-master-0.5.11 | python3.6.9    | ubuntu-18.04   |
| lutris-ppa-0.5.10.1  | python3.6.9    | ubuntu-18.04   |

- [lutris-master-0.5.11-steam-update-python3.6.9-ubuntu-18.04.log](https://github.com/lutris/lutris/files/9651552/lutris-master-0.5.11-steam-update-python3.6.9-ubuntu-18.04.log)
- [lutris-ppa-0.5.10.1-steam-update-python3.6.9-ubuntu-18.04.log](https://github.com/lutris/lutris/files/9651553/lutris-ppa-0.5.10.1-steam-update-python3.6.9-ubuntu-18.04.log)
- [lutris-master-0.5.11-steam-update-python3.10-ubuntu-22.04.log](https://github.com/lutris/lutris/files/9651554/lutris-master-0.5.11-steam-update-python3.10-ubuntu-22.04.log)

This pull request is merely an illustration of the observed behavior and an extremely simple remedy. It is not intended to be merged into the master branch unmodified. However, together with <https://github.com/lutris/lutris/pull/4527>, advanced users can easily create a temporary fix until a solution becomes available.

The segmentation fault happens in *LutrisWindow.update\_store* during the call to

``` py
self.game_store.store.clear()
```

since the "clear" debug output appears in the log file, while "hide\_overlay" from the modified method *LutrisWindow.update\_store* is never shown

``` py
def update_store(self, *_args, **_kwargs):
    logger.debug("clear")
    self.game_store.store.clear()
    logger.debug("hide_overlay")
    self.hide_overlay()
    games = self.get_games_from_filters()
    logger.debug("Showing %d games", len(games))
    self.view.service = self.service.id if self.service else None
    GLib.idle_add(self.update_revealer)
    for game in games:
        self.game_store.add_game(game)
    if not games:
        self.show_empty_label()
    self.search_timer_id = None
    return False
```

Modifying *SteamService.reload* by adding

``` py
self.emit("service-games-loaded")
return
```

makes sure, that nothing actually happens except for a call to *LutrisWindow.update\_store*

``` py
def reload(self):
    """Refresh the service's games"""
    self.emit("service-games-load")
    self.emit("service-games-loaded")
    return
    try:
        self.wipe_game_cache()
        self.load()
        self.load_icons()
        self.add_installed_games()
    finally:
        self.emit("service-games-loaded")
```

This already triggers the segmentation fault, so it is not necessary to look for a Steam specific error.

Playing with the number of games actually displayed

``` py
_limit = 50
#_limit = 0
for _i, game in enumerate(games):
    if _limit and  _i >= _limit:
        break
```

in *LutrisWindow.update\_store*

``` py
def update_store(self, *_args, **_kwargs):
    logger.debug("clear")
    self.game_store.store.clear()
    logger.debug("hide_overlay")
    self.hide_overlay()
    games = self.get_games_from_filters()
    logger.debug("Showing %d games", len(games))
    self.view.service = self.service.id if self.service else None
    GLib.idle_add(self.update_revealer)
    _limit = 50
    #_limit = 0
    for _i, game in enumerate(games):
        if _limit and  _i >= _limit:
            break
        self.game_store.add_game(game)
    if not games:
        self.show_empty_label()
    self.search_timer_id = None
    return False
```

shows that the segmentation fault already happens with approximately 50 games displayed.

Eliminating the call to *self.game\_store.store.clear*

``` diff
-        self.game_store.store.clear()
+        count = self.game_store.store.iter_n_children()
+        if count > 0:
+            self.redraw_view()
```

prevents the segementation faults reliably only if <https://github.com/lutris/lutris/pull/4527> is also applied. Otherwise, the segementation fault is only delayed for about 10 successive updates.

### DIFF
--- a/lutris/gui/lutriswindow.py
+++ b/lutris/gui/lutriswindow.py
@@ -440,7 +440,10 @@ class LutrisWindow(Gtk.ApplicationWindow):  # pylint: disable=too-many-public-me
                 self.show_label(_("No games found"))
 
     def update_store(self, *_args, **_kwargs):
-        self.game_store.store.clear()
+        count = self.game_store.store.iter_n_children()
+        if count > 0:
+            self.redraw_view()
+            return False
         self.hide_overlay()
         games = self.get_games_from_filters()
         logger.debug("Showing %d games", len(games))


danieljohnson2:
@strycore is amenable to a release soon, so I was thinking I should close this item since, seemingly, it is "merely an illustration".

But if I follow it, this could be merged- it switches from updating the GameStore to replacing it entirely.

Should this be merged for the next release? Does it fix the crash? Are there side effects we should be trying to work around?

wolfmanx:
@danieljohnson2 As documented, this patch alone does not prevent the seg-faults **reiliably**, you also need  https://github.com/lutris/lutris/pull/4527 for that.

How is it, that you do not experience seg-faults without this patch?

For me, on ubuntu 18.04, 20.04 and 22.04:
- with approx. 340 Steam games there are seg-faults **without this patch** (deduplications are not even reached)
- with less than 50 Steam games and no deduplications, this patch alone prevents the seg-faults,
. with 340 Steam games and 19 deduplications, this patch delays the segfaults, which then happen after approx. 10 deduplications
- wth the additional patch  https://github.com/lutris/lutris/pull/4527 the full refresh for each deduplication is suppressed and I no longer have any seg-faults for 340 Steam games and 19 deduplications.

danieljohnson2:
I was getting crashes on Lutris games. I do not have Steam.

With this patch. it would seg-fault every time. Without, it would misbehave after repeated reloading and eventually it might crash.

I do not understand it.

wolfmanx:
w@danieljohnson2 I just tested the latest changes and I no longer get a segmentation fault without this patch. It is sufficient to use #4527, which see. this PR can be closed.

PR #4544, opened 2022-10-08
Pass the parent dialog to Game.launch()

danieljohnson2:
So, I don't really like having UI code in Game, but I find that Lutris deliberately pops GTK dialogs from the command line to ask the user questions.

I could try to build a complex framework to move the dialogs that do this out of Game, but if Lutris is meant to pop dialogs from the command line, its hard to justify doing this.

Instead, this PR accepts a parent window as an argument to Game.launch() which allows those dialogs to be placed correctly. It must be passed through to Game.configure_game() via a closure, which is a little tricky but seems like the least bad choice here.

I still mean to minimize the error and warning dialogs in Game, but this PR will allow the question dialogs to work better.

### DIFF
--- a/lutris/game.py
+++ b/lutris/game.py
@@ -316,7 +316,7 @@ class Game(GObject.Object):
         )
         self.emit("game-updated")
 
-    def is_launchable(self):
+    def is_launchable(self, parent=None):
         """Verify that the current game can be launched."""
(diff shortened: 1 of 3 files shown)


danieljohnson2:
On second thought, this isn't worth the trouble. Too half-assed.

PR #4550, opened 2022-10-10
PyGIWarning webconnect_dialog.py

derfehler:
PyGIWarning: WebKit2 was imported without specifying a version first. Use gi.require_version before import to ensure that the right version gets loaded.


danieljohnson2:
Thanks, but I fixed this error like seconds ago. It should be okay now.

PR #4551, opened 2022-10-10
Attack of the Header Bars!

danieljohnson2:
I love GNOME header bars. Don't judge me.

But if you do too, here's a PR to put a header bar on the game configuration dialog, which is just crying out for one.
image
This does not apply to the preferences dialog; the way it hides and shows the action area does not work with a header bar, so It'll have to use the action area unless we want to redesign it.

The "advanced" option now appears in the header bar too as a switch:
image
It's only visible when an options tab is current. The shorter label is not as clear, but having it appear only on the tabs it applies on will help, and there's a tool-tip.

The save button is blue, but enter won't activate it. 'Ctrl-s' will, however. I figure that is safer.

### DIFF
--- a/lutris/gui/config/add_game.py
+++ b/lutris/gui/config/add_game.py
@@ -26,5 +26,4 @@ class AddGameDialog(GameDialogCommon):
         self.build_tabs("game")
         self.build_action_area(self.on_save)
         self.name_entry.grab_focus()
-        self.connect("delete-event", self.on_cancel_clicked)
         self.show_all()
--- a/lutris/gui/config/common.py
+++ b/lutris/gui/config/common.py
@@ -45,6 +45,8 @@ class GameDialogCommon(ModelessDialog):
         self.slug_change_button = None
         self.runner_dropdown = None
         self.image_buttons = {}
+        self.option_page_indices = set()
+        self.advanced_switch = None
         self.game_box = None
         self.system_box = None
         self.runner_name = None
@@ -68,6 +70,8 @@ class GameDialogCommon(ModelessDialog):
     def build_notebook(self):
         self.notebook = Gtk.Notebook(visible=True)
         self.notebook.set_show_border(False)
+        self.notebook.connect("switch-page", lambda _n, _p, index:
+                              self.update_advanced_switch_visibilty(index))
         self.vbox.pack_start(self.notebook, True, True, 0)
 
     def build_tabs(self, config_level):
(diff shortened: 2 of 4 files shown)


danieljohnson2:
It's "view-reveal-symbolic" and it's for showing the text in password fields. So here it reveals... advanced settings.

I'm open to suggestions here!

A6GibKm:
`settings-symbolic` might work better, but ultimately simply showing the text as it was plus a GtkSwitch might be clearer.

danieljohnson2:
It's rather crowded there for the text as it was, but perhaps something shorter could work?

A6GibKm:
Shorter I can think is replacing the "cancel" button with the regular `x` button (if they are equivalent) and use "Show Advanced"

danieljohnson2:
I like that idea. I've put in the switch. "Show advanced options" actually does just barely fit, but I feel it looks too crowded, so I shortened it to "Advanced". Clearer than the icon, and the tooltip is there anyway.

I've rigged it to appear only when the "Runner options" or "System options" tab is shown, since it affects only them.

I think the "Advanced" label is still overly prominent here, but I think on balance it is better to have it.

PalmTamino:
Hmm, does not work for me... 

I tried in "/usr/lib/python3.10/site-packages" typing 
curl https://patch-diff.githubusercontent.com/raw/lutris/lutris/pull/4551.diff | sudo git apply

And as i tried editing the files that were changed manually, the opening doesn't work anymore and it crashes (once trying to "configure" a game)

@danieljohnson2

danieljohnson2:
I think you just applied the PR to your distro's Lutris, @PalmTamino. It needs to be applied to the master branch instead. You'll have to repair your distro's Lutris to get it back; if you don't know how, you can probably just uninstall and reinstall it.

The easy way to try this PR is to download the branch, [here](https://github.com/lutris/lutris/tree/dj/header_bar_love)

Don't overwrite anything. Just run the script 'bin/lutris' in the branch.

PalmTamino:
> I think you just applied the PR to your distro's Lutris, @PalmTamino. It needs to be applied to the master branch instead. You'll have to repair your distro's Lutris to get it back; if you don't know how, you can probably just uninstall and reinstall it.
> 
> The easy way to try this PR is to download the branch, [here](https://github.com/lutris/lutris/tree/dj/header_bar_love)
> 
> Don't overwrite anything. Just run the script 'bin/lutris' in the branch.

Nice, it works, but how can i integrate this in my fedora's lutris?

danieljohnson2:
You can't do that without the rest of the master branch.

I believe that a release is coming soon, so maybe @strycore will smile on my header-bars and you'll get it that way.

danieljohnson2:
Yaaaaay header bars!

PR #4554, opened 2022-10-10
Use Webkit 4.1 and remove mentions of libnotify

A6GibKm:
See https://github.com/flathub/net.lutris.Lutris/pull/295


A6GibKm:
Alternatively one could use
```
try:
    gi.require_version("WebKit2", "4.1")
except ValueError:
    gi.require_version("WebKit2", "4.0")
```

danieljohnson2:
May I ask what this is about? Is this just for consistency with flatpak?

A6GibKm:
Webkit 4.0 uses libsoup 2 which is deprecated and not a part of GNOME 43, which will be completely be removed by 44, 4.1 has the same api but uses soup3 under the hood.

danieljohnson2:
OK, but I tried the change to 4.1 and it fails for me with the error:
```
  File "/home/danj/Projects/lutris/./bin/lutris", line 52, in <module>
    from lutris.gui.application import Application  # pylint: disable=no-name-in-module
  File "/home/danj/Projects/lutris/lutris/gui/application.py", line 45, in <module>
    from lutris.gui.installerwindow import InstallerWindow, InstallationKind
  File "/home/danj/Projects/lutris/lutris/gui/installerwindow.py", line 17, in <module>
    from lutris.installer import InstallationKind, get_installers, interpreter
  File "/home/danj/Projects/lutris/lutris/installer/interpreter.py", line 15, in <module>
    from lutris.installer.installer import LutrisInstaller
  File "/home/danj/Projects/lutris/lutris/installer/installer.py", line 13, in <module>
    from lutris.services import SERVICES
  File "/home/danj/Projects/lutris/lutris/services/__init__.py", line 5, in <module>
    from lutris.services.amazon import AmazonService
  File "/home/danj/Projects/lutris/lutris/services/amazon.py", line 16, in <module>
    from lutris.services.base import OnlineService
  File "/home/danj/Projects/lutris/lutris/services/base.py", line 16, in <module>
    from lutris.gui.dialogs.webconnect_dialog import WebConnectDialog
  File "/home/danj/Projects/lutris/lutris/gui/dialogs/webconnect_dialog.py", line 6, in <module>
    gi.require_version("WebKit2", "4.1")
  File "/usr/lib/python3/dist-packages/gi/__init__.py", line 129, in require_version
    raise ValueError('Namespace %s not available for version %s' %
ValueError: Namespace WebKit2 not available for version 4.1
```

It looks like your branch is built in Lutris 0.5.8, which is ancient. Can you update your repo and rebase onto the latest?

A6GibKm:
Done, by the way your error comes from not having 4.1 in your system.

danieljohnson2:
I see. I need to install the package gir1.2-webkit2-4.1 to use this.

Seems like we need this by the GNOME 44 release. That's to be in March 2023.

We don't absolutely need this now, but I think we should have it sooner rather than later.

But I think we should go with the 'alternate' version with the try-except, so if your distro has not got 4.1 yet you are not screwed. Maybe we also need to change the dependencies to allow for 4.0-or-4.1 somehow? I'd be happier if we were broadening compatibility, not just moving to a later target.

I'll merge it if you make that change.

PR #4564, opened 2022-10-18
Some additions to the changelog

danieljohnson2:
I would not presume to just stick stuff in the change-log, but I think there are just a couple more things it ought to have in it!


danieljohnson2:
@strycore told me to change the changelog myself, so I'll merge my own PR.

PR #4569, opened 2022-10-21
Remove "file://" prefixes if they are added to path-selection fields

semjon00:
Imagine this situation:
You had just downloaded a Windows game and now you add it as a local game into Lutris. In the game options, you need to pick the right executable ("Executable" field). One option is to "Browse..." for it. Another option is to just drag-and-drop the executable you just downloaded into this field. However, then it will have a "file://" prefix before the actual path (this happens at least on KDE). If you "Save" everything it as is, the game will not launch. Let's remove this prefix - it would be more convenient and less confusing this way.

### DIFF
--- a/lutris/gui/config/boxes.py
+++ b/lutris/gui/config/boxes.py
@@ -2,6 +2,7 @@
 # Standard Library
 # pylint: disable=no-member,too-many-public-methods
 import os
+import urllib
 from gettext import gettext as _
 
 # Third Party Libraries
@@ -423,10 +424,15 @@ class ConfigBox(VBox):
         file_chooser.entry.connect("changed", self._on_chooser_file_set, option_name)
 
     def _on_chooser_file_set(self, entry, option):
-        """Action triggered on file select dialog 'file-set' signal."""
-        if not os.path.isabs(entry.get_text()):
-            entry.set_text(os.path.expanduser(entry.get_text()))
-        self.option_changed(entry.get_parent(), option, entry.get_text())
+        """Action triggered when the field's content changes."""
+        text = entry.get_text()
+        if text.startswith('file:///'):
+            text = urllib.parse.unquote(text[len('file://'):])
+        if not os.path.isabs(text):
+            text = os.path.expanduser(text)
+        if text != entry.get_text():
+            entry.set_text(text)
+        self.option_changed(entry.get_parent(), option, text)
 
     # Directory chooser
     def generate_directory_chooser(self, option, path=None):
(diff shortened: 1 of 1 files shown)


lutris/gui/config/boxes.py
```diff
@@ -423,10 +423,15 @@ def generate_file_chooser(self, option, path=None):
         file_chooser.entry.connect("changed", self._on_chooser_file_set, option_name)
 
     def _on_chooser_file_set(self, entry, option):
-        """Action triggered on file select dialog 'file-set' signal."""
-        if not os.path.isabs(entry.get_text()):
-            entry.set_text(os.path.expanduser(entry.get_text()))
-        self.option_changed(entry.get_parent(), option, entry.get_text())
+        """Action triggered when the field's content changes."""
+        text = entry.get_text()
+        if text.startswith('file:///'):
```
semjon00, line 428: Note that there are three slashes. I wanted to avoid a situation where user types "file://" (for whatever reason), only to guess why Lutris had wiped all this text. With one slash remaining after the wipe, it is less confusing.

lutris/gui/config/boxes.py
```diff
@@ -423,10 +423,15 @@ def generate_file_chooser(self, option, path=None):
         file_chooser.entry.connect("changed", self._on_chooser_file_set, option_name)
 
     def _on_chooser_file_set(self, entry, option):
-        """Action triggered on file select dialog 'file-set' signal."""
-        if not os.path.isabs(entry.get_text()):
-            entry.set_text(os.path.expanduser(entry.get_text()))
-        self.option_changed(entry.get_parent(), option, entry.get_text())
+        """Action triggered when the field's content changes."""
```
semjon00, line 426: I believe that previously it was misleading.

lutris/gui/config/boxes.py
```diff
@@ -446,8 +451,13 @@ def generate_directory_chooser(self, option, path=None):
         self.option_widget = directory_chooser
 
     def _on_chooser_dir_set(self, entry, option):
-        """Action triggered on file select dialog 'file-set' signal."""
-        self.option_changed(entry.get_parent(), option, entry.get_text())
+        """Action triggered when the field's content changes."""
+        text = entry.get_text()
+        if text.startswith('file:///'):
+            text = text[len('file://'):]
```
danieljohnson2, line 457: I don't buy bit; this does not handle URL escaping ,like file:///dir/file%20with%20spaces -> /dir/file with spaces

I could fine no simple way to handle decoding a file URL, but apparently you can do something with urllib.parse.unquote.

semjon00: Interesting, it works on my machine even without the unquoting. But yes, unquoting would not hurt.

I have updated my commit

danieljohnson2: That seems very strange, but I expect the new version will be safer.

semjon00:
Merge please?
:)

semjon00:
I think it is safe enough to merge before beta2/release?

danieljohnson2:
I think it'd be safe, but I just merge fixes. @strycore will have to decide if he wants this.

strycore:
This is ok to merge

semjon00:
Thank you 😊

PR #4580, opened 2022-10-27
Adds logic to find a copy of zenity stashed in the winetricks runtime package

### LINKED ISSUE #4231: Winetricks does not open

danieljohnson2:
If it is not there, this should have no effect. There's a separate PR in buildbot to stick zenity into the winetricks package. This is its client-side counterpart.

Resolves #4231


danieljohnson2:
lutris/buildbot#167 has been merged, but that won't do any good without this PR also, so I'm going to go ahead and merge it.

The that PR does not work out, this one should still be harmless.

Issue #4610, opened 2022-11-24
Dont default to nvidia if nvidia exists


ImUrX:
I'm currently using an AMD GPU and an Nvidia GPU, both for different purposes. [Lutris detects if nvidia exists in the system and just defaults to it](https://github.com/lutris/lutris/blob/bd6810287c1b72a4c7fb8c4b9460d437280749eb/lutris/startup.py#L54), can this not happen so I can use my AMD GPU happily? I even added the Vulkan env variable so Nvidia's vulkan driver isnt loaded.

Also because of this, lutris outputs incorrect information on hardware information. It says my graphics is AMD which if you read the lutris logs it obviously isn't because it's defaulting to Nvidia
```
[System]
OS:              Manjaro Linux 22.0.0 Sikaris
Arch:            x86_64
Kernel:          6.0.8-1-MANJARO
Desktop:         GNOME
Display Server:  x11

[CPU]
Vendor:          AuthenticAMD
Model:           AMD Ryzen 5 5600 6-Core Processor
Physical cores:  6
Logical cores:   12

[Memory]
RAM:             31.2 GB
Swap:            0.0 GB

[Graphics]
Vendor:          AMD
OpenGL Renderer: AMD Radeon RX 6700 XT (navi22, LLVM 14.0.6, DRM 3.48, 6.0.8-1-MANJARO)
OpenGL Version:  4.6 (Compatibility Profile) Mesa 22.2.3
OpenGL Core:     4.6 (Core Profile) Mesa 22.2.3
OpenGL ES:       OpenGL ES 3.2 Mesa 22.2.3
Vulkan:          Supported
```

ImUrX:
There is also [this piece of function](https://github.com/lutris/lutris/blob/master/lutris/sysoptions.py#L94) that ignores ``VK_ICD_FILENAMES`` env variable, so it defaults to nvidia also

danieljohnson2:
Can you select AMD in the Preferences dialog, "Global Options" panel?

That should affect all games, including new installs.

ImUrX:
You mean the ICD option hidden in advanced configuration? Yeah, I ended up using that. Kind of sad that it ignores the ICD env that i have tho

danieljohnson2:
OK, so I think the check_driver() function is harmless; I think that is about complaining if your NVIDIA drivers are out of date.

[...]

PR #4652, opened 2022-12-13
sysoptions: search XDG_DATA_HOME and XDG_DATA_DIRS for vulkan ICDs

LunNova:
* Search XDG_*_DIRS for vulkan ICDs
  * This is needed to match Vulkan-Loader's behavior, otherwise on some platforms ICDs are not found.
  * See https://github.com/KhronosGroup/Vulkan-Loader/blob/v1.3.235/docs/LoaderDriverInterface.md#driver-discovery-on-linux for details of the vulkan loader search paths.
* Use vulkaninfo when available to find device info and only use glxinfo as a fallback.
  * This ensures the device printed for a vulkan ICD is actually the device it'll use.
* Ensure default/empty choice does not set a vulkan ICD override instead of selecting the first one.

I wanted to make it clear which GPU is associated with a given ICD where possible.

This approach does make the picker a bit cluttered for multi-gpu setups. If prime has an impact for a particular ICD both the default and DRI_PRIME GPU are shown which can result in a very long string.

image

### DIFF
--- a/lutris/sysoptions.py
+++ b/lutris/sysoptions.py
@@ -1,6 +1,9 @@
 """Options list for system config."""
+import functools
 import glob
 import os
+import re
+import shutil
 import subprocess
 from collections import OrderedDict, defaultdict
 from gettext import gettext as _
@@ -8,15 +11,20 @@ from gettext import gettext as _
 from lutris import runners
 from lutris.util import linux, system
 from lutris.util.display import DISPLAY_MANAGER, SCREEN_SAVER_INHIBITOR, USE_DRI_PRIME, has_graphic_adapter_description
+from lutris.util.log import logger
 
-VULKAN_DATA_DIRS = [
-    "/usr/local/etc/vulkan",  # standard site-local location
-    "/usr/local/share/vulkan",  # standard site-local location
-    "/etc/vulkan",  # standard location
-    "/usr/share/vulkan",  # standard location
-    "/usr/lib/x86_64-linux-gnu/GL/vulkan",  # Flatpak GL extension
-    "/usr/lib/i386-linux-gnu/GL/vulkan",  # Flatpak GL32 extension
-    "/opt/amdgpu-pro/etc/vulkan"  # AMD GPU Pro - TkG
+
+# vulkan dirs used by distros or containers that aren't
+# from https://github.com/KhronosGroup/Vulkan-Loader/blob/v1.3.235/docs/LoaderDriverInterface.md#driver-discovery-on-linux
+# don't include the /vulkan suffix
+FALLBACK_VULKAN_DATA_DIRS = [
+    "/usr/local/etc",  # standard site-local location
+    "/usr/local/share",  # standard site-local location
+    "/etc",  # standard location
+    "/usr/share",  # standard location
+    "/usr/lib/x86_64-linux-gnu/GL",  # Flatpak GL extension
+    "/usr/lib/i386-linux-gnu/GL",  # Flatpak GL32 extension
+    "/opt/amdgpu-pro/etc"  # AMD GPU Pro - TkG
 ]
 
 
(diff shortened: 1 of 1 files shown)


LunNova:
Marked as draft because not sure what to do about very long vendor strings for multi-gpu without losing info in the output.

lutris/sysoptions.py
```diff
@@ -99,77 +133,45 @@ def get_vk_icd_choices():
...
     nvidia_files = ":".join(nvidia)
     amdvlk_files = ":".join(amdvlk)
     amdvlkpro_files = ":".join(amdvlkpro)
 
-    # Start the 'choices' with an 'auto' choice. But which one?
-    auto_intel_name = _("Auto: Intel Open Source (MESA: ANV)")
-    auto_amdradv_name = _("Auto: AMD RADV Open Source (MESA: RADV)")
-    auto_nvidia_name = _("Auto: Nvidia Proprietary")
-
-    vk_icd_filenames = os.getenv("VK_ICD_FILENAMES")
```
LunNova, line 131: This removes the VK_ICD_FILENAMES logic as the new logic is able to show what Auto will do if this is already set without any special casing - vulkaninfo will run with it set and show which GPU will be used.

It also allows non-auto options to override it even if it is set system-wide. If that's not desired I can add a check back and have all values default to "" if it's set.

In that case I'd also add `VK_DRIVER_FILES` and make `VK_ICD_FILENAMES` the fallback per the vulkan loader spec.

lutris/sysoptions.py
```diff
@@ -184,7 +186,9 @@ def get_vk_icd_choices():
             choices.append(("AMDVLK Open source", amdvlk_files))
     if amdvlkpro_files:
         choices.append(("AMDGPU-PRO Proprietary", amdvlkpro_files))
-    choices.append((_("Unspecified (Use System Default)"), ""))
```
LunNova, line 187: This is gone as it'll now show `Auto: (your default GPU)` so it actually tells you what the default is.

LunNova:
This approach showing the GPU next to the ICD probably breaks translations too :/

LunNova:
This is marked draft as it needs some help/input, but is ready for review. Github doesn't distinguish between "don't merge" and "don't review" draft states.

lutris/sysoptions.py
```diff
@@ -82,13 +91,38 @@ def get_optirun_choices():
...
+    with subprocess.Popen(infocmd, shell=True, stdout=subprocess.PIPE, stderr=subprocess.STDOUT) as infoget:
+        result = infoget.communicate()[0].decode("utf-8").strip()
+    if "Failed to detect any valid GPUs" in result or "ERROR: [Loader Message]" in result:
+        return "No GPUs"
+    return result
+
+
+
+def get_vulkan_gpu_cmd(env):
+    """Run vulkaninfo (or glxinfo as fallback) command to get vendor based on certain conditions"""
```
danieljohnson2, line 114: This is not how Lutris usually handles environments.

'env' would normally be a dict, and passed as the 'env' parameter of Popen, and os.environ.copy() to obtain the parent enviroment first.

If we must prefix some shell stuff - I don't really like it - then at least the parameter name should not be 'env'. It is asking for a dict to be passed as per convention.

lutris/sysoptions.py
danieljohnson2: If vulkaninfo is not available, then I think VK_ICD_FILENAMES will again be ignored, and issue #4610 will be back.

So, how common is it vulkaninfo?

lutris/sysoptions.py
```diff
@@ -344,7 +348,7 @@ def get_vk_icd_choices():
     {
         "option": "vk_icd",
         "type": "choice",
-        "default": get_vk_icd_choices()[0][1],
+        "default": "",
```
danieljohnson2, line 351: No default? That seems unfortunate.

lutris/sysoptions.py
```diff
@@ -99,77 +133,45 @@ def get_vk_icd_choices():
     amdvlk = []
     amdvlkpro = []
     icd_files = defaultdict(list)
+
+    vulkan_data_dirs = set()
+    for env_var in VULKAN_DATA_DIR_ENV_VARS:
```
danieljohnson2, line 138: This logic does not entirely agree with [your link](https://github.com/KhronosGroup/Vulkan-Loader/blob/v1.3.235/docs/LoaderDriverInterface.md#driver-discovery-on-linux). That document describes how it falls back to other things when those environment variables are not available.

I think your logic falls back even if they are, I think.

lutris/sysoptions.py
```diff
@@ -99,77 +133,45 @@ def get_vk_icd_choices():
...
-                nvidia.append(loader)
-            elif "amd" in loader:
-                if "pro" in loader:
-                    amdvlkpro.append(loader)
-                else:
-                    amdvlk.append(loader)
+        all_icd_files.update(glob.glob(path))
+
+    for loader in sorted(all_icd_files):
+        icd_key = os.path.basename(loader).split(".")[0]
```
danieljohnson2, line 152: I don't like this sorted(); I think the link described a search through the directories in a well defined order and that was *not* alphabetical by path. I think alphabetical by path is very surprising and not desirable.

I think you should be using ordered sets throughout.

danieljohnson2:
I think @strycore would prefer to look at a finished product, that is ready to merge.

I don't think is, though. I found several issues that I'd like to see improvement on before I'd take this merge.

lutris/sysoptions.py
LunNova: Defaulting to "" makes the default Auto which leaves the env var unset.

IMO this makes more sense than Lutris defaulting to setting this env var, which is the current behavior.

lutris/sysoptions.py
LunNova: That's a good point, I'll try to make the search order match that more exactly.

lutris/sysoptions.py
LunNova: That particular issue wouldn't recur as Auto now leaves the env var unset, instead of trying to set it to whatever the default already was.

However if vulkaninfo isn't available and the system has multiple GPUs the text in the list showing which GPU corresponds with which ICD may be wrong.

Would it make more sense to not show which specific GPU corresponds to which ICD if vulkaninfo isn't available?

lutris/sysoptions.py
LunNova: The order here only impacts the order of multiple different options for a single vendor showing up in the dropdown.

It has no impact on the default selection since we default to leaving the env var unset.

lutris/sysoptions.py
LunNova: I could revert this line change - get_vk_icd_choices()[0][1] -> ("Auto: the GPU name", "")[1] -> "" so it makes no difference other than being slightly faster.

Or maybe add a comment noting that this corresponds to Auto / not setting the env var to allow default device selection to happen.

lutris/sysoptions.py
LunNova: Hmm, not sure the ordering matters here: the order matters for the vulkan loader selecting a default device.
It doesn't matter here because we just want to show all available ICDs for the user to pick between.

What about adding the extra fallback options of $HOME/.config and /etc/xdg and /etc and /usr/share that are currently missing but not trying to do anything about the ordering?

lutris/sysoptions.py
LunNova: Oh right, it can impact order for multiple ones for a single vendor so I have to fix that.

Will get back to you with that.

lutris/sysoptions.py
LunNova: Got it, I'll sort this out.

lutris/sysoptions.py
danieljohnson2: Oh, I see. Perhaps a comment then, I missed that.

lutris/sysoptions.py
danieljohnson2: It would be nice to know what is being selected, if that's possible.

But if vulkaninfo is universal anyway it may not really matter.

lutris/sysoptions.py
LunNova: Should be fixed now.

lutris/sysoptions.py
LunNova: Swapped it to use the env parameter.

LunNova:
Tried to address feedback by following the spec much more precisely.

Shortened the text in the dropdown by removing the dev/chip name part from the vulkaninfo output.

image

lutris/sysoptions.py
```diff
@@ -82,13 +90,97 @@ def get_optirun_choices():
...
+
+def get_vk_icd_files():
+    """Returns available vulkan ICD files in the same search order as vulkan-loader"""
+    all_icd_search_paths = []
+
+    def add_icd_search_path(paths):
+        if paths:
+            # unixy env vars with multiple paths are : delimited
+            for path in paths.split(":"):
+                path = path + "/vulkan"
```
danieljohnson2, line 143: Perhaps os.path.join would be better here?

danieljohnson2 (approved):
I like this better. It's not the same behavior as before, and that may cause compatibility issues. But on the whole I think this looks like an improvement.

Still, @strycore will have to make the final call on this.

strycore:
I don't really have a way to test this so it may be better if @GloriousEggroll reviews this one as he's written most of the code for Vulkan loaders.

PR #4653, opened 2022-12-13
Modify GitHub workflow to push releases and prereleases to the right PPA

Amaylia:
* Updated Workflow release types to `released`, and `prereleased`.
* Added conditional logic to push `released` releases to `ppa:lutris-team/lutris` and `prereleased` releases to `ppa:lutris-team/lutris-staging`
* Split the workflow out into a reusable workflow.
* Call the release PPA workflow only on GitHub release publications.
* Call the staging PPA workflow on all GitHub release & prerelease publications.
* Clean up comments, and make them a bit more consistent.
* Add autoincrement logic to the PPA version number when the version we're building already exists on the target PPA.
* Build Lunar and Kinetic packages on Jammy since GitHub only has LTS Ubuntu runners.


strycore:
Thanks! Can you also update the OS matrix to include 22.10 and 23.04 and then I can merge it.

strycore:
Also please remove the condition for the staging repo, all releases should be pushed to it, even final releases

Amaylia:
Okay, let me test those to make sure they build.. I know I had dependency trouble trying to build 18.04 in the past.. I was also thinking about adding a little logic to the build script that searches for an OS-Codename-Version specific script and uses that instead of the generic process if found.. like 

```bash
if [[ -f ".github/scripts/build-ubuntu-${OS_VERSION}.sh" ]]; then
    exec ".github/scripts/build-ubuntu-${OS_VERSION}.sh"
else
    exec ".github/scripts/build-ubuntu-generic.sh"
fi
```

So then it would be easier to add overrides for any build quirks a specific OS version might need.. up to you of course, but just a thought that I had when I was looking it over.  I'll stick to just testing the above versions and once I can verify that they build, I'll update my PR.

Thank you for your help with this!

Amaylia:
I've got the builds running now, but there seems to be a delay with the runners picking up the jobs.. I probably should have only included the builds that needed to be verified. :face_exhaling:  Oh well, I'll let you know the results when they finish.
image

Amaylia:
Okay.. these jobs I think will never be picked up.. it seems like GitHub only has runners for the LTS versions of Ubuntu ( https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources ), so what I may end up having to do then is have those versions built on their closest matching version, which I think in this case would be on Jammy for both 22.10 and 23.04.  I'll think about it a bit and do some research, but those versions can't be added in this fashion at the moment, it would seem. :disappointed:

Amaylia:
@strycore: I'm thinking I might implement the change I kinda laid out in https://github.com/lutris/lutris/pull/4653#issuecomment-1348787627 and then make a custom build script for Ubuntu 22.04 that tells it to produce builds for 22.10 and 23.04 as well, and then those will all get picked up when `dput` runs at the end of that workflow.  I'll test it on my personal PPA, and see if I can verify that works and also test that the .deb's produced actually install and run on those versions of Ubuntu.

Does that sound reasonable?  What are your thoughts?

strycore:
> Does that sound reasonable? What are your thoughts?

Yes, totally! Seems like the only option as well without having access to 22.10 and 23.04 in Github actions.

Amaylia:
Thank you @strycore, I'm working on that now, I was a little busy the past few days, I'm sorry!

strycore:
No worry at all, we have until the next release to get this right and the next release hasn't even started being developed :sweat_smile:

Amaylia:
Oh, okay. 🤭

I've got the build process producing packages now, and they're uploaded to my test PPA successfully, I'm just waiting for them to publish now so that I can add them to my Kinetic and Lunar Ubuntu VMs to test that they install and run correctly.  Once I can verify that, I'll update my PR and squash the commits.

image

Amaylia:
Installs and runs successfully on Lunar (23.04)
image

And same for Kinetic (22.10)
image

I'll work on getting my PR updated now.

Amaylia:
@strycore sorry, I accidentally did some weird things to my branch, but I think everything is ready now. 🤐

Amaylia:
Updating some comments, and I found a little artifact left behind from testing that I removed as well.

Amaylia:
Hey @strycore, if you're busy, please ignore this, but I was just wondering if you've had time to review these changes?  I'm going to be pretty busy again in the near future so I was hoping to have this finished soon.  If you'd like me to make any changes, I can get to them today or if you want to rethink how we approach the problem, I'm open to that discussion as well.

Thanks, and sorry for the poke!

Amaylia:
I've noticed a couple potential problems with the build process recently that I wanted to fix before they became a problem:

1. If a build is pushed where the Lutris version doesn't change, which can happen if we're just updating package dependencies, the version on the PPA won't automatically receive the version, and so I've created logic that will check the destination PPA for an existing same version and bump the `ppa<version>` (e.g. `ppa1` becomes `ppa2`) during the build process if so.
2. `debclean` in the github-ppa Makefile directive sometimes causes staging builds to fail if the version being built contains a revision number (I probably need to dig deeper into how to properly build when a revision number is present) but the runner environment is completely deleted after each run, so the cleaning process seems unnecessary there, and so I've opted to just `git clean -df && git reset --hard` instead between runs to clean up rather than using `debclean`.

I think in another PR, I will try to figure out how to automate revision builds properly and then bring debclean back.

Amaylia:
Builds are still processing correctly, and are now autoincrementing the ppa version when applicable.
image

Builds showing up on Launchpad.net:
image

Amaylia:
@danieljohnson2 I was wondering if I could trouble you for a review?  Just if anything looks bad/off/weird to you.. I know strycore (don't want to ping him again) should still be the one to decide on a merge, but I'm just looking for a bit of feedback now.  Thank you!

.github/scripts/build-ubuntu.sh
```diff
@@ -0,0 +1,112 @@
...
+# scheme like 0.5.12-0ubuntu1ppa1~ubuntu22.04, we trim everything up to
+# and including 'ppa' and everything after and including '~' to get the
+# PPA version number.
+if [[ "${APT_VERSION//ppa*/}" = "${LUTRIS_DEBIAN_VERSION}" ]]; then
+    echo "PPA (${PPA_URI}) has matching package version: ${LUTRIS_DEBIAN_VERSION}${PPA_VERSION}"
+    APT_PPA_VERSION="${APT_VERSION//*ppa/}"
+    APT_PPA_VERSION="${APT_PPA_VERSION//~*/}"
+    (( APT_PPA_VERSION++ ))
+    PPA_VERSION="ppa${APT_PPA_VERSION}~ubuntu${OS_VERSION}"
+fi
```
danieljohnson2, line 97: Fancy!

I think this is the code to handle the mess with 0.5.12, where @strycore reverted a change and replaced the .deb just after release, to fix a changed dependency that broke Lutris for Debian.

I should comment that this sort of thing seems to be quite rare, and a more manual process might be adequate.

.github/scripts/install-ubuntu.sh
```diff
@@ -0,0 +1,27 @@
...
+if [[ -e "${CODEBASE_ROOT}/.github/scripts/install-ubuntu-${OS_VERSION}.sh" ]]; then
+    echo "::group::Installing $OS_CODENAME ($OS_VERSION) build dependencies"
+    # shellcheck disable=SC1090
+    source "${CODEBASE_ROOT}/.github/scripts/install-ubuntu-${OS_VERSION}.sh"
+else
+    echo "::group::Installing generic build dependencies"
+    # shellcheck source=./install-ubuntu-generic.sh
+    source "${CODEBASE_ROOT}/.github/scripts/install-ubuntu-generic.sh"
+fi
+echo "::endgroup::"
```
danieljohnson2, line 27: Does this have some special meaning? "::endgroup::" is an odd thing to echo, and there's no "::begingroup::" anywhere that I can see.

Might just be a debugging assist, but it looks like nothing else here.

.github/workflows/publish-lutris-ppa.yml
```diff
@@ -4,11 +4,70 @@ name: Publish Lutris PPA
...
 jobs:
 
+  # This should never be true while our release types are defined as
+  # released and prereleased, but just in case we don't want to run on
+  # drafts at all.
   release-ppa:
+    if: |
+      github.event.release.draft == false
+      && github.event.release.prerelease == false
+
```
danieljohnson2, line 18: I know it's not your fault, but turning YML into a programming language is *gross*.

.github/workflows/publish-lutris-ppa.yml
```diff
@@ -48,10 +99,14 @@ jobs:
         env:
           PPA_GPG_PRIVATE_KEY: ${{ secrets.PPA_GPG_PRIVATE_KEY }}
           PPA_GPG_PASSPHRASE: ${{ secrets.PPA_GPG_PASSPHRASE }}
+          # This will be passed through to the build-specific script for
+          # Jammy and causes it to produce additional builds for the
+          # Ubuntu codenames listed below.
+          SERIES: "kinetic lunar"
         run: |
```
danieljohnson2, line 106: I see you are making the staging-ppa match release-ppa. Okay.

But this is a maintenance burden; any change must be reflected in both sections, yes?

Better if these bits could be shared. If not, a comment describing what needs to be match would at least be *something*. It's not clear to me what must match between the two sections, and what must differ.

danieljohnson2 (commented):
Well, my bash is quite weak and I know nothing of GitHub workflows, but I can mouth off nonetheless.

My first impression is that this is very complete and very well commented.

It appears to me that this supports back to Ubuntu 20.04- is that as far back as we wish to go?

.github/scripts/build-ubuntu.sh
```diff
@@ -0,0 +1,112 @@
...
+#       Codename of an Ubuntu series.  I.e. bionic, jammy, etc.
+#   OS_VERSION
+#       Version of an Ubuntu series.  I.e. 18.04, 22.04, etc.
+#   LUTRIS_VERSION
+#       Version of Lutris as defined in `lutris/__init__.py`.
+#   PPA_GPG_KEY_ID
+#       GPG Key ID of a key that is already added to the GPG keystore.
+
+# Go three directories up to get the codebase root path.
+[[ -z $CODEBASE_ROOT ]] && CODEBASE_ROOT="$(dirname "$(dirname "$(dirname "$(readlink -f "$0")")")")"
```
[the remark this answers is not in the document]
Amaylia: I guess I was just trying to keep it short, but I'll use `if` instead.. that one is a bit messy. 🤐

.github/scripts/install-ubuntu.sh
Amaylia: It's a special output that gets picked up by the GitHub runner.  It encapsulates all of the output between a
```bash
echo "::group::My Grouped Output"
# do some things.
echo "::endgroup::"
```

And it looks like this in the runner output:
image

.github/workflows/publish-lutris-ppa.yml
Amaylia: 🤭 Yeah, lol

.github/scripts/install-ubuntu.sh
danieljohnson2: Now I know.

I think I've learned more from this review than you have, really. :upside_down_face:

.github/workflows/publish-lutris-ppa.yml
Amaylia: Yeeeeah... the reason I had to split them was because the versions could be different, and joining them together was causing problems.. I was lazy and really need to comment about it though.. I'll do that with my next commit.

.github/scripts/install-ubuntu.sh
Amaylia: I've only learned these things recently myself!  I picked up on it from looking at what others were doing in their workflows, and I was like "oooh, nice!  That's clever!" I'm definitely not the originator of the ideas. 😊

Amaylia:
> Well, my bash is quite weak and I know nothing of GitHub workflows, but I can mouth off nonetheless.
> 
> My first impression is that this is very complete and very well commented.

Thank you!  I can be obsessive about finishing things once I start, and so having a direction to go in is very helpful to me, so thank you again. 🙂

> It appears to me that this supports back to Ubuntu 20.04- is that as far back as we wish to go?

I'm not sure.. I think the furthest we could go back would be 18.04, and there is a runner for that, so I could add it to the `os.matrix` list in `publish-lutris-ppa.yml` workflow.  I do think there was some trouble building on 18.04, but I don't recall the details.  I can try it in a VM and report back.

danieljohnson2:
If Lutris doesn't work on 18.04, the PPA surely need not support it. But it's probably worth a check.

.github/workflows/publish-lutris-ppa.yml
Amaylia: I'm trying to rethink this.. and I think a way that I could combine these into a single section would be if I store my builds under separate directories for each PPA I'm producing builds for and then just push out to any builds under those directories...

So I could set environment variables like:
```yml
    env:
      PPA_URI: "ppa:lutris-team/lutris"
      PPA_STAGING_URI: "ppa:lutris-team/lutris-staging"
      IS_PRERELEASE: ${{ github.event.release.prerelease }}
```

And then, in the build-ubuntu.sh script I could run a the first time against PPA_URI and store the builds under `../releases/` and if  IS_PRERELEASE is true, I could run a second time against the PPA_STAGING_URI and store the builds under `../prereleases/` and then `dput` everything under `../releases/` to PPA_URI and everything under `../prereleases/` to PPA_STAGING_URI.

I think that would work.. what do you think?

Amaylia: The pros and cons of the alternate approach would be..

**Pros**
- cleaner github workflow, and one shared section.
- easier maintenance?

**Cons**
- adds a bit more complication to the build script.
- the build process would take a bit longer under a shared runner (building both release and prerelease on the same runner) than they would have processing in parallel as separate jobs on separate runners.

Amaylia: Actually.. I think I can use this to make it cleaner:

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#defaults

danieljohnson2: That sounds like a good idea. I do not think concurrency is an issue- surely Lutris is cheap to build in any case and we don't need to throw cores at it.

Amaylia: I can make these changes to the `publish-lutris-ppa.yml` without any changes to the build scripts:
```yml
# Inspired by
# https://github.com/yuezk/GlobalProtect-openconnect/blob/master/.github/workflows/build.yml
name: Publish Lutris PPA

on:
  # See below for details about the 'release' event.
  # https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release
  release:
    types: [released,prereleased]

# These options are shared between all jobs.
defaults:
  strategy:
    # GitHub only has runners for LTS versions of Ubuntu.
    # https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources
    matrix:
      os: [ ubuntu-22.04, ubuntu-20.04 ]
  runs-on: ${{ matrix.os }}
  env:
    PPA_RELEASE_URI: "ppa:lutris-team/lutris"
    PPA_PRERELEASE_URI: "ppa:lutris-team/lutris-staging"
    # This will be passed through to the build-specific script for
    # Jammy and causes it to produce additional builds for the
    # Ubuntu codenames listed below.
    SERIES: "kinetic lunar"

jobs:
  # Only build the release PPA when the event is not a prerelease.
  release-ppa:
    if: github.event.release.prerelease == false
    steps:
      - name: Checkout the repository
        uses: actions/checkout@v2

      - name: Install build dependencies
        run: |
          ./.github/scripts/install-ubuntu.sh

      # This builds and signs the debian package for each OS in matrix.os
      # This script also recalculates dependencies based on the current
      # OS and can potentially produce different package control files,
      # but these files are not kept.
      #
      # Signing packages uses the secrets referenced below and passes
      # them through the build-ubuntu.sh script and into the "github-ppa"
      # Makefile directive.
      #
      # It would probably be a good idea to have a unique GPG key just
      # for this process, but whatever GPG key is passed here _must_ be
      # registered on the PPA as an authorized key.
      - name: Build the Debian Release Package
        env:
          PPA_URI: ${{ env.PPA_RELEASE_URI }}
          PPA_GPG_PRIVATE_KEY: ${{ secrets.PPA_GPG_PRIVATE_KEY }}
          PPA_GPG_PASSPHRASE: ${{ secrets.PPA_GPG_PASSPHRASE }}
        run: |
          ./.github/scripts/build-ubuntu.sh

      # Pushes the build to the Release PPA.
      - name: Publish to Release PPA
        run: |
          dput ${{ env.PPA_RELEASE_URI }} ../*.changes

  # This is the same as the release-ppa job, but the build process
  # environment variable PPA_URI targets PPA_PRERELEASE_URI instead.
  staging-ppa:
    steps:
      - name: Checkout the repository
        uses: actions/checkout@v2

      - name: Install build dependencies
        run: |
          ./.github/scripts/install-ubuntu.sh

      # Override PPA_URI with PPA_STAGING_URI so that our versions
      # are correctly targeted for that PPA.
      # PPA_GPG_PRIVATE_KEY & PPA_GPG_PASSPHRASE are required for
      # signing the packages.
      - name: Build Debian Prerelease Package
        env:
          PPA_URI: ${{ env.PPA_PRERELEASE_URI }}
          PPA_GPG_PRIVATE_KEY: ${{ secrets.PPA_GPG_PRIVATE_KEY }}
          PPA_GPG_PASSPHRASE: ${{ secrets.PPA_GPG_PASSPHRASE }}
        run: |
          ./.github/scripts/build-ubuntu.sh

      # Pushes the build to the Staging PPA.
      - name: Publish to Staging PPA
        run: |
          dput ${{ env.PPA_PRERELEASE_URI }} ../*.changes
```

Or,  I can combine them into a single job.
```yml
# Inspired by
# https://github.com/yuezk/GlobalProtect-openconnect/blob/master/.github/workflows/build.yml
name: Publish Lutris PPA

on:
  # See below for details about the 'release' event.
  # https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release
  release:
    types: [released,prereleased]

# These options are shared between all jobs.
defaults:
  strategy:
    # GitHub only has runners for LTS versions of Ubuntu.
    # https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources
    matrix:
      os: [ ubuntu-22.04, ubuntu-20.04 ]
  runs-on: ${{ matrix.os }}
  env:
    PPA_RELEASE_URI: "ppa:lutris-team/lutris"
    PPA_PRERELEASE_URI: "ppa:lutris-team/lutris-staging"
    # This will be passed through to the build-specific script for
    # Jammy and causes it to produce additional builds for the
    # Ubuntu codenames listed below.
    SERIES: "kinetic lunar"

jobs:
  # Only build the release PPA when the event is not a prerelease.
  publish-ppa:
    steps:
      - name: Checkout the repository
        uses: actions/checkout@v2

      - name: Install build dependencies
        run: |
          ./.github/scripts/install-ubuntu.sh

      # This builds and signs the debian package for each OS in matrix.os
      # This script also recalculates dependencies based on the current
      # OS and can potentially produce different package control files,
      # but these files are not kept.
      #
      # Signing packages uses the secrets referenced below and passes
      # them through the build-ubuntu.sh script and into the "github-ppa"
      # Makefile directive.
      #
      # It would probably be a good idea to have a unique GPG key just
      # for this process, but whatever GPG key is passed here _must_ be
      # registered on the PPA as an authorized key.
      - name: Build the Debian Release Package
        if: github.event.release.prerelease == false
        env:
          PPA_URI: ${{ env.PPA_RELEASE_URI }}
          PPA_GPG_PRIVATE_KEY: ${{ secrets.PPA_GPG_PRIVATE_KEY }}
          PPA_GPG_PASSPHRASE: ${{ secrets.PPA_GPG_PASSPHRASE }}
        run: |
          ./.github/scripts/build-ubuntu.sh

      # Pushes the build to the Release PPA.
      - name: Publish to Release PPA
        if: github.event.release.prerelease == false
        run: |
          dput ${{ env.PPA_RELEASE_URI }} ../*.changes

      # Override PPA_URI with PPA_STAGING_URI so that our versions
      # are correctly targeted for that PPA.
      # PPA_GPG_PRIVATE_KEY & PPA_GPG_PASSPHRASE are required for
      # signing the packages.
      - name: Build Debian Prerelease Package
        env:
          PPA_URI: ${{ env.PPA_PRERELEASE_URI }}
          PPA_GPG_PRIVATE_KEY: ${{ secrets.PPA_GPG_PRIVATE_KEY }}
          PPA_GPG_PASSPHRASE: ${{ secrets.PPA_GPG_PASSPHRASE }}
        run: |
          ./.github/scripts/build-ubuntu.sh

      # Pushes the build to the Staging PPA.
      - name: Publish to Staging PPA
        run: |
          dput ${{ env.PPA_PRERELEASE_URI }} ../*.changes
```

.github/scripts/build-ubuntu-generic.sh
```diff
@@ -0,0 +1,53 @@
...
+#   LUTRIS_DEBIAN_VERSION
+#       The Debian-specific version which follows the guidelines
+#       specified here: https://help.launchpad.net/Packaging/PPA/BuildingASourcePackage#Versioning
+#       Ex. "0.5.12-0ubuntu1"
+#
+#   PPA_VERSION
+#       The PPA release version which follows the same guide as mentioned
+#       in the LUTRIS_DEBIAN_VERSION variable.
+#       Ex. "ppa1~ubuntu22.04" or "ppa3~ubuntu22.04"
+#
```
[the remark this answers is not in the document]
Amaylia: This should be resolved in https://github.com/lutris/lutris/pull/4653/commits/f3cfb5f1742083284bb578461f60df139921516f

.github/scripts/build-ubuntu.sh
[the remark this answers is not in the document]
Amaylia: Resolved in https://github.com/lutris/lutris/pull/4653/commits/4398d28cff673d7297e2f1c4102ad98ccbea06cc

.github/scripts/build-ubuntu.sh
Amaylia: Commit https://github.com/lutris/lutris/pull/4653/commits/c14d45f79006df85c076afb52928da629a9fbbd3 should prevent autoincrement logic when the PPA number is either missing or in an unexpected format, in which case `ppa1` will be used.

.github/workflows/publish-lutris-ppa.yml
Amaylia: I went with the first example from above in https://github.com/lutris/lutris/pull/4653/commits/d06a5422a70291b57d98811d74c952357c86857e, but please let me know if the latter is more preferable, and I'll change it to that instead.

Amaylia:
@danieljohnson2: So after a lot of testing on 18.04 (bionic) it seems like the biggest obstacle is that meson is only at version 0.45.1 and the meson.python module only exists at or after >=0.46.0.. I managed to use 20.04 (focal) to build the source package for bionic.. it's currently building on the PPA now, so I'll test that build on bionic and see if it works, and if so, I'll implement the same JAMMY_BUILDS like script for Focal to build packages for bionic.

danieljohnson2:
Well, @Amaylia, I wouldn't kill yourself over it. 18.04 isn't the hill to die on; more a nice-to-have.

If it makes a mess of the build process, it's probably better to give it a skip.

Amaylia:
Well.... the source package build of it fails as welll.. _but_ a binary-only build worked and installed..
image

I'll have to make a `FOCAL_BUILDS="bionic"` and then a `build-ubuntu-20.04` script similar to `build-ubuntu-22.04` and a `build-ubuntu-18.04` script that produces a binary only package, and _then_ it will work.

Amaylia:
@danieljohnson2: I've split out the JAMMY_BUILDS functionality into a separate script to reduce code duplication, and the build scripts for both Jammy and Focal now reference that script.  I also moved  the `debuild -S` option into a BUILDS_OPT environment variable that is only set when the BINARY_ONLY environment variable is not set, and the builds script for Bionic then sets that variable before calling the generic build script.  So code replication should be at a bare minimum.

Amaylia:
@danieljohnson2 & @strycore: After testing the binary-only build process for 18.04, it worked... _buuuut_ according to launchpad.net they do not accept binary-only package uploads. 😭

image
https://help.launchpad.net/Packaging/PPA

image

I'll revert to before the changes for 18.04.

Amaylia:
I think the one remaining issue now is just making the GitHub Workflow more maintainable by turning it into a [reusable workflow](https://docs.github.com/en/actions/using-workflows/reusing-workflows) and then calling it once for both the release and staging PPAs..

I think once I do that, I might squash the commits and make a new PR since this one has gotten a bit messy.. :zipper_mouth_face:

strycore:
Don't bother with 18.04. People who still run Bionic in 2023 have bigger issues to deal with than installing Lutris.
Tell me when it's good to merge, so far it looks pretty great!

Amaylia:
> Don't bother with 18.04. People who still run Bionic in 2023 have bigger issues to deal with than installing Lutris.

🤭

> Tell me when it's good to merge, so far it looks pretty great!

Thank you, @strycore. 🙂

I want to finish adjusting the GitHub workflow to address the point @danieljohnson2 made about maintainability and then do one final test to make sure everything still builds properly, and then I'll ask for a review again.

This will likely be a separate PR, but I also wanted to add handling for a revision bit in the version string.  From what I've seen so far, .deb packages handle that in a special way where you include the original version source package and a patch file that is applied to that, and then upload those altogether to the PPA, and then the PPA builds the resulting package from there, but the current version of the PR handles it as a native package and I think it should be a quilt package instead.  I still need to read more about it.

Amaylia:
Okay, everything is split out and working now.  The builds are completing successfully and the packages are accepted on my test PPA.

image
image

I'm going to squash the commits down now and then I could use another review @strycore or @danieljohnson2.  Thank you both for all of your time and patience throughout this!

strycore:
Thank you so much, that automation is awesome!

I still have to create a new PGP key and I can test this on our PPAs.

Amaylia:
I'm glad I could contribute _something_ back to the project after being a Lutris user for the last couple of years. 🙂

Thank you both, again!

eli-schwartz:
> Don't bother with 18.04. People who still run Bionic in 2023 have bigger issues to deal with than installing Lutris.

Probably a good idea. But, if you really wanted to anyway, you should be able to backport Meson itself into your PPA and use that. Meson 0.61 installs fine on Ubuntu Bionic (Meson has no dependencies other than python 3.6, bumped to 3.7 in Meson 0.62).

PR #4676, opened 2022-12-29
Remove ResidualVM, as it was merged into ScummVM

RobLoach:
We no longer need ResidualVM as it was [merged into ScummVM](https://www.scummvm.org/news/20201009/).


strycore:
Thanks!

PR #4706, opened 2023-01-08
A Back Button for AddGamesWindow

### LINKED ISSUE #4708: How do I create a win32 prefix for my game?
I'm trying to install a game using "Install a Windows game from media". After I press "Install" button in the dialog, I see this in the console:

```
win64 Prefix created in /path/to/a/prefix
```

But I want my game to be installed in a win32 prefix. I was never given an option to select this for my game. I don't see such an option in the Wine runner options too. So, how do I do this?

danieljohnson2:
Or, "InstallerWindow is not enough!"

This PR is a facelift for the AddGamesWindow using the NavigationStack from InstallerWindow. It now looks like this:
imageThere's a back button, a cancel button, a border around the list, and now the one choice that directly opens another window gets a distinct icon. All others now go to a page, with further explanation on it.

The search page looks like this:
image
More buttons, and explanatory text. I think it worth saying that the installer window is coming.

The 'install from setup' page looks like this:
imageThis now features controls to specify the identifier, and to opt into a 32-bit Wine prefix (only on 64-bit systems). These settings are problematic to change after installation- you can't change what the Wine prefix is without rebuilding it, and the default directory name is the identifier. Resolves #4708. Settings you can conveniently alter after installation not included.

Also, the 'Continue' button has turned blue, says 'Install' and moved down to the lower right corner. And there's text to explain what this does.

When you click 'Scan folder', you get a new page:
Screenshot from 2023-01-13 19-23-44
You can key in a path, or click browse for the file selector. I think the explanatory text is badly needed here, as I found it hard to understand this. Lets hope I figured it out correctly!

The 'imported games' page looks like this:
image
Mostly cosmetic changes, but there's a 'Close' button now. This is to suggest that you're done now, though you can still go back too.

The 'Install from Script' option gets this page:
image
This tries to explain what a Lutris script is and where you could get one, and it explains that this is not a headless install; it just runs through the installer window.

And the dialog getrs a new "Import a ROM" option; here, which leads here:
image
This then leads to the dialog you can also get via drag & drop; but having this here adds a bit of discoverability.

### DIFF
--- a/lutris/gui/addgameswindow.py
+++ b/lutris/gui/addgameswindow.py
@@ -72,6 +72,9 @@ class AddGamesWindow(BaseApplicationWindow):  # pylint: disable=too-many-public-
         self.title_label = Gtk.Label(visible=True)
         self.vbox.pack_start(self.title_label, False, False, 0)
 
+        self.accelerators = Gtk.AccelGroup()
+        self.add_accel_group(self.accelerators)
+
         back_button = Gtk.Button(_("Back"), sensitive=False)
         back_button.connect("clicked", self.on_back_clicked)
         self.action_buttons.pack_start(back_button, False, False, 0)
@@ -79,6 +82,12 @@ class AddGamesWindow(BaseApplicationWindow):  # pylint: disable=too-many-public-
         self.continue_button = Gtk.Button(_("_Continue"), no_show_all=True, use_underline=True)
         self.action_buttons.pack_end(self.continue_button, False, False, 0)
         self.continue_handler = None
+
+        self.cancel_button = Gtk.Button(_("_Cancel"), use_underline=True)
+        self.cancel_button.connect("clicked", self.on_cancel_clicked)
+        key, mod = Gtk.accelerator_parse("Escape")
+        self.cancel_button.add_accelerator("clicked", self.accelerators, key, mod, Gtk.AccelFlags.VISIBLE)
+        self.action_buttons.pack_end(self.cancel_button, False, False, 0)
 
         self.stack = NavigationStack(back_button)
         self.vbox.pack_start(self.stack, True, True, 0)
@@ -112,6 +121,9 @@ class AddGamesWindow(BaseApplicationWindow):  # pylint: disable=too-many-public-
     def on_back_clicked(self, _widget):
         self.stack.navigate_back()
 
+    def on_cancel_clicked(self, _widget):
+        self.destroy()
+
     def on_watched_error(self, error):
         ErrorDialog(str(error), parent=self)
 
@@ -136,7 +148,7 @@ class AddGamesWindow(BaseApplicationWindow):  # pylint: disable=too-many-public-
     def present_inital_page(self):
         self.title_label.set_markup(f"<b>{self.title_text}</b>")
         self.stack.present_page("initial")
-        self.display_no_continue_button()
+        self.display_cancel_button()
 
     @watch_errors()
     def on_row_activated(self, listbox, row):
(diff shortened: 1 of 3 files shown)


hitesh2027 (approved):

strycore:
That's really nice to have!
Some few notes before I can merge this:

I wouldn't put "not recommended" next to the 32bit prefix checkbox. I'm considering having a drop-down with an OS selection with:
- Windows 10 - 64 bit (Default)
- Windows 7 - 64bit
- Windows 7 - 32bit
- Windows XP - 32bit
- Windows 98 - 32bit

For a lot of games requiring a 32bit prefix, it's also required to change the Windows version.


Installer scripts are mostly written in YAML by the users, not JSON as it is indicated in the dialog. JSON is only the transport format used by the API.

danieljohnson2:
Sure thing, @strycore, I've tweaked the text here and there as suggested: no more "(not recommended)" and "YAML" for "JSON". I've updated the screenshots above, and added screenies for the "Import a ROM" option.

The text wants further revisions, I think. I'm equivocating on how much credit to give TOSEC for the ROM thing; I'm not that solid on how the new Import a ROM feature really works.

I've not added a menu of Windows versions though- I do not see how to do it. There does not seem to be an installer option for this I could use (like "arch" for the 32-bit thing), and no game configuration option for it. I could build of this, I expect, but I don't think it belongs in this PR, which is already fairly large. ~~One~~five things at a time, yes?

strycore:
That's cool, the drop-drop thing can be added later. Basically, what it would do is set the prefix to the wanted arch and introduce a winetricks task to set the correct version in the automated installer.

For TOSEC, we do have data that comes from other databases such as No-Intro and Redump. I'll reword the message if need be.

danieljohnson2:
Feel free to correct my ramblings. I feel explanation is wanted here, but I'm explaining what I don't fully understand. Story of my life really.

I had figured I could do the windows-version setting the same way I did the DPI- by setting a registry value inside the prefix, with no winetricks dependency.

In any case, thanks for the mergitude @strycore!

realnc:
This has disappeared apparently? In 0.5.13, I don't get 32-bit choices anymore. Both Windows 10 and Windows 7 in the drop-down list are only 64-bit. When I try to install a game and then change to 32-bit later, it doesn't launch and I get a Wine error telling me that the prefix is 64-bit.

strycore:
Changing the architecture of an existing prefix isn't possible so you'll want to have the arch set to win32 from the start. I wonder why you're not seeing Win XP and Win 98 in the list. Have you tried with the latest code on git?

realnc:
> Changing the architecture of an existing prefix isn't possible so you'll want to have the arch set to win32 from the start. I wonder why you're not seeing Win XP and Win 98 in the list. Have you tried with the latest code on git?

XP 32-bit is there. But not 7 or 10.

danieljohnson2:
Well, I think the idea was that Windows 7 32 bit was very rare, and not a likely choice. It's strange to need that combination.

You can change the version after installing, though. Install using the XP 32-bit setting, and then afterwards use the "Wine configuration" command- in the Wine menu in the action bar at the bottom of the main Lutris window.

With that command you can change the Windows version, and get your Win7 32-bit prefix.

realnc:
Ah, OK. Thank you.

Just in case you're wondering, using 32-bit W10 is useful for very old 32-bit games (9x/XP) that happened to get a DX11 renderer by modding. Kinda like running games with dgVoodoo2. The renderer does require a modern Windows version, but the rest of the game code might still be from the stone age and thus behave better with a 32-bit prefix.

strycore:
Do you have one single example of such a game?
Also, special cases are still best handled with install scripts on the website.  
I would like to reach a point where install scripts *only* deal with special cases and the rest is automated.

PR #4711, opened 2023-01-14
Add ModDB download helper

### LINKED ISSUE #3918: Feature Request: Allow Dynamic -file: parameters in install scripts
So, there are instances, *especially* standalone mods like from moddb, where the game/installer file link is not permanent, but is always changing. For example, I recently wrote the new installer for the Anomaly standalone S.T.A.L.K.E.R. mod and it eventually got published. However, the link is dead because of the way they use their mirrors.

Here's how it works:

For each game, there's a base url. For Anomaly, it's https://www.moddb.com/downloads/mirror/207799/121 with 121 being the mirror number (there's any one of several mirrors, you can use any of them. But that alone gets you nowhere. The full link as of me typing this to download the file is  https://www.moddb.com/downloads/mirror/207799/114/458d9b0d207635c9d3f2be487355d61d/?referer=https%3A%2F%2Fwww.moddb.com%2F

That hash, `458d9b0d207635c9d3f2be487355d61d`, is what changes, and it changes every couple hours. So the original installer doesn't work unless the user downloads the Anomaly-1.5.1.2.7z file themselves and provides it during the installation. Sure, that's not a huge deal, but not ideal at all.

I actually contacted the devs over at moddb and explained the issue, and they provided me with a way to always get the current hash and correct download link. And I have a little bash script that will always grab the correct url and hash, turn it into a url that wget can parse, and download the file to `Anomaly-1.5.1.2.7z`.

```
#!/bin/bash
url="$(wget -q -O -  https:// www.moddb.com/downloads/start/207799 | grep /downloads/mirror/207799 | cut -d= -f2 | cut -d">" -f1 | cut -d";" -f1 | grep https | jq -r)"

wget -q $url -O Anomaly-1.5.1.2.7z
```

So, I edited the current install script (I haven't submitted the edits yet but I will shortly) and have tested it several times. 

Right now, it looks like this:

```
files:
- script: https://raw.githubusercontent.com/gardotd426/Anomalyscript/main/urlvar.sh
game:
  exe: drive_c/anomaly/AnomalyLauncher.exe
  prefix: $GAMEDIR
  working_dir: $GAMEDIR/drive_c/anomaly
installer:
- task:
    arch: win64
    description: Creating Wine prefix
    name: create_prefix
    prefix: $GAMEDIR
- execute:
    command: mkdir "$GAMEDIR/drive_c/anomaly"
- chmodx: script
- execute:
    file: script
- extract:
    arch: win64
    dst: $GAMEDIR/drive_c/anomaly
    file: $GAMEDIR/Anomaly-1.5.1.2.7z
    format: 7z
wine:
  dxvk: true
  esync: true
  version: lutris-fshack-6.14-4-x86_64
```

So what it's doing is just pulling in my script that I wrote with the help of the moddb guys that will download the `Anomaly-1.5.1.2.7z` file no matter what the current hash is, and then using that to download the archive and extracting it to the correct location. 

It works perfectly, 100% of the time. It will always work (until Anomaly is updated which will be months from now, that's a different issue, same as any other versioned file). But the problem is I can't provide it in the files section. 

So the script will always work, you just click install, let it do it's thing, then click launch and it works OOTB with zero issues, but I feel like I should be able to provide a sort of bash script or environment variable for the file instead. Something like:

```
files:
- script: https://raw.githubusercontent.com/gardotd426/Anomalyscript/main/urlvar.sh
- archive:
    url: $(wget -q -O -  https:// www.moddb.com/downloads/start/207799 | grep /downloads/mirror/207799 | cut -d= -f2 | cut -d">" -f1 | cut -d";" -f1 | grep https | jq -r)
    filename: Anomaly-1.5.1.2.7z
```
If it were able to parse that command, then the url would always be correct. I've written about 1 python patch in my life so I have no idea how this could be implemented or I'd file the PR itself, or if it's even something you guys would be willing to do. But I think it would be cool. 

As it stands though, the script I posted above does work, and will work regardless of when the current mirror link hash changes. So I'll submit the updated installer. 

I know this sounds  kind of confusing, so if any clarification is needed I'm happy to answer any questions.

For reference, here's the original that no longer works because of the expired link:

```
files:
- client: https://www.moddb.com/downloads/mirror/207799/123/fc30ca003ca55b47b54be96c75f24120/?referer=https%3A%2F%2Fwww.moddb.com%2Fmods%2Fstalker-anomaly%2Fdownloads%2Fstalker-anomaly-151
game:
  exe: drive_c/anomaly/AnomalyLauncher.exe
  prefix: $GAMEDIR
  working_dir: $GAMEDIR/drive_c/anomaly
installer:
- task:
    arch: win64
    description: Creating Wine prefix
    name: create_prefix
    prefix: $GAMEDIR
- execute:
    command: mkdir "$GAMEDIR/drive_c/anomaly"
- extract:
    arch: win64
    dst: $GAMEDIR/drive_c/anomaly
    file: client
    format: 7z
wine:
  dxvk: true
  esync: true
  version: lutris-fshack-6.14-4-x86_64
```

karthanistyr:
~~:warning:  *Please note that this should not be merged before https://github.com/ClementJ18/moddb/issues/25 is fixed*~~
*Now available: https://pypi.org/project/moddb/0.8.1/*

This helper facility will seamlessly swap a "File" url (e.g. "https://www.moddb.com/games/tribes/downloads/starsiege-tribes-retro-version") into a mirror download link.

This enables installer authors to specify files to be downloaded directly from ModDB, e.g.

```
files:
    - t1_setup:
        url: https://www.moddb.com/games/tribes/downloads/starsiege-tribes-retro-version
        filename: t1retro.exe
        referer: https://www.moddb.com/games/tribes/downloads
```

I guess it's never too late to come back to a long delayed project...
Reference discussion: https://github.com/lutris/lutris/discussions/3872
@intenscia for your information and thanking you for your support.

Also fixes #3918 

Signed-off-by: Antoine Mazeas <antoine@karthanis.net>


danieljohnson2:
I'm not a big fan of one aspect of this- it makes moddb URLs implicitly magic. They look normal, but do special things. I think there should be something explicit in the .json to say "go go moddb magic!"

The simplest way to do that I can see is to use a special fake URL schema, say "moddb:" or something to trigger the magic. The only allowed scheme now is 'https:', so that seems easy.

This would be best if we wanted to support such URLs in many places. If it is supposed to just be in the one spot in the installer json, we might instead change the key from "file" to "moddb_file" or something.

Either way, this should be documented in 'installers.rst'.

karthanistyr:
hi @danieljohnson2 , thanks for your comments. Would you consider a properly detailed "moddb url" section in the installer as sufficient? It looks a bit redundant to have a specially named token (be it the moddb: fake scheme or a special moddb_file key), because they can only, be, and ever will be, used with moddb.com urls anyway.

I've thought about the proposed schemes, applied to the above example

#### `moddb:` url scheme:
```
files:
    - t1_setup:
        url: moddb://www.moddb.com/games/tribes/downloads/starsiege-tribes-retro-version
```
This looks like it would push the installer author to have to modify the url (instead of just copy/paste it), so it's a bit more work and maybe a bit error prone, would you agree? Also it would deviate from prior art in lutris, which has a `N/A:` and `$STEAM:` special keys, but I find them less than obvious to use.
Alternative, but still redundant in my mind:
```
files:
    - t1_setup:
        url: MODDB:https://www.moddb.com/games/tribes/downloads/starsiege-tribes-retro-version
```
#### `moddb_file`
I don't really know what you are referring to as "key", looking at [schema/installer.schema.yml](https://github.com/lutris/lutris/blob/a40d5d7300680dd94c4eeb8dfde43f396268c59a/schema/installer.schema.yml), there doesn't seem to be such a thing as a key called `file`. Can you describe how it should look like ?

danieljohnson2:
Er, the key is 'url' I guess. You could have 'moddb_url: "https..."'.

But I am okay with your moddb: scheme, or with a "helper" key. Or emulating "N/A" or "STEAM" special keys- though I think that "N/A" is not very descriptive.

I 'spose there's something to be said for emulating the prior art here. More consistent.

I just want it to be an explicit thing you can see in the json. Something to say "here be dragons", if you see what I mean.

karthanistyr:
I've added the explicit `MODDB:` url prefix (see docs for details) as requested. I think the PR is ready to review now.

strycore:
Don't make users do weird shit, it's ok if Lutris automatically handles ModDB URLs. 
It should be transparent for the users and they shouldn't have to come up with made up URLs like `moddb:` or whatever.

strycore:
This needs a graceful fallback in case the moddb python module isn't installed.
Otherwise, this is awesome, one of the last big missing parts in Lutris!

karthanistyr:
@strycore thanks for this, I've reverted the url marker and added some guards; also, the handling of a missing lib.

karthanistyr:
@strycore thanks for this, I've reverted the url marker and added some guards; also, the handling of a missing lib. In this version, Lutris will  disable the moddb helper and pass the stated url in the installer file as is with no transform (fallback to current behaviour), with a console message.

karthanistyr:
I can see one surviving bug, is the helper's ctor can be abused by a developer by passing an explicit not None `parse_page_method` function and an explicit `moddb_lib=None` or any module other than moddb; the code will crash at the `isinstance` invocation further down.

karthanistyr:
I think I'm done now. Happy to have it reviewed.

strycore:
Looks good

PR #4720, opened 2023-01-22
Fix Steam Flatpak not appearing in sources

### LINKED ISSUE #4449: Flatpaked Steam Not Detected
steam is installed(flatpak). I am using lutris(flatpak) and steam is not showing up as a source.
image
file system permissions seem to be set correctly
image

Dwahgon:
- Fixes Steam not appearing in sources if Steam is flatpaked
- Fixes not being able to run games using flatpaked Steam
- ~~Adds a toggle to use either native or flatpak Steam~~

A Steam Flatpak Runner and Source might be better, but here's a workaround for now

Fixes #4449

### DIFF
--- a/lutris/runners/steam.py
+++ b/lutris/runners/steam.py
@@ -30,6 +30,7 @@ class steam(Runner):
     human_name = _("Steam")
     platforms = [_("Linux")]
     runner_executable = "steam"
+    runner_executable_flatpak = "com.valvesoftware.Steam"
     game_options = [
         {
             "option": "appid",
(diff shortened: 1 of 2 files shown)


strycore:
Why is there the need to have a toggle? Can't this be determined automatically?

Dwahgon:
> Why is there the need to have a toggle? Can't this be determined automatically?

If the user has both the native and flatpak version of Steam installed (pretty weird scenario but might happen), he can choose which one to use to run the game.
I could just remove the toggle and make it prefer the native version if both are installed.

strycore:
> I could just remove the toggle and make it prefer the native version if both are installed.

Yes, it's better to have a cleaner UI than to adapt to weird scenarios.

Victor239:
I use both versions of Steam because I have two accounts. 

I've also got a friend who uses both because they want to move to the Flatpak only but they had issues with a particular game so haven't fully migrated yet. 

It would be good for these kinds of scenarios that the user can launch games from all available sources, rather than having to choose.

strycore:
There are ways to handle this without adding options to the UI.

- Handling both transparently
- Detect that both are installed and show a popup asking the user which one to use

PR #4728, opened 2023-01-29
Fix failure to start a game if Steam is not installed and Proton-GE is used

semjon00:
Fixes a regression


danieljohnson2:
This looks like a good idea to me!

PR #4732, opened 2023-02-02
A Stop button for ImportGameDialog

danieljohnson2:
This is a face-lift for the ImportGameDialog; it removes the 'Launch Game' checkbox in favor of a 'Launch' button for each ROM- this way you can import multiple ROMs and pick one to launch.

It also adds a 'Stop' button at the bottom to interrupt the search and import process. This button turns to 'Close' upon completion. When in the 'Stop' mode, it does not close the dialog.

It's also a modeless dialog now, like InstallerWindow and AddGamesDialog.

image


strycore:
Nice!

PR #4744, opened 2023-02-10
Fix updates with itch.io

### LINKED ISSUE #4740: itch.io can't update games
Headline says all. 

It throws a 400 error and has issues downloading the files.
I wrote the itch.io service and make this issue mostly to let others know that I am already on it and will submit a merge request within the next couple days.

tbh I dropped the ball here as I barely tested this feature. Sorry about that.
In my only defense: I didn't have a game to update which has now changed as of a couple hours ago.

GoGoOtaku:
Some rewrites to fix updating.

As an aside: The sqlite database does not store when a game was updated. `installed` is set when a game is first installed and I couldn't find when `updated` is ever set. Not sure if this is wanted or not but it is outside the scope of the itch.io service and thus I made a "workaround" by saving the date in the meta file I already placed in the install directory.

This will close #4740

### DIFF
--- a/lutris/services/itchio.py
+++ b/lutris/services/itchio.py
@@ -350,16 +350,42 @@ class ItchIoService(OnlineService):
             }
         }
 
+    def _check_update_with_db(self, db_game, key, upload=None):
+        stamp = 0
+        if upload:
+            uploads = [upload["upload"] if "upload" in upload else upload]
+        else:
+            uploads = self.fetch_uploads(db_game["service_id"], key)
+            if "uploads" in uploads:
+                uploads = uploads["uploads"]
+
+        for _upload in uploads:
+            # skip extras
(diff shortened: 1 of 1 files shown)


danieljohnson2:
I have to say, this is the best line of code ever:
```
uploads = [upload["upload"] if "upload" in upload else upload]
```

danieljohnson2:
Still, looks like a legit fix for the 400 error, so I'll merge this.

danieljohnson2:
Pylint tells me the method '_check_outdated_in_db' is not there. I think you meant _check_update_with_db, so I'll make that change. Please advise if that's not correct.

GoGoOtaku:
Yes indeed. tbh I struggled a bit having a readable name for the function and renamed it. And it seems I forgot to rename it there.

PR #4748, opened 2023-02-12
Fix installers.rst formatting issue

canepan:
Also, make code block formatting more uniform

### DIFF
--- a/docs/installers.rst
+++ b/docs/installers.rst
@@ -210,7 +210,7 @@ Example: ``$GAMEDIR/path/to/game``
 that comes with a map editor, or that need to be launched with different arguments)
 you can specify them in this section. In this section, you can have a list of configurations
 containing ``exe``, ``args`` and ``working_dir`` plus a ``name`` to show in the launcher dialog.
-Example:
+Example::
 
   game:
     exe: main.exe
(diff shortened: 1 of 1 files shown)


danieljohnson2:
I'm lost. This commit re-arranges double-colons '::'; but I can't find a lick of documentation on what the heck '::' means and where it should go. Can you enlighten me?

canepan:
"::" in RestructiredText means the following spaced lines are a code block.
I figured it from the rest of the doc

⁣Nicola Canepa​

Il giorno 12 feb 2023, 10:13, alle ore 10:13, Daniel Johnson ***@***.***> ha scritto:

canepan:
Some example using `::` can be found [here](https://github.com/ralsina/rst-cheatsheet/blob/master/rst-cheatsheet.rst#blocks) or [here](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#literal-blocks).

danieljohnson2:
This is a weird format. Thank you for the documentation. It seems to endorse both forms, though.

So why is it better to have this with the '::' merged into the preceding ':'?

canepan:
I am a fan of uniformity, and both the main README file and (roughly) 20 out of 32 instances in the same file are using this format.  
Besides that, I prefer it :).

The main reason for this change was to fix a broken formatting at line `213`:

image

but then I got carried away

danieljohnson2:
Ah, I see what I mean. That's broken. I'll merge the PR to fix that.

PR #4766, opened 2023-03-06
Move PPA_URI for release and staging builds into repository variables.

Amaylia:
This is a very small change to the publish-lutris-ppa GitHub workflow that moves the PPA_URI into repository variables.  I also added a comment to make it more clear what content needs to go into each repository secret/variable just in case that helps in the future.

### DIFF
--- a/.github/workflows/publish-lutris-ppa.yml
+++ b/.github/workflows/publish-lutris-ppa.yml
@@ -9,8 +9,16 @@ on:
   # See below for details about the 'release' event.
   # https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release
   release:
-    types: [released,prereleased,published]
+    types: [released,prereleased]
 
+# Requires that PPA_GPG_PRIVATE_KEY & PPA_GPG_PASSPHRASE be defined
+# as repository secrets and that PPA_RELEASE_URI & PPA_STAGING_URI
+# be defined as repository variables.
+# Ex. PPA_GPG_PRIVATE_KEY: (( output of `gpg -a --export-secret-key <key_id>` ))
+# Ex. PPA_GPG_PASSPHRASE: (( password used to export the above key ))
+# Ex. PPA_STAGING_URI: ppa:lutris-team/lutris-staging
+# Ex. PPA_RELEASE_URI: ppa:lutris-team/lutris
+#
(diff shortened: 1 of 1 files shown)


Amaylia:
A few things need to be done to make this work correctly, but I've already done the first one:

- [ ] Add PPA_RELEASE_URI & PPA_STAGING_URI repository variables pointing to the lutris-team PPAs.
- [ ] Update repository secret PPA_GPG_PRIVATE_KEY at https://github.com/lutris/lutris/settings/secrets/actions with the output of `gpg -a --export-secret-key <key_id>` .
- [ ] Update repository secret PPA_GPG_PASSPHRASE at https://github.com/lutris/lutris/settings/secrets/actions with the passphrase needed to unlock the private key above.
- [ ] Remove the GitHub "Staging" environment, or limit it to a branch that isn't master.

The Staging environment GitHub environment is causing all branches in the repository to be considered part of Staging and so any GitHub workflows that don't specifically reference that environment won't run.  I don't think that was intended when creating that environment, but if it was, then we can rework it a bit so that only encapsulates branches outside of master.

strycore:
I've done steps 2, 3 and 4 with a newly generated GPG key (the previously used one didn't have a passphrase, which this setup doesn't seem to support)

I didn't really plan to create packages outside of the master branch so I removed the Staging environment entirely.

strycore (commented):
Looks good!

Amaylia:
Thank you, @strycore!  And sorry, I didn't think to include support for no password.  I see what you mean though.. in the event that the private key here was somehow compromised, the passphrase probably wouldn't be safe either.  If you ever do want to go back to using the environments in GitHub, the workflow would just need to specify which environment it should run under.  Environments _do_ seem like they could be useful.

strycore:
Ok, I've pushed a new tag with the GPG private key and secret filled in. 
Let's see if Github Actions works its magic.
If not I will release more betas, I don't care if we have to release 800 of them, we'll get this to work.

strycore:
Apparently, no action was run. So we're good for a beta 3.
@Amaylia Let me know what needs to be changed, or if you need more permissions on the repo so you can make changes yourself.

danieljohnson2:
I guess we're looking at a final release in 2156 or so!

Issue #3875, opened 2021-12-01
[Feature Request] add an option to disable game thumbnails


07416:
I have a large library and prefer not to download, store on disk and display game thumbnails in Lutris.

danieljohnson2:
What would you see in the UI with this option? Gray rectangles where the thumbnails should be?

07416:
> What would you see in the UI with this option? Gray rectangles where the thumbnails should be?

That's a decent placeholder, but the space formerly reserved for the images could perhaps be optimized (reduce the dimensions to some degree).

I don't know if Lutris offers a "compact" library UI akin to Steam, from the usability standpoint I find the interface quite good on a high-resolution screen:

![image](https://user-images.githubusercontent.com/14923168/144454545-36d3285c-a1a5-40eb-9a80-a46e256a9b53.png)

danieljohnson2:
Lutris offers a zoom slider (its in one of the drop downs); it's a little bugged right now but if you frob it enough you should get to a view with icons. Put it also in list mode and it's not entirely unlike that.

However, I believe Lutris will still download the banners all the same, as it's not related to what you view.

I think what would be better is if Lutris only downloaded the icons you view, and then being in the icon mode would mean that banners don't download.

strycore:
Buy a bigger hard drive.

07416:
> Buy a bigger hard drive.

@strycore I'm a wheelchair-bound US citizen without a health insurance, my income is mainly spent on medicine and the mental health care I can afford. Chronic pain can't be cured.

Does the reply represent the entire Lutris project?

danieljohnson2:
@strycore is in charge, but I think maybe he was kidding.

But, I wonder what the issue you are having actually is. Is it disk space usage? Just how big is your ~/.local/share/lutris/banners directory, for instance?

Mine is 9.3MB and contains 1,138 banners. That's not even as much space as one small game, these days.

If yours is actually very large, then something very strange is going on, I think. We should find out what's happening before implementing a solution.

You've proposed a solution- a switch to turn of banners- but that might not even help. It would not, if the issue is disk space- Lutris downloads the same stuff no matter what you display on the screen.

If the problem is not disk space usage, then what it is? Network bandwidth? Screen real estate usage? Ugly banners with no taste?

07416:
> And I wonder what the issue you are having actually is. Is it disk space usage? Just how big is your ~/.local/share/lutris/banners directory, for instance?
> 
[...]

PR #4790, opened 2023-03-28
Do we need stinking badges?

### LINKED ISSUE #220: There needs a way to differenciate different versions of a game

danieljohnson2:
Adds badges to indicate platforms on banners and cover-art:

Screenshot from 2023-03-28 19-26-15

Now, in #220 you asked for consistent, good-looking 32x32 icons. These seem consistent, look pretty crap, and are 16x16. But they are the icons we have in the sidebar, so there's that. 16x16 is still too big to work on the icon view, but 8x8 looks *awful*, so I just omit the badges there.

The badges are omitted from the  platform view, because they are pretty redundant there.

Notice that you can have more than one platform. Installed games should always show  the platform of the installation, but uninstalled games may show multiple if that information is there. I implemented this for Lutris, GOG and the Dolphin service, but it looks like every services has its own 'details' JSON, so each service that knows about platforms will need to handle this separately 

The cover art media get larger badges which look better:
Screenshot from 2023-03-29 04-21-18

There are no badges in the list view, but multiple platforms can appear there:
image

You can turn the badges off if they offend your sensibilities:
Screenshot from 2023-04-01 06-31-10
I've made this update the view as soon as you toggle it, and I threw in the same for 'Hide text under icons', so no settings require restarts anymore.

There may be a slight performance cost here, as this PR no longer caches 'faded' uninstalled media; rather the media is composited with any badges, and that is faded as a whole. I think this should be okay.

Resolves #220

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -9,6 +9,7 @@ lutris (0.5.13) jammy; urgency=medium
   * Group configuration options into sections
   * Added checkbox to stop asking for the launch config for a game
   * Support for launch-configs in shortcuts and the command line
+  * Show platform badges on banners and cover-art
   * The add-games window can now create 32-bit WINE prefixes
   * Add filter field to runner list
   * Show game count in search bar
--- a/lutris/gui/config/preferences_box.py
+++ b/lutris/gui/config/preferences_box.py
@@ -9,7 +9,7 @@ from lutris.gui.widgets.common import VBox
 class PreferencesBox(VBox):
     settings_options = {
         "hide_client_on_game_start": _("Minimize client when a game is launched"),
-        "hide_text_under_icons": _("Hide text under icons (requires restart)"),
+        "hide_text_under_icons": _("Hide text under icons"),
         "hide_badges_on_icons": _("Hide badges on icons"),
         "show_tray_icon": _("Show Tray Icon"),
         "dark_theme": _("Use dark theme (requires dark theme variant for Gtk)"),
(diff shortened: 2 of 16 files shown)


QmwJlHuSg9pa:
Offtopic, but if we're getting a toggle for badges, it would be nice to have a toggle for banners, too (and subsequently give LUTRIS_HIDE_MEDIA #3875 some more love .e.g. adjust white space). I say this because I can't bear to have banners enabled due to inconsistencies between platforms (and the smallest size is generally too big).

strycore:
I feel that if we're having a tpggle to hide the images, it should only apply to the list view and not the grid view. The grid view heavily relies on images and wouldn't really be usable. 
I understand how the various image sizes on the list view can be disrupting.

danieljohnson2:
The Platform column should be working better in the latest master; have you tried that?

(No need to install, just download it and run the bin/lutris script within)

QmwJlHuSg9pa:
I did a fresh RPM install into a Silverblue Toolbox (dependencies), followed by `git clone`, and ran the `bin/lutris` script. I then synced my GOG.com account.

I can now see Platform listings for uninstalled games, and it is listing multiple platforms where appropriate (e.g. "Windows, Mac, Linux"). I have not tried installing any games. All the same, it seems to be working well!

Thank you.

PR #4791, opened 2023-03-30
added greek translation

Pyrofanis:
for any issues please contact me

best regards
pyrofani


danieljohnson2:
Well, looking at this what I see is a grammar error in the English text. "By default, it use the directories in the ..." should be "By default, it will use the directories in the ...". I'll fix that, but that'll break your .po file.

@strycore has told me not to tamper with .po files to 'fix' them, so I'll leave this to people who know what they are doing.

strycore:
@danieljohnson2 the grammar error is likely in the code itself and that's where it has to be fixed. Then the po files can be re-generated with the fixed string.

danieljohnson2:
I fixed it in the code.

Pyrofanis:
ok guys
so what steps should I take to correct the translation after the grammatical error fix ?

danieljohnson2:
It looks like the README.md file describes 'Updating a translation', which says you do this:
```sh
meson builddir
ninja lutris-update-po -C builddir
```

And that updates all .po files.

Pyrofanis:
ok going to resolve it asap thanks 👍🏼

Pyrofanis:
ok two strings needed to be updated 
i only pushed the updated el.po file to not mess with other translations 
also deleted the first 5 lines containing the copyright

Pyrofanis:
nope is certainty not thanks daniel for pointing  them out

Pyrofanis:
found some more strings that needed to be adjusted

PR #4795, opened 2023-04-03
Distinguish between service type and service ID

ntninja:
Distinguish between service type and service ID, where the ID contains extra information to distinguish between different service accounts

This is the first, and apparently most controversial, commit from #4487 (multi-account support).

Terms used in the following:

  * *service type*: A Lutris service implementation that can communicate in a certain way with some kind of remote service (ie: GOG, HumbleBundle, itch.io, …). Importantly, while the credentials used to communicate with the remote service may change, the employed API/RPC calls and exchanged data structures will otherwise remain the same.
  * *service connection*: An individual instance of a Lutris service type implementation. Each service connection will typically use different credentials (access a different account), but will otherwise reuse all logic of the underlying Lutris service type.

As you can see it renames the `id` property on services to `type` and adds a new `id` property that uniquely identifies a single *service connection*. As of this PR, this infrastructure isn’t used yet (`id` is always just set to `type` are always the same), but subsequent commits in #4487 change `id` to be of the form `<type>~<username>` for GOG.

You might be wondering why renaming the current `id` to `type` and then adding a new `id` property would be preferable to just adding new property with a different name and keeping the current `id` as-is? Well, looking at Lutris service source code it turned out that most instances that refer to the service ID currently actually use to to distinguish between things that happen to be *per-service-connection* information, rather than *per-service-type* (in my reading at least). So the diff is actually smaller this way.

Also this PR does not touch any database stuff, since the current database just refers to the `service.id` and that is exactly what we want, since available/installed games are *per-service-connection* things anyways.

### DIFF
--- a/lutris/gui/application.py
+++ b/lutris/gui/application.py
@@ -54,7 +54,7 @@ from lutris.util.http import HTTPError, Request
 from lutris.util.log import logger
 from lutris.util.steam.appmanifest import AppManifest, get_appmanifests
 from lutris.util.steam.config import get_steamapps_dirs
-from lutris.services import get_enabled_services
+from lutris.services import get_service
 from lutris.database.services import ServiceGameCollection
 
 from .lutriswindow import LutrisWindow
@@ -663,7 +663,7 @@ class Application(Gtk.Application):
         if service:
             service_game = ServiceGameCollection.get_game(service, appid)
             if service_game:
-                service = get_enabled_services()[service]()
+                service = get_service(service)
                 service.install(service_game)
                 return 0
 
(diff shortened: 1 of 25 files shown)


ntninja:
Rebased on top of latest Lutris

strycore:
I've reviewed this PR again, and while there are thing that still make me uncomfortable, the changes made here don't cause changes in the functionality so I agree to merge this if it can move things forward.

> Also this PR does not touch any database stuff

I really wish it did. This composite naming with <service_type>-<account_id> is really going to make querying things harder. Why can't there be a new account column in the database?

The current priority is handling multiple Steam accounts because that's a use case that comes up a lot with the Steam Deck. Steam is likely a lot simpler to handle than GOG (we don't have to handle auth). So I can work on that can and we can find a solution that fits every service.

strycore:
I ended up reverting the merge because it's too disrupting. 
Please make the following changes:
- Add an account_id field to the database
- Do not create composite names with "~" or "-"
- All existing game libraries should not be disrupted
- Accounts should be optional, if we don't have support or need for multiple accounts in a service, it should still do the right thing by default

PR #4802, opened 2023-04-07
Reuse licensing complaince

CordBlaster:
The scope of this pull request is to make this repo complaint with a good license notice specification called [REUSE](https://reuse.software) from the [FSFE](https://fsfe.org). Code and other files can be more distinguishable for us humans and machines regarding licensing.

The Licensing should be clear (GPLv3-or-later) however. I looked at some files more thoroughly regarding the copyright notices (Authors/Contributors).


danieljohnson2:
This seems very enterprisey. Not a fan of that.

Who would want to use your REUSE tool on Lutris? Not many corporate IT departments will be interested in Lutris, I think. What's it good for? Who is going to use it?

We seem to be doing this GitHub-style now. Will GitHub recognize your new thing? You are removing the usual LICENSE file here; not just adding new stuff.

I don't like smearing licensing information all over the project like this at any rate; you've even hidden some of it *inside* .svg files, and that seems very strange and unfriendly.

I notice also that the new 'dep5' thing seems to list some files under a different license, rather than GPL. That seems off. Why are those files special?

This is not a PR that fills me with joy.

CordBlaster:
> What's it good for? Who is going to use it?

Copyright and License notices in each work was and should be the standard. It's nothing new in the history of Free (Libre) and Open Source Software. Even the GNU General Public License recommends this good practice and REUSE tries to simplify the mentioning in the file headers.

> We seem to be doing this GitHub-style now. Will GitHub recognize your new thing?

Stating the authors and licenses is not a new thing and network services like GitHub and GitLab will only check `LICENSE` or `COPYING` for the license. (I will correct that check for this PR) The bad part about e.g. GitHub and GitLab is that they don't distinguish between version X-only and X-or-later of the GNU license family, but that's another topic.
If 'GitHub-style' means 'just put a license file into the folder and be done with it', there would be better ways or 'styles' for this.

> I don't like smearing licensing information all over the project like this at any rate

There are some files with multiple contributors. If you don't like too many in the list: Just credit the team, relevant authors and/or e.g. the original author could be a way to mention the copyright holder(s) if the people in question are okay with it.

> you've even hidden some [licensing information] inside .svg files, and that seems very strange and unfriendly.

The spec recommends it. It's not the same as with .jpeg and .png. An .svg file is just a plain text XML file. The same thing could apply e.g. to .html and .md where your browser or book reader program could open it. Alternatives would be the `.license` extension or `.reuse/dep5`.

> I notice also that the new 'dep5' thing seems to list some files under a different license, rather than GPL. [...] Why are those files special?

Sorry, my fault. I would revert it. This non-special mostly uncopyrightable config or list files could be under the public domain, however. Only copyright holders of copyrightable work can license something under another license like e.g. the CC0.

danieljohnson2:
I am not persuaded. This is plainly no simplification. I think there's no audience for this kind of thing, not for Lutris.

We have a source control system that tracks authorship; polluting source files with that information is high maintenance and useless.

I also do not think it makes much sense to put individual bits of Lutris under a different license. It seems to me the bits you've picked out have little or no value separated from Lutris itself anyway. It's the strangest thing the PR- the rest of my complaints are really about all the ceremony, but this part is a licensing change.

I am also a bit concerned that SVGs with comments in them might not be as well supported as normal ones; though on paper you are correct, and the hidden licensing data should do no real harm. Perhaps I am just being a nervous Nelly here. But *hidden* licensing data seems like a strange choice.

CordBlaster:
> This is plainly no simplification.

In case you made proper attributions before: REUSE makes it easier, if you had this in a file e.g.:
```
/*
Copyright (C) 2023 Daniel Jonson

This program is free software: you can redistribute it and/or modify
it under the terms of the GNU General Public License as published by
the Free Software Foundation, either version 3 of the License, or
(at your option) any later version.

This program is distributed in the hope that it will be useful,
but WITHOUT ANY WARRANTY; without even the implied warranty of
MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
GNU General Public License for more details.

You should have received a copy of the GNU General Public License
along with this program.  If not, see <http://www.gnu.org/licenses/>.
*/
```

You can reduce it to this instead following the REUSE specification:

```
/*
Copyright (C) 2023  Daniel Jonson

SPDX-License-Identifier: GPL-3.0-or-later
*/
```
If you don't want to (or can't) remove the license information (not the Copyright holder), you can just add the SPDX-License-Identifier under it to be complaint.

> We have a source control system that tracks authorship

https://reuse.software/faq/#vcs-copyright
There can be big differences between the Author and the Copyright Holders, when it comes to the version control system. If I made one contribution to each projects I know of, where I only change a typo or add a not relevant code line, I would probably not hold any copyright on my changes. (The copyright laws only covers originality.) The VCS would still mention my name though. I don't think this Pull Request is copyrightable too.

> polluting source files with that information is high maintenance and useless.

A copyright and license notice in a file doesn't add 200 lines at the top. You don't even need to change it at all if it was correct in the past in my opinion (you don't even need to maintain the years). Some authors also prefer an attribution.

> I also do not think it makes much sense to put individual bits of Lutris under a different license.

There are no other licenses in this moment. It's technically possible to have multiple (GPL compatible) licensed code in the repo however.

> I am also a bit concerned that SVGs with comments in them might not be as well supported as normal ones

SVGs can have comments. This shouldn't be a problem (according to the XML specification) if you don't put anything above the XML declaration (if there is one).

>  the hidden licensing data should do no real harm.

They aren't _hidden_ (in the SVGs). The same thing would apply for any other file format which you can look at with a text editor. It's plain text from top to bottom.
I made a separate `.license` file for your concerns.

I also changed the notes to a more familiar `Copyright (C) [Author]`

PR #4828, opened 2023-04-26
Update pcsx2.py to match command line of PCSX2 1.7+

7HEPOW:
- Removed libgdk-x11-2.0.so.0 dependency as it's not required by PCSX2 1.7+
- Corrected command line switches from double dashes to single dashes (e.g. -nogui instead of --nogui)
- Removed --configpath and --configfile as they've been removed and replaced by per game configuration files created through the game properties in the PCSX2 GUI

Comes a little early maybe, but 1.7 is what I get from the AUR. 1.7 is still considered a nightly. They've moved the GUI to Qt6 and have a new Vulkan based renderer which fixed a lot of games already.

I figured that I fix this for a future release of Lutris when PCSX2 gets updated accordingly.


danieljohnson2:
If we're going to switch to the QT version of PCSX, should we not be adding the dependency on QT, not just removing GTK?

7HEPOW:
True. My mistake :)

legluondunet:
PCSX2 runner has been recently updated, it now use pcsx2-qt AppImage. 
As you know AppImage does not need dependencies.
This changes are only available in latest Lutris dev code.

PR #4834, opened 2023-04-29
add LD_LIBRARY_PATH to dosbox.py

legluondunet:
Hello,
users encounter regularly issues about Dosbox-staging does not launch from Lutris because of missing libraries.
Until dosbox-staging team purposes an AppImage, we could provides needed Dosbox-staging libraries in a lib folder and use LD_LIBRARY_PATH variable. That's already the case for ScummVM runner and I took code from scummvm.py.


legluondunet:
concerning this line:
    runner_executable = "dosbox/bin/dosbox" -->     runner_executable = "dosbox/dosbox"
dosbox-staging has a different structure folder and executable are not stored in a bin folder.
In order to remain compatible with older Lutris version, dosbox-staging archive contains executable in root archive and in a bin folder too.

PR #4875, opened 2023-05-28
Provide explicit compatibility checking for components

danieljohnson2:
The recent addition of a DLL to VKD3D caused problems for older versions of Lutris as soon as it was downloadable. These older Lutris versions tried to use the new VKD3D, but did not know about the new DLL, so it did not work.

It's a bit hard to expect everyone to upgrade to latest instantly; we should have some better behavior here.

So, this PR adds a check to the version selection and download process for components. If a component has a JSON file named '.lutris_compatibility.json'. It looks like this:
```
{"min_lutris_version": "0.5.14"}
```

It's JSON in case we want to get fancier about this.

What this PR does is check for this, and if the version downloaded is found to be incompatible with your Lutris, we'll go on to download the next version (and so on until we find a compatible one). The incompatible versions remain and can be selected, but won't be the default if any compatible version is found.

### DIFF
--- a/lutris/runners/wine.py
+++ b/lutris/runners/wine.py
@@ -16,7 +16,7 @@ from lutris.util.display import DISPLAY_MANAGER, get_default_dpi
 from lutris.util.graphics import vkquery
 from lutris.util.log import logger
 from lutris.util.steam.config import get_steam_dir
-from lutris.util.strings import parse_version, split_arguments
+from lutris.util.strings import split_arguments
 from lutris.util.wine.d3d_extras import D3DExtrasManager
 from lutris.util.wine.dgvoodoo2 import dgvoodoo2Manager
 from lutris.util.wine.dxvk import REQUIRED_VULKAN_API_VERSION, DXVKManager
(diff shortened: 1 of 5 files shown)


strycore:
Instead of blocking certain versions on older releases, couldn't we expose the needed DLLs from the API so if there's an addition in the future, it can be handled by any version?

danieljohnson2:
That is also a a good idea, but this PR provides a option when something *else* causes compatibility breakage, too. I think we should do this PR, even if we also provide the DLL-lists somewhere.

A design question though. Would it be better to provide the DLL-list inside the component packages (like this compat thing) or as part of the versions-JSON, or somewhere else?

strycore:
It's likely a good idea to have both.

We could possibly deduct the DLL list by just looking at the contents of the archive.

strycore:
I'm super nervous about the .lutris_compatibility.json file :cold_sweat:

danieljohnson2:
Why is that?

strycore:
It is a hidden file stored in an arbitrary folder, which is close to being evil.
It is so ultra-specialized that is has no hope/no future outside of its own ultra-specialized purpose.

danieljohnson2:
Well, these components are just for Lutris, are they not?

I had also made ".lutris_dpi_assignment" and ".lutris_destkop_integration" because I can't spell either. Evil also?

But these things can be absolved of their evil by dropping the '.', perhaps. Not a super hard change.

strycore:
> I had also made ".lutris_dpi_assignment" and ".lutris_destkop_integration" because I can't spell either. Evil also?
Yes, very

> But these things can be absolved of their evil by dropping the '.', perhaps. Not a super hard change.

Good start, then maybe consider having more general purpose config files

strycore:
Traumatic.

image

danieljohnson2:
You want to look at .DS_Store files?

Whatever. Perhaps a visible "lutris.json" in each relevant location would be ~~spelled correctly~~ more suitable?

strycore:
That seems a lot more reasonable!

danieljohnson2:
I'll do that, then. I'll also replaced those previous dotfiles with a single 'lutris.json' at the prefix root.

danieljohnson2:
Okay, renamed.

I notice that the itch.io runner uses .lutrisgame.json for something I don't understand. I won't touch that now, but I'll fix the two I mentioned before in the master.

strycore:
Great, thanks!
I'll merge this one.
The itch.io file probably needs to be addressed at some point too

PR #4877, opened 2023-05-29
Add Italian Translation

Domefemia:
Almost translated all the strings, only 80 strings remaning.


Domefemia:
I think it's better to merge it right away, I don't that much free time to translate them and most of them use specialized terminology I don't know.

danieljohnson2:
Okay. I'll push the button.

PR #4866, opened 2023-05-21
Make Nvidia info not depend on /proc access

lengau:
Some environments have limited access to /proc, but we can still retrieve the same info in other ways.

(Fixes #4858)

PR #4879, opened 2023-05-31
Sample mypy configuration (with baselines)

lengau:
This was brought up by discussion in #4866 about use of type annotations - here's how I'd handle mypy to hopefully have it provide utility without causing headaches.

I've marked this as a draft since its purpose is demonstrative rather than requesting this change.


strycore:
I would like to get started with type annotations, is it possible to get this in a mergable state?

PR #4881, opened 2023-05-31
Updated Spanish translation

mbarrio:
I have updated the Spanish translation file. There are some missing strings in the .po file though, so there still remains untranslated labels and texts here and there, like the "Hide badges on icons" preference, and all the section labels in the system preferences pane ("Display", "Audio", "Input" and so on...).

### DIFF
--- a/po/es.po
+++ b/po/es.po
@@ -3,7 +3,7 @@ msgstr ""
 "Project-Id-Version: lutris\n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2023-02-04 23:57+0300\n"
-"PO-Revision-Date: 2022-04-02 16:32+0200\n"
+"PO-Revision-Date: 2023-05-31 14:21+0200\n"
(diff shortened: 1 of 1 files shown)


danieljohnson2:
Do you want your partial effort merged now, or should we wait for you to finish?

I confess, "Hide badges on icons" is one of mine, so I feel a little slighted here. Just a little.

strycore:
No need to wait if the PR isn't marked as draft. The rest can be completed in another patch.

danieljohnson2:
Slighted I say! Slighted!

mbarrio:
Thanks for merging the patch. I will translate the missing strings in another patch after they get included in the .po file.

mbarrio:
> ... I will translate the missing strings in another patch after they get included in the .po file.

It was my fault after all as I forgot to run meson and ninja before altering the es.po file. I will provide another patch in the following days. I'm sorry for the hassle @danieljohnson2.

PR #4884, opened 2023-06-02
Updated Spanish translation (again)

mbarrio:
This patch should add all the missing strings in the prior patch (because I forgot to run the meson and ninja commands before doing anything).


danieljohnson2:
Ahhh.. no longer slighted!

PR #4913, opened 2023-06-22
Update es.po

tridoxx:
The "enable discord rich presence for avalible games" line is modified, since "rich presence" was being translated, which is a proper English word "it would be like translating software name like "facebook" which should not be done, it is also I add in parentheses information about what requires installing pypresence for it to work

### DIFF
--- a/po/es.po
+++ b/po/es.po
@@ -1107,8 +1107,8 @@ msgstr "Usar tema oscuro (requiere la variante oscura del tema para Gtk)"
 #: lutris/gui/config/preferences_box.py:16
 msgid "Enable Discord Rich Presence for Available Games"
 msgstr ""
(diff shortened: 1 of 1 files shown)


danieljohnson2:
I don't know a word of Spanish, so I'll just assume you know what you are talking about!

tridoxx:
@danieljohnson2 

hahahah, With this it is better understood, although another example was that people translated "easy anticheat", which is a program that is called like that literally and does not translate it into Spanish,  By the way, if lines in Spanish are added, you could ping me on the discord server, to help translate it or here too

danieljohnson2:
We're changing the English text all the time, I'm afraid. I just did some changes just yesterday.

PR #4962, opened 2023-08-05
Update German translation

WebSnke:


danieljohnson2:
Yep, looks like German all right. Checks out, merging it now!

PR #4970, opened 2023-08-09
File Collection allow user to select main installer exe for Gog files

RiqueFR:
Allow the user to select main installer file for Gog games. Amazon games will fail to install if user try to select a file, there is no way of knowing the service in file_box, so I was not able to filter only Gog.

### DIFF
--- a/lutris/gui/installer/file_box.py
+++ b/lutris/gui/installer/file_box.py
@@ -89,14 +89,11 @@ class InstallerFileBox(Gtk.VBox):
             url_label = InstallerLabel("In cache: %s" % self.installer_file.get_label(), wrap=False)
             box.pack_start(url_label, False, False, 6)
             return box
-        # InstallerFileCollection should not have user or steam providers
         if self.provider == "user":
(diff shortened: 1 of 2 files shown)


danieljohnson2:
I'll take this as an improvement, but we're not shippable yet.

The combo box shows and 'Use Cache' option works, but the 'Select File' option also appears and it does not work. It seems to think all files after the first should come from cache, but they aren't there. There's only one text-box to select a file, and no way to pick any others.

The 'Select File' option needs to work, or else not be there.

RiqueFR:
> the 'Select File' option also appears and it does not work. It seems to think all files after the first should come from cache, but they aren't there.

The problem is that I miss write "goginstaller", so it was not setting the right file. It does not matter where it think the other files are, the user can provide the location to the main installer and it will check the same directory to the complementary files, so the fix is simple and I already tested it, soon I will make a new pull request with this fix.

The Select File option can be used on Gog installer but not for Amazon Games, since the amazon installer provide all game files, can't provide a main installation file, but I'm still trying yo figure out how to do not show this option for Amazon.

Issue #4968, opened 2023-08-09
Can't refresh the list of games for Dolphin (Sources)


awsms:
### Bug description

I've added Dolphin as a Source, but it doesn't show anything when refreshing the list.

Similar bug: https://github.com/lutris/lutris/issues/4094#issuecomment-1215331152

### How to Reproduce

1. Set a library path in Dolphin
2. Add Dolphin as a source in Lutris
3. Click refresh
4. Nothing happens

### Expected behavior

Lutris should display the list of games in my Dolphin library

### Log output

```shell
2023-08-09 09:06:09,898: GPU: 10DE:2560 17AA:3A81 (nvidia drivers)
2023-08-09 09:06:09,898: GPU: 1002:1638 17AA:3A81 (amdgpu drivers)
INFO     2023-08-09 09:06:10,172 [startup.init_lutris:159]:Starting Lutris 0.5.14
INFO     2023-08-09 09:06:10,173 [startup.get_drivers:57]:Using NVIDIA drivers 535.86.05 for x86_64
INFO     2023-08-09 09:06:10,174 [startup.get_drivers:61]:GPU: NVIDIA GeForce RTX 3060 Laptop GPU
INFO     2023-08-09 09:06:10,174 [display.get_gpus:77]:GPU: 10DE:2560 17AA:3A81 (nvidia drivers)
INFO     2023-08-09 09:06:10,174 [display.get_gpus:77]:GPU: 1002:1638 17AA:3A81 (amdgpu drivers)
WARNING  2023-08-09 09:06:10,257 [libretro.get_platform:121]:'desmume' not found in Libretro cores
WARNING  2023-08-09 09:06:10,258 [game.set_platform_from_runner:417]:The Libretro runner didn't provide a platform for Rentiangougou (libretro)
INFO     2023-08-09 09:06:10,308 [api.get_default_runner_version:109]:Getting runner information for wine
DEBUG    2023-08-09 09:06:10,308 [http._request:88]:GET https://lutris.net/api/runners/wine
DEBUG    2023-08-09 09:06:10,519 [xrandr._get_vidmodes:16]:Retrieving video modes from XrandR
DEBUG    2023-08-09 09:06:10,858 [lutris.get_missing_game_ids:202]:Checking for missing games
DEBUG    2023-08-09 09:06:12,237 [base.wipe_game_cache:174]:Deleting games from service-games for dolphin
[...]

PR #4971, opened 2023-08-10
Dolphin: refreshing the gamelist in 'Source' menu and 'Batch mode' work again

awsms:
Added a new value (from GameFile.h in Dolphin) + replaced the type of some other values to their corresponding one.
Fixes https://github.com/lutris/lutris/issues/4968

The "no-gui" mode was removed a couple of months ago (https://github.com/lutris/lutris/commit/2290d2a42786e17a825302adca30c2d605682d25), but traces of it remain in the code, which prevented the "Batch" mode to work properly.

### DIFF
--- a/lutris/runners/dolphin.py
+++ b/lutris/runners/dolphin.py
@@ -57,7 +57,7 @@ class dolphin(Runner):
         command = self.get_command()
 
         # Batch isn't available in nogui
-        if self.runner_config.get("batch") and not self.runner_config.get("nogui"):
+        if self.runner_config.get("batch"):
             command.append("--batch")
 
         # Custom Global User Directory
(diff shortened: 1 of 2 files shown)


danieljohnson2:
This looks good but I do have one question. This updates SUPPORTED_CACHE_VERSION from 20 to 24. What happens to people who have the older cache version? Do they just need an updated runner from Lutris or is there more to it?

strycore:
>This looks good but I do have one question. This updates SUPPORTED_CACHE_VERSION from 20 to 24. What happens to people who have the older cache version? Do they just need an updated runner from Lutris or is there more to it?

They need to update the runner in Lutris and click refresh on the Dolphin source, that's it.

PR #4975, opened 2023-08-11
File Collection fix Gog 'Select File' and remove it for Amazon

RiqueFR:
Possible to install Gog games independent of the location. Removed 'Select File' option for Amazon games.


danieljohnson2:
Ah, the joys of spelling. This looks like an improvement, but I wonder if all this will affect services other than GOG and Amazon. Do any others use InstallerFileCollections?

Anyway, I'll merge this guy.

RiqueFR:
> Ah, the joys of spelling.

It always happen hahahhaha.

> I wonder if all this will affect services other than GOG and Amazon. Do any others use InstallerFileCollections?

No other services use InstallerFileCollection, at least for now. Maybe in the future we need to add more, and if they do not support the 'Select File', filter them like Amazon.

PR #4987, opened 2023-08-15
Remove one progress bar from file collection downloads

RiqueFR:
Add InstallerFile size support and InstallerFileCollection full download size support.
Removed one progress bar from the widget.

Separate Gog extra files from FileCollection.

Amazon and Gog using new file size system.

Please, check the UI if it is correct. I do not have a lot of games on Gog and Amazon, I tested downloading almost all my games, but it is best to test with more games, if possible.

### DIFF
--- a/lutris/gui/widgets/download_collection_progress_box.py
+++ b/lutris/gui/widgets/download_collection_progress_box.py
@@ -1,3 +1,4 @@
+import time
 from gettext import gettext as _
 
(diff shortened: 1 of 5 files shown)


danieljohnson2:
OK, this seems to make the remaining bar fill continuously instead of in big chunks. I think it's a reasonable compomise, though I'll miss my big rack of progress bar.

Merging away.

Issue #4985, opened 2023-08-14
CLI options to query service_games table


culain-45124:
Hello,

May I ask about a feature I’m interested in please? I've tried reaching out on your Discord but not had any luck.

For a bit of background, I’ve got a Steam Deck and I’d like a way to bulk import all my games across GOG, Origin, EGS etc, into Steam, whether they’re installed or not, using something like BoilR or Steam ROM Manager. Then I can see all my games, plus my ROMs in one place, set art for them, and use deck-roulette to choose games from across my entire library.

Ideally this process would be repeatable and automated. There is an option in BoilR, to import Lutris games, but it is limited to installed games. I was poking around your CLI commands to see how possible it was with the existing commands to list all my games, but there didn’t seem to be any difference between `-l` for list all games in the DB, and `-o` for just list installed games. It seemed to me from a quick scan of the codebase that games only seem to make it in to the games table in the DB if they’re installed, so `installed=1` pretty much everywhere other than in the pga tests. I can see the data I probably want is in the `service_games` table instead.

Would you have any problem with a feature request to either change the existing `-l` option to point at the `service_games` table instead of the `games` table or, probably less risky given things might be using `-l` already, a new CLI option to list the service_games?

I've actually already done the work for the latter, to add a `-a` option to list all games in `service_games` and a `--list-service-games` to list all games for a specified service, but I can see in your contributing guide that you prefer things to be roadmapped first before just receiving a PR out of the blue.

Its probably similar to https://github.com/lutris/lutris/issues/4770, but approaching it from displaying everything in SteamOS instead. It might also make this issue in BoilR more viable, https://github.com/PhilipK/BoilR/issues/290

Would love to discuss this further if anyone has time.

Thanks

strycore:
> Would you have any problem with a feature request to either change the existing -l option to point at the service_games table instead of the games table 
Yes, we have a problem with that. Why would you even suggest such a thing??????
Gosh.
Ok, now let me finish reading your long ticket.

strycore:
Ok, you can submit a PR to list service games but please, do not and I mean DO NOT break existing command line flags that people depend on. 
Do not code a list-service-games command line flag that list every game from every service. Instead you can add a list-services and a `--list-service-games <gog,humble,whatever>`. If you list every service game from every service, in your use case you'll reimport Steam games into Steam and who wants that.

PR #4990, opened 2023-08-16
Added new cli options to list service games

culain-45124:
Resolves https://github.com/lutris/lutris/issues/4985

- Updated readme to detail new options
- Created two new CLI options `-a, --list-all-service-games` and `--list-service-games`
- One lists all games associated with a provided service
- The second lists all service games
- Both options work with the existing `-o` and `-j` options to return just those marked as installed, or as a json
- Added generic query to service_games table

I know you said not to list all service games, but it seemed best if this didn't cater to my niche use case and was more generic. Games from the steam source are already filtered on the Boilr side, which is likely where I'll move on to next to achieve my dream, but this way it isn't misleading for others and provides everything somebody might want.

Happy to discuss this further, or add additional options to take a list of services, rather than just one.

Looking forward to your comments.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -85,25 +85,27 @@ Command line options
 
 The following command line arguments are available::
 
--v, --version              Print the version of Lutris and exit
--d, --debug                Show debug messages
--i, --install              Install a game from a yml file
--b, --output-script        Generate a bash script to run a game without the client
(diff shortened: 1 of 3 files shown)


strycore:
Looks good!

PR #4993, opened 2023-08-20
French translation & POTFILES updated

Astrobald:
Did my best :cat:


danieljohnson2:
I see accents in there. That's French enough for me!

PR #4999, opened 2023-08-24
Added condition for downloader progress cleanup

dusty73:
This fixes the error `[Errno 21] Is a directory: '/home/dusty/.cache/lutris/installer/name-of-the-game'` that is triggered for Amazon games with multiple files.


danieljohnson2:
I'm trying to fix this by enabling it to replace directories, not just files, in this case. Let's not merge this unless that effort fails.

strycore:
> Let's not merge this unless that effort fails.

Was the effort successful ?

danieljohnson2:
As far as I know it was, but I could not get @dusty73 to test it.

strycore:
Ok, well it's a directory creation, not rocket science so I'm sure it's fine!

PR #5012, opened 2023-09-01
Detect config from Steam snap

techtonik:
https://github.com/canonical/steam-snap


capecrusader-121:
This will most probably break the snap variant of lutris as different snaps aren't allowed to access each others files and configs within the confinement.

 > For e.g. Media stored under the ~/snap/xyz/downloads of a snap isn't accessible by snapped media players.

Not sure although whether lutris enforces this strictly if not, then this won't create a problem.

strycore:
We have something similar in place for the Steam flatpak. Ideally we will need access to all config folders associated to Snap runners.

PR #5022, opened 2023-09-09
Add the ability to change a game's playtime from the edit dialog

AdivonSlav:
So, recently I had to reinstall Lutris and I unfortunately lost my playtime information. Instead of having to resort to using a SQLite DB editor to change this, I thought it would be nice to have an addition to the game edit dialog window where you could change the playtime for each game.

I also made sure to validate the input as a separate *FloatEntry* class. The value is just the number of hours as a floating-point value, the same way it is held in the database.

### DIFF
--- a/lutris/gui/config/common.py
+++ b/lutris/gui/config/common.py
@@ -607,6 +607,13 @@ class GameDialogCommon(SavableModelessDialog, DialogInstallUIDelegate):
         if self.runner_name == "steam" and not self.lutris_config.game_config.get("appid"):
             ErrorDialog(_("Steam AppID not provided"), parent=self)
             return False
+        if self.playtime_entry.get_text():
+            try:
+                float(self.playtime_entry.get_text())
+            except ValueError:
+                ErrorDialog(_("The entered playtime is invalid"), parent=self)
+                return False
+            
         invalid_fields = []
         runner_class = import_runner(self.runner_name)
         runner_instance = runner_class()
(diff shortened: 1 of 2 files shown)


dobval (approved):

AdivonSlav:
If somebody has the time, I'd appreciate it if they reviewed the change to see if it's okay

danieljohnson2:
I took a quick look; mostly okay.

But... what happens if the user types an invalid float in there? I see you filter the input, but I think the user could enter something like "2...22" with too many decimal points.

AdivonSlav:
> I took a quick look; mostly okay.
> 
> But... what happens if the user types an invalid float in there? I see you filter the input, but I think the user could enter something like "2...22" with too many decimal points.

The user can't manually input that sequence due to the filter, but it's still possible to paste in an invalid value. In that case a conversion exception from string to float is thrown. I'll try and catch it manually and display a better error message

danieljohnson2:
I mainly would wish to make sure you can't crash Lutris this way. I don't trust this sort of filtering to be 100% effective, in any case.

AdivonSlav:
Rebased master and added a validation step in the is_valid() method that checks to see if the playtime input can actually be converted to a float. If not, it spawns an error dialog and the save process discontinues. This way, even if the initial input filtering passes, the app won't crash due to a malformed value

AdivonSlav:
Any further edits that need to be made or is this okay?

strycore:
This is ok, I just wanted to wait until after the 0.5.14 release to merge it

MujoBabajic:
nicely done

PR #5030, opened 2023-09-21
Added mGBA to Runners

Quinn64:
Since it includes a GUI, I didn't see much need to add options. Users can configure settings by launching it standalone.


strycore:
Cool, thanks!

PR #5067, opened 2023-10-10
ScummVM: fix of last-minute-before-commit bug

bell07:
"game_id" is available in the scummvm runner only, but not in database. My last fix at this place was garbage. Now switch back to "appid" and remove unused game_id assignment.


danieljohnson2:
Looks like a new flatpak update is coming soon, so we better grab this fix!

PR #5069, opened 2023-10-10
ScummVM: don't set LD_LIBRARY_PATH if runtime disabled and custom command used

mwilck:
More often than not, the "custom executable" set for the ScummVM runner will be the system's scummvm binary. If this binary is used, the libraries shipped with Lutris will most probably not match. Therefore, if a custom executable is used and the Lutris runtime is disabled, don't set LD_LIBRARY_PATH.

This avoids the following error on openSUSE Leap 15.5 with the latest ScummVM runner (2.8.0):
```
DEBUG    2023-10-11 00:01:59,594 [grid.on_item_activated:95]:Item activated: 3
INFO     2023-10-11 00:01:59,630 [runner.use_runtime:386]:Runtime disabled by system configuration
lutris-wrapper: The Secret of Monkey Island
Started initial process 32173 from /usr/bin/scummvm --extrapath=/usr/share/scummvm --themepath=/usr/share/scummvm --aspect-ratio --fullscreen --scaler=normal --scale-factor=3 --path=/home/konrad/Games/the-secret-of-monkey-island monkey
Start monitoring process.
/usr/bin/scummvm: error while loading shared libraries: libreadline.so.8: cannot open shared object file: No such file or directory
Monitored process exited.
Initial process has exited (return code: 32512)
```

### DIFF
--- a/lutris/runners/scummvm.py
+++ b/lutris/runners/scummvm.py
@@ -474,6 +474,10 @@ class scummvm(Runner):
 
     @property
     def libs_dir(self):
+        if ("runner_executable" in self.runner_config
+            and self.runner_config["runner_executable"] == self.get_executable()
+            and not self.use_runtime()):
+            return ""
         path = os.path.join(settings.RUNNER_DIR, "scummvm/lib")
         return path if system.path_exists(path) else ""
 


mwilck:
Note that (on openSUSE Leap at least) ScummVM 2.8.0 doesn't work with the Lutris runtime enabled, either, as `libreadline.so.8` is missing in the Ububtu 18.04 runtime as well. This patch was the only way I got ScummVM to run without ugly hacks.

strycore:
Too specific, I don't think ScummVM should have custom code when that's a feature all runners have.

mwilck:
So how do you propose to fix this issue? 

The scummvm runner *today* has specific code modifying `LD_LIBRARY_PATH` which does not respect the user's choice about use of the runtime, causing breakage. My code is admittedly specific, but fixes this specific bug.

strycore:
I've added a patch for this that isn't based on runtime settings here: https://github.com/lutris/lutris/commit/e75e477109e81eda2b39c85f263ee4e0f1101524

It should fix the issue.

mwilck:
Thanks, this is looking good. I'll test it.

mwilck:
```
ERROR    2023-10-11 22:19:58,916 [exceptions.wrapper:103]:The Secret of Monkey Island (scummvm) has encountered an error: 'scummvm' object has no attribute 'libs_dir'
  File "/home/konrad/lutris/lutris/exceptions.py", line 97, in wrapper
    result = function(*args, **kwargs)
  File "/home/konrad/lutris/lutris/game.py", line 605, in configure_game
    gameplay_info = self.get_gameplay_info(launch_ui_delegate)
  File "/home/konrad/lutris/lutris/game.py", line 583, in get_gameplay_info
    gameplay_info = self.runner.play()
  File "/home/konrad/lutris/lutris/runners/scummvm.py", line 525, in play
    return {"command": command, "ld_library_path": self.libs_dir}
AttributeError: 'scummvm' object has no attribute 'libs_dir'
```

`libs_dir` is still referenced in `play()`.

It works if I add the following patch on top:
```
diff --git a/lutris/runners/scummvm.py b/lutris/runners/scummvm.py
index 73fe8e44..5fb1ef4f 100644
--- a/lutris/runners/scummvm.py
+++ b/lutris/runners/scummvm.py
@@ -522,7 +522,7 @@ class scummvm(Runner):
         for arg in split_arguments(args):
             command.append(arg)
         command.append(self.game_config.get("game_id"))
-        return {"command": command, "ld_library_path": self.libs_dir}
+        return {"command": command}
 
     def get_game_list(self):
         """Return the entire list of games supported by ScummVM."""
```

strycore:
Right, I had only fixed the run standalone code. I've pushed a patch that updates the play() method

mwilck:
Thanks, works now.

PR #5074, opened 2023-10-15
Runner updates and added Rosalie's Mupen GUI as runner

Quinn64:
I went through and updated some of the runners.

Added Rosalie's Mupen GUI as a standalone runner. I also removed its flatpak ID from the mupen64plus runner to avoid confusion or conflicts.

Updated the ruffle link to the current nightly since it was a few months out-of-date.

Updated the mGBA runner to include GB(C) ROMs as a possible option in its help.

Changed http links to https links to avoid the, albeit very minor, concern of possible MITM attacks. I checked each of the links to make sure https is viable.


strycore:
Nice, thanks!

PR #5086, opened 2023-10-20
scummvm: handle runner not being available in get_command, add/fix test

K900:
Before this change, the scummvm tests would fail on a system where scummvm is not available, because https://github.com/K900/lutris/blob/7efc304f595a9547bb4f01a823250d95b7722506/lutris/runners/scummvm.py#L487 attempted to access an element from an empty list. Fix that by propagating the empty list, and use mocks to ensure we test both cases.

### DIFF
--- a/lutris/runners/scummvm.py
+++ b/lutris/runners/scummvm.py
@@ -482,6 +482,8 @@ class scummvm(Runner):
 
     def get_command(self):
         command = super().get_command()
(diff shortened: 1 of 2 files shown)


Antiz96:
I confirm the bug and the fix!

Thanks :slightly_smiling_face:

strycore:
Thanks!

PR #5087, opened 2023-10-21
po: Add Georgian translation

NorwayFun:
Hello,
This is initial, unfinished translation to Georgian.
Will finish it, eventually.

Temuri


danieljohnson2:
Hey, some Georgian is better than no Georgian. And that script is cool!

Merging posthaste!

PR #5126, opened 2023-11-09
Fix launching games from tray icon

daniele-bondi:
The tray icon class calls Application.get_launch_ui_delegate, which is not defined in Application.
I wasn't sure if it was better to make the icon class use directly Application.launch_ui_delegate or add the getter method.
I opted for the getter method so that the field can be treated as an implementation detail.

EDIT: Sorry for the messy PR, making the change directly via github might have been a bad idea.


danieljohnson2:
Oops, I think that one was my bug. I'll take your fix.

PR #5127, opened 2023-11-11
Add more options to get service_game and return appid

### LINKED ISSUE #5116: Issues using lutris: protocol link followed by a game identifier
Hello,

I've been trying to use the `lutris:` protocol link followed by a game identifier to either install or run a game via Lutris from within Steam and I've run into some issues.

It seems that the identifier isn't 100% consistent for service games e.g. Anno 1503 from gog:

| Identifier | Value |
| --------- | ----------- |
| appid      | 1438074791 |
| name      | Anno 1503 A.D. |
| slug      | anno_1503_ad |
| lutris_slug      | anno-1503-ad |

None of those identifiers return an installer, or an error. It turns out the correct identifier here would be `anno-1503`. There are also issues where some of these identifiers aren't set for some services, e.g. `lutris_slug` isn't set for egs games.

The issue is that I'm trying to take values from the `-aj` output to generate Steam shortcut commands, and the working value doesn't match any of the values returned by the service game list CLI.

I have discovered in the code that you can also do something like `lutris:service:appid` but this only works with the `appid` for the service game, though it does seem to work consistently. Unfortunately `appid` isn't currently returned by the `-aj` CLI command.

I've made some code changes myself to add `appid` to the `-aj` CLI output, but also to be more lenient and allow Lutris to try and match the other identifiers above for service games, similar to how you try to find `db_game` with `game_slug`. I've also attempted to fix the error when unable to find an installer for a game, as it currently just hangs if you use an identifier that a game can't be found for via the API.

Would you be willing to take a PR for this? Or discuss it further?

Thanks

culain-45124:
Resolves #5116.

- Adds more options to grab `service_game` so it isn't just relying on the `appid`, now accepts `slug`, `lutris_slug` or `name` as well as `appid`
- Adds `get_game_by_field` function to `services`
- Returns `appid` as part of the json response for list all games for all services CLI command so there is a consistent way to obtain it
- Adds an example of how to specify games with their service for the `lutris` protocol link to the `README`
- Potentially fixes error handling for when an installer isn't returned

Let me know what you think. These changes shouldn't impact the existing functionality.

Apologies for submitting this without prior approval.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -112,6 +112,10 @@ identifier on the command line such as::
 
     lutris lutris:quake
 
+or if you want to specify the game's service too::
+
+    lutris lutris:gog:quake
+
 This will install the game if it is not already installed, otherwise it will
 launch the game. The game will always be installed if the ``--reinstall`` flag is passed.
 
--- a/lutris/api.py
+++ b/lutris/api.py
@@ -257,7 +257,7 @@ def get_game_installers(game_slug, revision=None):
     request = http.Request(installer_url)
     request.get()
     response = request.json
-    if response is None:
+    if response["count"] == 0:
         raise RuntimeError("Couldn't get installer at %s" % installer_url)
 
     # Revision requests return a single installer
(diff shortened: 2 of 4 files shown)


strycore:
I don't think this is acceptable. There is some major confusion in how we treat service games and keep them related to the Lutris database. 

If  this is to make up for the imperfections of our databases then the solution is to improve our database, not write hacks to circumvent our shortcomings.

Issue #5135, opened 2023-11-18
Lutris stores user config'd media in XDG_CACHE_HOME


twowTurtReynolds:
### Bug description

Lutris stores auto-downloaded banners and icons as well as user uploaded ones to XDG_CACHE_HOME

XDG_CACHE_HOME is, as I understand it, meant to be only for temporary content that will be able to regenerate itself. As such, I trash my ~/.cache folder from time to time as it fills up and i need some extra disk space. 

This is where I learn that the cache i cleared and later deleted from trash bin has now lost me all the hard work I did setting up icons and banners. 

Instead, user submitted data should be kept in XDG_DATA_HOME or maybe XDG_CONFIG_HOME, as adding artwork for games is often painstakingly done on purpose by users, and should never be at risk of data loss from cache clears.

Also, if lutris' cache folder is ever cleared, it should start regenerating the artwork upon the first reload. In my case, after a day it still hasn't regenerated.

### How to Reproduce

1. Add a custom game, Upload custom artwork
2. Add a game from lutris website, let it download the artwork
3. turn off lutris
4. delete ~/.cache or XDG_CACHE_HOME 
5. start lutris, see games all there but all artwork gone

### Expected behavior

Preserve users' hard work in downloading editing configuring

### Log output

```shell
~]$ lutris --debug
INFO     2023-11-17 22:32:14,449 [application.do_command_line:436]:Starting Lutris 0.5.14
INFO     2023-11-17 22:32:14,452 [startup.get_drivers:70]:Running AMD Mesa driver 23.2.1 on AMD Radeon RX 580 Series (polaris10, LLVM 16.0.6, DRM 3.54, 6.6.1-arch1-1) (0x67df)
INFO     2023-11-17 22:32:14,452 [display.display_gpu_info:78]:GPU: 1002:67DF 1458:22FC (amdgpu drivers)
WARNING  2023-11-17 22:32:14,511 [libretro.get_platform:116]:Game don't have a core set
WARNING  2023-11-17 22:32:14,511 [game.set_platform_from_runner:421]:The Libretro runner didn't provide a platform for Conker's Bad Fur Day (libretro)
WARNING  2023-11-17 22:32:14,515 [libretro.get_platform:116]:Game don't have a core set
WARNING  2023-11-17 22:32:14,515 [game.set_platform_from_runner:421]:The Libretro runner didn't provide a platform for Dr. Mario 64 (libretro)
[...]

PR #5136, opened 2023-11-18
Automatically sync media at startup- just sometimes

danieljohnson2:
So this PR tries to bring back Lutris's limited self-repair of the media directories. That it, it calls sync_media() from RuntimeUpdater, once per 10 days like in the good old days.

To do this, I had to break a circular reference. I did this by moving RuntimeUpdater in its own .py file.

Also, this PR adds code to notice when Lutris creates its 'banner' or 'coverart' directories in ~/.cache; if Lutris does this, the runtimer updater will notice an run sync_update(). This means if you clear your cache directory altogether (as some people do), Lutris will notice and try to fix it.

It's still quite limited- sync_media() only downloads media for your PGA games, not for any services.

This also does not preserve custom user media when the cache is cleared.

This should help with #5135, but it's not a complete fix.

### DIFF
--- a/lutris/gui/application.py
+++ b/lutris/gui/application.py
@@ -49,7 +49,7 @@ from lutris.gui.installerwindow import InstallerWindow, InstallationKind
 from lutris.gui.widgets.status_icon import LutrisStatusIcon
 from lutris.migrations import migrate
 from lutris.startup import init_lutris, run_all_checks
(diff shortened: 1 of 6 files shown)


strycore:
We are eventually getting rid of the init dialog, so,no. If you want to put sync_media back, it will have to be done at a later stage.

PR #5137, opened 2023-11-18
Allow to select multiple games

### LINKED ISSUE #1711: Allow to select mutliple games
It would be great if you could select multiple games with help of shift and control key.
`Remove` is the most crucial action for multiple games. (e.g. delete all steam games and reimport)
But `Create desktop shortcut`/`Delete desktop shortcut`, `Create application menu  shortcut`/`Delete application menu  shortcut`, `Browse files` and `View on lutris.com` could also be useful for multiple games.

pcoder59:
Adds the ability to select multiple games and remove them by right clicking

Closes #1711


danieljohnson2:
Not a huge fan of how this is done; you've duplicated a whole bunch of code into separate multi/single select code paths.

I sure don't think there should separate signals ("game-selected"/"games-selected"). ContextalMenu.popup() is basically two functions separated by a flag parameters, though the two paths duplicate a lot of identical code.

I think it's okay to have a separate GameActions class, but it *should not* have a different interface; rather the base class ought to accommodate multiple selection somehow.

Nobody should be using 'games_actions.game' directly anyway, so I will correct the master branch so it does not do this anymore.

danieljohnson2:
OK, I've removed the runner contextual menu items code. It was actually dead anyway, those items would never appear, and if we want those items to appear, we can add them in GameActions anyway.

pcoder59:
Will look into into and make the necessary changes.

pcoder59:
Updated the code as recommended by @danieljohnson2 

Removed extra signal games-selected and kept game-selected only
Removed on_games_selection_changed and implemented it in on_game_selection_changed
Removed duplicates and made update_revealer to work with multiple selections
No changes necessary in popup after rebase, removing unnecessary flag that I added
Removed get_multiple_game_actions from base.py and implemented it into get_game_actions itself
Kept separate GameActions class, but removed the second interface as advised. Base class now accomodates multiple selections

lutris/gui/lutriswindow.py
```diff
@@ -996,28 +996,39 @@ def on_sidebar_changed(self, widget):
...
-        if not game:
-            game = {
+        for game_id in game_ids:
+            if self.service:
+                game = ServiceGameCollection.get_game(self.service.id, game_id)
+            else:
+                game = games_db.get_game_by_field(int(game_id), "id")
+            games.append(game)
+        if not games:
+            games = [{
```
danieljohnson2, line 1022: This 'if not games' bit is wrong; this needs to be inside the look, to generate a placeholder for *each* game that is not found. As written I think it never runs, as you will wind up with a list-of-one-element, containing only None, if you select a game that is not found in the database.

lutris/gui/lutriswindow.py
```diff
@@ -996,28 +996,39 @@ def on_sidebar_changed(self, widget):
...
     def on_game_selection_changed(self, view, selection):
+        game_ids = []
+        games = []
         if not selection:
             GLib.idle_add(self.update_revealer)
             return False
-        game_id = view.get_model().get_value(selection, COL_ID)
-        if not game_id:
+        if type(selection) != list:
+            game_id = view.get_model().get_value(selection, COL_ID)
```
danieljohnson2, line 1005: This is over complicated; just use a list here *all the time*. A one element list is perfectly fine when only one game is selected.

lutris/gui/lutriswindow.py
```diff
@@ -996,28 +996,39 @@ def on_sidebar_changed(self, widget):
...
+            game_ids.append(game_id)
+        else:
+            for path in selection:
+                iterator = view.get_model().get_iter(path)
```
danieljohnson2, line 1009: What is 'iterator' for?

lutris/game_actions.py
```diff
@@ -87,14 +93,18 @@ def on_locate_installed_game(self, *_args):
...
-        else:
-            RemoveGameDialog(game_id=self.game.id, parent=self.window).run()
+        for game in self.games:
+            if game.is_installed:
+                UninstallGameDialog(game_id=game.id, parent=self.window).run()
+            else:
+                RemoveGameDialog(game_id=game.id, parent=self.window).run()
 
     def on_view_game(self, _widget):
         """Callback to open a game on lutris.net"""
```
danieljohnson2, line 111: You missed this one; this should be a loop too.

lutris/game_actions.py
```diff
@@ -29,20 +29,26 @@
...
 
 
 class BaseGameActions:
-    def __init__(self, game, window, application=None):
+    def __init__(self, games, window, application=None):
         self.application = application or Gio.Application.get_default()
         self.window = window  # also used as a LaunchUIDelegate
-        self.game = game
+        self.game = games[0]
+        self.games = games
```
danieljohnson2, line 52: Let's not have both self.game and self.games at the same time like this, in the base class.

Instead, set self.game in the sub-classes that support one game only (GameActions, ServiceGameActions)- they can check that there's only one. If there is exactly one, having self.game and self.games at the same time is safe; it won't matter which the code uses.

As is, you'll have code that has been overlooked (like on_game_view below) just targeting a single game, whichever was first in the selection.

lutris/game_actions.py
```diff
@@ -391,3 +401,14 @@ def get_displayed_entries(self):
             "add": self.is_installable,
             "view": True
         }
+
+class MultiGameActions(BaseGameActions):
+    def get_multiple_game_actions(self):
```
danieljohnson2, line 408: This needs to be named get_game_actions to override the base class method, which it absolutely should be doing.

lutris/game_actions.py
```diff
@@ -29,20 +29,26 @@
...
+        return GameActions(games, window, application)
 
     if game.service:
-        return ServiceGameActions(game, window, application)
+        return ServiceGameActions(games, window, application)
 
-    return BaseGameActions(game, window, application)
+    return BaseGameActions(games, window, application)
+
+def get_multiple_game_actions(games, window):
```
danieljohnson2, line 42: Suggest you combine this get_multiple_game_actions() and get_game_actions() into one method that takes a list. Then, if len(games)==1, it does the single-game code, else the multi-case.

lutris/gui/lutriswindow.py
```diff
@@ -537,13 +537,13 @@ def get_service_media(self, icon_type):
...
+    def update_revealer(self, games=None):
+        if games:
             if self.game_bar:
                 self.game_bar.destroy()
-
-            self.game_bar = GameBar(game, self.application, self)
-            self.revealer_box.pack_start(self.game_bar, True, True, 0)
+            if len(games) == 1 and games[0]:
+                self.game_bar = GameBar(games[0], self.application, self)
+                self.revealer_box.pack_start(self.game_bar, True, True, 0)
```
danieljohnson2, line 546: Need an else: around here. If you do not replace self.game_bar, set it to None at the least.

lutris/gui/views/base.py
```diff
@@ -3,15 +3,15 @@
...
+from lutris.game_actions import get_game_actions, get_multiple_game_actions
 from lutris.gui.views import COL_ID
 from lutris.gui.widgets.contextual_menu import ContextualMenu
 
 
 class GameView:
     # pylint: disable=no-member
     __gsignals__ = {
-        "game-selected": (GObject.SIGNAL_RUN_FIRST, None, (Gtk.TreeIter,)),
+        "game-selected": (GObject.SIGNAL_RUN_FIRST, None, (object,)),
```
danieljohnson2, line 14: I think this should be list, not object - if that will work with GTK signals anyway.

lutris/gui/views/base.py
```diff
@@ -28,35 +28,57 @@ def popup_contextual_menu(self, view, event):
...
-            view.select()
-            model = self.get_model()
-            _iter = model.get_iter(view.current_path[0])
-            if not _iter:
-                return
-            col_id = str(model.get_value(_iter, COL_ID))
-
-            game_actions = self.get_game_actions(col_id)
+        game_ids = []
+        if self.view_type == "grid":
```
danieljohnson2, line 32: If you want to check the type of the view, you can use isinstance(self, GameGridView); no need to track object types with a field like this. But it would be better to use methods on the view classes- you can have matching methods on both view classes, that provide the selection in whatever way is required. Polymorphism!

lutris/gui/views/base.py
```diff
@@ -28,35 +28,57 @@ def popup_contextual_menu(self, view, event):
...
+            selection = self.get_selected_items()
+        else:
+            selection = self.get_selection().get_selected_rows()
+            selection = selection[1]
+        if len(selection) < 2:
+            view.current_path = view.get_path_at_pos(event.x, event.y)
```
danieljohnson2, line 38: I think we should handle the single selection and multi-selection with the same code. I don't think we need get_path_at_pos(), not unless there is no selection- and maybe we should not just have a context menu in that case. Having a selection of one, but then opening the menu for *some other game* would be confusing.

I think we could just use the 'else' block code only.

lutris/gui/views/base.py
```diff
@@ -28,35 +28,57 @@ def popup_contextual_menu(self, view, event):
...
+                    contextual_menu.popup(event, game_actions)
+        else:
+            for path in selection:
+                iterator = view.get_model().get_iter(path)
+                game_id = view.get_model().get_value(iterator, COL_ID)
+                game_ids.append(game_id)
+            game_actions = self.get_game_actions(game_ids)
             if game_actions:
-                contextual_menu = ContextualMenu(game_actions.get_game_actions())
+                contextual_menu = ContextualMenu(game_actions.get_multiple_game_actions())
```
danieljohnson2, line 58: No need for a changed method name here. get_game_actions() for the win!

lutris/gui/views/base.py
```diff
@@ -28,35 +28,57 @@ def popup_contextual_menu(self, view, event):
...
+                    db_game = ServiceGameCollection.get_game(self.service.id, game_id)
+                    game = Game.create_empty_service_game(db_game, self.service)
+            elif game_id:
+                game = self.get_game_by_id(game_id)
             else:
-                db_game = ServiceGameCollection.get_game(self.service.id, game_id)
-                game = Game.create_empty_service_game(db_game, self.service)
-        elif game_id:
-            game = self.get_game_by_id(game_id)
+                return None
```
danieljohnson2, line 75: I think we had the empty service games in there to handle games that are in the service (GOG, Steam) but which have not yet been loaded into Lutris as a proper Game; this happens when the game has not yet been installed. You can install it from the menu- that's what this is supposed to enable.

Bring back the empty service games!

lutris/gui/views/base.py
```diff
@@ -28,35 +28,57 @@ def popup_contextual_menu(self, view, event):
...
+            games.append(game)
+        if len(games) == 1:
```
danieljohnson2, line 77: This len check should live in get_game_actions() so it's harder to forget to do it.

lutris/gui/views/grid.py
```diff
@@ -18,6 +18,9 @@ def __init__(self, store, hide_text=False):
         Gtk.IconView.__init__(self)
         GameView.__init__(self, store.service)
 
+        Gtk.IconView.set_selection_mode(self, Gtk.SelectionMode.MULTIPLE)
+        self.view_type = "grid"
```
danieljohnson2, line 22: Le'ts not restate the object type like this.

lutris/gui/views/grid.py
```diff
@@ -97,8 +100,12 @@ def on_item_activated(self, _view, _path):
 
     def on_selection_changed(self, _view):
         """Handles selection changes"""
-        selected_items = self.get_selected_item()
-        if selected_items:
+        selected_items = self.get_selected_items()
```
danieljohnson2, line 103: No, just pass the list to the signal all the time. Don't make the handlers write special checks for list-vs-not-list.

lutris/gui/views/list.py
```diff
@@ -25,6 +25,8 @@ def __init__(self, store):
         Gtk.TreeView.__init__(self)
         GameView.__init__(self, store.service)
 
+        self.view_type = "list"
```
danieljohnson2, line 28: Let's not restate the object type like this.

lutris/gui/views/list.py
```diff
@@ -116,13 +118,13 @@ def set_sort_with_column(self, col, sort_col):
         self.model.set_sort_func(col, sort_func, sort_col)
 
     def get_selected_item(self):
```
danieljohnson2, line 120: Isn't it get_selected_items() now? If not, what happens when multiple items are selected?

danieljohnson2 (commented):
This has improved, but needs some more work. I've given a bunch of detailed feedback; hope that helps.

lutris/gui/lutriswindow.py
pcoder59: It's supposed to generate placeholder if games list is empty after the loop. It's how it is done right now in the master branch too.

danieljohnson2: No, the master branch has no list here; it does the placeholder if the (single) game is not found.

I think that logic should be per game, rather than meaning something like 'if no games are selected, make a placeholder somehow'.

pcoder59: Ok. Will make it work like that then

danieljohnson2: Thank you.

lutris/gui/lutriswindow.py
pcoder59: Turning to only list type is ok for the grid view, but list view returns a single treeiter object, not a list.

lutris/game_actions.py
pcoder59: Did not understand. Uninstalling and removing is already in a loop.

lutris/gui/lutriswindow.py
danieljohnson2: This needs to be consistent. Both versions should generate lists, perhaps lists of iterators or of game-ids. But the same thing whether it is a list or a grid.

lutris/game_actions.py
danieljohnson2: on_view_game() is not a loop, that's what I meant.

lutris/game_actions.py
pcoder59: Ok. Will need to modify all the places where it uses self.game then.

danieljohnson2: That's the idea. Find all the places that should have loops!

lutris/game_actions.py
pcoder59: Oh... Missed that. Will override the parent method instead

lutris/game_actions.py
pcoder59: Tried that. But will need to modify the gamebar code too then. It doesn't pass a list and will cause problems.

lutris/gui/lutriswindow.py
pcoder59: Isn't it already destroyed in the above if?

lutris/game_actions.py
danieljohnson2: Yes indeed, but that should be a minimal change: get_game_actions([game]) instead of ...(game). Easy!

lutris/gui/views/base.py
pcoder59: Again, the list view of games return a treeiter, not a list. Also, llist won't work with GTK signals it seems.

danieljohnson2: It needs to be consistent. I think you used a list sometimes, so I guess you can pass a list even if you say object here.

lutris/gui/views/base.py
pcoder59: Thank you. Will keep that in mind

lutris/gui/lutriswindow.py
danieljohnson2: No. The field 'self.game_bar' is not the same thing as the game bar; it is but a reference to it, and you've left a reference to a destroyed control. That's not a great thing to do; clear the field too, via 'self.game_bar = None'

lutris/gui/views/base.py
pcoder59: Tried to use isinstace(self, GameGridView). GameGridView is undefined in GameView it seems.

lutris/gui/lutriswindow.py
pcoder59: Ok. Will need to work on that

lutris/game_actions.py
pcoder59: Yeah. Did not make all of them loops. Only worked on the code to uninstall/remove. Will make them all loops then.

lutris/game_actions.py
pcoder59: Ok

lutris/game_actions.py
pcoder59: Ok. Will try

lutris/gui/lutriswindow.py
pcoder59: Ok

lutris/gui/views/base.py
pcoder59: Will fix with the above issue for consistancy

lutris/gui/views/base.py
danieljohnson2: Hmm. I see the problem. You can't import GameGridView because that forms a cycle, a circular pattern of imports, and Python does not allow that. I'm used to languages that do.

There are several ways to fix that. The correct way is to move to move the code to the data. Let's say: create a get_selected_game_actions() method on both GameGridView and GameListView, then call it via 'self.get_selected_game_actions()' in GridView. That is a classic refactoring.

This method would do all the selection jiggering and would return the suitable GameActions object.

However, if you find that too challenging to tackle, leave it the way it is and I will sort this out after the merge.

(I'll probably want to make more commands work in the multi-select case, like 'Add to Favories', too. But lets get 'Remove' done first.)

lutris/gui/views/base.py
pcoder59: Is there any change I need to do here? I did not understand. there is Game.create_empty_service_game right?

lutris/gui/views/list.py
pcoder59: Modified already existing get_selected_items() to do the job. Will rename it to item

pcoder59:
Made list view also return only a list
The above eliminates the need for list type checking, removing the overcomplication
Fixed Game placeholder generation
Converted all GameActions to Loops and Removed MultiGameActions class. Will help prevent rewrite all the actions for other actions which will be implemented in the future.
Combined multiple_game actions and get_game_actions. Also modified gamebar to pass only a list here
Set self.game_bar to none if destroyed
Removed restating grid or list view property and Refactored the code to include a get_selected method instead of checking for Grid or List View
Removed unnecessary len check from popup_contextual_menu

lutris/gui/views/base.py
danieljohnson2: I've gone over it again, I think I was a little confused here.

But this logic does abandon the actions altogether if any one game is not resolved. I don't know that this can happen really- it would have to be a non-installed game, but not in a service view.

Still, perhaps it would be better to return None only if *all* games are not resolved, rather than if *any* is. But again, I don't see how this ever happens, so maybe we can just leave it.

lutris/game_actions.py
danieljohnson2: Hmm. I didn't mean that the MultipleGameActions class should go; just loops in the base class. But this resolution is okay; multi-select support for more commands should be easier this way.

lutris/gui/views/list.py
danieljohnson2: Well, list-returning methods should be plural to make it clearer that they return a list. But you've done that; I think there's one stray 'selected_item' variable, no big deal.

lutris/game_actions.py
```diff
@@ -68,154 +72,179 @@ def on_game_stop(self, *_args):
...
-            for game_id in ids:
-                if str(game_id) == str(self.game.id):
-                    return self.game
-            logger.warning("Game %s not in %s", self.game.id, ids)
+        for game in self.games:
+            if game and game.is_db_stored:
+                ids = self.application.get_running_game_ids()
+                for game_id in ids:
+                    if str(game_id) == str(game.id):
+                        return game
```
danieljohnson2, line 244: If this class will support multiple games, this method should return a list, not one game or None.

This is because we'd really like to have a multi-select 'Stop' command, just below.

lutris/gui/views/base.py
```diff
@@ -28,45 +28,48 @@ def popup_contextual_menu(self, view, event):
...
+        return get_game_actions(games, window=self.get_toplevel())
 
     def get_game_by_id(self, game_id):
         application = Gio.Application.get_default()
         game = application.get_running_game_by_id(game_id) if application else None
         return game or Game(game_id)
 
     def get_selected_game_id(self):
-        selected_item = self.get_selected_item()
+        selected_item = self.get_selected()
```
danieljohnson2, line 67: This is not a big thing, but it's better to use plural names for lists, like ```selected_items``` here. Alternately, use the neutral name you picked for the method: ```selected = self.get_selected()```.

lutris/gui/views/list.py
```diff
@@ -145,7 +143,7 @@ def on_row_activated(self, widget, line=None, column=None):
         self.emit("game-activated", selected_id)
 
     def on_cursor_changed(self, widget, _line=None, _column=None):
-        selected_item = self.get_selected_item()
+        selected_item = self.get_selected()
```
danieljohnson2, line 146: Again, ```selected_items``` or ```selected```.

danieljohnson2 (approved):
Looking a lot better. I'd ask for a couple of name changes, and one functional change I'd like- let's make the Stop command a loop too. We're going to want this.

But I would merge even without these changes. Approve!

Now, a bunch of these actions do not support multi-select *well* yet, as they will pop dialog after dialog, one per game. But this does not matter much since user's can't execute them yet. I do not think you should try to rewrite all those dialogs in this PR. We'll deal with them later, if we want to multiple-select for those actions.

All that said, I don't merge new features like this; I leave it to @strycore to decide if he wants them. I think we should have this, but I'll let the boss decide.

pcoder59:
Done. Implemented the changes.

Make get_running_games return list
change selected_item to selected_items

pcoder59:
Resolved conflict with new get_running_games()

strycore:
Looks good to me and it's a much needed feature. Thanks for putting some work into this!

strycore:
Well the code looks good but the feature doesn't work. I don't seem to be able to select multiple games.

danieljohnson2:
I just grabbed it and I can. Maybe you had Lutris running already so that copy just got activated?

Or maybe you were trying to drag-select. It's done by holding control or shift and clicking on each game in turn, not with a drag.

Just guessing, but it works on my machine:tm:.

danieljohnson2:
I've taken a pass over this; I tried to make the behavior more like Nautilus, and also added several more commands that are simple to support- 'Stop', 'Hide',' 'Unhide', 'Add to Favorites' and 'Remove from Favorites'.

These seem to be kinda slow on many games, but they are better than doing them one at a time by hand, so they are a win even if they could be better optimized.

danieljohnson2:
I've been testing and debugging this, and have found an awkward problem- though maybe this should have been obvious to me.

Things like Stop and Add to Favorites work well. I really like that we can now now do ctrl-a, ctrl-break to stop all games without needing a mouse anymore; that was always an issue.

But... the original Remove command pops up a long string of 'do you really want to do it' dialogs. We need a consolidated dialog for this before we ship this, I think.

I'll see what I can do.

strycore:
nevermind, it works now.

danieljohnson2:
I've rigged up a new dialog for uininstalling or removing (or both) multiple games. It's not as rich as one might wish- it does not list the games- but it shares a lot of code with the old dialog.

basxto:
Just a `15 games (50% of installed games)`/`all games` might help a lot to quickly see when you selected more than intended

danieljohnson2:
I think he's asking for feedback on what's *selected*.

Though I'm afraid my quick and dirty confirm-multiple-uninstall dialog does *not* list the games, just gives a count. Maybe that needs to improve.

strycore:
I haven't looked at your new uninstall dialog but you made it sound like it was a separate one from the existing one.
It shouldn't be separate. Uninstalling a single game is just a special case of uninstalling multiple games.

Same goes for categories, we should add multi-game support for the existing dialog (although in this case, displaying the name of the games would look awkward)

danieljohnson2:
OK, I've put together that fancy dialog you wanted and merged it. It looks like this:

image

It has a header bar because of course it does. I stuck the giant count-of-games text into the subtitle, because I always wanted to subtitle something.

This actual offers some bonus functionality- you can chose to take a game with playtime and files out completely (untick 'Keep playtime') rather than having to uninstall and then remove it. That always annoyed me; but the default is as it ever was- delete files, keep playtime.

I tried to keep the existing checks. For instance, if a game uses the same game-directory as another game that you are not removing at the same time, you can get this:

image

There's a message at the top, and the 'Keep files' check is disabled but still ticked.

strycore:
Well now we're exposing our dirty internals to the end user. That's not a way to treat the end user. How confusing must it be to be asked if you want to keep the playtime?

I also think we should use the term "Delete" instead of "Keep" because it's a dialog for deletion and keeping is the opposite of this which requires a context switch from the user. Doubly confusing!

strycore:
Of course, the whole thing wouldn't be necessary if we add library sync with the website to keep track of game playtimes. 
We keep improving the hacks around that instead of actually building the correct thing.

danieljohnson2:
The playtime thing was already hanging out; I think I made it *less* confusing.

Previously, "Removing" a game usually did *not* remove it- the icon would remain, you had to remove it again. That was very confusing. Playtime is something the user can reasonable know about, even if "keep playtime = keep icon" is not really obvious.

I'd be fine with dropping "Keep playtime" if we could address the double-remove issue; if not, I think it's the lesser of two evils- it calls what why "Remove" didn't remove.

I can swap the sense of the check-boxes to "Discard playtime" and "Delete files", if you like that better. No big deal.

basxto:
I mostly asked for the percent and number for bigger uninstalls, where lists become of little use. I wanna remove duplicate and triplicate entries of >200 steam games. So far it was too much of a hassle via UI and I was too lazy to dig directly into the database for that.

strycore:
That sounds more like a bug that should be resolved in the Steam integration than something to be handled by the remove dialog.

PR #5140, opened 2023-11-20
Minor read_setting fixes

m-bdf:
- Fix `read_setting` calls on lines 55-56 of lutris/settings.py
  - Second parameter is not `default` but `section`, so use the `or` operator instead
  - Convert returned value to int (fixes "TypeError: a real number is required, not str")
- Make `read_setting` parameters keyword-only to avoid the first mistake in the future
- Fix or remove keyword-only arguments in every call to `read_setting`
- Fix style by running `make sc` as suggested in CONTRIBUTING.md


danieljohnson2:
There's a whole lot of stylistic change here, not comfortable with merging this. Also, I think your fix for the TypeError bug is unsafe.. or at least confusing.

It would make sense for the DEFAULT_RESOLUTION_HEIGHT and _WIDTH to be ints, but there are *many* places that depend on resolutions being made of strings. They are used to build command lines a lot. I think what we need to do is tolerate strings everywhere, and I've tried to take adjust the code to do this.

We could have ints in some places, strings in others for resolution, as your PR does, but I feel that's extra confusing and I don't like it.

Also, I see the bug where the default value is passed as 'section'; I think the fix here is to make 'section' the last parameter, since it is almost never used. I've done that; that makes DEFAULT_RESOLUTION_HEIGHT and _WIDTH pass the default value as they should, but I put the name in anyway.

PR #5141, opened 2023-11-20
Fix for 5132 - Fixes a failed comparison on game ids when launching v…

ahungry:
…ia KEY_Return


danieljohnson2:
Good work, but I'm not taking this PR. Everywhere else we deal with ids (or compare them) we convert them to *strings*, not ints. There's a case where a slug is used for an ID, and that has to be a string.

I've done a commit to convert IDs to strings where they are a stored in game and store-item objects, so we do not have to convert just before comparisons like this. This should also fix your problem.

PR #5142, opened 2023-11-21
Shifts 'fluidsynth' and 'xdg-desktop-portal' to be 'Recommends'

### LINKED ISSUE #5138: Too strict Debian dependencies
### Bug description

With 0.5.14 two new hard dependencies have been added to the Debian packaging (0f813cdcc24ec86cc549749758304c8d42d5cebb, 470213829de8dc631677445ebbdfdd2076ec6023).
But they don't seem to be actually needed to run Lutris. At least I have not noticed anything broken without having them installed in 0.5.13.

But as they are declared as Depends, the package manager requires that they are installed as well.

### How to Reproduce

1. Install the lutris Debian package (or upgrade an old version to 0.5.14)
2. The package manager attempts to install fluidsynth, xdg-desktop-portal (and their dependencies)

### Expected behavior

The package only has hard dependencies on components that are actually needed for the program to run. Programs/libraries that are only needed for some use-cases should be a softer dependency (Recommends), to give the user the choice if they want to install them. Having them as Recommendation will still install them by default.

See also Debian policy: https://www.debian.org/doc/debian-policy/ch-relationships.html#binary-dependencies-depends-recommends-suggests-enhances-pre-depends

> The Depends field should be used if the depended-on package is required for the depending package to provide a significant amount of functionality.
> ...
> The Recommends field should list packages that would be found together with this one in all but unusual installations.

### Log output

```shell
N/A
```

### System Information

```shell
N/A
```

### Media (optional)

_No response_

danieljohnson2:
'fluidsynth' is not really required, and 'xdg-desktop-portal' appears to just be for flatpak support, so I think we do not need hard dependencies on these.

I am popping a PR for this as I do not feel confident that I understand Debian packaging, still less Flatpak. I'd like to get someone else's eyes on this one, but this change seems correct to me.

Resolves #5138


Boulotaur2024:
Is there a reason we need Fluidsynth at all ? 
I'm just wondering. 
Last I checked Fluidsynth is a "real-time software synthesizer based on SoundFont"
I don't think I need that in my games !

It always runs in the background eating 1% of my CPU (?)
Why would I need this ? Especially since my CPU is not so great
Screenshot from 2023-11-21 21-22-46

strycore:
Fluidsynth is needed for games that have MIDI music. I'm not sure why it would be running in the background though.

Boulotaur2024:
> Fluidsynth is needed for games that have MIDI music. I'm not sure why it would be running in the background though.

I don't know too
It might be a Ubuntu thing... The service is started at boot (I've just come back home from work and checked the process is still running in the background, and yes it is, eating that 1% of my CPU while Lutris is not yet launched

Nevermind I'll just revert to Lutris 0.5.13 for now

Oh and thanks for making Lutris ! (I just checked your profile, saw that you're the author)
Merci beaucoup mon vieux !

strycore:
> Nevermind I'll just revert back to Lutris 0.5.13 for now

This is an extremely bad idea. 0.5.13 is unsupported!

> Merci beaucoup mon vieux !

Avec plaisir :D

at46:
@Boulotaur2024 "sudo systemctl --global disable fluidsynth.service" should disable it.

PR #5146, opened 2023-11-25
Background download for new Wine updates

danieljohnson2:
This is minimal, but may quell the demands to stop updates to the runtime. I think the big issue is that the downloads block Lutris startup.

With this PR, Wine version updates (only!) do not. The Wine updates are big and frequent, and represent the biggest pain point, so I focused on them.

The download is deferred and runs in the background while you play.

1. The archive downloads to ~/.cache/lutris/tmp/whatever in the background, while you blast aliens.
2. The archive extracts to ~/.local/share/lutris/runners/wine/whatever in the background, while you stealth past mooks.
3. We clear the wine version cache, and your new Wine becomes available.

I'm relying on the extraction function to not leave a half-extracted Wine where it can be used; it appears to do this.

There is no UI *at all*. I'm working on that.

### DIFF
--- a/lutris/gui/application.py
+++ b/lutris/gui/application.py
@@ -320,7 +320,7 @@ class Application(Gtk.Application):
         else:
             pci_ids = [" ".join([gpu["PCI_ID"], gpu["PCI_SUBSYS_ID"]]) for gpu in gpu_info["gpus"].values()]
             runtime_updater = RuntimeUpdater(pci_ids=pci_ids, force=self.force_updates)
-            if runtime_updater.has_updates:
+            if runtime_updater.has_updates():
                 init_dialog = LutrisInitDialog(runtime_updater)
                 init_dialog.run()
 
--- a/lutris/gui/dialogs/__init__.py
+++ b/lutris/gui/dialogs/__init__.py
@@ -331,7 +331,7 @@ class LutrisInitDialog(Gtk.Dialog):
         self.show_all()
 
         self.connect("destroy", self.on_destroy)
-        AsyncCall(self.runtime_updater.update_runtimes, self.init_cb)
+        AsyncCall(self.runtime_updater.update_runtimes_at_startup, self.init_cb)
 
     def show_progress(self):
         self.progress.set_fraction(self.runtime_updater.percentage_completed())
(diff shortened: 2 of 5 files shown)


strycore:
So you've taken the responsibility to create the global download queue. 

(We're not doing background downloads)

strycore:
We don't need to restart lutris to use Wine updates.
If that's the case then our code is bugged.

danieljohnson2:
I have no idea what your download queue is supposed to be, if it's not background downloads. I'm just trying to help, not trying to take responsibility for features I don't even understand.

This is a PR, no matter how badly it's needed, because that way you can just close it if it doesn't fit your vision, whatever that may be.

danieljohnson2:
I've spoken to @strycore off GitHub; we spent a million years on the terms "notification" and "download queue", it was great.

It looks like the big issue here is that there's no UI while the download is going on. I dunno when 0.5.15 will drop, but this PR may have to wait until after that, if it needs that much UI. But I think I can do it, given the time.

danieljohnson2:
Since @strycore assures me this will work, I've removed the 'staging' directory altogether, and the notification too. It extract right to the final directory, and the new Wine becomes available to you as soon as its extracted- no restart required.

Also I removed the notification; there's no UI at all. I'm working on a UI separately.

danieljohnson2:
So I've got an early work-in-progress branch for the UI for this. It's no 'queue', but it looks like this:

image

I hope this is more or less what is wanted.

strycore:
Yes, this is great!

danieljohnson2:
The improved UI is not ready yet, but I'll merge that in when it is.

strycore:
It looks like some stuff is sent to the main window while the init dialog remains.
It would be cool to kill the init dialog entirely.

danieljohnson2:
It can easily be made to send everything to the main window (once the UI for that is merged), but is it safe?

I know that the icons that are downloaded after startup won't be used- but it turns out icons downloaded during startup aren't used either, you have to restart for them. I have not been able to fix that.

Are things like DXVK and such safe to install while Lutris runs? If so, I'll do that.

strycore:
if it's not safe then we should fix our broken code.

danieljohnson2:
I don't know if 0.5.15 is going to drop soon. Dare I drop the LutrisInitDialog, and load everything behind the LutrisWindow like that? We may well have broken code... but if you say we should do it, I'll do it.

danieljohnson2:
We have QA?

Oh, you mean *end-users*. Magic of open source and all.

Sure, I'll switch everything over once the progress reporting stuff is ready. By the end of the weekend at latest.

PR #5150, opened 2023-11-28
Linux Runner - Add option to set custom platform per game

bell07:
Added advanced setting to the Linux runner, that allow to set platform per game. 
That is useful if the game is not Linux native but launched by any wrapper or any custom not yet supported emulator.

By the way I removed some not necessary linebreaks for working_dir and ld_library_path definitions


strycore:
Sorry, but this seems too hackish. 
If we make the platform field editable, that doesn't seem to be the right place.

PR #5165, opened 2023-12-03
Allow user to set cache directory in settings and also delete existing cache files

### LINKED ISSUE #3695: Allow user to set cache directory in settings and change the language around it
I have just found myself in the rather unpleasant situation of having my /home partition completely full due to Lutris saving the installer file in /home/user/.cache/lutris. This led to the installation process failing and to me having to download 30+ GB of files _again_.
Now, there is indeed a "cache" option, but it is rather difficult to understand what it does if one doesn't click on it and read the text, which is rather poor UX. I only found it out _after_ the download failed, as it was _not_ clear from the language that this would be the place where the downloaded files would be saved. I also noticed that that option is nowhere to be fond in the general "settings" page, so one can't set it as a general rule.
I would therefore suggest moving the option there or at least making it available there as a general, application-wide option that can be then overridden with the button in the installation phase. I would also suggest renaming the feature from "cache", which has little meaning to most users in this context, to "set installer download path", "save installer to" or similar verbiage, so that the feature is clearer in its scope and definition.

pcoder59:
Adds the ability to set cache folder, i.e. installer download location globally in global settings.

Also added a button that on click will delete all the cached installer files.

Closes #3695

### DIFF
--- a/lutris/config.py
+++ b/lutris/config.py
@@ -9,6 +9,7 @@ from lutris.runners import InvalidRunner, import_runner
 from lutris.util.log import logger
 from lutris.util.system import path_exists
 from lutris.util.yaml import read_yaml_from_file, write_yaml_to_file
+from lutris.cache import save_cache_path
 
 
 def make_game_config_id(game_slug: str) -> str:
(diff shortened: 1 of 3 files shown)


strycore:
The system options is not the place for this setting.
A purge cache option isn't desired.

pcoder59:
> The system options is not the place for this setting. A purge cache option isn't desired.

Why not close the related issue then? What exactly does the issue demand?

strycore:
We're going to add a section in the preferences windows to set paths. We need to make the cache path more visible. 

The feature is wanted, it's just that I don't want a button that potentially erases hundreds of GB of data so easily accessible. Especially now that we have a button to open folder locations in the file browser. It's just as easy to click the browse button, selecting all and pressing delete.

strycore:
If you want to take a shot at it, you can take a look at how UpdatesBox and AccountsBox are implemented. Those are 2 recent additions to the preferences window.

PR #5177, opened 2023-12-08
Hidden Runner Switches

danieljohnson2:
This is a simple feature, but I've wanted this for a while. It's an option to hide runners from the sidebar using switches, similar to what you get in 'Sources'. Like this:

image

You only get switches on installed runners, but that includes the Linux and Flatpak runners and those runners that are found though flatpak. The main reason I want this is because I can't really uninstall these guys, but I don't want them to clutter up my sidebar.

This affects only the sidebar; games can still use hidden runners.

These settings are kept in ~/.config/lutris/lutris.conf; it's meant to be very like the sources switches.

I'm not wholly sold on this UI, since these switches aren't labelled. I did put tool-tips on the switches and buttons. But it resembles the 'Sources' tab, and it does not interfere with the install/remove button.

I also considered presenting this as a 'remove' option on flatpak, linux, and so on, and that would actually just hide the things. But while I want to hide the flatpak provided runners, there is also an option to install the native version instead, and I would not wish to displace that. So, separate switches.

### DIFF
--- a/lutris/gui/config/common.py
+++ b/lutris/gui/config/common.py
@@ -368,7 +368,7 @@ class GameDialogCommon(SavableModelessDialog, DialogInstallUIDelegate):
         """Build a ListStore with available runners."""
         runner_liststore = Gtk.ListStore(str, str)
         runner_liststore.append((_("Select a runner from the list"), ""))
-        for runner in runners.get_installed():
+        for runner in runners.get_installed_runners():
             description = runner.description
             runner_liststore.append(("%s (%s)" % (runner.human_name, description), runner.name))
         return runner_liststore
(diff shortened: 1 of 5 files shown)


strycore:
> I'm not wholly sold on this UI, since these switches aren't labelled. 

I'm not sold either. Something like this should go in the sidebar, via a contextual menu or in the runner options. 

Everyone is going to be confused why there's a toggle button there.

danieljohnson2:
I did think of adding a runner option, but it needs to be a runner option that does *not* appear in a game's 'runner options'. Do we have anything like that now?

Of course, I could do it even if it's the first option like that.

A context menu for this discoverability problems; we do this for games, and we have to have special text that says "No visible games matching '%s' found. Press Ctrl+H to show hidden games."; I see no equivalent for that I could do for runners.

strycore:
> that does not appear in a game's 'runner options'

Where else would it appear then? We also have the ability to hide options with `scope` and `condition`

>  this discoverability problems

This is the kind of feature I don't want to be discoverable. I'm willing to tolerate it as long as it's as hidden as it can be. The ideal being hiding the unwanted runners via editing lutris.conf manually, I would immediately approve this.

danieljohnson2:
OK, I've got a much less discoverable version up, as #5182. Closing this one.

PR #5179, opened 2023-12-10
po: Add Polish translation

mkljczk:


danieljohnson2:
You do not seem to have updated the ``LINGUAS`` file. Did you read [this file](https://github.com/lutris/lutris/blob/master/po/README.md) of instructions about translations?

PR #5182, opened 2023-12-13
Hidden Runner Option

danieljohnson2:
At @strycore request, a new version of the hidden runners... a hidden version of it. Hidden hidden runners. It's meta! Let's look at it:
image
Wait, where is it? Oh, it's way down here:
image
At the very bottom of the "Runner Options", only in the runner config dialog, an only if advanced options are turned on.

When this is turned off, the runner disappears from the side panel even if it is installed or available from flatpak. It's still in the runners drop down in the game configuration though- this is just about the side panel.

Replaces #5177

### DIFF
--- a/lutris/gui/config/runner.py
+++ b/lutris/gui/config/runner.py
@@ -1,5 +1,7 @@
 from gettext import gettext as _
 
+from gi.repository import GObject
+
 from lutris.config import LutrisConfig
 from lutris.gui.config.common import GameDialogCommon
 from lutris.runners import get_runner_human_name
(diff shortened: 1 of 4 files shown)


strycore:
It's pretty weird to have it here. Why not have it in the Runner tab?

danieljohnson2:
It's not controlling the runner, but the side-panel. I'd put it in the "Lutris" group at the top of "System options" if that didn't make it more visible.

However, if you would prefer the Runner tab, I'll try to move it.

strycore:
> It's not controlling the runner, but the side-panel.

It's controlling the runner's visibility in the side panel, it would make more sense to have it as part of the runner options.

With this implementation, it's not clear at all what is visible in the side panel. 
Also, even if it's one option, I would put it in a frame since we're trying to put every option in a frame.

danieljohnson2:
Well, okay, I'll change it.

We can put every option in a frame; every option has a section, but we have logic to not show the frame if there is only one option in it. We just need to remove that, and poof, frames for everybody.

danieljohnson2:
OK, I've moved the option over the to Runner options; makes it much more visible for stuff like the Linux runner, but it is still 'Advanced'.

Also, over in master I've removed the auto-hiding frames. If you wan't everything to have a frame, give everything a section and that should work now. You can see that 'Show in Side Panel' now has a frame in my refreshed screenie above.

danieljohnson2:
Yay!

PR #5183, opened 2023-12-14
Game Starting Animation

danieljohnson2:
Just a little UI nicety- an animation to indicate that a game is starting. It looks like this:

[Screencast from 2023-12-13 18-59-15.webm](https://github.com/lutris/lutris/assets/6507403/71fa16c4-8976-4a01-9e0f-14f5e3987bd3)

This is an exaggeration though, since the pulse animation is usually much shorter- just one cycle. It will continue as long as the game is in the Launching state, but since that often goes by very fast, it will do a minimum of one cycle

I'd like to keep it going longer, until the game is really up and runner, but I can see no way to know when that is. As it is, the Launching state lasts while the game is running the prelaunch stuff, but once the game proper gets launched, it ends.

I think the in-and-out animation looks nice, but part of it is that I can draw the banner *smaller* easily, but growing it out of its cell is harder.

This is really kinda redundant with the text on the "Play" button, which updates to show the same state. But users may not understand that, and this may be more obvious for them.

Resolves #5081

### DIFF
--- a/lutris/game.py
+++ b/lutris/game.py
@@ -706,7 +706,7 @@ class Game(GObject.Object):
         return True
 
     def start_game(self):
-        """Run a background command to lauch the game"""
+        """Run a background command to launch the game"""
         self.game_thread = MonitoredCommand(
             self.game_runtime_config["args"],
             title=self.name,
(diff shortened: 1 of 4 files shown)


strycore:
I like it, let's give it a try

danieljohnson2:
BoingBoingBoingBoingBoingBoingBoingBoingBoingBoingBoingBoingBoingBoingBoingBoingBoing!

PR #5187, opened 2023-12-16
Replace watch_errors decorator with method swizzling

danieljohnson2:
Swizzling. Not monkey patching. No monkeys around here.

This PR replaces the repetitive use of the @watch_errors() decorator with automatic error handling provided by replacement methods for connect, add_emission_hook, idle_add and timeout_add. The DBus stuff is not covered; that appears very complicated to swizzle, and we do not use it much.

The cool/scary part is at the end of exceptions.py, where GLib's methods are replaced with our own. We just call the original after wrapping the callback function in an error handler.

The handler logs the exception and pops an ErrorDialog. Most of the ``on_watched_errors()`` methods are gone; the generic handler works in all cases.. but one.

InstallerWindow needs to use an error *page* not a dialog when the installation has begun, which will not let you go 'Back' afterwards. To support this, the handler looks for special methods like ``on_signal_error()`` and defers to them if found. I am a bit disappointed I had to support this, but it's not an enormous amount of code and we might need it again.

But the good news is that this blocks a lot of potential crashes; we have lots of signal handlers and idle-functions and whatnot, and this provides protection for them all.

*Ook ook*

### DIFF
--- a/lutris/exceptions.py
+++ b/lutris/exceptions.py
@@ -106,6 +106,8 @@ def watch_game_errors(game_stop_result, game=None):
 
 
 def _handle_callback_error(error_objects, error, error_method_name):
+    """Handles an error from a callback. This will call the error method named,
+    or if this cannot be found, it will show an ErrorDialog."""
     first_toplevel = None
 
     for error_object in error_objects:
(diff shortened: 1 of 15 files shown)


strycore:
It sure does look like monkey patching to me but this is also probably the first time this has been done so it's pioneer monkey patching.

PR #5193, opened 2023-12-18
Update el.po

loveisfoss:
Correct a grammatical error in the greek translation.
"Διακοπή" has to be accentuated at the end of the word.


danieljohnson2:
It's not a one-character PR. It's *less than* a whole character. I love it.

Issue #4725, opened 2023-01-26
Option to disable Comptom/Picom for better performance


Swivelgames:
This is to add support for Picom and Compton to `Disable Desktop Effects`, introduced in #881 given the popularity of these compositors.

o-schneider:
I saw that picom/compton handling has been added by reading the code (it wasn't in the changelog!) but it isn't working properly. The function `_get_compositor_commands` in `util/display.py` is called to find which commands to use for both disabling and enabling the compositor. Picom or compton commands will be used if a process with their name can be found. This works fine for disabling, but when the compositor should be re-enabled, there's no picom or compton process, so no restart. The functions used should be kept in memory I guess (and probably rather found at startup or first-run, there's little chance that it changes while lutris is running).

danieljohnson2:
I see what you mean; and I think you are right. I've tweaked the code.

Since you can test this, please download the latest master code and run the 'bin/lutris' script within- no need to install anything. Activate the compositor-disable feature and try it; does it work better now?

o-schneider:
Yup, the fix works! Nice! :)

Though I read the original PR comments and strycore has a point saying this feature is kinda outdated, made me try a bit more picom's `unredir-if-possible` option and it seems to work ok, probably the way to go rather than killing and restarting processes like that.

danieljohnson2:
Yeah, it is outdated. I guess we'll remove it just as soon as we've got it working right!

Thanks for the report though.

PR #5198, opened 2023-12-21
Add compton/picom to disable desktop effects: #4725

ChrisWin22:
Fixes: #4725 

Uses `pgrep` to determine if compton/picom is running.

Stops by running `pkill`

Starts again by starting app again in background

I'm not sure if we care to support compton since it has been completely replaced by picom but I'm not sure if it is actually still used in the wild and since the ticket asked for it I thought I would add it. If we don't want it anymore I can remove it.


strycore:
wow, this  is hackish!

Merging it because it's such a small edge case that it should affect almost no one.

strycore:
Disabling the desktop compositor is such an outdated concept, I wonder why we still do that.

ChrisWin22:
Yeah it is really hacky. I think it picom is still used for window managers. I am using AwesomeWM with Picom and GSync doesnt work unless you disable the compositor. This is basically what I have been doing in a pretty/post script for awhile.

PR #5206, opened 2023-12-30
Retire all use of @lru_cache in favor of a simpler substitute

danieljohnson2:
I don't know that this PR really has a lot of value, but it seemed like an interesting exercise. If you hate having ``lru`` pasted all over everything even when it's not meaningful, well, this PR fixes that.

We've been using ``@lru_cache`` for our simple caching needs, and it is a lot more than we need. We are only caching single values from no-arg functions (almost!).

So, this PR adds ``@cache_single``, a minimal caching decorator that only works on no-arg functions. It has a ``cache_clear()`` attached-function like ``@lru_cache``, but that's the only frill. No ``dict``, no stats, nothing fancy.

It's a drop in replacement for ``@lru_cache`` in every case we have... but one. That one case is ``get_default_runner_version_info``, which takes args. But I believe I've found a workaround- we almost always call it with the same args, ("wine", None) and I've added a no-args function to do that- with``@cache_single``. The one exception is in a place where the lack of caching should not be noticed.

### DIFF
--- a/lutris/api.py
+++ b/lutris/api.py
@@ -1,5 +1,4 @@
 """Functions to interact with the Lutris REST API"""
-import functools
 import json
 import os
 import re
@@ -174,7 +173,6 @@ def format_runner_version(version_info: Dict[str, str]) -> str:
     return version
 
 
-@functools.lru_cache()
 def get_default_runner_version_info(runner_name: str, version: str = None) -> Dict[str, str]:
     """Get the appropriate version for a runner
 
--- a/lutris/gui/config/updates_box.py
+++ b/lutris/gui/config/updates_box.py
@@ -5,7 +5,7 @@ from typing import Callable
 from gi.repository import Gio, Gtk
 
 from lutris import settings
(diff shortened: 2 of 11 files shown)


strycore:
Yeah, I like it. Hot single cache in your area!

strycore:
It's not totally clear why it's `cache_single` instead of just `cache` but at least it's less confusing than using a caching strategy we don't actually need.

danieljohnson2:
It's not ``@cache`` so people do not think it's the one from the standard library, which has more features and requires Python 3.9.

Maybe ``@memoize`` would have had more of an 'I meant to do that!' energy, but it sounds so hoity-toity!

strycore:
> people do not think it's the one from the standard library, which has more features and requires Python 3.9.

But, of the few people who will read the source code, I can pretty much assure you no one will be thinking that :stuck_out_tongue: 

And even if they did, since it does the exact same thing as the standard lib, it's really not a big deal

PR #5234, opened 2024-01-14
Reduce the use of signals in Lutris

danieljohnson2:
Been sitting on this for the release, but I think Lutris uses too many signals. Many are used to essentially call a method, and maybe capture the default application object on the way there.

This PR removes many signals!

- icons-changed

This one is just unused altogether.

- game-launch
- game-install
- game-install-update
- game-install-dlc

These are all used to just run some method on Application; they can be methods on Game itself and we can just call them.

- view-updated

This is just a call to self.update_store(), so we can just call that.

- game-stop

Is a dupe of game-stopped, but fired a little earlier so the running-games list can be updated before use. I've dropped it, and use a method to get the running games that filters out the stopped ones instead.

- service-games-load

This is used to update the UI of the sidebar, which it can do itself. It means the sidebar has to handle service-login instead of LutrisWindow.

- game-start
- game-started
- game-stopped

These are handled via a replacement 'game-state-changed' event, which is raised by a property setter. The handlers wind up with if statements in them, but I think it's okay- the branches are closely related stuff like hiding and showing a sidebar row. Pulling that together seems okay.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -344,7 +344,7 @@ max-branches=12
 max-locals=15
 
 # Maximum number of parents for a class (see R0901).
-max-parents=7
+max-parents=15
 
 # Maximum number of public methods for a class (see R0904).
 max-public-methods=20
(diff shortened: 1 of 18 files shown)


danieljohnson2:
This is too conflicty to merge, but I've cherry picked what I can from it. The rest can be redone later.

PR #5253, opened 2024-01-18
Added the Supermodel Emulator

telanus:
Added the Sega Model 3 Emulator: Supermodel. Needs all it's support files in ~/.local/share/supermodel/ (IE Assets) else it won't work.


strycore:
This is cool, but please note that this runner is 100% unusable because there is no download link.

strycore:
I'm going to disable the runner until we get a link for a build

strycore:
Wait there's no way to disable an individual JSON runner.... 
@telanus please provide a link to a build or this runner will be removed.

PR #5254, opened 2024-01-19
Add 'Recommended' dependencies on app-indicators for DEB and RPM.

### LINKED ISSUE #5249: Tray icon not showing on Ubuntu 23.10 (Wayland session) and incorrect window size.
Lutris does not show the icon in the system tray when running in a Gnome Wayland session.

It threw this critical GTK error when running it from the terminal:

`2024-01-17 22:28:44,003: Starting Lutris 0.5.16
2024-01-17 22:28:44,004: Running AMD Mesa driver 23.2.1 on AMD Radeon Graphics (gfx1102, LLVM 15.0.7, DRM 3.56, 6.7.0-x64v3-xanmod1) (0x7480)
2024-01-17 22:28:44,004: GPU: 1002:7480 1462:5029 (amdgpu drivers)`

`(lutris:41540): Gtk-CRITICAL **: 22:28:56.522: gtk_widget_get_scale_factor: assertion 'GTK_IS_WIDGET (widget)' failed`

Wayland session screenshot:
Captura desde 2024-01-17 22-29-08

X11 session screenshot:
Captura desde 2024-01-17 22-36-41

danieljohnson2:
I am not too confident about packaging, so I'll pop this as a PR and won't just merge it.

The thing is, we offer status icons via AppIndicators or Gtk.StatusIcon. AppIndicators are often not installed, and Gtk.StatusIcon does nothing on Wayland, except to log a strange error. Wayland users are finding the the status icon just doesn't work because they do not happen to have the AppIndicators package installed.

I've already committed code to not even try to use Gtk.StatusIcon except on X11, and to hide the option in Preferences if it won't be possible to implement it. But that reduces the status icon to a magic hidden feature on Wayland- you have to know what package to install. Not discoverable.

This PR should prompt DEB and RPM package managers to install the AppIndicators package we need, if it is available. If not, we should be installable anyway - without the status-icon option, on Wayland.

Resolves #5249

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -46,6 +46,7 @@ Recommends: python3-evdev,
             fluidsynth,
             gamescope,
             gamemode,
+            gir1.2-appindicator3-0.1,
             xdg-desktop-portal,
             xdg-desktop-portal-gtk | xdg-desktop-portal-kde,
 Description: video game preservation platform
--- a/lutris.spec
+++ b/lutris.spec
@@ -31,6 +31,7 @@ Requires:       cabextract
 Requires:       mesa-vulkan-drivers
 Requires:       vulkan-loader
 Recommends:     wine-core
+Recommends:     libappindicator-gtk3
 
 %ifarch x86_64
 Requires:       mesa-vulkan-drivers(x86-32)


strycore:
Not merging this. Our tray icon works. On KDE there is nothing to do, it just work. Gnome has murder fantasies about getting rid of tray icons. You have to jump through hoops to get something working. 

It's not our job to fix this shit. 

Not only that but `gir1.2-appindicator3-0.1` seems to be a very specific package version. Have you checked this is the correct package name for *each and every* deb based distro we currently support?

strycore:
Note that I will change my mind about this if I am shown an example of software using this dependency. Not something Gnome specific, something generic like Steam, Discord or Nextcloud.

strycore:
I will also change my stance if a Gnome developer says "We're sorry for the confusion, here's the official way to support tray icons in Gnome."

strycore:
Am I upset with Gnome? Yes, somewhat. I've been verrrry patient. I've put up with their shit for a loooong time. I try KDE for a minute and realize I can get back everything that has been taken away over the years? This made me ditch Gnome in the fastest way possible.

strycore:
I downloaded the Discord deb and it has

`Recommends: libappindicator1 | libayatana-appindicator1`

strycore:
gir1.2-appindicator3-0.1 isn't a valid Debian package. The closest one would be gir1.2-ayatanaappindicator3-0.1

https://packages.debian.org/search?suite=stable&section=all&arch=any&searchon=names&keywords=gir1.2-appindicator3-0.1

danieljohnson2:
Hmm. I thought we *preferred* app-indicators over Gtk.StatusIcon, even when the later works. But as you like.

strycore:
It's not a matter of preference,  it's a matter that your patch points to a package that doesn't exist

danieljohnson2:
I'd try to fix it, but I think you don't want to do this at all. So we won't.

strycore:
If there is a missing dependency then we should add it. But it seems the vast majority of users have no problem seeing the tray icon. So I guess it's more an edge case of particular setups. And we're not our mission to solve that.

danieljohnson2:
I think the situation is fairly clear: GNOME is against these status icons. GTK4 does not have them at all; GTK3's version only works on X11. Fedora is not going to install AppIndicators by default, *and* you need an extension to even see them if you do install that.

Kinda like Lutris and Virtual Desktops in Wine. I may like this feature, but it is not where Lutris wishes to go. So it is with status icons and GNOME, except I don't like status icons.

For now, our users mostly run X11 or have AppIndicators set up so it works. If they don't, then we just don't have the status icon.

I think it's fine.

strycore:
Or they can use KDE which has status icons without any fiddling, even on Wayland.

When we migrate to Gtk4, we'll have to deprecate the feature anyway.

For the virtual desktops, it's a consequence of using something Proton based. I'm still on the lookout for games that require it.

danieljohnson2:
I expect KDE just ships AppIndicators out of the box. We can continue to use that on GTK4. It's fine.

PR #5274, opened 2024-01-25
Fixed empty context menu error

AinoSpring:
Right-clicking on empty space in list mode caused a
`cannot unpack non-iterable NoneType object` popup.

This was simply fixed by adding a basic `if .. is None` in `lutris/gui/views/list.py`.

This is actually expected behaviour because it is already checked for None at 
`lutris/gui/views/base.py:63`.


danieljohnson2:
Good catch; I'll take this.

Issue #5276, opened 2024-01-26
Ability to quickly obtain game ID for CLI launch


drws:
There seems to be no easy/quick way to obtain Lutris' internal ID of a game, needed for launching from CLI. One has to first create a `.desktop` file from which the ID can then be read. Hopefully I didn't miss some obvious way, but there should be one.

AinoSpring:
Right-click on the game -> open logs
shows the game ID basically at the start of every line.

AinoSpring:
But you are right this feature should be properly implemented.

danieljohnson2:
Might be preferable to use the form "lutris:rungame/<slug>" rather than "lutris:rungameid/<id>"

Confusingly, the slug is what appears as "Identifier" in the game config dialog.

strycore:
Slugs are non unique, we need to show the numerical ID in the game properties while still keeping the presentation elegant (it shouldn't be a read only full width text entry)

strycore:
Fixed

PR #5283, opened 2024-01-27
Closes issue #5276 | showing game id in configure menu

AinoSpring:
Currently a game can be run from the cli using `lutris:rungameid/<id>`.

The problem is that this game id was not easy to get: a desktop shortcut
had to be created which then contained the command including the game id
to run the game.

This has been fixed by adding a `Game ID` field on top
of the configure menu as seen in the image.

image

This contribution would close issue #5276.

### DIFF
--- a/lutris/gui/config/game_common.py
+++ b/lutris/gui/config/game_common.py
@@ -190,22 +190,33 @@ class GameDialogCommon(SavableModelessDialog, DialogInstallUIDelegate):
         return box
 
     def _get_slug_box(self):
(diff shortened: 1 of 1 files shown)


strycore:
* It needs to be in a smaller font to indicate it's not  vital information
* It shouldn't be at the top
* It shouldn't be called Game ID because that makes the "Identifier" field below confusing

I would suggest adding the ID to the Identifier label like "Identifier (internal ID: 1)" with (internal ID) in font size 8 or 9
but other presentations are welcome.

AinoSpring:
I personally would be satisfied with this approach.

image

PR #5305, opened 2024-02-10
A Sluggish Duplicate Game Dialog

danieljohnson2:
This PR extends the 'Duplicate Game' dialog to also ask for a slug, thus:
Screenshot from 2024-02-10 09-23-05

Also, there's a header bar for the duplicate game dialog. Did you ever doubt it would come?

Anyway, the default behavior here is to preserve the old slug even if you change the name; the theory is that in most cases, you are creating a new configuration for an old game, not writing a new game inside Lutris. But you can see this, and can override it if you want to.

### DIFF
--- a/lutris/game_actions.py
+++ b/lutris/game_actions.py
@@ -16,7 +16,7 @@ from lutris.gui import dialogs
 from lutris.gui.config.add_game_dialog import AddGameDialog
 from lutris.gui.config.edit_game import EditGameConfigDialog
 from lutris.gui.config.edit_game_categories import EditGameCategoriesDialog
-from lutris.gui.dialogs import InputDialog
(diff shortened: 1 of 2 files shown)


strycore:
I dont like this

danieljohnson2:
Well, if that's all the feedback I get, then this PR is done. Closing.

strycore:
> Well, if that's all the feedback I get, then this PR is done. Closing.

This was already discussed in the ticket where this feature was suggested. This would only cater to weird users who used Lutris to do weird things while adding to the already clumsiness and inelegance of our UI.

This would be greatly confusing for 99.999% of users, it's just bad UX that ties the internals of our software that should remain internal to the user interface.

PR #5312, opened 2024-02-15
new preference setting for having hidden games always hidden on lutris launch

kiselina2:
## What PR adds:
An option to have hidden games be always hidden when you launch lutris.

## Why:
You forget to hide them before closing lutris, then x time later you launch it while someone sees your screen and sees the "hidden" games, which defeats the purpose of them being hidden.

_let me know, if anything needs to be changed :)_

### DIFF
--- a/lutris/gui/config/preferences_box.py
+++ b/lutris/gui/config/preferences_box.py
@@ -14,6 +14,7 @@ class InterfacePreferencesBox(BaseConfigBox):
         "show_tray_icon": _("Show Tray Icon"),
         "dark_theme": _("Use dark theme (requires dark theme variant for Gtk)"),
         "discord_rpc": _("Enable Discord Rich Presence for Available Games"),
+        "hide_hidden_games_on_launch": _("Always keeps hidden games hidden on lutris launch"),
     }
 
     settings_accelerators = {
--- a/lutris/gui/lutriswindow.py
+++ b/lutris/gui/lutriswindow.py
@@ -272,8 +272,8 @@ class LutrisWindow(Gtk.ApplicationWindow,
         """Load the initial filters when creating the view"""
         # The main sidebar-category filter will be populated when the sidebar row is selected, after this
         filters = {
-            "hidden": settings.read_setting("show_hidden_games").lower() == "true",
-            "installed": settings.read_setting("filter_installed").lower() == "true"
+            "hidden": self.show_hidden_games,
+            "installed": self.filter_installed
         }
         return filters
 
@@ -285,9 +285,10 @@ class LutrisWindow(Gtk.ApplicationWindow,
     def hidden_state_change(self, action, value):
         """Hides or shows the hidden games"""
         action.set_state(value)
-        settings.write_setting("show_hidden_games", str(value).lower(), section="lutris")
         self.filters["hidden"] = bool(value)
         self.emit("view-updated")
+        if not self.hide_hidden_games_on_launch:
+            settings.write_setting("show_hidden_games", str(value).lower(), section="lutris")
 
     @property
     def current_view_type(self):
@@ -324,7 +325,13 @@ class LutrisWindow(Gtk.ApplicationWindow,
 
     @property
     def show_hidden_games(self):
+        if self.hide_hidden_games_on_launch:
+            return False
         return settings.read_setting("show_hidden_games").lower() == "true"
+
+    @property
+    def hide_hidden_games_on_launch(self):
+        return settings.read_bool_setting("hide_hidden_games_on_launch")
 
     @property
     def sort_params(self):
@@ -1063,6 +1070,8 @@ class LutrisWindow(Gtk.ApplicationWindow,
         self.on_settings_changed(None, not state, "hide_badges_on_icons")
 
     def on_settings_changed(self, dialog, state, setting_key):
+        if setting_key == "hide_hidden_games_on_launch":
+            settings.write_setting("show_hidden_games", "false", section="lutris")
         if setting_key == "hide_text_under_icons":
             self.rebuild_view("grid")
         else:


strycore:
@kiselina2  hi, this is question time! Please answer the following questions and I will review your patch.

* Do you like cluttered software? 
* Do you like have 3 gazillion options in your software?
* Did you write this thinking the lutris devs would be totally cool merging this?
* What kind of games are you playing that absolutely need hiding at launch?
* Have you ever used software that had a similar option in your lifetime or a past one?
* Would you be happy if your cornercase was covered and every lutris user in the world had to deal with yet one more option?
* Would you be happy if 10,000 people opened 10,000 cornercase PRs for Lutris and we endend up with 10,000 options?
* What is your opinion on Microsoft Office 365?
* Gnome or KDE?
* Have you ever installed Slackware?
* On a scale from 1 to 10, how well can you make enchiladas?
* Is the game that is supposed to be hidden HuniePop?

danieljohnson2:
I think this has potential! If we have a 'set show-hidden on startup' option, that's just the beginning!

We can then add a 'set set-show-hidden-on-startup on startup' option. But- stay with me here- we can then add 'set set-set-show-hidden-on-startup-on-startup on startup'.

The possibilities are endless!

More seriously, the strikes me as being like 'very hidden' columns in Excel; hidden, but also the normal method of showing hidden doesn't show them. You have go through extra dialogs to do it.

It might be useful for... certain games. You know. *Those*. But you would not want a global switch like this, it would be for specific games.

kiselina2:
Well, I just thought that since there is the option to hide a game, that it is something an average user wants for whatever reason, but since every time they have to unhide the game and then **dont forget** to hide it again once they start it, that that defeats the purpose of having that option even there. I would be more that happy with there not being another preference option and the games being hidden on startup always. But since there is probably a large amount of people already accustomed to this behavior I didnt want to change it and instead opted to add the option to change it. I am very sorry I made you guys react this way.

kiselina2:
@strycore 
> What kind of games are you playing that absolutely need hiding at launch?
> Is the game that is supposed to be hidden HuniePop?

NGL kinda weird you are judging me/ making fun of me, when I tried to make a feature **already in the program** more usable, but because its a hidden games feature, you had to make these comments.

strycore:
> NGL kinda weird you are judging me/ making fun of me, when I tried to make a feature already in the program more usable, just because its a hidden games feature.

I'm not judging you based on the types of games you play. I'm judging people who think they can add every problem by adding a checkbox to an UI. 
We don't want more checkboxes. We don't want more options. We want to make software that works the right way by default with no user intervention.
I'm more than open to discuss how we handle hidden hidden but the checkbox you've added in your patch crosses the boundaries of what's acceptable in an UI. 
**This** is the reason we are poking fun at this PR, it really has nothing to do with what games you play.

I personally use the hide game feature to hide games/tools that don't have a coverart. I would not be mad if we just removed "show hidden games" as a settings and toggling it on and off would only affect the current session. We don't have any visual indicator showing that we are displaying hidden games other than the checkbox in the view menu which isn't open when "Ctrl+H" is used. If we removed the setting, then I wouldn't be left wondering if I have hidden games displayed or not, I would just expect them to be hidden by default.

@danieljohnson2 what do you think? should we remove the "show_hidden_games" setting?

danieljohnson2:
I think removing the saved 'show_hidden_games' setting would be a good compromise- get @kiselina2's feature by making Lutris simpler rather that more complicated.

But I would rather find out what problem we're trying to solve. I fear a yak shaving exercise here.

strycore:
Steam treats hidden games as a special category (like we have the favorites category). Maybe we should convert "hidden" to a category instead of having a boolean on the game table.
image

strycore:
Ubisoft Connect treats hidden games as a category

image

danieljohnson2:
Hmm. That might not address @kiselina2's concern, unless we go one further. We'd need to make Lutris *not save* that you're in the Hidden view, so that when you start Lutris up again you aren't seeing the Hidden games right away.

Might address your issue- you'd know if you were viewing hidden games more easily.

strycore:
GOG Galaxy also treats hidden games as a category:

image

>  We'd need to make Lutris not save that you're in the Hidden view

I wouldn't even put the Hidden category with the rest of the categories. In Steam, you have to find a separate menu entry to access hidden games. This, of course, should never be saved.

danieljohnson2:
Hmm. I don't understand how you view your Hidden games; would the checkbox for 'show hidden games' now show and hide the entire category, ie, the view option in the sidebar? Or would there be no such view option?

strycore:
> Hmm. I don't understand how you view your Hidden games; would the checkbox for 'show hidden games' now show and hide the entire category, ie, the view option in the sidebar? Or would there be no such view option?

That would be more in line with Steam's behavior. It would indeed stop being a checkbox and just let you see hidden games. This would remove the ability to see regular games + hidden games but it would let users know which ones of their games are hidden. Currently, with my library of 2400 games (yes, I have more games now that I implemented library sync) I have no way of knowing which games are hidden.

I'm also going to implement category syncing and having hidden as a category would make it easier to deal with syncing hidden games (I could add a hidden boolean to the LibraryGame model on the website but I don't wanna)

danieljohnson2:
Sounds like you've got your own reasons to want this, then. Do you want me to take a stab at it, or would you rather handle it yourself?

strycore:
> Sounds like you've got your own reasons to want this, then. 

I don't have any other reason than wanting to have a simple to maintain system. 

> Do you want me to take a stab at it, or would you rather handle it yourself?

It's not a priority for me right now so feel free to work on this.

danieljohnson2:
OK, I've got a PR for you, #5314. I think it came out well.

The 'Show Hidden Games' command takes you to a new view of only hidden games, which be visible only while selected, and will disappear again at once when you leave it. It will not be re-selected on start-up in any case. I think that should handle @kiselina2's desire as well.

One concern with it is that the code to test if any hidden games exist might slow down Lutris start-up for a large library. If so, we can remove the test and just leave 'Show Hidden Games' enabled at all time.

PR #5314, opened 2024-02-17
Convert the 'Hidden' flag to a category

danieljohnson2:
This PR does two things; it migrates the current 'hidden' flag on games into a category named ".hidden"- this should provide technical advantages for the Lutris website sync feature.

Also, this PR provides a more category-like UI for the beast.

You still hide games with the context menu, but the hamburger menu looks a little different:
Screenshot from 2024-02-17 08-33-51
No more checkbox. I moved it down one item to keep the remaining checkboxes together, just for cosmetics.

When you click 'Show Hidden Games', you get this:
Screenshot from 2024-02-17 08-34-03
The hidden category is revealed! Only hidden games appear, which makes finding your hidden games easy. You can use multi-select to unhide them all. I love it when a plan comes together!

Selecting any other view will make this hidden category disappear again at once. Also, when you start Lutris it will refuse to select a view that is hidden- this new one, and also this applies to the 'Running' view. Instead it will select the first visible one, which should be 'Games'. Hidden means hidden!

This PR includes a migration for the hidden flag, and drops the 'hidden' column from the schema, but does not try to drop the column from existing databases- doing so would break the migration. If that column is missing, however, the migration just does nothing.

Resolves #5312

### DIFF
--- a/lutris/database/categories.py
+++ b/lutris/database/categories.py
@@ -33,23 +33,30 @@ def get_category(name):
 
 def get_game_ids_for_categories(included_category_names=None, excluded_category_names=None):
     """Get the ids of games in database."""
-    query = (
-        "SELECT games.id FROM games "
-        "LEFT OUTER JOIN games_categories ON games.id = games_categories.game_id "
-        "LEFT OUTER JOIN categories ON categories.id = games_categories.category_id"
-    )
-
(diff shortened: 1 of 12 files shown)


kiselina2:
nice

PR #5332, opened 2024-02-28
Add .desktop file's path variable to the working directory field on local source games

### LINKED ISSUE #5325: Lutris can not launch a game through a .desktop file
Hello, first of all thank you for the amazing work you have done in the linux gaming scene.

When Lutris imports a game from the local source (a desktop file in .local/share/applications), it can not be launched.
I am using ULWGL for launching the game through the desktop file.

[Desktop Entry]
Type=Application
Categories=Game
Name=Grim Dawn
Icon=/usr/share/icons/Papirus-Dark/16x16/places/folder-games.svg
Path=/HDD/GranAlmacen/Juegos/Grim Dawn
Exec=env GAMEID="" WINEPREFIX="/HDD/GranAlmacen/Juegos/Grim Dawn/wine" ulwgl-run-cli "x64/Grim Dawn.exe"

Cheers !

Aqa-Ib:
This is working. Telling the truth I don't write python code, so I had help from chatgpt, so review the code before committing.
Fixes #5325

### DIFF
--- a/lutris/services/xdg.py
+++ b/lutris/services/xdg.py
@@ -115,6 +115,7 @@ class XDGService(BaseService):
                 "game": {
                     "exe": details["exe"],
                     "args": details["args"],
+                    "working_dir": details["path"],
                 },
                 "system": {"disable_runtime": True},
             },
@@ -152,13 +153,26 @@ class XDGGame(ServiceGame):
         service_game.appid = get_appid(xdg_app)
         service_game.slug = cls.get_slug(xdg_app)
         exe, args = cls.get_command_args(xdg_app)
+        path = cls.get_desktop_entry_path(xdg_app)
         service_game.details = json.dumps(
             {
                 "exe": exe,
                 "args": args,
+                "path": path,
             }
         )
         return service_game
+
+    @staticmethod
+    def get_desktop_entry_path(xdg_app):
+        """Retrieve the Path variable from the .desktop file"""
+        desktop_entry = xdg_app.get_filename()
+        with open(desktop_entry, 'r') as f:
+            contents = f.read()
+            match = re.search(r'^Path=(.*)$', contents, re.MULTILINE)
+            if match:
+                return match.group(1)
+        return None
 
     @staticmethod
     def get_command_args(app):


strycore:
This is such an ugly solution, probably explains why I didn't want to implement it in the first place. The rest of the code uses the Gio.AppInfo class and for this one bit we have to resort to reading the .desktop file to make up for the shortcomings of Gio. 

We might as well consider not using the Gio.AppInfo class at all and build our integration by manually parsing .desktop file. If we're doing that anyway why not go all the way.

Anyway. Thanks for the patch. Ugly and working is better than good looking and not working. It's not like the XDG integration gets a lot of use anyway.

danieljohnson2:
Well, I don't like the use of regular expressions for this. I think we should call ``xdg_app.get_string("Path")`` rather than reading the file. It would look like this:

```python3
    @staticmethod
    def get_desktop_entry_path(xdg_app):
        """Retrieve the Path variable from the .desktop file"""
        return xdg_app.get_string("Path")
```

Can you make that change?

danieljohnson2:
I guess not; @strycore just merged this. I'll change it in master.

strycore:
> I think we should call xdg_app.get_string("Path") rather than reading the file.

Where is this documented?

danieljohnson2:
I found it [here)(https://lazka.github.io/pgi-docs/Gio-2.0/classes/DesktopAppInfo.html)

But let me see if I can find something more official.

danieljohnson2:
I think [this](https://docs.gtk.org/gio/method.DesktopAppInfo.get_string.html) is the official docs.

strycore:
How did you figure out that this was the method used to get the Path value? Do you have some kind of divination power?

danieljohnson2:
It is not specifically for the path value, it pulls out one of the values by key. I found no more specific method than this; but at least it beats a RegEx on the file content.

strycore:
> I think [this](https://docs.gtk.org/gio/method.DesktopAppInfo.get_string.html) is the official docs.

Gtk really has the worst documentation any API has ever had. I'm so ashamed of using a toolkit that is that bad.

danieljohnson2:
Yeah, it's not great that way. It's painfully short of samples, and the references docs aren't 100% complete either. And they keep changing it.

danieljohnson2:
I've updated this in master. Thank's for the PR @Aqa-Ib.

PR #5338, opened 2024-03-01
Partial adding types to utils folder

qarmin:
Should decrease mypy errors by ~200/300

I found some possibly invalid types and I marked this with `TODO MYPY` so this needs to be fixed in different PR

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,7 +91,7 @@ snap:
 
 req-python:
 	pip3 install PyYAML lxml requests Pillow setproctitle python-magic distro dbus-python types-requests \
-	 types-PyYAML evdev PyGObject pypresence protobuf moddb
+	 types-PyYAML evdev PyGObject pypresence protobuf moddb webdav4
 
 dev:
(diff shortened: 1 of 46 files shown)


danieljohnson2:
@strycore is not fond of all the ``Optional[str]`` stuff, so I turned on implicit_optional so we can do optional parameters without it; ``p: str = None`` is understood as ``p: Optional[str] = None``.

I see you've put accurate types on all the parameters like that, which is more correct in a way, but noisier.

qarmin:
I started adding them out of habit.

I'm not sure, but adding them rather allowed more errors to be displayed in mypy(but it is also possible that adding the types to all the parameters allowed mypy to find them better)

I may change them back to var: type = None, but this seems less readable to me than the current solution

strycore:
It's fine to have some Optional stuff. Maybe if we have too many of them, it's a sign we should write better, more deterministic code.

strycore:
damn, I was about to merge this and there are so many conflicts

PR #5344, opened 2024-03-04
Fix shortcuts for launch_config options

MadByteDE:
There where multiple issues with shortcuts for `launch_config` options:
- The `name` tag didn't change for any `launch_config` and displayed the main `game_name` instead
- `Exec` links where broken:
```py
Exec=env LUTRIS_SKIP_INIT=1 lutris lutris:rungameid/1/Game%20-%Config
```
instead of : 
```py
Exec=env LUTRIS_SKIP_INIT=1 lutris lutris:rungameid/1/Game - Config
```
- the `game.name` var not being passed correctly in `game_actions.py`, resulting in the `name` tag missing which caused the shortcut to display the file name (e.g `net.lutris.game-config-1.desktop`) instead of the actual game name.

### DIFF
--- a/lutris/api.py
+++ b/lutris/api.py
@@ -501,7 +501,6 @@ def format_installer_url(installer_info):
     if launch_config_name:
         parts.append(launch_config_name)
 
-    parts = [urllib.parse.quote(str(part)) for part in parts]
     path = "/".join(parts)
 
     if revision:
--- a/lutris/game_actions.py
+++ b/lutris/game_actions.py
@@ -333,7 +333,7 @@ class SingleGameActions(GameActions):
         game = self.game
         launch_config_name = self._select_game_launch_config_name(game)
         if launch_config_name is not None:
-            xdgshortcuts.create_launcher(game.slug, game.id, launch_config_name, menu=True)
+            xdgshortcuts.create_launcher(game.slug, game.id, game.name, launch_config_name, menu=True)
 
     def on_create_steam_shortcut(self, *_args):
         """Add the selected game to steam as a nonsteam-game."""
--- a/lutris/util/xdgshortcuts.py
+++ b/lutris/util/xdgshortcuts.py
@@ -68,7 +68,7 @@ def create_launcher(game_slug, game_id, game_name, launch_config_name=None, desk
         Icon={}
         Exec=env LUTRIS_SKIP_INIT=1 {} {}
         Categories=Game
-        """.format(game_name, "lutris_{}".format(game_slug), lutris_executable, shlex.quote(url))
+        """.format(launch_config_name or game_name, "lutris_{}".format(game_slug), lutris_executable, shlex.quote(url))
     )
 
     launcher_filename = get_xdg_basename(game_slug, game_id)


danieljohnson2:
Well, we should fix that missing ``game.name`` parameter, but the rest of this does not seem so good to me.

This PR produces a ``lutris:`` that does not escape things in the URL; that'll work until it doesn't. Spaces are okay, but if the launcher option name contains ``/`` things won't be so great. Shell quoting is not a complete solution.

Also, this changes the name in the shortcut to the launcher option name, and these are often unhelpful things like 'Settings' or 'GameTask' or something like that. In any case, Lutris only makes one shortcut for the menu and one for the desktop, *not* one per launcher option. I think the game's own name is better for this.

lutris/api.py
```diff
@@ -501,7 +501,6 @@ def format_installer_url(installer_info):
     if launch_config_name:
         parts.append(launch_config_name)
 
-    parts = [urllib.parse.quote(str(part)) for part in parts]
```
MadByteDE, line 504: Here we might need some code to check for specific runners. I'm unsure if this format might be needed by shortcuts for other runners then `wine` and `steam` to properly work. *.desktop files break when this code is used.

danieljohnson2: This should not be runner specific; the thing that needs to be quoted is the launcher option name. We don't want that to be understood as multiple path segments in the URL.

I do not understand the bug here; we should be unquoting everything before use.

MadByteDE: > This should not be runner specific; the thing that needs to be quoted is the launcher option name. We don't want that to be understood as multiple path segments in the URL.
> 
> I do not understand the bug here; we should be unquoting everything before use.

You're right. 

The bug is that *.desktop files currently won't work when we create a shortcut of an launch config option using the `LaunchConfigSelectDialog`.

MadByteDE:
You're probably right.

I'm using the config options for a local `installer script` because it's the only way to create multiple shortcuts using a single installer - and for the game to work it's necessary to launch a server before the actual game is launched. 
Sharing a single file installer makes it easier for ppl not wanting to manually configure all workarounds for a game by themself.

[Example Gist](https://gist.github.com/MadByteDE/a3eafbf713f78dd3c0bdeca376e82cb4)

In this use case it's helpful to have named config shortcuts, even if it might be a bit out of bounds for what they are usually used for. 

Either way, feel free to close any time. :+1:

danieljohnson2:
OK, I think I've figured out the broken shortcuts. First, it wasn't really escaping ``/``; that's apparently considered safe by default. Second, desktop files consider % special on their own, and that character needs to be escaped by doubling it.

So if you have a change, grab the latest master code and try it to see if it creates better shortcuts now. No need to install, just run the bin/lutris script inside.

MadByteDE:
> OK, I think I've figured out the broken shortcuts. First, it wasn't really escaping `/`; that's apparently considered safe by default. Second, desktop files consider % special on their own, and that character needs to be escaped by doubling it.
> 
> So if you have a change, grab the latest master code and try it to see if it creates better shortcuts now. No need to install, just run the bin/lutris script inside.

Unfortunately it's not fixed yet.
```py
Exec=env LUTRIS_SKIP_INIT=1 lutris lutris:rungameid/1/BSG%%20Launcher
```

danieljohnson2:
Hmm. It works on my machine!

Try running this from the terminal:
```
lutris lutris:rungameid/1/BSG%20Launcher
```
Only one % there. Does that command work at least?

MadByteDE:
yeah works just fine, sorry I'm tired. Thanks for the quick fix.

danieljohnson2:
OK. Well,it works from the terminal but not in the .desktop file. Mysterious.

Maybe edit the file to remove the double %% and see if that works?

MadByteDE:
No, no desktop file as well.. all fine now

danieljohnson2:
OK. I think we can declare victory then.

PR #5365, opened 2024-03-19
Update ULWGL paths

R1kaB3rN:
The current model expects clients to install ULWGL as a system package (e.g., flatpak, system package) then execute, and this patch fixes the problem where the path ~/.local/share/ULWGL was the first path candidate which would cause lutris to run outdated files while the updated files were pulled from the runtime.

To ensure the ULWGL launcher and runtime files are synchronized or updated, lutris should never starts ulwgl_run.py in ~/.local/share/ULWGL and that lutris prioritizes the system package then the lutris runtime directory. 

However, depending on the team's preference, the runtime directory can be preferred to account for distributions or users that do not update their system packages frequently.

### DIFF
--- a/lutris/util/wine/proton.py
+++ b/lutris/util/wine/proton.py
@@ -24,14 +24,14 @@ def get_ulwgl_path():
     if system.can_find_executable("ulwgl-run"):
         return system.find_executable("ulwgl-run")
     path_candidates = (
-        os.path.expanduser("~/.local/share"),
+        "/app/share",  # prioritize flatpak due to non-rolling release distros
         "/usr/local/share",
         "/usr/share",
         "/opt",
         settings.RUNTIME_DIR,
     )
     for path_candidate in path_candidates:
-        script_path = os.path.join(path_candidate, "ULWGL", "ulwgl_run.py")
+        script_path = os.path.join(path_candidate, "ulwgl", "ulwgl_run.py")
         if system.path_exists(script_path):
             return script_path
 


lutris/util/wine/proton.py
```diff
@@ -24,17 +25,20 @@ def get_ulwgl_path():
...
-        os.path.expanduser("~/.local/share"),
+        "/app/share",  # prioritize flatpak install due to non-rolling release distros
         "/usr/local/share",
         "/usr/share",
         "/opt",
-        settings.RUNTIME_DIR,
     )
     for path_candidate in path_candidates:
-        script_path = os.path.join(path_candidate, "ULWGL", "ulwgl_run.py")
+        script_path = os.path.join(path_candidate, "ulwgl", "ulwgl_run.py")
```
R1kaB3rN, line 34: For distributions that build via the Makefile, the directory will be all lowercase.

See https://github.com/Open-Wine-Components/ULWGL-launcher/blob/main/Makefile.in#L75

lutris/util/wine/proton.py
```diff
@@ -24,17 +25,20 @@ def get_ulwgl_path():
...
         if system.path_exists(script_path):
             return script_path
 
+    script_path = os.path.join(settings.RUNTIME_DIR, "ULWGL", "ulwgl_run.py")
```
strycore, line 38: no need to have a special case for the runtime dir, I can re-upload ULWGL in lower case

danieljohnson2:
@strycore asked me to merge this for him, so here we go!

PR #5375, opened 2024-03-22
Added Wezterm terminal support.

CrynTox:


strycore:
The Lutris terminal selection dropdown. Where people can learn about new terminals every day!

Issue #5339, opened 2024-03-02
[Feature Request] "Uncategorized" category


hirak99:
Now that we are able to assign categories to games, it will be really nice if we have an "Uncategorized" automatic category.
This will help us to know exactly which games we miss when we only look at games in the categories, which right now is not very easy to do.

(The idea can be extended in the future, to have automatic categories definable by the user based on some query.)

strycore:
> The idea can be extended in the future, to have automatic categories definable by the user based on some query.

I would rather go for that feature first.

dobval:
Hey, I find this idea pretty useful too. I think I managed to implement something along the lines of it. After every successful installation, the game gets added into "Uncategorized". Let me know if there are any bugs

hirak99:
Thanks! I tested it, it works when installing games - have a couple of observations -

1. It does not work when a game is added in any other way. Say for example if an already installed game is re-added. Thus it wouldn't work on locally installed games.
2.  This also means that previously added games (that are already in the library) will not get the label Uncategorized. This can make this commit a bit confusing for existing users.
3. It wouldn't recategorize if a game is manually put in a category. In other words, the "Uncategorized" category has to be manually removed. This can be confusing too.

I think this commit makes a lot of sense if the category name is changed to "Installed" instead of "Uncategorized"? And it should perhaps come with an option to turn on or off.

For Uncategorized on the other hand, I think it should be a smart category. By smart I mean the category always keeps track of all games that are _currently_ not in any other categories. This is also how Steam's Uncategorized works - so it will make Steam users feel right at home.

hirak99:
I think it's actually fairly easy to add an UNCATEGORIZED smart-category, that stays up to date automatically.

All the code can be encapsulated in categories.py.

I made this change, and invite you to try it here - https://github.com/hirak99/lutris

danieljohnson2:
I think if you are going to have a hard-coded "Uncategorized" category it should not present with the user-defined categories; it should be up in the top section with "Recent". And have a cool icon.

If you can make these things user-defined, that might be a bit different. But then you would want real category IDs instead of fake ones.

I think @strycore would prefer the user-defined approach, but I do not see how to define these things without users keying in SQL (which we should avoid).

hirak99:
Good points on UX, mostly agree, and here are some of my thoughts -

[...]

PR #5394, opened 2024-03-31
"Uncategorized" automatic creation of category on game install

dobval:
#5339 had the cool idea, also ran "ruff format" according to CONTRIBUTING.md

### DIFF
--- a/lutris/api.py
+++ b/lutris/api.py
@@ -1,4 +1,5 @@
 """Functions to interact with the Lutris REST API"""
+
 import json
 import os
 import re
--- a/lutris/cache.py
+++ b/lutris/cache.py
@@ -1,4 +1,5 @@
 """Module for handling the PGA cache"""
+
 import os
 import shutil
 
--- a/lutris/database/categories.py
+++ b/lutris/database/categories.py
@@ -2,6 +2,7 @@ import abc
 import re
 from collections import defaultdict
 from itertools import repeat
+from typing import List, Dict, Union
 
 from lutris import settings
 from lutris.database import sql
@@ -15,7 +16,7 @@ class _SmartCategory(abc.ABC):
         pass
 
     @abc.abstractmethod
-    def get_games(self) -> list[int]:
+    def get_games(self) -> List[int]:
         pass
 
 
@@ -25,7 +26,7 @@ class _SmartUncategorizedCategory(_SmartCategory):
     def get_name(self) -> str:
         return "UNCATEGORIZED"
 
-    def get_games(self) -> list[int]:
+    def get_games(self) -> List[int]:
         query = (
             "SELECT games.id FROM games "
             "LEFT JOIN games_categories ON games.id = games_categories.game_id "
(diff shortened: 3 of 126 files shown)


danieljohnson2:
I've some thoughts about this.

First, oh god look at all the reformatting. ``ruff format .`` doesn't do anything like this much for me. What happened?

Second, I'd like to suggest that your new category is not named right; it should be "Newly Installed"; games remain in the category no matter how many other categories they are put in. If we really want "Uncategorized" we'd need a very different and much more dynamic implementation (Like "Recent" has).

And it is "Newly Installed" because it covers only what is installed via our install scripts; if you add an existing game it will not be included.

And I think there's a further thing here- you can have a game that you can categorized, but which goes into the new category anyway because you have just installed it. Games that are not installed can be categorized- in the new Lutris release, with library sync on, everything in your Lutris library now appears in "Games" and *can be categorized*. But they only get put into your new category when you install them, I think.

If you really want "Uncategorized", you need to take a very different approach. But maybe this behavior is what you want?

dobval:
I agree with the suggestion of "Newly installed" and I think a better functionality for "Uncategorized" would be, as mentioned, a dynamic category that automatically includes all the games without a user-set category in order to help list any games that a user might have missed, but still wants to categorize all of them. Any thoughts? Should I only implement a dynamic "Uncategorized" or both with "Newly Installed" About `ruff format`, weird behaviour but I can revert all the formatting if needed.

strycore:
I've pushed a commit that reformats the whole project with the latest version of Ruff. The project had previously been reformatted with Ruff 0.2.2, that's why the change (mostly whitespace) was so large.

That being said, having to manage an "uncategorized" category will end up being a nightmare and it certainly not compatible with library syncing on the Lutris.net website. Please update your PR to use @hirak99 's suggestion which doesn't create a database entry for games without a category.

dobval:
Alright, later today I'll add @hirak99's suggestion and change the hard-coded entry to "Newly Installed".

strycore:
There is no need for a hard-coded entry. If we have dynamic filter, we can query all kinds of stuff. 
While we don't have  a custom UI for this kind of filtering, I think we can easily add that functionality in the search bar. 

Some examples of queries we could support:
 - installed at: last 2 months  (show games installed in the past 2 months)
 - runner: mame (show games using the mame runner)
 - category: none (show games without a category)
 - last played: 2022 (show games last played in 2022)
 - last played: >2022 (show games played after 2022)
 - year: <2007 (show games released before 2007)
 - doom year: <1996 runner: gzdoom last played: last 6 months (show games matching the string "doom" released before 1996 using the gzdoom runner last played in the past 6 months)

dobval:
I managed to add an option in Preferences>Interface to toggle "Uncategorized", looks good.

danieljohnson2:
A whole lot of reformatting noise still. Can you rebase this onto the current master? Might help us read it.

strycore:
Yes, rebase against the master branch. Your PR should be changing 3 files at most. 
Also, don't create an option to disable the feature you just implemented. You wanted it, you own it.

strycore:
Closing this PR. It will be simpler to just create a new, cleaner one.

PR #5399, opened 2024-04-03
Added Uncategorized SmartCategory

dobval:
No switch button to disable the option


danieljohnson2:
Surprised to see this merged. Now there's always an "Uncategorized" category in there; I thought we wanted to have fewer sidebar rows. I'd started working on a search-based alternative to this.

danieljohnson2:
I'll go ahead and fix this up a little; it's a fixed category, so it should have an icon and be in the top section. And it should have a reserved name so it won't collide with user-defined categories.

PR #5400, opened 2024-04-04
Kill wineserver via UMU Proton

R1kaB3rN:
Kills the wineserver process associated with the running prefix via the `wineserver` command in the UMU-Proton directory instead of the `wineboot` command, which had been reported to lead to unreliable results and had relied on the system wineboot. 

The user's current UMU-Proton directory will be resolved via the UMU-Latest symbolic link, which is guaranteed to exist for new installations since https://github.com/Open-Wine-Components/umu-launcher/commit/fd1798f75c80c001bc51f9b1acd6465698333e63. However, umu currently does not recreate this link each run, so the new kill implementation will not work for current UMU users. There will be patch for this shortly, to restore the link each launch.

### DIFF
--- a/lutris/runners/commands/wine.py
+++ b/lutris/runners/commands/wine.py
@@ -178,11 +178,14 @@ def winekill(prefix, arch=WINE_DEFAULT_ARCH, wine_path=None, env=None, initial_p
     """Kill processes in Wine prefix."""
 
     initial_pids = initial_pids or []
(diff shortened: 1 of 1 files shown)


R1kaB3rN:
The patch to recreate the UMU-Latest link has been merged in https://github.com/Open-Wine-Components/umu-launcher/commit/13613175bed7a6f356ae61dff6b475e94c843bd4

PR #5405, opened 2024-04-06
Change umu description in update channel

R1kaB3rN:
Applies small stylistic changes to the heading text/description for umu in the wine update channel section.

- Prefer 'UMU' over 'Umu'. To me, capitalizing or lowercasing it appears better here for some reason...
- Removes Pressure Vessel from description. I don't think users really should know about this technology specifically. Just mentioning Proton is enough to sell it to them, and Proton already encompasses it.
- Prefer 'experimental' than 'heavy development'. 'Heavy development' gives a dangerous impression that'd likely make the user avoid it. I don't know, but perhaps that's more appropriate here?
- Moves UMU as the last section. This is experimental and features like this are more appropriately hidden or located in the bottom of the section.


R1kaB3rN:
cc @danieljohnson2 for feedback

R1kaB3rN:
lutris

danieljohnson2:
I do not know the preferred orthography for "Umu", but I do know that it is not an acronym. It's an obscure word, a kind of oven I think.

I do think that we'd rather have users on Umu/umu/uMu/whatever than on the unsupported 'channel'.

R1kaB3rN:
umu2


> I do not know the preferred orthography for "Umu", but I do know that it is not an acronym. It's an obscure word, a kind of oven I think.
> 
> I do think that we'd rather have users on Umu/umu/uMu/whatever than on the unsupported 'channel'.

I would say the preferred orthography would be lowercase and the only instance the term is capitalized is in the Proton release where it's UMU-Proton. But if it's the middle section, it looks a bit off to me because it's lowercase compared to uppercase.

Putting that aside though, I think this description is better since its the same length as the other two. I could also include the updates are downloaded automatically bit, if that's important to mention.

R1kaB3rN:
Description with the automatic download bit:

umu3

If this bit is included though, I think this looks worse compared to the above because of the line length...

R1kaB3rN:
Latest:

umu4

R1kaB3rN:
@strycore for review and whether you think auto downloading is important to mention in the description.

strycore:
Umu is not an acronym so I don't get why it's uppercase all of a sudden. 

This whole update selection is to let users know how their wine versions are updated so we should probably keep a mention of umu keeping proton up to date.

Also, I'm not sure the current state of umu is good enough to have this option in the client for 0.5.17. I think the whole umu section should be hidden on release and then put back when we resume development on 0.5.18.

danieljohnson2:
I'm sure it's not stable, but it does Umu basically work in Lutris? Can it install 'Proton' without ProtonUp-QT to help?

If so, I think we should show the option. That is how we'll get the most feedback.

strycore:
Having umu to work and having it as a update channel like Wine-GE are 2 different things. 
Everyone will have umu, it's shipped in the lutris runtime. The question is, when will it be usable enough to have it as the default for everything in Lutris. I don't think the current state justifies enabling the umu update channel.

strycore:
I would HATE having people think this Updates thing is how you use umu.

danieljohnson2:
> I would HATE having people think this Updates thing is how you use umu.

It isn't?

strycore:
NO. It's never been the way to use umu. This was added as an experiment after umu was already fully implemented and usable in Lutris. It tries to change the defaults of things.

danieljohnson2:
I had expected that radio button meant Lutris would provide versions of Umu or Proton *and* make that the default. Maybe the option should be hidden for 0.5.17.

strycore:
Yes, it also makes umu and Proton the default, which makes sense since you may not have Wine-GE at all in this scenario. 
And also yes, we should hide this option in this release. Until at least we're sure umu can be used for game installs.

PR #5408, opened 2024-04-08
Add a GAME_DIRECTORY environment variable

### LINKED ISSUE #5407: Pre-launch/post-exit environment variables for wrapper script to detect game
For context, I'm the author of [Ludusavi](https://github.com/mtkennerly/ludusavi). It has a feature to be used as a wrapper to back up your save data, like `ludusavi wrap --name "Game Title" -- ./game --foo`. To avoid writing out the title, it also supports `--infer [steam|heroic]`, which will check for some environment variables in order to figure out which game is running.

I'd like to add support for `--infer lutris`, and I was wondering if you'd be open to adding some environment variables to support that. It would just need enough information so it can uniquely identify the game in the database and/or YAML files. Would `LUTRIS_GAME_SLUG` (table: `games`, column: `slug`) be sufficient, or would `LUTRIS_GAME_ID` (table: `games`, column: `id`) be better?

These were the closest I could find, but I don't think they cover this use case:

https://github.com/lutris/lutris/blob/3866cf21bd522b546a6f3f75653cf82dd3156c46/lutris/game.py#L689-L691

https://github.com/lutris/lutris/blob/3866cf21bd522b546a6f3f75653cf82dd3156c46/lutris/monitored_command.py#L133

danieljohnson2:
This env-var is for the user of pre-launch and post-exit scripts, and it set only for games that have a directory.

We already had WINEPREFIX and game_name.

Resolves #5407

### DIFF
--- a/lutris/game.py
+++ b/lutris/game.py
@@ -666,15 +666,20 @@ class Game(GObject.Object):
         This method sets the game_runtime_config attribute.
         """
         gameplay_info = self.get_gameplay_info(launch_ui_delegate)
-        if not gameplay_info:  # if user cancelled- not an error
+        if not gameplay_info:  # if user cancelled - not an error
             return False
         command, env = get_launch_parameters(self.runner, gameplay_info)
 
         if env.get("WINEARCH") == "win32" and "umu" in " ".join(command):
             raise RuntimeError("Proton is not compatible with 32bit prefixes")
-        env["game_name"] = self.name  # What is this used for??
         env["GAMEID"] = proton.get_game_id(self)
         env["STORE"] = self.get_store_name()
+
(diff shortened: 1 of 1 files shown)


strycore:
> We already had game_name.

If no one explains why we have it, we're not going to keep it for long.

strycore:
In the ticket that was just closed we were discussing better integration between Ludusavi and Lutris. With Ludusavi being one of the possible backends for backing up saves.

What I don't approve of is having environment variables exposed for no apparent reason. I've changed "game_name" to "GAME_NAME" to maybe get some feedback of "game_name" users.

Hopefully we will have an integration with Ludusavi that doesn't require pre-launch and post-exit scripts.

PR #5413, opened 2024-04-10
Add a switch for "Remove from library" in preferences

OlegAckbar:
Removing game from the library will erase playtime and "last start" time

### DIFF
--- a/lutris/gui/config/preferences_box.py
+++ b/lutris/gui/config/preferences_box.py
@@ -14,6 +14,7 @@ class InterfacePreferencesBox(BaseConfigBox):
         "show_tray_icon": _("Show Tray Icon"),
         "dark_theme": _("Use dark theme (requires dark theme variant for Gtk)"),
         "discord_rpc": _("Enable Discord Rich Presence for Available Games"),
+        "remove_from_library": _("Remove games from library when uninstalling"),
     }
 
     settings_accelerators = {"hide_badges_on_icons": "<Primary>p"}
@@ -35,5 +36,5 @@ class InterfacePreferencesBox(BaseConfigBox):
                 list_box_row = Gtk.ListBoxRow(visible=True)
                 list_box_row.set_selectable(False)
                 list_box_row.set_activatable(False)
-                list_box_row.add(self.get_setting_box(setting_key, label))
+                list_box_row.add(self.get_setting_box(setting_key, label, setting_key=="remove_from_library"))
                 listbox.add(list_box_row)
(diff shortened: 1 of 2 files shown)


strycore:
We're obviously all disagreeing on that matter.

Initially, Lutris would keep uninstalled games in the library.
Then we implemented library sync.
Then @danieljohnson2 switched the default to keep games when library sync is enabled
Then I realized that I was changing that default setting almost every time I wanted to remove a game. So I made it enabled by default.
And now we're back to square one, with not deleting games by default.

Technically, we have 1 solution left to try out, which would be to remove games when library sync is enabled. But I don't think this makes a lot of sense.

danieljohnson2:
There are so many options here!

I say: start the check-boxes in the [-] indeterminate state, and then if left that way, we flip a coin to decide. That covers everybody's desires! Well, sometimes anyway,

OlegAckbar:
What about a checkbox in Lutris settings which will decide default behavior?

danieljohnson2:
Can *that* checkbox start of as [-] and flip a coin?

strycore:
> What about a checkbox in Lutris settings which will decide default behavior?

Ah, you're one of those people who want to add one gazillion checkboxes to the Lutris preferences to make it the most bloated piece of software on earth. 
Unfortunately, we don't have the kind of workforce to even attempt aiming to be in the Guiness Book of World Records.
I'm not sure it's a viable strategy.

OlegAckbar:
I've added a switch that will control default state of "Remove from library". It's on by default

strycore:
Sorry, we've run out of space to add another checkbox so we can't take that patch

strycore:
Let's say that my motivation for switching the default to True was partly motivated by the fact that I was working on library sync and removal was wanted 100% of the time. Now that library sync is implemented, this would no longer be the case but the change to true was also partly motivated by the installation of games for testing only which is also not the typical use case when you're not a Lutris developer. 

The change that @danieljohnson2 made previously kept the game only if sync was activated. That means people without sync enabled lose their play times by default.

If we default to False like suggested in the PR, games are kept by default and we go back to having to remove a game twice to remove it completely.

Having a toggle in the preferences would be understood by no one. This is not something we will consider. 

Let's eliminate True as the default because that doesn't satisfy the typical use case. If we use False instead of not Library sync enabled, then we treat every user as if they had sync enabled. And it's a safer option too, there's no undelete for games while having the game a second time isn't the end of the world.

I've now switch the default to False

danieljohnson2:
Well, that's simple. Remove twice by default makes sense to me if you *are* using library sync, which we are encouraging firmly. So I guess it's okay.

PR #5415, opened 2024-04-11
Honor user's umu environment variables if set

R1kaB3rN:
For non-steam games, umu depends on the `GAMEID` and the `STORE` environment variables to apply fixes to the WINE prefix, and currently the Lutris client does not allow overriding the default values of these variables after setting their own in the _Environment Variables_ section. 

umu

This pull request makes it so the client can allow the user to override them to apply an existing umu protonfix to the running game.


danieljohnson2:
I think this could would be nicer using the 'or' operator:

```python3
env["STORE"] = env.get("STORE") or self.get_store_name()
```

R1kaB3rN:
Applied feedback

PR #5421, opened 2024-04-13
Separate Game and GameLauncher classes

strycore:
In order to avoid growing the Game class beyond what's reasonable, I've separated it into 2 classes. The Game class handles what is related to the game's data and also its management on the filesystem. The GameLauncher class handles the game execution and what's related to it. 

This is quite a substantial change and there are possibly still some bugs. I tried to add a maximum of type hints to avoid as many errors as possible.  I did some testing, native games and umu are working but there is still an issue with wine games which I suspect is something I missed related to PID tracking.

### DIFF
--- a/lutris/exception_backstops.py
+++ b/lutris/exception_backstops.py
@@ -9,7 +9,7 @@ from lutris.gui.dialogs import ErrorDialog
 from lutris.util.log import logger
 
 
-def watch_game_errors(game_stop_result, game=None):
+def watch_game_errors(game_stop_result, game_launcher=None):
     """Decorator used to catch exceptions and send events instead of propagating them normally.
     If 'game_stop_result' is not None, and the decorated function returns that, this will
     send game-stop and make the game stopped as well. This simplifies handling cancellation.
@@ -19,24 +19,28 @@ def watch_game_errors(game_stop_result, game=None):
     If you do not provide a game object directly, it is assumed to be in the first argument to
     the decorated method (which is 'self', typically).
     """
-    captured_game = game
+    captured_game = game_launcher
 
     def inner_decorator(function):
         @wraps(function)
         def wrapper(*args, **kwargs):
             """Catch all exceptions and emit an event."""
-            game = captured_game if captured_game else args[0]
+            game_launcher = captured_game if captured_game else args[0]
 
             try:
                 result = function(*args, **kwargs)
-                if game_stop_result is not None and result == game_stop_result and game.state != game.STATE_STOPPED:
-                    game.stop_game()
+                if (
+                    game_stop_result is not None
+                    and result == game_stop_result
+                    and game_launcher.state != game_launcher.STATE_STOPPED
+                ):
+                    game_launcher.stop_game()
                 return result
             except Exception as ex:
-                logger.exception("%s has encountered an error: %s", game, ex, exc_info=ex)
-                if game.state != game.STATE_STOPPED:
-                    game.stop_game()
-                game.signal_error(ex)
+                logger.exception("%s has encountered an error: %s", game_launcher, ex, exc_info=ex)
+                if game_launcher.state != game_launcher.STATE_STOPPED:
+                    game_launcher.stop_game()
+                game_launcher.signal_error(ex)
                 return game_stop_result
 
         return wrapper
--- a/lutris/game.py
+++ b/lutris/game.py
@@ -1,44 +1,31 @@
 """Module that actually runs the games."""
 
-# pylint: disable=too-many-public-methods disable=too-many-lines
+from __future__ import annotations
+
 import json
 import os
-import shlex
 import shutil
-import signal
-import subprocess
-import time
 from gettext import gettext as _
 from typing import cast
 
-from gi.repository import Gio, GLib, GObject, Gtk
+from gi.repository import Gio, GObject
 
 from lutris import settings
 from lutris.config import LutrisConfig
 from lutris.database import categories as categories_db
 from lutris.database import games as games_db
 from lutris.database import sql
-from lutris.exception_backstops import watch_game_errors
-from lutris.exceptions import GameConfigError, InvalidGameMoveError, MissingExecutableError
+from lutris.exceptions import GameConfigError, InvalidGameMoveError
+from lutris.game_launcher import GameLauncher
 from lutris.installer import InstallationKind
-from lutris.monitored_command import MonitoredCommand
-from lutris.runner_interpreter import export_bash_script, get_launch_parameters
+from lutris.runner_interpreter import export_bash_script
 from lutris.runners import import_runner, is_valid_runner_name
 from lutris.runners.runner import Runner
-from lutris.util import discord, extract, jobs, linux, strings, system, xdgshortcuts
-from lutris.util.display import DISPLAY_MANAGER, SCREEN_SAVER_INHIBITOR, disable_compositing, enable_compositing
-from lutris.util.graphics.xephyr import get_xephyr_command
-from lutris.util.graphics.xrandr import turn_off_except
-from lutris.util.linux import LINUX_SYSTEM
+from lutris.util import extract, jobs, strings, system, xdgshortcuts
 from lutris.util.log import LOG_BUFFERS, logger
-from lutris.util.process import Process
 from lutris.util.steam.shortcut import remove_shortcut as remove_steam_shortcut
 from lutris.util.system import fix_path_case
-from lutris.util.timer import Timer
-from lutris.util.wine import proton
 from lutris.util.yaml import write_yaml_to_file
-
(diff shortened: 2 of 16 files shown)


lutris/exception_backstops.py
```diff
@@ -19,24 +19,28 @@ def watch_game_errors(game_stop_result, game=None):
     If you do not provide a game object directly, it is assumed to be in the first argument to
     the decorated method (which is 'self', typically).
     """
-    captured_game = game
+    captured_game = game_launcher
```
danieljohnson2, line 23: Maybe rename captured_game to captured_launcher?

lutris/game_actions.py
```diff
@@ -79,7 +79,7 @@ def on_game_stop(self, *_args):
         """Stops the game"""
         games = self.get_running_games()
         for game in games:
-            game.force_stop()
+            game.game_launcher.force_stop()
```
danieljohnson2, line 82: Don't like this. get_running_games() should return the launchers- that is what is running, not the Game object.

This design should support running the same game twice, even if we don't want to surface that in the UI.

lutris/game_actions.py
```diff
@@ -300,7 +300,7 @@ def on_game_launch(self, *_args):
         game = self.game
         if game.is_installed and game.is_db_stored:
             if not self.application.is_game_running_by_id(game.id):
-                game.launch(launch_ui_delegate=self.window)
+                game.game_launcher.launch(launch_ui_delegate=self.window)
```
danieljohnson2, line 303: Maybe we should create the launcher object at this point?

lutris/gui/application.py
```diff
@@ -77,23 +78,21 @@ def __init__(self):
...
         self.force_updates = False
         self.css_provider = Gtk.CssProvider.new()
-        self.window = None
+        self.window: LutrisWindow = None
+        self.style_manager: StyleManager = None
         self.launch_ui_delegate = LaunchUIDelegate()
         self.install_ui_delegate = InstallUIDelegate()
-
-        self._running_games = []
+        self._running_games: List[Game] = []
```
danieljohnson2, line 92: This should be GameLaunchers, not games.

We should not create GameLaunchers on demand, we should create them on launch, and *only* on launch. Games, on the other hand, we load from the DB whenever we feel like it- a key difference.

This change would make it safe for GameLauncher to track the running game; right now we have to use _running_games to have special blessed Game objects for that purpose, and that's been a source of bugs. Let's fix that.

I might argue that _running_games might be better moved out of application.py to GameLauncher's module instead. I see no reason to tie this to the UI, and if you really do want to go to QT, you'll want to separate this from GTK.

lutris/gui/application.py
```diff
@@ -758,10 +757,11 @@ def on_error(game, error):
                 return True
 
             game = Game(db_game["id"])
-            game.connect("game-error", on_error)
-            game.launch(self.launch_ui_delegate)
+            game_launcher = game.game_launcher
```
danieljohnson2, line 760: Again, we're launching the game here, so lets create the launcher here.

lutris/gui/application.py
```diff
@@ -774,22 +774,23 @@ def on_error(game, error):
...
+    def on_game_start(self, game_launcher: GameLauncher):
+        self._running_games.append(game_launcher.game)
         if settings.read_setting("hide_client_on_game_start") == "True":
             self.window.hide()  # Hide launcher window
         return True
 
-    def on_game_stopped(self, game):
+    def on_game_stopped(self, game_launcher: GameLauncher):
         """Callback to quit Lutris is last game stops while the window is hidden."""
+        game = game_launcher.game
```
danieljohnson2, line 793: This, however, I like better. We want to go from the launcher to the game, not vice versa.

lutris/gui/dialogs/game_import.py
```diff
@@ -203,19 +203,19 @@ def import_rom(self, rom_set, filename):
...
         launch_button.set_sensitive(True)
         launch_button.connect("clicked", self.on_launch_clicked, game)
 
-    def on_launch_clicked(self, _button, game):
+    def on_launch_clicked(self, _button, game: Game) -> None:
         # We can't use this window as the delegate because we
         # are destroying it.
         application = Gio.Application.get_default()
-        game.launch(application.launch_ui_delegate)
+        game.game_launcher.launch(application.launch_ui_delegate)
```
danieljohnson2, line 215: Another place to create the launcher object!

lutris/gui/lutriswindow.py
```diff
@@ -1128,7 +1128,11 @@ def is_game_displayed(self, game):
 
         if row:
             # Stopped games do not get displayed on the running page
-            if row.type == "dynamic_category" and row.id == "running" and game.state == game.STATE_STOPPED:
+            if (
+                row.type == "dynamic_category"
+                and row.id == "running"
+                and game.game_launcher.state == game.game_launcher.STATE_STOPPED
+            ):
```
danieljohnson2, line 1135: Oh, tricky. Here's a place where it's not as easy to just use the launcher not the game.

I'd still do it though. You need a way to find all the "running" launchers for a game, which should be doable, but it will make this code more complex.

lutris/gui/lutriswindow.py
```diff
@@ -1198,7 +1202,7 @@ def on_game_activated(self, _view, game_id):
         if game_id:
             game = Game(game_id)
             if game.is_installed:
-                game.launch(launch_ui_delegate=self)
+                game.game_launcher.launch(launch_ui_delegate=self)
```
danieljohnson2, line 1205: Here's another place to create a launcher. We do launch in a bunch of places, don't we?

lutris/gui/widgets/game_bar.py
```diff
@@ -206,11 +213,11 @@ def get_play_button(self, game_actions):
 
         if self.game.is_installed:
             game_buttons = self.get_game_buttons(game_actions)
-            if self.game.state == self.game.STATE_STOPPED:
+            if self.game.game_launcher.state == self.game.game_launcher.STATE_STOPPED:
```
danieljohnson2, line 216: This is like the GameActions, and should be accessing the running games like that does. Or maybe it should actually be the same code somehow?

lutris/gui/widgets/game_bar.py
```diff
@@ -269,7 +276,10 @@ def on_install_clicked(self, button):
         """Handler for installing service games"""
         self.service.install(self.db_game)
 
-    def on_game_state_changed(self, game):
+    def on_game_launcher_state_changed(self, game_launcher: GameLauncher):
```
danieljohnson2, line 279: Are we going to change the signal and handler names to say 'launcher'? If so, you missed a bunch.

lutris/runners/linux.py
```diff
@@ -145,12 +145,12 @@ def get_launch_config_command(self, gameplay_info, launch_config):
 
         return command
 
-    def get_command(self):
+    def get_command(self) -> list:
```
danieljohnson2, line 148: I think the element type is important, and here we know it- this should be List[str].

lutris/runners/runner.py
```diff
@@ -18,10 +18,14 @@
 from lutris.util.linux import LINUX_SYSTEM
 from lutris.util.log import logger
 
+if TYPE_CHECKING:
```
danieljohnson2, line 21: Oh, I wish I had known about TYPE_CHECKING! I'll definitely be able to use that.

lutris/game.py
```diff
@@ -363,8 +284,8 @@ def on_error(_game, error):
                 return True
 
             game = Game(game_id)
-            game.connect("game-error", on_error)
-            game.launch(launch_ui_delegate)
+            game.game_launcher.connect("game-error", on_error)
```
danieljohnson2, line 287: Another place to create the GameLauncher.

lutris/game.py
```diff
@@ -1,67 +1,40 @@
 """Module that actually runs the games."""
 
-# pylint: disable=too-many-public-methods disable=too-many-lines
+from __future__ import annotations
+
```
danieljohnson2, line 4: Oh, can we do this?

If so, can we use it to say ``str | None`` and the like? In Python 3.7, this works in mypy but not at runtime, I read.

danieljohnson2 (approved):
I think you've made a lot of non-relevant type annotations here, adding noise. But I'e done it too, those things are addictive.

I have one major change I would make- no 'game_launcher' member on Game. We should create a GameLauncher only when launching, and at other times we will need to search for the launchers (plural) that are running.

Non-running launchers should be a very ephemeral thing.

There's a lot of signals and methods that say "game" instead of "game_launcher"; I could live with that, but it feels a little inconsistent right now.

However, even as it is I think this a great improvement, and I'd merge it promptly. We can get rid of Game.game_launcher afterward easily enough over time.

:+1:

PR #5423, opened 2024-04-13
Fix NoneType error in get_game_id function

CKolle:
Fixes a NoneType error with `game.runner.get_env()`. The previous code did not check if the `game` object was None this resulted in an error when `get_game_id` was called from [winexec](https://github.com/lutris/lutris/blob/b3914c6be78f9dd1da2c8c769eabef9c4a43b300/lutris/runners/commands/wine.py#L338C1-L341C75). Also did a small refactor to make it more readable.

### DIFF
--- a/lutris/util/wine/proton.py
+++ b/lutris/util/wine/proton.py
@@ -100,14 +100,22 @@ def get_proton_path_from_bin(wine_path):
 
 
 def get_game_id(game) -> str:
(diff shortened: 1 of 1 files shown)


danieljohnson2:
Looks clean to me. Thanks for the fix!

R1kaB3rN:
Yes, thank you for catching this!

PR #5425, opened 2024-04-14
Ensure that LC_ALL gets treated as HOST_LC_ALL for proton compatibility

CKolle:
In the previous [pull request](https://github.com/lutris/lutris/pull/5420), I ensured that the HOST_LC_ALL environment variable is set when the locale option is specified under game execution and the runner is Proton. This works, however a many install scripts set LC_ALL directly. This pull request addresses this issue by automatically setting HOST_LC_ALL to LC_ALL if LC_ALL is defined but HOST_LC_ALL is not, and the wine version is Proton. 

To achieve this I first changed wine runner's `get_env` to do the necessary logic to check if there is an LC_ALL and set the HOST_LC_ALL. 

Additionally, modifications were made to the `wineexec` function, as it is called directly in install scripts and does not retrieve the locale environment variable from the runner object, but rather the script itself through the env parameter. Similarly the `wineexec` function now checks that if it got an LC_ALL and the wine version is proton, then set HOST_LC_ALL. 

Lastly a bug fix was made in the `runner_interpreter.py` to ensure that if there is no locale set don't set HOST_LC_ALL to None but leave it untouched.

### DIFF
--- a/lutris/runner_interpreter.py
+++ b/lutris/runner_interpreter.py
@@ -31,13 +31,13 @@ def get_launch_parameters(runner, gameplay_info):
         env["LC_ALL"] = ""
 
     # Set correct LC_ALL depending on user settings
-    if system_config["locale"] != "":
-        env["LC_ALL"] = system_config["locale"]
-
+    locale = system_config.get("locale")
+    if locale:
+        env["LC_ALL"] = locale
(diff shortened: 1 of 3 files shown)


danieljohnson2:
Looks legit to me.

PR #5435, opened 2024-04-18
Add UMU-Proton to dropdown menu

R1kaB3rN:
Currently in the Lutris client when using umu, UMU-Proton is used as the wine version instead of GE-Proton despite the dropdown menu listing 'GE-Proton (Latest)'. This is misleading and for users to use the real GE-Proton, the environment variable `PROTONPATH=GE-Proton` needs to be added in the System Options menu. This pull request adds UMU-Proton to the wine version dropdown menu and ensures the default value for the `STORE` environment variable is 'none' to ensure the UMU database is referenced for umu gamefixes such as [umu-starcitizen](https://github.com/Open-Wine-Components/umu-protonfixes/blob/master/gamefixes-umu/umu-starcitizen.py).

### DIFF
--- a/lutris/util/wine/proton.py
+++ b/lutris/util/wine/proton.py
@@ -9,7 +9,8 @@ from lutris import settings
 from lutris.util import system
 from lutris.util.steam.config import get_steamapps_dirs
 
-GE_PROTON_LATEST = _("GE-Proton (Latest)")
+GE_PROTON_LATEST = _("UMU-Proton (Latest)")
+PROTON_LATEST = _("GE-Proton (Latest)")
 DEFAULT_GAMEID = "umu-default"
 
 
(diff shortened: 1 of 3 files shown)


lutris/util/wine/proton.py
```diff
@@ -9,7 +9,8 @@
 from lutris.util import system
 from lutris.util.steam.config import get_steamapps_dirs
 
-GE_PROTON_LATEST = _("GE-Proton (Latest)")
+GE_PROTON_LATEST = _("UMU-Proton (Latest)")
+PROTON_LATEST = _("GE-Proton (Latest)")
```
R1kaB3rN, line 13: This is confusing because the identifier contradicts the values. But unfortunately, if we don't want to break user's wine prefixes in the next release, the identifiers for the values can't be changed. 

Unless Lutris wants to accept breaking prefixes, I can change them though.

strycore:
I have removed umu-proton, I do not see a need for it and I find it confusing. Lutris will stick with Proton GE.

PR #5442, opened 2024-04-20
Fancy Game Search

danieljohnson2:
This PR adds fancy search terms to the game search, and the runner search to boot. There are logical operators and 'tagged' searches that can match attributes of games. Some of them can override 'built in' filtering, so you can get views otherwise impossible.

There's almost no UI, just some tooltips. But hardcore Linux kernel hackers can probably manage a UI that works rather like Google search.

It's my hope that this will be the foundation for future features. I intend that you should sometime be able to save one o these searches a user-defined dynamic category, for instance. And we want a better UI to build fancy searches that just typing them in. But this seems to me to be a good start.

Stand by for screenshots! All the screenshots!

Here we see OR, which must be in all-caps, to find two sets of games:
image

Let's get fancier, with AND and the 'installed' attribute and some parentheses:
image

AND is mostly redundant though, and we assume AND if you have no conjunction. Like Google, we use "-" for negation, but there's some subtlty here.

Here's a search leaving out that time when there's only war:
image

But this still works:
image
Because the - is not initial here. To get it to treat - as negation, an explicit AND can help:
image

Because these tags override built in filters, we can do tricks. We can show hidden games in a category:
image

Just to be extra cute, you can conveniently get both hidden and visible games at once, like this:
image
You could do this with ``hidden:true OR hidden:false`` as well, but this is cuter.

Some tags are a bit fancier. Here's the games that took way too long:
image

But here's a more specific search:
image
Here, the search is matching games where the playtime rounds to 21 hours (ie, nearer that than 20 or 22 hours). This is about the only way to make that useful; exactly 21 hours is very unlikely. But change the search to ``playtime:20 hours 52 minutes`` and it rounds to the minute instead.

Here's one for the last-played time:
image
This is a bit limited. I could find no way to parse a date-time from a user (ie, without a specific format string), so you can only specify last-played in relative terms. These are the games I last played 3 months ago.

Here's a list of the tags supported, taken from the tooltip:
<b>installed:</b><i>true</i>			Only installed games.
<b>hidden:</b><i>true</i>			Only hidden games.
<b>favorite:</b><i>true</i>			Only favorite games.
<b>categorized:</b><i>true</i>			Only games in a category.
<b>category:</b><i>x</i>			Only games in category <i>x</i>.
<b>source:</b><i>steam</i>			Only Steam games
<b>runner:</b><i>wine</i>			Only Wine games
<b>platform:</b><i>windows</i>	Only Windows games
<b>playtime:</b><i>&gt;2 hours</i>	Only games played for more than 2 hours.
<b>lastplayed:</b><i>&lt;2 days</i>	Only games played in the last 2 days.
<b>directory:</b><i>game/dir</i>	Only games at the path.

Finally, this stuff also works in the runners tab of the preferences. Just so I can do this:
image

### DIFF
--- a/lutris/database/categories.py
+++ b/lutris/database/categories.py
@@ -69,6 +69,15 @@ def get_category(name):
         return categories[0]
 
 
+def normalized_category_names(name: str) -> List[str]:
+    """Searches for a category name case-insensitively and returns all matching names;
+    if none match, it just returns 'name' as is."""
+    query = "SELECT name FROM categories WHERE name=? COLLATE NOCASE"
+    parameters = (name,)
+    names = [cat["name"] for cat in sql.db_query(settings.DB_PATH, query, parameters)]
+    return names or [name]
+
+
 def get_game_ids_for_categories(included_category_names=None, excluded_category_names=None):
     """Get the ids of games in database."""
     filters = []
--- a/lutris/database/categories.py
+++ b/lutris/database/categories.py
@@ -69,12 +69,22 @@ def get_category(name):
         return categories[0]
 
 
-def normalized_category_names(name: str) -> List[str]:
+def normalized_category_names(name: str, subname_allowed: bool = False) -> List[str]:
     """Searches for a category name case-insensitively and returns all matching names;
-    if none match, it just returns 'name' as is."""
+    if none match, it just returns 'name' as is.
+
+    If subname_allowed is true but name is not a match for any category, we'll look for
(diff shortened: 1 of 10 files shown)


danieljohnson2:
Thanks for the early merge!

Now I need to figure out a UI for saving these as 'dynamic categories'. ``Gtk.SearchEntry`` has been a bit of a pain.

PR #5481, opened 2024-05-23
Update ruffle version and its runner options

tobil4sk:
Updated ruffle to a recent nightly version and added the following options: `--fullscreen`, `--force-align`, and `--no-gui`.

### DIFF
--- a/share/lutris/json/ruffle.json
+++ b/share/lutris/json/ruffle.json
@@ -14,6 +14,20 @@
     ],
     "runner_options": [
         {
+            "option": "fullscreen",
+            "type": "bool",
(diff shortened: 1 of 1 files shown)


tobil4sk:
- `--fullscreen` is self explanatory.
- `--no-gui` is useful because it hides the menu bar at the top of the Window. I've turned that on by default and marked that as advanced because it's only really useful for debugging purposes.
- `--force-align` is needed to fix some alignment issues with certain swf files. For example, this happens with: https://lutris.net/games/deep-sleep/.

tobil4sk:
Many thanks!

PR #5499, opened 2024-06-01
DuckStation as a runner.

mintymoons:
- Addition of DuckStation as a runner.

### DIFF
--- a/lutris/runners/__init__.py
+++ b/lutris/runners/__init__.py
@@ -31,6 +31,7 @@ __all__ = [
     "yuzu",
     "cemu",
     # Sony
+    "duckstation",
     "pcsx2",
     "rpcs3",
     "vita3k",
--- a/lutris/runners/atari800.py
+++ b/lutris/runners/atari800.py
@@ -21,7 +21,7 @@ class atari800(Runner):
     human_name = _("Atari800")
     platforms = [_("Atari 8bit computers")]  # FIXME try to determine the actual computer used
     runner_executable = "atari800/bin/atari800"
-    bios_url = "https://netactuate.dl.sourceforge.net/project/atari800/ROM/Original%20XL%20ROM/xf25.zip?viasf=1"
+    bios_url = "http://kent.dl.sourceforge.net/project/atari800/ROM/Original%20XL%20ROM/xf25.zip"
     description = _("Atari 400, 800 and XL emulator")
     bios_checksums = {
         "xlxe_rom": "06daac977823773a3eea3422fd26a703",
--- a/lutris/runners/atari800.py
+++ b/lutris/runners/atari800.py
@@ -21,7 +21,7 @@ class atari800(Runner):
     human_name = _("Atari800")
     platforms = [_("Atari 8bit computers")]  # FIXME try to determine the actual computer used
     runner_executable = "atari800/bin/atari800"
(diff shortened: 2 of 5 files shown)


danieljohnson2:
I see this code places the downloaded BIOS files in ~/.local/share/duckstation/..., not inside ~/.local/share/lutris anywhere. Why is that?

mintymoons:
Greetings, @danieljohnson2.

Because DuckStation detects BIOS files in this directory when it is opened by the first time: ~/.local/share/duckstation/.
Nonetheless, end users can change it in the mentioned emulator.

danieljohnson2:
Careful there; those changes to the Atari 800 and MicroM8 runners maybe should be separate from this PR; you can do that by using separate branches for unrelated changes and submitting more than one PR.

mintymoons:
I haven't submitted this PR yet. it appeared automatically before creating it.
Could you close this PR, please?

danieljohnson2:
Yes, but can't you do that? And this is the DuckStation PR; I thought you wanted this in.

mintymoons:
I only submitted this PR about DuckStation.

Atari800 and MicroM8's links will be in another PR with a new branch.

danieljohnson2:
I think what you should do is reset the DuckStation branch back to commit 50b86d74eecb0716bb8a2eb3b065a677113c2cf2, which is the original DuckStation commit, and then force-push that to fix this PR.

Then do a new branch off master and do the other stuff that way.

mintymoons:
I did this and created a new branch for the new links. Could you check it again, please?

danieljohnson2:
That revert has put it back where it was.

mintymoons:
It is a relief. Thanks for your revision.

strycore:
I torn about the inclusion of PS1 bios file URLs. 
We've traditionally avoided doing that, with some exceptions, like Atari800. 
But put this into consideration: Atari800 support the XE130 released in 1983. Lutris added support for Atari800 around 2010. So it was 27 years later. We're in 2024 and the PS1 was released in 1995, so 29 years ago. 
Based on that, we should consider shipping links to BIOS files from a machine that is almost 30 years old.

mintymoons:
Greetings, @strycore.

Thank you for your feedback. I really appreciate it.

mintymoons:
@strycore. BIOS files URLs were removed.

strycore:
> @strycore. BIOS files URLs were removed.

Thanks, I think we should come up with a more generic solution for dealing with BIOS files. One that's not specific for Duckstation. The same BIOS files would be needed for the PSX Retroarch cores. And we need a way to handle BIOS files on a more generic level for other types of platforms (Amiga, Dreamcast, PS2, ....)

mintymoons:
It was a pleasure, @strycore.

mintymoons:
Could I create a new discussion to write about how we should handle BIOS generically for many emulators? It is to refer to emulators in general instead of DuckStation or this PR.

I tried that before and I found some ways. I also want to learn about this.

strycore:
Yes, you totally can!

mintymoons:
It is appreciated. I will create the new discussion.

mintymoons:
Do you recommend me to categorise the new discussion as "General" or "Q&A"? I try to be precise about it.

strycore:
General is fine

strycore:
Merging this before releasing 0.5.18

PR #5536, opened 2024-07-18
Saved Searches

danieljohnson2:
There seems to be little interest in this, but here it is if you want it. Saved searches!

You create them with a menu command in the hamburger menu:
Screenshot from 2024-07-17 20-11-44
And there's a dialog to edit the search, rather than typing it in:
Screenshot from 2024-07-17 19-55-46
And the result appears in the sidebar:
Screenshot from 2024-07-17 19-56-00

They do not appear as a kind of category, because they are not part of the library sync feature, and I expect they won't become so. This should be teachable- categories are attributes of a game, but searches aren't.

The main downside of this PR, I think, is significant complexity. The search predicates aren't functions any more but much more complex objects. That is how searches are editable, but if you are feeling the code is getting out of hand... maybe it's not worth it.

### DIFF
--- a/lutris/database/categories.py
+++ b/lutris/database/categories.py
@@ -2,7 +2,7 @@ import abc
 import re
 from collections import defaultdict
 from itertools import repeat
-from typing import Dict, List, Union
+from typing import Dict, List, Optional, Union
 
 from lutris import settings
 from lutris.database import sql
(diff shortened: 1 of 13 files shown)


strycore:
I will make a follow up PR with adjustments to this feature

PR #5539, opened 2024-07-19
Update umu path candidates

R1kaB3rN:
Since https://github.com/Open-Wine-Components/umu-launcher/pull/140, finding `umu_run.py` in a `/share` subdirectory is no longer valid as umu-launcher is now being installed as a Python module in a site-packages directory. Instead, for the system installed umu-launcher, Lutris must search for `umu-run` in PATH and its runtime directory.

### DIFF
--- a/lutris/util/wine/proton.py
+++ b/lutris/util/wine/proton.py
@@ -40,7 +40,7 @@ def is_umu_path(wine_path: Optional[str]) -> bool:
     PROTONPATH, but if this is omitted it will default to the latest Proton it
     downloads."""
     if wine_path:
-        return wine_path.endswith("/umu_run.py") or wine_path.endswith("/umu-run")
+        return wine_path.endswith("/umu-run")
     return False
 
 
@@ -52,35 +52,18 @@ def get_umu_path() -> str:
     The path that this returns will be considered an Umu path by is_umu_path().
 
     If this script can't be found this will raise MissingExecutableError."""
-
-    # We can access Umu via it's main .py file, or via the public name which is just
-    # symbolic link to the same file. We'll check both for compatibility, just in case
-    # Umu gets translated into Perl or something.
-    entry_points = ["umu_run.py", "umu-run"]
-
+    entry_point = "umu-run"
     custom_path = settings.read_setting("umu_path")
-    if custom_path:
-        for entry_point in entry_points:
-            entry_path = os.path.join(custom_path, entry_point)
-            if system.path_exists(entry_path):
-                return entry_path
 
-    # We only use 'umu-run' when searching the path since that's the command
-    # line entry point.
-    if system.can_find_executable("umu-run"):
-        return system.find_required_executable("umu-run")
-    path_candidates = (
-        "/app/share",  # prioritize flatpak due to non-rolling release distros
-        "/usr/local/share",
-        "/usr/share",
-        "/opt",
-        settings.RUNTIME_DIR,
-    )
-    for path_candidate in path_candidates:
-        for entry_point in entry_points:
-            entry_path = os.path.join(path_candidate, "umu", entry_point)
-            if system.path_exists(entry_path):
-                return entry_path
+    if custom_path and system.path_exists(umu_path := f"{custom_path}/{entry_point}"):
+        return umu_path
+
+    if system.can_find_executable(entry_point):
+        return system.find_required_executable(entry_point)
+
+    if system.path_exists(umu_path := f"{settings.RUNTIME_DIR}/umu/{entry_point}"):
+        return umu_path
+
     raise MissingExecutableError("Install umu to use Proton")
 
 


danieljohnson2:
What's the status on this? GitHub call this a draft, but it seems like something we might need. Is there still an issue with this PR?

R1kaB3rN:
The zipapp functionality is merged, which bundles umu as a standalone executable. 

**Important**: I don't know when Lutris plans to release, but in the meantime, the following files must be included in runtime directory to ensure those who compile from source and those who are on the latest stable version (0.5.17) will not be broken:
```

filelock                   __pycache__                 umu_consts.py                 umu_proton.py   umu_util.py
filelock-3.15.4.dist-info  python_xlib-0.33.dist-info  umu_launcher-0.1.0.dist-info  umu-run         umu_version.json
__init__.py                six-1.16.0.dist-info        umu_log.py                    umu_run.py      Xlib
__main__.py                six.py                      umu_plugins.py                umu_runtime.py
```

Where `umu-run` is a relative symbolic link to umu_run.py

danieljohnson2:
Hmm. Looks like new versions of the Umu runner will just have a different list of files.

I see you will have both ``umu-run`` and ``umu_run.py``; Lutris seems to prefer using ``umu_run.py`` directly right now. Is that correct, or should I change it?

lutris/util/wine/proton.py
```diff
@@ -40,7 +40,7 @@ def is_umu_path(wine_path: Optional[str]) -> bool:
     PROTONPATH, but if this is omitted it will default to the latest Proton it
     downloads."""
     if wine_path:
-        return wine_path.endswith("/umu_run.py") or wine_path.endswith("/umu-run")
+        return wine_path.endswith("/umu-run")
```
danieljohnson2, line 43: I don't like that you are dropping support for the .py entry points- this means older Umu runtimes we've had won't work.

lutris/util/wine/proton.py
```diff
@@ -52,35 +52,12 @@ def get_umu_path() -> str:
     The path that this returns will be considered an Umu path by is_umu_path().
 
     If this script can't be found this will raise MissingExecutableError."""
-
-    # We can access Umu via it's main .py file, or via the public name which is just
-    # symbolic link to the same file. We'll check both for compatibility, just in case
-    # Umu gets translated into Perl or something.
-    entry_points = ["umu_run.py", "umu-run"]
```
danieljohnson2, line 59: Again, I'd like to keep umu_run.py support.

lutris/util/wine/proton.py
```diff
@@ -52,35 +52,12 @@ def get_umu_path() -> str:
...
-
-    custom_path = settings.read_setting("umu_path")
```
danieljohnson2, line 61: You've removed support for the umu_path setting altogether. Why?

lutris/util/wine/proton.py
```diff
@@ -52,35 +52,12 @@ def get_umu_path() -> str:
...
     if system.can_find_executable("umu-run"):
         return system.find_required_executable("umu-run")
-    path_candidates = (
-        "/app/share",  # prioritize flatpak due to non-rolling release distros
-        "/usr/local/share",
-        "/usr/share",
-        "/opt",
-        settings.RUNTIME_DIR,
-    )
-    for path_candidate in path_candidates:
```
danieljohnson2, line 79: We're searching the PATH still, but not most of the specific directories. Is that right? Will flatpak installs still work this way?

lutris/util/wine/proton.py
```diff
@@ -52,35 +52,12 @@ def get_umu_path() -> str:
...
-        for entry_point in entry_points:
-            entry_path = os.path.join(path_candidate, "umu", entry_point)
-            if system.path_exists(entry_path):
-                return entry_path
+
+    if os.path.exists(entry_point := f"{settings.RUNTIME_DIR}/umu/umu-run"):
```
danieljohnson2, line 58: The walrus operator (:=) was introduced in Python 3.8, and we support by back to 3.7. You can't use this feature. Sorry.

Also, I think system.path_exists() should be used here since it supports '~'.

Also also, os.path.join() should probably be preferred over using an f-string. Though maybe that just my Windows experience talking, where paths are extra crazy.

danieljohnson2 (changes requested):
This latest version of this PR seems problematic to me. We're now dropping the umu_path setting, and also support for the current Umu runtime, and it's not Python 3.7 compatible. Among other things.

I wouldn't merge this.

lutris/util/wine/proton.py
R1kaB3rN: If Lutris decides to subscribe to using a standalone executable, this problem can resolved by simply creating a relative umu_run.py symbolic link to umu-run in its build process

lutris/util/wine/proton.py
R1kaB3rN: I'll revert this change as this was careless of me and will look into how it works

lutris/util/wine/proton.py
R1kaB3rN: I'll revert the syntax as I wasn't aware and will make the changes. Also, if that's the case, is there any specific reason why the Python version defined in the project's pyproject.toml was not downgraded to reflect that?

lutris/util/wine/proton.py
danieljohnson2: No, that's not enough. We've already shipped an Umu runtime- we should not break it until we've actually replaced it. It has no 'umu-run' in it, just 'umu_run.py'.

lutris/util/wine/proton.py
danieljohnson2: I believe this is a setting you'd have to edit your configuration files to set, but I'd expect it to grow a GUI setting in due course.

lutris/util/wine/proton.py
danieljohnson2: Hmm, I see what you mean. This is surprise to me, and it arrived in a seemingly unrelated PR #5237. I'll ping @strycore about this on Discord. Maybe we're okay with the walrus now and I'm the one whose out of date here.

danieljohnson2: OK, he's told me that we're go for 3.8, so the walrus is our friend and you don't need to change this.

lutris/util/wine/proton.py
danieljohnson2: Thank you.

danieljohnson2 (approved):

R1kaB3rN:
@danieljohnson2 Can we expect this to be merged? The runtime just got updated, and users in Lutris v0.5.17 cannot run their games because the client cannot find the `umu-run` executable in its runtime directory:

> Traceback (most recent call last):
>   File "/usr/lib/python3.12/site-packages/lutris/exception_backstops.py", line 31, in wrapper
>     result = function(*args, **kwargs)
>              ^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/usr/lib/python3.12/site-packages/lutris/game.py", line 668, in configure_game
>     gameplay_info = self.get_gameplay_info(launch_ui_delegate)
>                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/usr/lib/python3.12/site-packages/lutris/game.py", line 607, in get_gameplay_info
>     gameplay_info = self.runner.play()
>                     ^^^^^^^^^^^^^^^^^^
>   File "/usr/lib/python3.12/site-packages/lutris/runners/wine.py", line 1216, in play
>     command = self.get_command()
>               ^^^^^^^^^^^^^^^^^^
>   File "/usr/lib/python3.12/site-packages/lutris/runners/wine.py", line 1188, in get_command
>     raise MissingExecutableError("Install umu to use Proton")
> lutris.exceptions.MissingExecutableError: Install umu to use Proton

Steps to reproduce:

1. Build Lutris 0.5.17
2. Uninstall, if exists, the system wide umu-launcher package
3. Delete `$HOME/.local/share/lutris`
4. Open a terminal emulator
5. Run `LUTRIS_EXPERIMENTAL_FEATURES_ENABLED=1 lutris`
6. Update the Lutris runtime in the Preferences menu
7. Wait for umu to be downloaded
8. Confirm the executable umu-run at  `$HOME/.local/share/lutris/runtime/umu` exists
9. Run a game with the `GE-Proton (Latest)`
10. Observe the above traceback

Step to fix:

- Create a umu_run.py symbolic link to umu-run

System information:
OS: Arch Linux
Package: 0.5.17-5
[PKGBUILD](https://gitlab.archlinux.org/archlinux/packaging/packages/lutris/-/blob/cafdf609f83ab0c915bab2b3ed7ead515f2bcc8c/PKGBUILD)

R1kaB3rN:
Updated some logic to only find `umu-run`.

R1kaB3rN:
Well, actually, this PR won't fix the problem because it was addressed already some time in the master branch. Instead, it just simplifies the logic since Lutris is only looking for umu-run.

R1kaB3rN:
So if Lutris wants to resolve this issue for their userse, a hotifx would need to be pushed where a umu_run.py symbolic link is included or the umu-run executable is renamed to umu_run.py.

danieljohnson2:
If all this PR does is break existing Lutris installs from 0.5.17, we should reject it. What does this PR do for us, really?

strycore:
We can easily add a symlink to umu_run.py if that's needed.

danieljohnson2:
It's needed to get 0.5.17 working- master knows to look for ``umu-run``.

strycore:
We don't care about master. The only thing important is 0.5.17 and 0.5.18

strycore:
current umu now has the symlink to umu_run.py so it should work again

PR #4690, opened 2023-01-01
Add itch.io service

GoGoOtaku:
This adds itch.io support - this took me a bit of work since I didn't find a full API documentation (open-api/swagger or just regular documentation) for itch.io. So I extracted the API from https://github.com/itchio/go-itchio.
**Just like the itch electron client this doesn't show games/apps bought in big bundles until they are manually added to the library (by clicking download on the bundle page)**
This is a limitation of itch.io. At this point I do not know how to even check which bundles the user owns and neither is there any indicator that a key is available through a previously purchased bundle through the game search function. I know this is awkward but I can't really do anything against this atm and itch.io has stated repeatedly that they do not want this to work differently.

I also added the ability to give the Downloader cookies since I needed cookies for authentication. Unless I did something wrong.

### TODO
- [x] Download and install patches
- [x] Download assets (if separate)
- [x] Download extras
- [x] Handle multiple files
- [x] ~~Handle Flash, HTML, etc. based games (currently just Linux and Wine) (optional)~~
- [x] ~~Uploading play time to itch.io servers (optional)~~
- [x] Probably more stuff I forgor...
- [x] Code Cleanup

Things that ARE implemented:
- Get library of purchased games
- Install Linux and Windows games
- Install (free) games with lutris installer script
- Update games
- Install extras like manual, soundtrack, art, etc...

Things that are NOT implemented:
- "embedded" games for the browser (HTML, Unity, Flash, Java) - Playable through the itch.io website
- "Offline" games like RPG manuals, printable card games, etc (mostly just a PDF anyway)
- Wharf Patching (I am very sorry itch.io :cry:)
- Collections

Things that are not implemented for external reasons:
- DLC support - itch.io doesn't have any facilities for DLC. So games and DLC can not automatically linked. There are only a handful of "DLC" on itch that have to be manually installed anyway. We COULD implement a list of known DLC which would be manually kept up to date. There is a DLC tag on itch.io (tags can not be queried over the API as far as I can tell)
- Bundle games - Like mentioned above. This is a thing with itch. To add games to the library the "Download" button on the bundle page has to be clicked which leads to a page with download links and installation instructions. This will _activate_ the game. The game does not need to be actually downloaded.

This will close #243

PR #5549, opened 2024-07-30
Itch.io: use a collection in addition to bought games

### LINKED ISSUE #4757: [itch.io] Feature Request: Collection(s)
As stated in https://github.com/lutris/lutris/issues/243#issuecomment-1436054841 and in #4690 Lutris currently does not support collections. Collections are a nice way on itch.io to have a smaller library (itch.io libraries can be quite massive) that also holds free games (itch.io ties the library to purchase keys).
Also as stated in the issue you can add free games via scripts (either local or from the Lutris website) while that works most itch.io games aren't on there. I don't know much about the Lutris internal database but adding all of the itch.io games might be overkill as it's free to upload a game and there are probably magnitudes more games on itch.io than on steam or gog. I'm not checking.

The real issue here is that there is afaik no way of having settings for a source and while I'm not saying that it is impossible seemed a lot for just collections.

My "low effort" solution would be to make a second source that inherits from the main source but overrides the `load` function to pull in from specific collections (for example collections named Lutris or lutris). This would by too cryptic for a main source but for a secondary quality of life source I could see this being an alternative.

I am aware that this is slightly clunky and I would understand it if the maintainers would veto this. I make this issue partially to discuss if this is ok or if there are alternative ideas on how to do it. From proper sources settings up to just saying "no".

quantenzitrone:
closes #4757

i haven't tested this thoroughly, but at first glance it works

Reloading of the library also magically works, even though i haven't implemented any function to use the `force_load=True` on `get_collected_games()`. Someone more knowledgeable about the codebase please help.

<details>
<summary>screenshot and api json the data from the screenshot is coming from</summary>

image

i haven't bought any games on itch.io yet, so this would probably also need to be tested by someone with a more versatile itch.io library

this is my `https://api.itch.io/profile/collections`:
```json
{
  "collections": [
    {
      "updated_at": "2024-07-30T18:10:02.000000000Z",
      "title": "my collection",
      "created_at": "2023-09-04T14:55:32.000000000Z",
      "games_count": 5,
      "id": 3614632
    }
  ]
}
```
as you see, i only have one collection, so lutris uses that one
if I had more collections, lutris would the first collection called "lutris", if there is none it wouldn't use any collections

this is my `https://api.itch.io/collections/3614632/collection-games`:
```json
{
  "page": 1,
  "per_page": 100,
  "collection_games": [
    {
      "game": {
        "id": 2840340,
        "traits": [
          "p_windows",
          "p_linux",
          "p_osx",
          "can_be_bought"
        ],
        "type": "default",
        "url": "https://vividev.itch.io/kiosk",
        "title": "Kiosk",
        "published_at": "2024-07-22T12:34:53.000000000Z",
        "created_at": "2024-07-18T17:26:32.000000000Z",
        "classification": "game",
        "short_text": "Short horror experience in a fast food kiosk",
        "cover_url": "https://img.itch.zone/aW1nLzE3MDI4MDY4LnBuZw==/315x250%23c/Hkkkc9.png",
        "min_price": 0
      },
      "created_at": "2024-07-30T18:10:02.000000000Z",
      "user_id": 8767225,
      "blurb": null,
      "position": 5
    },
    {
      "game": {
        "still_cover_url": "https://img.itch.zone/aW1hZ2UvMjA2MTE0Lzk3MzA0OC5naWY=/315x250%23cm/NuyrK7.gif",
        "id": 206114,
        "traits": [
          "p_windows",
          "p_osx"
        ],
        "short_text": "A basketball themed visual novel and dating sim with stat management!",
        "url": "https://intothehoop.itch.io/demo",
        "title": "Into The Hoop Demo",
        "published_at": "2017-12-14T14:08:22.000000000Z",
        "created_at": "2017-12-14T11:53:48.000000000Z",
        "type": "default",
        "classification": "game",
        "cover_url": "https://img.itch.zone/aW1hZ2UvMjA2MTE0Lzk3MzA0OC5naWY=/315x250%23c/5J8k0f.gif",
        "min_price": 0
      },
      "created_at": "2024-07-30T15:07:08.000000000Z",
      "user_id": 8767225,
      "blurb": null,
      "position": 4
    },
    {
      "game": {
        "embed": {
          "height": 685,
          "fullscreen": false,
          "width": 960
        },
        "still_cover_url": "https://img.itch.zone/aW1hZ2UvMTgzMTY1Lzg2MTUzNC5naWY=/315x250%23cm/qnXs09.gif",
        "id": 183165,
        "url": "https://zephyo.itch.io/hi-no-homo",
        "short_text": "realistic simulator of awkward people saying hi",
        "traits": {},
        "title": "hi no homo",
        "published_at": "2017-10-15T17:55:16.000000000Z",
        "created_at": "2017-10-10T21:57:55.000000000Z",
        "type": "html",
        "classification": "game",
        "cover_url": "https://img.itch.zone/aW1hZ2UvMTgzMTY1Lzg2MTUzNC5naWY=/315x250%23c/tDaRZE.gif",
        "min_price": 0
      },
      "created_at": "2023-11-19T01:36:48.000000000Z",
      "user_id": 8767225,
      "blurb": null,
      "position": 3
    },
    {
      "game": {
        "still_cover_url": "https://img.itch.zone/aW1nLzIxNTI3NTAuZ2lm/315x250%23cm/b1m9g%2F.gif",
        "id": 415035,
        "traits": [
          "p_windows",
          "p_linux",
          "p_osx",
          "can_be_bought"
        ],
        "short_text": "‘goth gf’s iPhone’ airdrops you - accept or decline? A love/horror story about life, death, & memes.",
        "url": "https://zephyo.itch.io/missed-message",
        "title": "missed messages.",
        "published_at": "2019-05-01T05:44:05.000000000Z",
        "created_at": "2019-04-30T00:55:09.000000000Z",
        "type": "default",
        "classification": "game",
        "cover_url": "https://img.itch.zone/aW1nLzIxNTI3NTAuZ2lm/315x250%23c/JW5rGE.gif",
        "min_price": 0
      },
      "created_at": "2023-11-19T01:35:28.000000000Z",
      "user_id": 8767225,
      "blurb": null,
      "position": 2
    },
    {
      "game": {
        "still_cover_url": "https://img.itch.zone/aW1hZ2UvMTA2NTk5LzU4NjAxMi5naWY=/315x250%23cm/PKts8V.gif",
        "id": 106599,
        "traits": [
          "p_windows",
          "p_linux",
          "p_osx",
          "can_be_bought",
          "in_press_system"
        ],
        "short_text": "Will you write the way into her heart?",
        "url": "https://teamsalvato.itch.io/ddlc",
        "title": "Doki Doki Literature Club!",
        "published_at": "2017-09-22T16:53:10.000000000Z",
        "created_at": "2016-12-17T15:34:50.000000000Z",
        "type": "default",
        "classification": "game",
        "cover_url": "https://img.itch.zone/aW1hZ2UvMTA2NTk5LzU4NjAxMi5naWY=/315x250%23c/ZFe38m.gif",
        "min_price": 0
      },
      "created_at": "2023-09-04T14:55:32.000000000Z",
      "user_id": 8767225,
      "blurb": null,
      "position": 1
    }
  ]
}
```
If you compare this json with the above screenshot, you may notice that "hi no homo" is not shown in the screenshot. This is because there aren't any supported platforms for that game.
</details>

Things still to do (i think):
- [x] Testing
  - [x] no collections
  - [x] only one collection
  - [x] multiple collections, none with the title `"lutris"`
  - [x] multiple collections, one with the title `"lutris"`
  - [x] mutliple collections with the title `"lutris"` (which collection is chosen is kinda random, maybe we should sort the result by some metric before choosing one)
  - [x] combination of collections and bought games
    - I don't know what will happen if games are bought and in a collection, will they be shown multiple times?
      - maybe its necessary to remove duplicates (prefering games with `download_key_id`s) 
  - [x] installation of non-buyable (free as in price/gratis) game
  - [x] installation of buyable game that costs 0
  - [x] installation of unbought game in collection ⇾ it asks for the installation file
- [x] find out why reloading works magically ~~(this might be a bug)~~

Also, is it ok that I committed the `flake.nix` I created for development? 
You might not want the repository clutter, but it may help other contributors that also use Nix.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,8 @@ pga.db
 tests/coverage/*
 /dist
 /lutris.egg-info
+/flake.nix
+/flake.lock
 
 # meson builddirs
 builddir
--- a/lutris/services/itchio.py
+++ b/lutris/services/itchio.py
@@ -111,7 +111,9 @@ class ItchIoService(OnlineService):
     cache_path = os.path.join(settings.CACHE_DIR, "itchio/api/")
 
     key_cache_file = os.path.join(cache_path, "profile/owned-keys.json")
+    collection_list_cache_file = os.path.join(cache_path, "profile/collections.json")
     games_cache_path = os.path.join(cache_path, "games/")
+    collection_cache_path = os.path.join(cache_path, "collections/")
     key_cache = {}
 
     supported_platforms = ("p_linux", "p_windows")
@@ -182,6 +184,18 @@ class ItchIoService(OnlineService):
     def fetch_owned_keys(self, query=None):
         """Do API request to get games owned by user (paginated)"""
         return self.make_api_request("profile/owned-keys", query)
+
+    def fetch_collections(self, query=None):
+        """Do API request to users collections"""
+        return self.make_api_request("profile/collections", query)
+
+    def fetch_collection(self, collection_id):
+        """Do API request to get info about a collection"""
+        return self.make_api_request(f"collections/{collection_id}")
+
+    def fetch_collection_games(self, collection_id, query=None):
+        """Do API request to get the list of games in a collection"""
+        return self.make_api_request(f"collections/{collection_id}/collection-games", query)
 
     def fetch_game(self, game_id):
         """Do API request to get game info"""
@@ -268,9 +282,118 @@ class ItchIoService(OnlineService):
             self._cache_games(games)
         return games
 
+    def get_collection_cache(self, collection_id):
+        """Create basic cache key based on collection slug and collection_id"""
+        return os.path.join(self.collection_cache_path, f"{collection_id}.json")
+
+    def _cache_collection(self, collection):
+        """Store information about collections in cache"""
+        os.makedirs(self.collection_cache_path, exist_ok=True)
+        filename = self.get_collection_cache(collection["id"])
+        key_path = os.path.join(self.collection_cache_path, filename)
+        with open(key_path, "w", encoding="utf-8") as cache_file:
+            json.dump(collection, cache_file)
+
+    def get_games_in_collections(self, collection_list: list, force_load=False):
+        """Get all games from a list of collections"""
+
+        games = []
+
+        known_appids = set()
+
+        # fetch collected games for each collection
+        for collection in collection_list:
+            fresh_data = True
+
+            collection_cache_path = self.get_collection_cache(collection["id"])
+
+            if (not force_load) and os.path.exists(collection_cache_path):
+                with open(collection_cache_path, "r", encoding="utf-8") as key_file:
+                    collection = json.load(key_file)
+                fresh_data = False
+            else:
+                # get the list of games in that collection
+                collection["games"] = []
+                query = {"page": 1}
+                # Basic security; I'm pretty sure itch.io will block us before that tho
+                safety = 65507
+                while safety:
+                    response = self.fetch_collection_games(collection["id"], query)
+                    if isinstance(response["collection_games"], list):
+                        collection["games"] += response["collection_games"]
+                        if len(response["collection_games"]) == int(response["per_page"]):
+                            query["page"] += 1
+                        else:
+                            break
+                    else:
+                        break
+                    safety -= 1
+
+                # filter out bad data for safety
+                collection["games"] = list(
+                    filter(lambda col_game: "game" in col_game and "id" in col_game["game"], collection["games"])
+                )
+
+                # try to get download keys from cache
+                for col_game in collection["games"]:
+                    game = col_game["game"]
+                    game_cache_path = self.get_game_cache(game["id"])
+                    if (
+                        "can_be_bought" in game.get("traits", [])
+                        and game.get("min_price", 0) > 0
+                        and os.path.exists(game_cache_path)
+                    ):
+                        with open(game_cache_path, "r", encoding="utf-8") as key_file:
+                            cached_game = json.load(key_file)
+                            if "download_key_id" in cached_game:
+                                game["download_key_id"] = cached_game["download_key_id"]
+
+                # cache the resulting collection
+                self._cache_collection(collection)
+
+            if fresh_data:
+                self._cache_games([col_game["game"] for col_game in collection["games"]])
+
+            for col_game in collection["games"]:
+                game = col_game["game"]
+                if game["id"] not in known_appids:
+                    known_appids.add(game["id"])
+                    games.append(game)
+        return games
+
+    def get_collection_list(self, force_load=False):
+        collections = []
+        if (not force_load) and os.path.exists(self.collection_list_cache_file):
+            with open(self.collection_list_cache_file, "r", encoding="utf-8") as key_file:
+                collections = json.load(key_file)
+        else:
+            collections = self.fetch_collections().get("collections", [])
+            with open(self.collection_list_cache_file, "w", encoding="utf-8") as key_file:
+                json.dump(collections, key_file)
+        return collections
+
     def get_games(self):
         """Return games from the user's library"""
-        games = self.get_owned_games()
+        # get and cache owned games
+        owned_games = self.get_owned_games()
+
+        # get all collections
+        collections = self.get_collection_list()
+
+        # if there is only one collecion, use owned games and this collection
+        if len(collections) == 1:
+            games = owned_games + self.get_games_in_collections(collections)
+        else:
+            # if there are collections titled "lutris" (case insestitive) we use only these
+            lutris_collections = list(
+                filter(lambda col: col.get("title", "").lower() == "lutris" and "id" in col, collections)
+            )
+            if len(lutris_collections) > 0:
+                games = self.get_games_in_collections(lutris_collections)
+            # otherwise we just use all owned games
+            else:
+                games = self.get_owned_games()
+
         filtered_games = []
         for game in games:
             traits = game.get("traits", {})
--- a/lutris/util/wine/wine.py
+++ b/lutris/util/wine/wine.py
@@ -24,11 +24,12 @@ WINE_PATHS: Dict[str, str] = {
 
 # Insert additional system-wide Wine installations.
 try:
-    for _candidate in os.listdir("/usr/lib/"):
-        if _candidate.startswith("wine-"):
-            _wine_path = os.path.join("/usr/lib/", _candidate, "bin/wine")
-            if os.path.isfile(_wine_path):
-                WINE_PATHS["System " + _candidate] = _wine_path
+    if system.path_exists("/usr/lib"):
+        for _candidate in os.listdir("/usr/lib/"):
+            if _candidate.startswith("wine-"):
+                _wine_path = os.path.join("/usr/lib/", _candidate, "bin/wine")
+                if os.path.isfile(_wine_path):
+                    WINE_PATHS["System " + _candidate] = _wine_path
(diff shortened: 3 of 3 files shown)


danieljohnson2:
I've given this some initial testing, and it looks pretty good!

I have purchased games and a collection; the sync gave me all my purchased games plus all other games in the collection, without duplicates (and I had some). This is because of the title check on line 156 in itchio,py: ``if game["title"] in seen:``

I don't think you need ``force_load`` at all- ``get_owned_games`` is never called with this = True either. Instead a base class (``OnlineService``) implements ``wipe_game_cache`` to delete the cache directory for the service altogether  before the load. This all seems correct and it works in my testing.

I am quite curious how you found out about the API you are using; I tried to implement this but could find no documentation.

It'd take this patch as is, but I leave feature calls to @strycore.

danieljohnson2:
I do think there's a fault here, which is the way this selects a collection to load. This PR selects the one you have if you have only one, or the one named "lutris" (case sensitive?) if you have more than one. That's kind of obscure, and nothing tells the user about this.

Is there some way we can make this more visible? I'd favor having a setting somehow to let the user select which collection to use; maybe the ability to omit the owned games too.

If there's no way to do a setting (and I don't see one right now), I'd favor a rule like "use the 'Lutris' collection if present, owned games if not". Not as complicated, and it gives the user more control- the ability to omit owned games they don't want to see. Still not too obvious though.

quantenzitrone:
> I am quite curious how you found out about the API you are using; I tried to implement this but could find no documentation.

this (i found it through the initial itch.io support pr #4690):
https://github.com/itchio/go-itchio/blob/1811474a71ea28178cb96ca1c522eabda7904867/endpoints_profile.go#L64-L69
https://github.com/itchio/go-itchio/blob/1811474a71ea28178cb96ca1c522eabda7904867/endpoints_collections.go
and a bit of trial and error

i don't think there is any documentation

quantenzitrone:
> I don't think you need `force_load` at all- `get_owned_games` is never called with this = True either. 

it actually is in `get_key()`
https://github.com/lutris/lutris/blob/85082abc8026540517194a494437b135c15d5e9d/lutris/services/itchio.py#L309-L311

> Instead a base class (`OnlineService`) implements `wipe_game_cache` to delete the cache directory for the service altogether before the load happens. This all seems correct and it works in my testing- and this is why reloading magically works.

that makes sense

quantenzitrone:
> Is there some way we can make this more visible? I'd favor having a setting somehow to let the user select which collection to use; maybe the ability to omit the owned games too.

If I was to add a preference for the used collection, it would go in the 'Sources' tab in the preferences, I think.
For that, I would need to add a 'settings' property for `BaseService` in `lutris/services/base.py`, implement that for `ItchIoService` and show these settings in `_get_service_box()` in `lutris/gui/config/services_box.py`.

I could probably do that, but it may be a bit overkill. What do you think? Do we need feedback from @strycore for interface changes?

danieljohnson2:
You are right, get_key() looks like it needs to access the collections too, and that's where you'd pass True.

I do not think this PR ought to try to add settings for sources; that would be too big and unfocused a PR. I think we should go with this PR and do another for the setting.

quantenzitrone:
> I'd favor a rule like "use the 'Lutris' collection if present, owned games if not". Not as complicated, and it gives the user more control- the ability to omit owned games they don't want to see. Still not too obvious though.

I think this would significantly complicate things, because we would also need to find the download keys for all owned paid games in that collection.
I think most users would just have one collection called whatever itch.io suggested to them (usually `<username>'s collection`), and add all the games they are interested in to that one. Not showing games in that collection would be more unintuitive.
Furthermore, afaik no other source has this control over which games are shown in Lutris, so why do it for itch?

I agree that the way I currently choose the collection to get games from is not perfect:
Some ideas to maybe improve that:
- use multiple collections, so we don't have to pick one if there are multiple with the title "lutris"
- choose the lutris collection with the most games
- match "lutris" case insensitively
- choose the collection with the most games if there are multiple collections and none titled "lutris"

The end goal of itch.io support would probably be either:
- sub-menus for bought games and each collection in the lutris ui
- or a setting to choose which collection(s) to show

So it would probably be more future-proof to make lutris handle multiple collections now.

quantenzitrone:
> I think this would significantly complicate things, because we would also need to find the download keys for all owned paid games in that collection.

its actually ok, i just get the keys from cache

What do you think would be the most intuitive way to choose which collections to show?

danieljohnson2:
I think the *right thing* is to provide source level settings just as we do runner level settings.

But this seems quite complicated to do well, since you'd expect settings of this kind to include the "System options" and now there's a priority question- if an option is set for the runner and the source, which one takes effect?

quantenzitrone:
so what do you think, is this mergable like this?

danieljohnson2:
Yes, I'd merge this.

I haven't done so because I leave the call on feature work to @strycore; I only merge bug-fixes myself.

He tends to merge things near to doing the next release, which I don't think is especially close right now.

Patience, I think he'll take this in due course.

quantenzitrone:
`flake.nix` and `flake.lock` have to be at the root of the project. They are technically not NixOS specific, as Nix can be installed on most Linux distributions as a standalone package manager.

danieljohnson2:
But what are they for? Why do they belong in Lutris?

I wouldn't think *any* lock file would be under source control ever, but maybe I don't understand this stuff.

danieljohnson2:
Looking at these files- is this support for delivering Lutris as a Nix package?

If that's what this is, I think it should be a separate PR.

quantenzitrone:
> what are they for?

The `flake.nix` defines a "development shell", an environment with every needed dependency for developing Lutris.
If you have nix installed (and enabled flakes), you can then just run `nix develop` to enter a shell with this environment. The dependencies are dropped automatically when exiting this shell.

> I wouldn't think any lock file would be under source control ever

Lock files should almost always be under source control, because they ensure reproducibility, defining the exact versions of dependencies that are known to work. I only know rust and nix lockfiles tho, so maybe that's language specific.

> Looking at these files- is this support for delivering Lutris as a Nix package?

No, there already is a nix package for Lutris. The nix flake here is not needed for anything, it's only a convenience for everyone that is using the nix package manager and wants to contribute to lutris.

danieljohnson2:
I see. This is (kinda) like the ``.idea`` directory for PyCharm, or the ``.vs`` directory for Codium; it's the configuration for your development environment.

I notice that my ``.idea`` directory has its own ``.gitignore``; I think  JetBrains intends this to be under source control, except for the bits listed there.

But we don't do that; we list ``.idea`` in our top level ``.gitignore`` to keep the JetBrains cooties out of GitHub. I think we might list these files Nix in ``.gitignore`` too; you could continue to use them without sharing them then.

But in any case they don't belong in *this* PR. By putting them into git like this, you are asking the rest of us to maintain these files when we change the dependencies that Lutris has. But I don't know a thing about Nix and I'm not competent to do this.

Let's drop the files from this PR, and either add them to ``.gitignore`` or alternately pop a separate PR to add these files to Lutris.

quantenzitrone:
Alright, I removed the nix stuff other than my fix for developing on NixOS and added `flake.nix` and `flake.lock` to `.gitignore`.

I also combined the other 2 commits into one and added an explanation on how the current logic works.

danieljohnson2:
OK, this looks like the code I tested, and the flakey files are out, so I think I can merge this for you now.

PR #5555, opened 2024-08-03
Improve error handling

JassonCordones:
Improves error handling in some places, mainly in steam service. There was a bug when trying to reload games.
Minor typo fix.

### DIFF
--- a/bin/lutris
+++ b/bin/lutris
@@ -47,10 +47,10 @@ try:
         gettext.bindtextdomain("lutris", optional_settings.LOCALE_DIR)
         locale.textdomain("lutris")
         gettext.textdomain("lutris")
-    except:
+    except Exception as ex:
         sys.stderr.write(
             "Couldn't bind gettext domain, translations won't work.\n"
-            "LOCALE_DIR: %s\n" % optional_settings.LOCALE_DIR
+            "LOCALE_DIR: %s\nError: %s\n" % (optional_settings.LOCALE_DIR,ex)
         )
 except ImportError:
     pass
@@ -58,7 +58,11 @@ except ImportError:
 if "WEBKIT_DISABLE_DMABUF_RENDERER" not in os.environ:
     os.environ["WEBKIT_DISABLE_DMABUF_RENDERER"] = "1"
 
-from lutris.gui.application import Application  # pylint: disable=no-name-in-module
+try:
+    from lutris.gui.application import Application  # pylint: disable=no-name-in-module
+except ImportError as ex:
+    sys.stderr.write("Error importing Lutris application module: %s\n" % ex)
+    sys.exit(1)
 
 app = Application()  # pylint: disable=invalid-name
 sys.exit(app.run(sys.argv))
--- a/lutris/services/steam.py
+++ b/lutris/services/steam.py
@@ -201,8 +201,8 @@ class SteamService(BaseService):
                 try:
                     steam_game.uninstall()
                     stats["removed"] += 1
-                except Exception as e:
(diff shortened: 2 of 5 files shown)


lutris/util/jobs.py
```diff
@@ -15,12 +15,12 @@ def __init__(self, func, callback, *args, **kwargs):
...
         self.stop_request = threading.Event()
-
-        super().__init__(target=self.target, args=args, kwargs=kwargs)
         self.function = func
         self.callback = callback if callback else lambda r, e: None
-        self.daemon = kwargs.pop("daemon", True)
 
+        super().__init__(target=self.target, args=args, kwargs=kwargs)
+        #self.daemon = kwargs.pop("daemon", True)
+        self.setDaemon(kwargs.pop("daemon", True))
```
danieljohnson2, line 23: I think this change is wrong; we support back to Python 3.7, and by then setDaemon had been deprecated. You are supposed to set the self.daemon property.

So why setDaemon?

Also, moving the call to ``super().__init__`` later seems harmless, but why?

bin/lutris
```diff
@@ -47,18 +47,22 @@ try:
...
 
 if "WEBKIT_DISABLE_DMABUF_RENDERER" not in os.environ:
     os.environ["WEBKIT_DISABLE_DMABUF_RENDERER"] = "1"
 
-from lutris.gui.application import Application  # pylint: disable=no-name-in-module
+try:
+    from lutris.gui.application import Application  # pylint: disable=no-name-in-module
+except ImportError as ex:
+    sys.stderr.write("Error importing Lutris application module: %s\n" % ex)
+    sys.exit(1)
```
danieljohnson2, line 65: Why do we want to catch this exception? We're going to exit if this fails in any case. What does this handler add?

lutris/services/steam.py
```diff
@@ -127,29 +127,33 @@ def install_from_steam(self, manifest):
...
+                installed=1,
+                installer_slug=lutris_game_id,
+                configpath=configpath,
+                platform="Linux",
+                service=self.id,
+                service_id=appid,
+            )
+            return slug
+        except Exception as e:
+            logger.error("Failed to install from Steam: %s", str(e))
```
danieljohnson2, line 155: I don't think we need ``str(ex)`` here- just ``ex`` should work. That's what we usually have for ``logger.error``.

danieljohnson2 (changes requested):
I like most of this- it's mostly log-and-continue-to-the-next-game handling, which seems helpful. But I have a few concerns before I'll merge this.

The big one is switching us over to a deprecated method in jobs.py. I don't like that.

bin/lutris
JassonCordones: Here I was just trying to make the import error explicit, but it's true the program will exit anyways if you want I can revert it back.

lutris/util/jobs.py
JassonCordones: I'll change it back sorry, I was not aware. 

For moving the call to `super().__init__` later , honestly I just liked how it looked better. I can revert it back too if you want.

lutris/services/steam.py
JassonCordones: I'll make the change

danieljohnson2:
I will merge if we just go back to self.daemon instead of self.setDaemon; but I'd make those other changes too if it were my PR.

danieljohnson2 (approved):

danieljohnson2:
Merged it. Thanks for your contribution!

Issue #4882, opened 2023-06-01
Remove all display configuration options


strycore:
With Wayland taking over Xorg, the options we have for setting resolutions or turning off some displays will no longer work.

Users who have specific needs for display configurations would be better creating pre-launch and post-launch scripts that are tailored for their displays.

For everyone else, there is Gamescope. 

Options to remove:

- Restore resolution on game exit
- Restore gamma on game exit
- SDL 1.2 fullscreen monitor
- Turn off monitors except
- Switch resolution to

Also to deprecate, Xephyr support (again, in favor of Gamescope)

- Use Xephyr
- Xephyr resolution
- Xephyr fullscreen

danieljohnson2:
Yikes! I realize that Wayland is the future, but it is really the present?

My efforts to use it in the past always ended with a conclusion of 'not good enough'- and I have not been able to get the current Pop!_OS to use Wayland *at all*.

I think it's too soon to be dropping X11 support.

strycore:
This isn't about dropping support for X11, it's about dropping options that are, in most cases, broken.
Gamescope is perfectly usable in Xorg and I don't remember when was the last time any of the soon to be deprecated options were useful.

danieljohnson2:
I've used some of those settings for older games.

Looks to me like gamescope isn't available on Pop!_OS, and it appears to have problems with NVidia graphics from what I read.

May I ask what distro you use? Maybe Pop!_OS is no longer a good choice for gaming on Linux?

strycore:
Managing to get a build of Gamescope can be a struggle on Pop OS. I remember I managed to get a deb from here https://launchpad.net/~ar-lex/+archive/ubuntu/gamescope because I couldn't get recent versions of Gamescope to build.

I had Pop on my Surface Pro but after struggling to get gamescope to work on Intel GPUs, I gave Nobara a try.
I also had Pop on my main gaming machine but after installing a NVMe drive in it, I also installed Nobara.
I was really happy with Nobara so I ordered another NVMe, installed it in the Lutris demo machine and put Nobara on it, replacing Manjaro.
I have a workstation running Pop, but I've transferred all its content to a Ubuntu Qemu VM running on the above Nobara machine. I haven't used the workstation since then.

(short answer: Nobara)

strycore:
Here's Cyberpunk 2077 running on Nvidia, Wayland and Gamescope

[...]

PR #5566, opened 2024-08-13
Hide various X11-only  settings when not on X11

danieljohnson2:
This is in support of #4882; these settings are (I understand) never going to work on Wayland and will eventually disappear. They should disappear for Wayland users right now, since it's just confusing to offer options that don't work. Like so:

image

The settings that disappear are:

- Restore resolution on game exit
- Disable desktop effects
- SDL 1.2 Fullscreen Monitor
- Turn off monitors except
- Switch resolution to
- All Xephyr settings

These are the ones mentioned as doomed in #4882, plus 'disable desktop effects'- I'm pretty sure we can't kill the compositor on Wayland, so let's hide that too. "Restore gamma on game exit" is no longer included in Lutris at all, so no need to hide it.

I suspect there are other settings that won't work on Wayland too, but I don't know what they are.

I think this PR is a simple, safe step to take on the road to actually removing all these settings.

### DIFF
--- a/lutris/gui/config/boxes.py
+++ b/lutris/gui/config/boxes.py
@@ -149,6 +149,13 @@ class ConfigBox(VBox):
                 option_key = option["option"]
                 value = self.config.get(option_key)
 
+                if "visible" in option:
+                    if callable(option["visible"]):
+                        option["visible"] = option["visible"]()
+
+                    if not option["visible"]:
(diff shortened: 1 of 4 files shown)


strycore:
Great, this means those options are mostly gone now. Except for the retro ppl who are still running X

PR #5582, opened 2024-09-02
Display Trace-Back and Logs for Errors in the InstallerWindow and ErrorDialog, and provide links to GitHub and Discord.

danieljohnson2:
When the InstallerWindow displays errors in a page, instead of a separate dialog, we will now show the trace-back and the log:

Screenshot from 2024-09-02 13-09-10

This gives you the trace-back and then the Lutris log (which usually contains the trace-back, but better safe that sorry). The message directs the user to GitHub or Discord, and this way they can easily provide the information we need to help them.

I've also updated the ErrorDialog to include the log in the same way, but no new links in there yet.

The downside here is just that all this tech stuff may be scary for users.  The ErrorDialog hides this stuff initially- you have to ask for it.

Still, I think Linux gamers can handle it, and getting better error feedback from users is worthwhile.

### DIFF
--- a/lutris/gui/config/sysinfo_box.py
+++ b/lutris/gui/config/sysinfo_box.py
@@ -1,4 +1,3 @@
-import os
 from gettext import gettext as _
 from typing import Dict, Iterable, List
 
--- a/lutris/gui/config/sysinfo_box.py
+++ b/lutris/gui/config/sysinfo_box.py
@@ -8,7 +8,7 @@ from lutris.gui.config.base_config_box import BaseConfigBox
 from lutris.gui.widgets.log_text_view import LogTextView
 from lutris.util import linux, system
 from lutris.util.linux import gather_system_info_dict
-from lutris.util.log import LOG_FILENAME
+from lutris.util.log import get_log_contents
 from lutris.util.strings import gtk_safe
 from lutris.util.wine.wine import is_esync_limit_set, is_fsync_supported, is_installed_systemwide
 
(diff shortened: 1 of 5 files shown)


strycore:
Can we add a more user friendly message than "Error details" for all the people who are not Python developers (there are quite a lot) and have no idea what a stacktrace is or what to do with it?

danieljohnson2:
Sure; that's the text used in the ``ErrorDialog``, so I just used that.

What would be better here? Should I change ``ErrorDialog`` also?

I could just say "trace back" but few users would know what that is, I expect.

strycore:
We should also share the same wording with the other error popup and avoid coding terms like stacktrace, traceback or similar. 
Something along the lines of "And unexpected error has occured, please share the details below with the Lutris team on Github or Discord"

danieljohnson2:
I see. I think that will work here. I can provide a clickable link to Lutris on GitHub here; should I? And is there a way to do that for Discord?

This style won't work ``ErrorDialog`` which is used all over the place for everything in sight. I'll leave that for later.

strycore:
For Discord you can use the URL we have on the website, it's a permanent invite to the server.

danieljohnson2:
I've revised the error message and added links to it, and I noticed we can provide the log in the Preferences dialog- so I've stuck that on the end here, and also in ErrorDialog. That's a lot of what we want users to provide us for tech support.

I'll think about how to rephrase the ErrorDialog text so it can also include these links.

danieljohnson2:
OK, I've also updated the ``ErrorDialog`` with some very generic "You can get support" text that says nothing much about the error, but provides the links. See the PR description for a screenie.

This stuff is linked to the details expander; an ``ErrorDialog`` that just displays a (string) message will show none of this, but these uses are mostly things like "Please fill in the name".

strycore (approved):

danieljohnson2:
Sempai approves! Yay!

strycore:
yes, good to merge

PR #5595, opened 2024-09-10
If there is no banner, then replace it with cover image

DisDis:
I think it looks better,
(New version)
image

than the old version
image

### DIFF
--- a/lutris/services/base.py
+++ b/lutris/services/base.py
@@ -42,9 +42,7 @@ class LutrisBanner(ServiceMedia):
 
     def get_possible_media_paths(self, slug: str) -> List[str]:
         paths = super().get_possible_media_paths(slug)
-        cover_paths = [
-            os.path.join(settings.COVERART_PATH, pattern % slug) for pattern in LutrisCoverart.file_patterns
-        ]
+        cover_paths = [os.path.join(settings.COVERART_PATH, pattern % slug) for pattern in LutrisCoverart.file_patterns]
         return paths + cover_paths
 
 
--- a/lutris/services/base.py
+++ b/lutris/services/base.py
@@ -43,7 +43,7 @@ class LutrisBanner(ServiceMedia):
     def get_possible_media_paths(self, slug: str) -> List[str]:
         paths = super().get_possible_media_paths(slug)
         cover_paths = [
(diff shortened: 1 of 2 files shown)


danieljohnson2:
Well, I have two concerns here. One is that if we do this, people will report this as a defect; it looks rather like it's broken to my eye.

But the other is that this PR downloads the cover art (a second time) to the banners directory; I think it should instead display the cover when the banner is missing, using the file we have already.

The different his make is what happens if Lutris later does supply the banner. If we've downloaded the cover-as-a-banner, Lutris won't even try to download the banner again. I'd like it to do so, which would happen if we didn't have the file in the banners directory.

DisDis:
@danieljohnson2 I agree and this was my first solution, but I couldn't find the logic where the renderer chooses the image :(. Maybe you can just point me to the file and I'll change the MR

danieljohnson2:
I think you are looking for the method ``get_possible_media_paths`` of ``ServiceMedia``; this returns a list of paths to media files, and the cell-renderer picks the first file that it finds.

You can implement this on ``LutrisBanner``, something like this:
```python
    def get_possible_media_paths(self, slug: str) -> List[str]:
        paths = super().get_possible_media_paths(slug)
        cover_path = something something
        return paths + [cover_path]
```

DisDis:
@danieljohnson2 thx, I changed the way

danieljohnson2:
Almost there.

But strictly speaking ``self.patterns`` are the patters for the banner, which just happen to be exactly the same as ``LutrisCoverart.patterns`` - but really should just use the later form in case we change it in future.

You don't need to get an instance of ``LutrisCoverart`` - this is a class field, so ``LutrisCoverart.patterns`` ought to work fine.

It wouldn't be wrong to use ``LutrisCoverart.dest_path`` instead of ``settings.COVERART_PATH`` for the same reason, though I'm skeptical we'll change that one.

DisDis:
@danieljohnson2 done

danieljohnson2:
Alright, I think it's okay now. Not sure I like what it does, but I'll let the boss decide.

strycore:
We are absolutely not doing that. I'm sorry you did all the work but squeezing covers inside the banner space just makes them unreadable and makes the visual experience inconsistent.

PR #5602, opened 2024-09-14
Add libayatana-appindicator based tray icon

jackj1:
Gentoo dropped the libappindicator package from its main repository in favour of libayatana-appindicator which broke the status tray, I simply added a check for libayatana-appindicator if libappindicator is not found.

### DIFF
--- a/lutris/gui/widgets/status_icon.py
+++ b/lutris/gui/widgets/status_icon.py
@@ -1,4 +1,4 @@
-"""AppIndicator based tray icon"""
+"""AppIndicator/AyatanaAppIndicator based tray icon"""
 
 from gettext import gettext as _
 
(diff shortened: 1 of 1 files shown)


strycore:
Thank you for this clear and concise patch!

godlike64:
I found this issue as I was looking into why games started via Lutris (namely, Battle.net) show a small window with the tray icon now, instead of the tray icon being in the, you know, tray :). The last update I did was to update to KDE 6 but I don't know if it is related to this. Anyway I tested the patch, and it works (assuming you previously started xembedsniproxy, which is provided by plasma-workspace).

strycore:
I haven't had any problems with the tray icons on Plasma 6 and Nobara but I do remember the Wine tray icons showing in their own separate window on Gnome based distros without the correct extensions installed. Hope this patch improves tray icon for Gnome as well!

PR #5615, opened 2024-09-22
Add a "Show in games" command in the context menu for service games

### LINKED ISSUE #5606: Ability to add categories to games which aren't currently installed
When going through my list of Steam games I don't want to install a game just to be able to list it under a category.

danieljohnson2:
This command creates the (uninstalled) Lutris game, and links the games to each other.

This won't overwrite an existing game; if a game with the right slug is found, we'll just link them. If the existing game already is linked to a service game for some other service, we'll show an error and not change anything.

This command is supported for multi-selection, so you can do it for your whole library. That is a bit dicey if you have the same game from different services- this code won't create two games for the same slug, so you'll have an awkward time adding games one at a time. A more elaborate UI could mitigate this, but I think this simple version is still useful.

The point of all this is to help users categorize games that are just coming from their services. To this end, I've enabled the 'Categories' command for service games, and it prompts you with this:
image
If the user clicks 'Yes', then we add or link the games, and then when that is done we pop the Categories UI as usual.

Resolves #5606.

### DIFF
--- a/lutris/database/services.py
+++ b/lutris/database/services.py
@@ -29,3 +29,23 @@ class ServiceGameCollection:
         if len(results) > 1:
             logger.warning("More than one game found for %s on %s", appid, service)
         return results[0]
+
+    @classmethod
+    def link_service_game(cls, service, appid, lutris_slug):
+        """Updates a service-game to have the corresponding Lutris slug."""
+        sql.db_update(
+            settings.DB_PATH,
+            "service_games",
+            {"lutris_slug": lutris_slug},
+            conditions={"appid": appid, "service": service},
+        )
+
+    @classmethod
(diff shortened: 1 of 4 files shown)


strycore:
This is totally unacceptable. We can't expose our internal database structure to users and ask them question about it. 
This prompt only makes sense to people who have spent several years coding on Lutris, to everyone else, it's gibberish.

If we can't make good decisions and handle things without asking users weird questions then we shouldn't be coding features at all.

strycore:
Just remove the prompt and your patch will be fine.

strycore:
I mean, remove the "Show in games" as well, obviously.

danieljohnson2:
Guess it's not as funny when I do it!

PR #5616, opened 2024-09-24
Convert saved searches to advanced search / dynamic collections

strycore:
It's cool that we're now able to do more advanced search but I'd like to adjust the recent saved searches: 

- to have a more general purpose, making it easier to do advanced search without saving the search
- to be more discoverable. Having the advanced search in the search bar is where users will expect this feature to be. Users will have a hard time finding / figuring what is "Add saved search" in a menu
- to be more clear and concise by removing terms like "omit from search" and "maybe", which may be confusing to the user.

The goal is to have something that works in a similar fashion as the Steam advanced filters. 

image

Creating a Dynamic Collection is proposed but it's not the main feature of this dialog. The main feature is just searching.

TODO:

- [ ] Show advanced search when a dropdown on the right hand side of the search bar is clicked.
- [ ] Remove "Add Saved search" from menu
- [ ] Maybe simplify the UI more, some of those Yes/No dropdowns could be checkboxes?
- [ ] Move the action buttons from the dialog to the box widget so we don't depend on  being in the dialog
- [ ] Show the dynamic categories with the other categories but add an icon next to it. 
- [ ] Center the search on the dialog and try to group the 2 entry + menu button widgets

### DIFF
--- a/lutris/gui/config/edit_saved_search.py
+++ b/lutris/gui/config/edit_saved_search.py
@@ -17,17 +17,17 @@ from lutris.search_predicate import AndPredicate, SearchPredicate, format_flag
 
 
 class EditSavedSearchDialog(SavableModelessDialog):
-    """A dialog to edit saved searches."""
+    """A dialog to edit dynamic categories"""
 
     def __init__(self, parent, saved_search: SavedSearch) -> None:
         self.saved_search = copy(saved_search)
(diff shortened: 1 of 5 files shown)


strycore:
No time to finish this. We need to ship. This will be completed in next release.

Issue #3519, opened 2021-04-22
Deb extraction when zstd compressed


stephanlachnit:
Currently Lutris hard-codes the compression method for `.deb` files here:
https://github.com/lutris/lutris/blob/5637f25c1802c4b26f7868c78ca1877897f21f87/lutris/util/extract.py#L210-#L224

However, `.deb`s may also be compressed with zstd, see https://bugs.launchpad.net/ubuntu/+source/lutris/+bug/1923845.

Lutris should detect the compression method used and use the according extractor.

noahboa27:
Hello @strycore I would like to pick this up and work on it if it's still wanted. Thanks!

strycore:
not sure if it's still wanted but go on, I don't think anyone is working on that right now.

TriMoon:
Does this still need to stay open?

strycore:
It is up to the community to close their own tickets. I've personally stopped trying to do any kind of ticket management, it was too draining.

strycore:
it is the mark of a great open source project to have thousands of open tickets most of which referring to issues that have been fixed years ago 😎

TriMoon:
> it is the mark of a great open source project to have thousands of open tickets most of which referring to issues that have been fixed years ago 😎

Yea sure, that way it is obvious that the project is a spaghetti that other developers won't touch because they don't know what to pursue.
"It's full of bugs, just leave it and move to another more maintained project to help.."

danieljohnson2:
> Yea sure, that way it is obvious that the project is a spaghetti that other developers won't touch because they don't know what to pursue. "It's full of bugs, just leave it and move to another more maintained project to help.."

It's not so bad. It's got a few idiosyncrasies. All that callback-based asynchrony is a bit tricky to follow.

But the thing is, it is *someone else's code*. Reading someone else's code is one of the really key skills a developer needs to have, and many people struggle with it.

Customer support is a different thing. Users often expect to be supported as if they were customers even when they are not, but no-one is all that motivated to actually do it. I know I let a lot of GitHub issues just go by.

strycore:
I do try to understand other peoples code in other projects. In Lutris, if I don't understand a piece of code, it gets instantly deleted or rewritten. I don't want code I have a hard time understanding in this project.

Hopefully, this also makes the code easier to understand to everyone else.
[...]

PR #5627, opened 2024-10-01
Reformatting

noahboa27:
Solution for issue https://github.com/lutris/lutris/issues/3519 was already implemented here https://github.com/noahboa27/lutris-fork/commit/49355ba5e74d4244b76b985796db6223254a5112.

Decided to do a little reformatting while I was here anyway. Removed an unused method, reordered code to make it read better, and added an underscore prefix to all private functions.


strycore:
Thanks!

Issue #5611, opened 2024-09-21
Downloading GOG game has several issues for me


BileDemon:
I'm trying to download Cyberpunk 2077 (2.13) from GOG for hours and hours on end and it never works unfortunately. This has worked in the past (mostly) without problems.

The download always starts at part 27 and it proceeds until parts 15 (or 13), where it gets a 403 error and the downloader just breaks with an exception. There is no way to make it continue, it always shows the exception. I tried downloading those files from the GOG website directly and it works just fine.

Also setting the download cache folder doesn't do anything, because it always replaces the files that have successfully been downloaded previously, again starting with part 27 and progressing backwards until it gets a 403 error. It never recognizes the earlier download progress.

Maybe it has something to do with another issue that it has: The downloaded parts all have wrong filenames, like:
"setup_cyberpunk_2077_2.13_%2864bit%29_%2876205%29-27.bin"
instead of
"setup_cyberpunk_2077_2.13_(64bit)_(76205)-27.bin"

I know for certain that in the past Lutris allowed me to select which files I wanted to download separately and in which order, so this must have been removed in favor of downloading one file at a time and not being able to easily select the files.

This is the log output:
`DEBUG    2024-09-21 09:42:47,613 [installerwindow.present_extras_page:522]:Showing extras page
INFO     2024-09-21 09:42:50,982 [installer.prepare_game_files:168]:Getting files for setup
INFO     2024-09-21 09:42:50,982 [gog.get_game_details:284]:Getting game details for 1423049311
DEBUG    2024-09-21 09:42:51,004 [http._request:88]:GET https://api.gog.com/products/1423049311?expand=downloads&locale=de-DE
INFO     2024-09-21 09:42:51,059 [gog.get_download_info:291]:Getting download info for https://api.gog.com/products/1423049311/downlink/installer/en1installer0
DEBUG    2024-09-21 09:42:51,091 [http._request:88]:GET https://api.gog.com/products/1423049311/downlink/installer/en1installer0
INFO     2024-09-21 09:42:51,286 [gog.get_download_info:291]:Getting download info for https://api.gog.com/products/1423049311/downlink/installer/en1installer1
DEBUG    2024-09-21 09:42:51,326 [http._request:88]:GET https://api.gog.com/products/1423049311/downlink/installer/en1installer1
INFO     2024-09-21 09:42:51,531 [gog.get_download_info:291]:Getting download info for https://api.gog.com/products/1423049311/downlink/installer/en1installer2
[...]

Issue #5599, opened 2024-09-12
U Boat fails to donwload


phodina:
### Bug description

I bought on GoG the game [U-boat](https://www.gog.com/en/game/uboat).

It's listed in my account. I wanted to install it so clicked on it, selected Wine install. And then the download started ~24G.

However around 20 GB it stops and throws a dialog.

```
1 parameters needed for signal error; 0 given

Traceback (most recent call last):

  File "/nix/store/f8ql8n2a075r3rzkd7j1l6m8gha33sgi-lutris-unwrapped-0.5.17/lib/python3.12/site-packages/lutris/exception_backstops.py", line 108, in error_wrapper
    return handler(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^

  File "/nix/store/f8ql8n2a075r3rzkd7j1l6m8gha33sgi-lutris-unwrapped-0.5.17/lib/python3.12/site-packages/lutris/gui/widgets/download_collection_progress_box.py", line 161, in _progress
    self.emit("error")

TypeError: 1 parameters needed for signal error; 0 given1 parameters needed for signal error; 0 given
```

### How to Reproduce

1. Go to 'GoG'
2. Click on 'Uboat'
3. Do 'Install'
4. See 'Dialog with exception during download'

### Expected behavior

Game installs

### Log output

```shell
$ lutris -d
Error: unable to open display 
2024-09-12 17:32:40,026: ['glxinfo', '-B'] command failed: Command '['glxinfo', '-B']' returned non-zero exit status 255.
2024-09-12 17:32:40,026: No available glxinfo output
2024-09-12 17:32:40,027: Invalid glxinfo received
INFO     2024-09-12 17:32:40,826 [application.do_command_line:479]:Starting Lutris 0.5.17
ERROR: [Loader Message] Code 0 : vkCreateInstance: Found no drivers!
Cannot create Vulkan instance.
This problem is often caused by a faulty installation of the Vulkan driver or attempting to use a GPU that does not support Vulkan.
ERROR at /build/source/vulkaninfo/./vulkaninfo.h:458:vkCreateInstance failed with ERROR_INCOMPATIBLE_DRIVER
ERROR    2024-09-12 17:32:40,975 [system.read_process_output:188]:['vulkaninfo', '--summary'] command failed: Command '['vulkaninfo', '--summary']' returned non-zero exit status 1.
[...]

PR #5631, opened 2024-10-03
Skip downloading cached files

DanBmh:
Adds a check if an installation file is already existing in the cache and skips downloading it then.

My problem was that the download failed midways and I could not resume it anymore. I always needed to restart the complete download, wait until it retrieved all the already existing files, only to find out that it failed again after some hours. So this was wasting time and energy/bandwidth...

Note that the current approach might cause problems if a cached file is broken/incomplete because it is skipped as well.
This can happen for example if Lutris is started from the command line and the download is aborted by canceling the program directly (if the download is aborted via the GUI, the intermediate file is deleted correctly).

### DIFF
--- a/lutris/gui/widgets/download_collection_progress_box.py
+++ b/lutris/gui/widgets/download_collection_progress_box.py
@@ -119,10 +119,16 @@ class DownloadCollectionProgressBox(Gtk.Box):
             self.start()
             return
 
+        # Use a temporary file to avoid problems with partially downloaded files
+        tmp_path = file.dest_file + ".tmp"
+        file.tmp_file = tmp_path
+        if os.path.exists(tmp_path):
+            os.remove(tmp_path)
+
         self.update_download_file_label(file.filename)
         if not self.downloader:
             try:
-                self.downloader = Downloader(file.url, file.dest_file, referer=file.referer, overwrite=True)
+                self.downloader = Downloader(file.url, file.tmp_file, referer=file.referer, overwrite=True)
             except RuntimeError as ex:
                 display_error(ex, parent=self.get_toplevel())
                 self.emit("cancel")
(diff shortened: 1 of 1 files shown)


DanBmh:
This should also help with the installation problems in https://github.com/lutris/lutris/issues/5611, https://github.com/lutris/lutris/issues/5599. \
It doesn't fix the error, but allows to restart the download, without having to re-download everything again. \
I had it as well, and around after the same time again on my second try, so maybe its some timeout problem.

BileDemon:
This is great! Thanks for implementing this feature.
I hope it gets merged before the next milestone release.

danieljohnson2:
You seem to have accidentally added some unrelated commits that update readme files. Maybe those should be in another PR.

I am also concerned about robustness in the face of a crash- a half downloaded file would not be re-downloaded, so if we fail to delete it (and we would in a crash) we'd wind up unable to install that game afterwards.

I think we need a way to recognize incomplete files; say, we could download to a temp file and then rename on completion- something like that.

DanBmh:
> I think we need a way to recognize incomplete files; say, we could download to a temp file and then rename on completion- something like that.

Good idea, I've updated it.

danieljohnson2:
Better, but I do not feel good using a fixed name ('download.tmp') for this- if we do parallel downloads this will fail.

The ``tempfile`` module seems awfully awkward though. How about we just take ``file.dest_file`` and pop a suffix on the end? That is simple, and it lets you see what the temp file will become when it grows up.

It can still fail if somehow we wind up downloading the same file twice at the same time, but I don't think that's going to work out too well anyway.

danieljohnson2:
I've tested this and it seems to work; I think we can merge this.

I may be asking for a paddlin' from @strycore, but I think I will merge this as a 'fix' for #5611 and #5599.

DanBmh:
> I may be asking for a paddlin' from @strycore, but I think I will merge this as a 'fix' for https://github.com/lutris/lutris/issues/5611 and https://github.com/lutris/lutris/issues/5599.

I think a 'solution' for those could be to reauthenticate or regenerate the download links before each new file. Since I only needed to restart the download, I don't think the problem is within lutris, but rather that the links/downloads are only valid for a certain amount of time (I normally was able to download about 4 files (~18gb, 1h) before it failed).

PR #5634, opened 2024-10-05
Small Tweaks to Desktop/Application Launcher entries

mihawk90:
This incorporates two small changes:

1. Use the `TryExec` key for generated Desktop Entries
2. Remove the executable bit from Desktop Entries, since it is not required.

---

## 1. `TryExec` key
Generated desktop entries are non-functional when Lutris is not installed (e.g. after user uninstalls Lutris or reinstalls their OS).
We can use `TryExec` to hide such entries. Upon reinstallation of Lutris they will automatically be shown again with no further action by users.

> [`TryExec`](https://specifications.freedesktop.org/desktop-entry-spec/latest/recognized-keys.html)
>
> Path to an executable file on disk used to determine if the program is actually installed. If the path is not an absolute path, the file is looked up in the $PATH environment variable. If the file is not present or if it is not executable, the entry may be ignored (not be used in menus, for example).

Note this only takes a single executable as a value. For that reason it is not possible to use this key for Flatpaks. The presence of `flatpak` does not guarantee the Lutris Flatpak actually being installed.
The use of `flatpak run net.lutris.Lutris` validates in `desktop-file-validate`, however KDE at least does not show the entry.

An empty value also validates and doesn't hide the entry, so leaving it for Flatpaks appears fine.

## 2. Executable bit
Desktop entries don't need to be executable since they are not actually executed by the application launcher, but rather metadata for an installed binary.

Executing a desktop files also merely results in a lot of junk output on the terminal.

Additionally, any permission that is not required should not be granted, see Principle of least Privilege.

### DIFF
--- a/lutris/util/xdgshortcuts.py
+++ b/lutris/util/xdgshortcuts.py
@@ -65,6 +65,8 @@ def create_launcher(game_slug, game_id, game_name, launch_config_name=None, desk
     # field code in the Exec key.
     command = f"{lutris_executable} {shlex.quote(url)}".replace("%", "%%")
 
+    try_exec = "" if LINUX_SYSTEM.is_flatpak() else "lutris"
+
     launcher_content = dedent(
         """
         [Desktop Entry]
@@ -73,7 +75,8 @@ def create_launcher(game_slug, game_id, game_name, launch_config_name=None, desk
         Icon={}
         Exec=env LUTRIS_SKIP_INIT=1 {}
         Categories=Game
-        """.format(game_name, f"lutris_{game_slug}", command)
+        TryExec={}
+        """.format(game_name, f"lutris_{game_slug}", command, try_exec)
     )
 
     launcher_filename = get_xdg_basename(game_slug, game_id)
--- a/lutris/util/xdgshortcuts.py
+++ b/lutris/util/xdgshortcuts.py
@@ -86,7 +86,7 @@ def create_launcher(game_slug, game_id, game_name, launch_config_name=None, desk
         tmp_launcher.close()
     os.chmod(
         tmp_launcher_path,
-        stat.S_IREAD | stat.S_IWRITE | stat.S_IEXEC | stat.S_IRGRP | stat.S_IWGRP | stat.S_IXGRP,
+        stat.S_IREAD | stat.S_IWRITE | stat.S_IRGRP | stat.S_IWGRP,
     )
 
     if desktop:


RayJW:
Wouldn't the best way to handle this be the [DynamicLauncher portal](https://docs.flatpak.org/en/latest/portal-api-reference.html#gdbus-org.freedesktop.portal.DynamicLauncher) to create the desktop entries.

Also, the `TryExec` field can be handled in Flatpak. This is how the MS Edge Flatpak does it for “installed” web apps: `TryExec=/var/lib/flatpak/exports/bin/com.microsoft.Edge`.

> It is recommended to include a TryExec= line with either a binary name or an absolute path. The launcher will be deleted if the TryExec binary cannot be found on session start.

mihawk90:
> Wouldn't the best way to handle this be the [DynamicLauncher portal](https://docs.flatpak.org/en/latest/portal-api-reference.html#gdbus-org.freedesktop.portal.DynamicLauncher) to create the desktop entries?

Possibly, but that would have to be implemented separately from the ground up. Knowing how other portals work it's probably more involved and not a simple change to make (i.e. requesting permission and all that fun stuff). I certainly have no idea where to even start implementing that.

> Also, the TryExec field can be handled in Flatpak. This is how the MS Edge Flatpak does it for “installed” web apps: `TryExec=/var/lib/flatpak/exports/bin/com.microsoft.Edge`.
> 
> > It is recommended to include a TryExec= line with either a binary name or an absolute path. The launcher will be deleted if the TryExec binary cannot be found on session start.

Hm interesting... It seems that quote is [from the DynamicLauncher portal documentation](https://docs.flatpak.org/en/latest/portal-api-reference.html#gdbus-org.freedesktop.portal.DynamicLauncher:~:text=It%20is%20recommended%20to%20include%20a%20TryExec%3D%20line%20with%20either%20a%20binary%20name%20or%20an%20absolute%20path.%20The%20launcher%20will%20be%20deleted%20if%20the%20TryExec%20binary%20cannot%20be%20found%20on%20session%20start.), but they don't go into detail what you're supposed to put there. "a binary or absolute path" sure, I already knew that, but it doesn't explain what path I should use for the Flatpak in question. It should also be noted though that while Portals were initially developed for Flatpaks, they are multi-purpose and any application can use them, be they Flatpak or native (or god forbid, Snap). So this paragraph might just not apply to Flatpaks.

"The launcher will be deleted if the TryExec binary cannot be found on session start" is just not true though. A spec compliant launcher should not (and in most cases *can* not) delete a desktop entry, it can only hide it. According to the spec (see above):

>  If the file is not present or if it is not executable, the entry **may be ignored** (not be used in menus, for example).

But that aside, absolute paths are problematic in that we can't guarantee a user installs their Flatpaks globally. If a user is doing a per-user install (i.e. `flatpak install --user <remote> <name>`), that absolute path in `TryExec` will fail and the entry would be hidden, making the appearance as if the feature did not work.
Even if the user does install globally, we cannot guarantee that this is the path their distro is using. The Flatpak documentation says:
> System-wide installations are installed in the system's installation directory, which is **usually** `/var/lib/flatpak`.

AFAIK the spec has no provision on where the install has to go, it's configurable for the distro (although I don't know any distro that does, but that's besides the point). Users can even set up [their own custom directory](https://docs.flatpak.org/en/latest/tips-and-tricks.html#adding-a-custom-installation).

Between the two, IMO it is preferable to *always* show the entry (current behaviour) then *never* show it.

RayJW:
> Possibly, but that would have to be implemented separately from the ground up. Knowing how other portals work it's probably more involved and not a simple change to make (i.e. requesting permission and all that fun stuff). I certainly have no idea where to even start implementing that.

It actually seems quite easy compared to the handling of the different cases, but I do not have time to make a PR right now, so obviously I'm not in a position to tell others what to code, just so you know that's not my intention. From [this](https://flatpak.github.io/xdg-desktop-portal/docs/convenience-libraries.html) it would seem like [libportal](https://github.com/flatpak/libportal/) would be the way to go as it has Python bindings and supports GTK 3, and also GTK 4 depending on the future of Lutris.

> Hm interesting... It seems that quote is [from the DynamicLauncher portal documentation](https://docs.flatpak.org/en/latest/portal-api-reference.html#gdbus-org.freedesktop.portal.DynamicLauncher:~:text=It%20is%20recommended%20to%20include%20a%20TryExec%3D%20line%20with%20either%20a%20binary%20name%20or%20an%20absolute%20path.%20The%20launcher%20will%20be%20deleted%20if%20the%20TryExec%20binary%20cannot%20be%20found%20on%20session%20start.), but they don't go into detail what you're supposed to put there. "a binary or absolute path" sure, I already knew that, but it doesn't explain what path I should use for the Flatpak in question. It should also be noted though that while Portals were initially developed for Flatpaks, they are multi-purpose and any application can use them, be they Flatpak or native (or god forbid, Snap). So this paragraph might just not apply to Flatpaks.

Definitely, the reason I brought this up is that I think the portal should completely replace the handling to ensure mitigating possible future compatibility problems with different desktops / versions. From the [Arch Wiki](https://wiki.archlinux.org/title/XDG_Desktop_Portal): 

> Portals were designed for use with applications sandboxed through [Flatpak](https://wiki.archlinux.org/title/Flatpak), but any application can use portals to provide uniform access to features independent of desktops and toolkits.

And the [XDG-Desktop-Portal docs](https://flatpak.github.io/xdg-desktop-portal/docs/for-app-developers.html):

> Portal interfaces can be used by sandboxed and unsandboxed apps alike, but sandboxed apps benefit the most since they don’t need special permissions to use portal APIs. XDG Desktop Portal safeguards many resources and features with a user-controlled permission system.

> But that aside, absolute paths are problematic in that we can't guarantee a user installs their Flatpaks globally. If a user is doing a per-user install (i.e. `flatpak install --user <remote> <name>`), that absolute path in `TryExec` will fail and the entry would be hidden, making the appearance as if the feature did not work.
> Even if the user does install globally, we cannot guarantee that this is the path their distro is using. The Flatpak documentation says:

There must be some way to detect how the Flatpak was installed. I'll try to find out how Chromium for example handles it for webapps that are installed since it seems like they have `TryExec` since MS Edge has it.

> AFAIK the spec has no provision on where the install has to go, it's configurable for the distro (although I don't know any distro that does, but that's besides the point). Users can even set up [their own custom directory](https://docs.flatpak.org/en/latest/tips-and-tricks.html#adding-a-custom-installation).

I mean to be fair, it's also possible a user installs Lutris without adding it's binary to the path, but yes I agree that would have to be somehow detected.

RayJW:
> So from what I can tell it tries loading an existing desktop file and if none exists (or one exists but has no `TryExec`), it hardcodes the `TryExec` to `epiphany` which would run into the exact issue I was describing in the PR description.
> 
> I'll have to test it later but my guess is that the "Installed" WebApps aren't going to show in the launcher unless you have epiphany installed natively.
> 
> I also can't find any info right now whether the Flatpak is even official and therefore supported, because a lot of the time on Flathub, it isn't. There is a verification program so application developers can "adopt" their own Flatpaks at some point but for a large part of Flathub packages, they are inofficial.

Also had time to check on a laptop now, and I see what you mean, but it must work somehow because I just tried with the Epiphany Flatpak, and it works installing webapps. 

I added example.org as a webapp and for the user install there is the file `~/.local/share/applications/org.gnome.Epiphany.WebApp_13b9a55dd478a012b5d26b0fa3c4e09a71d221f6.desktop` which is a symlink to `../xdg-desktop-portal/applications/org.gnome.Epiphany.WebApp_13b9a55dd478a012b5d26b0fa3c4e09a71d221f6.desktop` and looks like this:

```
[Desktop Entry]
Exec=flatpak run --command=epiphany 'org.gnome.Epiphany' --application-mode --profile=/home/user/.var/app/org.gnome.Epiphany/data/org.gnome.Epiphany.WebApp_13b9a55dd478a012b5d26b0fa3c4e09a71d221f6 http://example.org/
StartupNotify=true
Terminal=false
Type=Application
Categories=GNOME;GTK;
StartupWMClass=org.gnome.Epiphany.WebApp_13b9a55dd478a012b5d26b0fa3c4e09a71d221f6
X-Purism-FormFactor=Workstation;Mobile;
Name=Example Domain
Icon=/home/user/.local/share/xdg-desktop-portal/icons/scalable/org.gnome.Epiphany.WebApp_13b9a55dd478a012b5d26b0fa3c4e09a71d221f6.svg
TryExec=/home/user/.local/share/flatpak/exports/bin/org.gnome.Epiphany
X-Flatpak=org.gnome.Epiphany
```

Uninstalling the user, Flatpak keeps the .desktop file, but as expected the application won't show up in the launcher any more. Reinstalling it correctly displays the webapp again without any action on my side, and it even gives me a nice notification that Example Domain is ready to be launched as an app.

I also checked what's happening with a system install. Creating a webapp puts the desktop file in the same location with the same symlink, but now it looks like this:

```
[Desktop Entry]
Exec=flatpak run --command=epiphany 'org.gnome.Epiphany' --application-mode --profile=/home/user/.var/app/org.gnome.Epiphany/data/org.gnome.Epiphany.WebApp_13b9a55dd478a012b5d26b0fa3c4e09a71d221f6 http://example.org/
StartupNotify=true
Terminal=false
Type=Application
Categories=GNOME;GTK;
StartupWMClass=org.gnome.Epiphany.WebApp_13b9a55dd478a012b5d26b0fa3c4e09a71d221f6
X-Purism-FormFactor=Workstation;Mobile;
Name=Example Domain
Icon=/home/user/.local/share/xdg-desktop-portal/icons/scalable/org.gnome.Epiphany.WebApp_13b9a55dd478a012b5d26b0fa3c4e09a71d221f6.svg
TryExec=/var/lib/flatpak/exports/bin/org.gnome.Epiphany
X-Flatpak=org.gnome.Epiphany
```

I even checked what happens if I switch between the two. Epiphany still recognizes the installed webapp, but obviously it isn't "runnable" since `TryExec` fails the check. But theoretically it would even be possible to handle the switch between different install methods if that would be desired.

> I also can't find any info right now whether the Flatpak is even official and therefore supported, because a lot of the time on Flathub, it isn't. There is a verification program so application developers can "adopt" their own Flatpaks at some point but for a large part of Flathub packages, they are inofficial.

Do you mean the Epiphany one or for Lutris? Epiphany definitely is and Lutris also is, I saw a lot of work from @strycore, and it's verified on Flathub. However, it seems like there is some frustration with it from Lutris's side, so maybe that will change in the future.

mihawk90:
I mean the Epiphany Flatpak, it was in regards to whether it actually worked.

But regardless I don't know how it's done then. I haven't tracked down what existing keyfile they are loading, but it's probably related to that then (otherwise `tryexec` would be `NULL` and therefore set to `epiphany`).

strycore:
I kinda have the memory that is was required to set the executable bit on .desktop files but that was a long time ago and I'm glad it's no longer required.

In its current state, is this safe to merge for 0.5.18? If not it can wait until next release. 

> Epiphany definitely is and Lutris also is, I saw a lot of work from @strycore, and it's verified on Flathub. However, it seems like there is some frustration with it from Lutris' side, so maybe that will change in the future.

The Lutris Flatpak is annoying to maintain because we shove every dependency possible in the Flatpak to get as much compatibility with all runners as possible. The end goal is to have a very slim Lutris Flatpak with just the Gtk UI and dependencies on umu, other Flatpaks or Flatpak extensions to get the games running. This will make the Lutris Flatpak a lot easier to maintain and a lot lighter.

RayJW:
> I mean the Epiphany Flatpak, it was in regards to whether it actually worked.
> 
> But regardless I don't know how it's done then. I haven't tracked down what existing keyfile they are loading, but it's probably related to that then (otherwise `tryexec` would be `NULL` and therefore set to `epiphany`).

I found the [PR responsible for adding portal support to Epiphany](https://gitlab.gnome.org/GNOME/epiphany/-/merge_requests/1067) which specifically targets the Flatpak as well. The commit messages might help to get to the bottom of this. I still have this on my radar occasionally, but time is very sparse at the moment.

mihawk90:
> In its current state, is this safe to merge for 0.5.18?

Yes, unless you *really* want Flatpak support to be added first. As is it works as expected for native installations.

strycore:
I'll merge this but we need both native and Flatpak to work.

danieljohnson2:
So I guess 0.5.18 will he held up until the Flatpak is fixed?

mihawk90:
I honestly don't even know where to start looking for how Epiphany did it because the changeset is quite large. From skimming it, it does seem to be done via the dynamic launcher portal, but implementing that would require quite a large rewrite (provided we can even interact with it in Python in the first place).

PR #5641, opened 2024-10-06
Check for Umu/Proton *only* by version, not by path.

danieljohnson2:
This PR removes the ``is_proton_path()`` jank; it deletes that function altogether.

The only way to decide if a Wine version is Protonic by looking at the version string. This means it is done consistently (or so I hope). At least it does not depend on the version sometimes, and the directory containing Wine sometimes

To do this, there's a new ``wine_version`` argument passed in to almost all of the ``command/wine.py`` commands; this is provided automatically just like ``wine_path`` is. In some cases these commands create a runner, but they now give it an that explicit ``wine_version`` to use instead of the usual configuration stuff.

I've done some testing, but this is substantial enough (and close enough to release) that I'd like some more eyes on it, rather than just merging the fix up front.

But I am convinced that deciding when to use Umu in two different ways that can disagree just *can't* be right. Let's try to get this fixed for 0.5.18.

### DIFF
--- a/lutris/installer/commands.py
+++ b/lutris/installer/commands.py
@@ -8,6 +8,7 @@ import shlex
 import shutil
 from gettext import gettext as _
 from pathlib import Path
+from typing import Tuple
 
 from lutris import runtime
 from lutris.cache import get_cache_path, has_custom_cache_path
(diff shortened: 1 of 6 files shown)


danieljohnson2:
After discussion with @GloriousEggroll, this PR is incorrect. We should be looking for a file names ``proton`` at the root of the Wine directory, so I'm closing this and moving onto that strategy instead. That will be a new PR.

PR #5646, opened 2024-10-09
fixed 'failed to set text' - Portuguese translation

victorsouzaleal:
Oops, it seems i forgot to close an italic tag and Lutris was spamming errors in the terminal.


danieljohnson2:
Yay! Translation fixes!

PR #5653, opened 2024-10-15
pt_BR translation: fixed search terms that shouldn't be translated.

victorsouzaleal:
the book is on the table


danieljohnson2:
> the book is on the table

I'm not sure why we care where the book is. Is this some Brazilian idiom?

Anyway, the changes look good. I'll merge this.

victorsouzaleal:
"The book is on the table" is the equivalent of "Hello World" in English courses in Brazil xD

PR #5666, opened 2024-10-20
fix: kill wine procs through umu when using GE-Proton (Latest)

R1kaB3rN:


danieljohnson2:
I might merge this if I knew what bug this fixed. It looks legit, but I don't really understand it.

GloriousEggroll:
this is ok, merging

PR #5668, opened 2024-10-21
The Flathub Capitalization in the Changelog is wrong

RayJW:
I'm sorry, I know usually small PRs like this are a bit shunned, but I also didn't want to create the overhead of a new issue just for something minor like this. The capitalization in the changelog for the unreleased 0.5.18 version is wrong, Flathub is not spelled like that.


danieljohnson2:
If that is the only word I misspelled, I was doing well (for me anyway)!

PR #5673, opened 2024-10-23
refactor proton/umu detection and usage

GloriousEggroll:
complete refactor to allow UMU:
1. full detection of proton path and wine executable as well as proton version
2. full winetricks support
3. full wine program support (winecfg, regedit, wineconsole, etc)
4. full installations for wine games using UMU
5. we do -not- replace the wine executable with umu in get_executable. Instead we safely get the proper wine executable from the proton path and use updated is_proton_path to determine if it's proton, then if so we perform an umu swap in only 5 locations -- 

runners/wine.py `get_command`:
```
    def get_command(self) -> List[str]:
        command = super().get_command()
        if command:
            exe = command[0]

            if proton.is_proton_path(exe) and not proton.is_umu_path(exe):
                command[0] = proton.get_umu_path()
```
runners/commands/wine.py `createprefix`:
```
    if proton.is_proton_path(wine_path):
        # All proton path prefixes are created via Umu; if you aren't using
        # the default Umu, we'll use PROTONPATH to indicate what Proton is
        # to be used.
        wineenv["PROTON_VERB"] = "run"

        proton.update_proton_env(wine_path, wineenv)

        command = MonitoredCommand([proton.get_umu_path(), "createprefix"], env=wineenv)
        command.start()
    else:
        wineboot_path = os.path.join(os.path.dirname(wine_path), "wineboot")
```

runners/commands/wine.py `winekill`:
```
def winekill(prefix, arch=WINE_DEFAULT_ARCH, wine_path=None, env=None, initial_pids=None, runner=None):
    """Kill processes in Wine prefix."""

    initial_pids = initial_pids or []
    steam_data_dir = os.path.expanduser("~/.local/share/Steam/compatibilitytools.d")
    if not env:
        env = {"WINEARCH": arch, "WINEPREFIX": prefix}
    if proton.is_proton_path(wine_path):
        command = [proton.get_umu_path(), "wineboot", "-k"]
```

runners/commands/wine.py `wineexec`:
```
    command_parameters = []
    if proton.is_proton_path(wine_path):
        command_parameters.append(proton.get_umu_path())
        if winetricks_wine is not "" and wine_path not in winetricks_wine:
            command_parameters.append("winetricks")
    else:
        command_parameters.append(wine_path)
```

runners/commands/wine.py `open_wine_terminal`:
```
def open_wine_terminal(terminal, wine_path, prefix, env, system_winetricks):
    winetricks_path, _working_dir, env = find_winetricks(env, system_winetricks)
    if proton.is_proton_path(wine_path):
        wine_path = proton.get_umu_path() + " wine"
        env["PROTON_VERB"] = "waitforexitandrun"
        proton.update_proton_env(wine_path, env)

    aliases = {
        "wine": wine_path,
        "winecfg": wine_path + "cfg",
        "wineserver": wine_path + "server",
        "wineboot": wine_path + "boot",
        "winetricks": winetricks_path,
    }
```

Basically, wineexec, and functions that use wine but do not call wineexec directly. `open_wine_terminal` is the only one we modify the actual wine_path, the rest wine_path is completely untouched.

### DIFF
--- a/lutris/installer/commands.py
+++ b/lutris/installer/commands.py
@@ -21,7 +21,6 @@ from lutris.util import extract, linux, selective_merge, system
 from lutris.util.fileio import EvilConfigParser, MultiOrderedDict
 from lutris.util.jobs import schedule_repeating_at_idle
 from lutris.util.log import logger
-from lutris.util.wine import proton
 from lutris.util.wine.wine import WINE_DEFAULT_ARCH, get_default_wine_version, get_wine_path_for_version
 
 
@@ -38,10 +37,7 @@ class CommandsMixin:
         version = runner.get_installer_runner_version(self.installer, use_runner_config=False)
         if version:
             wine_path = get_wine_path_for_version(version)
-            if proton.is_proton_path(wine_path):
-                return proton.get_umu_path()
-            else:
-                return wine_path
+            return wine_path
 
         # Special case that lets the Wine configuration explicit specify the path
         # to the Wine executable, not just a version number.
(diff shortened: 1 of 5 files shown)


AliensInc:
Just a thought,
wouldn't "...detection of proton path..." break what danieljohnson2 did, thus NOT to look at the path for proton which created an error before?

danieljohnson2:
I've closed my PR on this, because we're going a different way- we'll detect Proton by looking for a specific file in a specific place- not exactly by the path itself, or the version string.

This does mean reversing out some of my work. That's the way the cookie crumbles.

PR #5681, opened 2024-10-27
translation: pt_BR updated

victorsouzaleal:
100% translation of .po file and some translations have been fixed :brazil:


danieljohnson2:
That's great - you managed to get the message about using Proton with 32-bit prefixes that I put in this morning!

Too bad I misspelled it. '32bit' instead of '32-bit'. I've fixed it in master. Perhaps you could update your PR in that sense?

The message was:
> Proton is not compatible with 32bit prefixes.

Now it has a hyphen in it.

PR #5682, opened 2024-10-27
translation: pt_BR updated

victorsouzaleal:
100% translation of .po file and some translations have been fixed

Looks like I discarded the commit I had made haha my bad :clown_face:


danieljohnson2:
Round and round we go! Thanks for coping with my, um, development style today.

Merging away.

PR #5683, opened 2024-10-27
Fixed 3 missing texts and translated to pt-BR

victorsouzaleal:
I fixed 3 texts that I couldn't translate and applied the translation.

I think that's all for now :+1:


danieljohnson2:
Ah, you've found some forgotten ``_(  )`` thingies. That's a bug-fix, so I'll go ahead and merge it.

Issue #4868, opened 2023-05-23
Tooltips are not displayed in KDE Plasma


potatoleek:
### Bug description

In KDE Plasma, tooltips are not displayed when hovering over options. I believe the issue is isolated to KDE Plasma because I cannot replicate the issue in a different desktop environment in the same distro.

### How to Reproduce

1. Go to the Configure menu of any runner.
2. Hover over any option. No tooltip is displayed.

### Expected behavior

A tooltip explaining the option should be displayed.

### Log output

```shell
INFO     2023-05-22 19:02:30,630 [startup.init_lutris:209]:Starting Lutris 0.5.13
INFO     2023-05-22 19:02:30,632 [startup.check_driver:65]:Using NVIDIA drivers 530.41.03 for x86_64
INFO     2023-05-22 19:02:30,632 [startup.check_driver:69]:GPU: NVIDIA GeForce GTX 745
INFO     2023-05-22 19:02:30,632 [startup.check_driver:85]:GPU: 10DE:1382 10DE:1065 (nvidia drivers)
INFO     2023-05-22 19:02:30,633 [startup.check_driver:85]:GPU:   (simple-framebuffer drivers)
DEBUG    2023-05-22 19:02:30,881 [xrandr._get_vidmodes:16]:Retrieving video modes from XrandR
DEBUG    2023-05-22 19:02:30,946 [lutris.get_missing_game_ids:191]:Checking for missing games
DEBUG    2023-05-22 19:02:33,537 [xrandr._get_vidmodes:16]:Retrieving video modes from XrandR
DEBUG    2023-05-22 19:02:33,596 [application.show_window:339]:Showing window RunnerConfigDialogwine
```

### System Information

```shell
[System]
OS:              Fedora Linux 38 KDE Plasma
Arch:            x86_64
Kernel:          6.2.15-300.fc38.x86_64
Desktop:         KDE
Display Server:  x11

[CPU]
Vendor:          GenuineIntel
Model:           Intel(R) Core(TM) i7-4790 CPU @ 3.60GHz
Physical cores:  4
Logical cores:   8

[Memory]
RAM:             11.6 GB
Swap:            8.0 GB

[Graphics]
Vendor:          NVIDIA Corporation
OpenGL Renderer: NVIDIA GeForce GTX 745/PCIe/SSE2
OpenGL Version:  4.6.0 NVIDIA 530.41.03
OpenGL Core:     4.6.0 NVIDIA 530.41.03
[...]

PR #5705, opened 2024-11-04
Fix tooltips on some Gtk3 themes on X11

B--B:
As reported in lutris/lutris#4868 , under some circumstances Lutris custom tooltips are completely broken. More specifically this happens only on X11 with KDE 5/6, and only with some themes. I honestly don't know why there are these differences between themes or X11/Wayland, but the problem is caused by the tooltip (which is a Gtk Box) that steal window focus when appear. For this reason I switched tooltips to Gtk EventBox. The advantage of EventBox compared to Box is that EventBox cannot gain focus and ignore hover events unless explicitly configured, so we are 100% sure that tooltips won't interfere with Lutris windows. Please note that this PR has been tested only on KDE on both X11/Wayland in Ubuntu Oracular, even if should work fine on Gnome and other DEs it would be probably better to try it before merge, at least to check if there are aesthetic changes. Last thing, this bug does not seem related to some specific GPU/driver, I can reproduce it on Intel, AMD and NVidia GPUs on X11, and all of them works fine on Wayland.
PS: while i was at it I've also fixed some pylint warns, is the line limit to 100 characters good for you guys or you prefer a different length?

EDIT: I answered myself when I saw ruff check fail, removed the offending changes and ran a format, now ruff it's happy again

### DIFF
--- a/lutris/gui/config/boxes.py
+++ b/lutris/gui/config/boxes.py
@@ -679,10 +679,10 @@ class ConfigBox(VBox):
         label = Label(text)
         label.set_use_markup(True)
         label.set_max_width_chars(60)
-        hbox = Gtk.Box()
-        hbox.pack_start(label, False, False, 0)
-        hbox.show_all()
-        tooltip.set_custom(hbox)
+        event_box = Gtk.EventBox()
+        event_box.add(label)
+        event_box.show_all()
+        tooltip.set_custom(event_box)
         return True
 
     def option_changed(self, widget, option_name, value):


lutris/gui/config/boxes.py
```diff
@@ -732,7 +732,7 @@ def on_reset_button_clicked(self, btn, option, _widget, wrapper):
     def set_style_property(property_, value, wrapper):
         """Add custom style."""
         style_provider = Gtk.CssProvider()
-        style_provider.load_from_data("GtkHBox {{{}: {};}}".format(property_, value).encode())
+        style_provider.load_from_data(f"GtkHBox {{{property_}: {value};}}".encode())
```
danieljohnson2, line 735: Hang on, does this work? Don't you have to double up the curly braces that do not signify interpolation?

I would prefer this line be left as a ``.format`` call just to avoid that business all together.

danieljohnson2:
I'm not sure the f-string refactoring belongs in this PR, and I think it may be bugged too.

lutris/gui/config/boxes.py
danieljohnson2: OK, I've tested this, and your code works. The doubled {} are already doubled in the original version, so I guess this is fine.

B--B:
Removed from the PR, i will take a look at them and will send a different PR.

PR #5706, opened 2024-11-04
Translation: pt_BR update

victorsouzaleal:
and a small fix to be able to translate "Theme" :+1:


danieljohnson2:
That someone was me. But I guess you can translate that word *if you have to, I guess*.

Issue #4621, opened 2022-11-28
Unsupported architecture aarch64


TYeniyayla:
Is lutris arm64 not compatible? When I run it from console I get the following error. when I run it as an application, only the error report screen appears.

```
$ lutris
2022-11-28 19:18:24,700: Unsupported architecture aarch64
Traceback (most recent call last):
  File "/usr/games/lutris", line 52, in <module>
    from lutris.gui.application import Application  # pylint: disable=no-name-in-module
  File "/usr/lib/python3/dist-packages/lutris/gui/application.py", line 38, in <module>
    from lutris.command import exec_command
  File "/usr/lib/python3/dist-packages/lutris/command.py", line 13, in <module>
    from lutris import runtime, settings
  File "/usr/lib/python3/dist-packages/lutris/runtime.py", line 12, in <module>
    from lutris.util.linux import LINUX_SYSTEM
  File "/usr/lib/python3/dist-packages/lutris/util/linux.py", line 434, in <module>
    LINUX_SYSTEM = LinuxSystem()
  File "/usr/lib/python3/dist-packages/lutris/util/linux.py", line 123, in __init__
    self.populate_libraries()
  File "/usr/lib/python3/dist-packages/lutris/util/linux.py", line 359, in populate_libraries
    for req in self.requirements:
  File "/usr/lib/python3/dist-packages/lutris/util/linux.py", line 258, in requirements
    return self.get_requirements()
  File "/usr/lib/python3/dist-packages/lutris/util/linux.py", line 287, in get_requirements
    if drivers.is_amd():
  File "/usr/lib/python3/dist-packages/lutris/util/graphics/drivers.py", line 91, in is_amd
    for card in get_gpus():
  File "/usr/lib/python3/dist-packages/lutris/util/graphics/drivers.py", line 69, in get_gpus
    for cardname in os.listdir("/sys/class/drm/"):
PermissionError: [Errno 13] Permission denied: '/sys/class/drm/'
```

uname -m
`aarch64`

uname -a
`Linux localhost 5.4.0-faked #1 SMP PREEMPT Tue Nov 15 06:59:43 UTC 2022 aarch64 GNU/Linux
`

strycore:
This has nothing to do with the ARM architecture. Lutris is trying to read available GPUs from /sys/class/drm but your distribution doesn't allow it. 
If added error handling for this case, so it should go beyond that point now.
Please note that while the Lutris client should run without issue on ARM, we don't have runners for ARM yet.

Issue #3988, opened 2022-01-14
Preferences window doesn't open when running on aarch64 (RPI4/400, Manjaro)


midwan:
### Bug description

Testing on a Pi400 running the latest Manjaro XFCE (aarch64).
I installed lutris using `pacman -S lutris`, which also installed the dependencies. I can start it up normally, but I cannot open the preferences window.

Starting it from the console showed some more details:
If I try to open the Prefs window, it crashes with the following piece of the traceback:
```
"/usr/lib/python3.9/site-packages/lutris/util/linux.py", line 449, in gather_system_info_str 
cpu_dict["Vendor"] = system_info["cpus"][0]["vendor_id"] 
KeyError: 'vendor_id'
```

As the author of Amiberry, I was looking at how I could integrate it to Lutris so it can launch Amiga games (I know FS-UAE is already available). I could set up a "Linux" game hardcoding the emulator path and parameters to a game, and it worked as expected, so it would be interesting to fully integrate the emulator the "proper" way.
But the prefs window problem is a bit of a blocker for now. :)

### How to Reproduce

1. Use a Raspberry Pi 4 or Pi400, running the latest Manjaro XCFE
2. Install lutris using `pacman`
3. Run lutris from the console to see the messages
4. Try to open the Preferences window

### Expected behavior

The aarch64 CPU should be detected and the application should function normally (Prefs window should open).

### Log output

```shell
midwan@pi400 ~]$ lutris -d
2022-01-14 11:22:47,571: Unsupported architecture aarch64
2022-01-14 11:22:48,184: Starting Lutris 0.5.9.1
2022-01-14 11:22:48,273: No cores found
2022-01-14 11:22:49,327: Startup complete
INFO     2022-01-14 11:22:49,330 [startup.check_driver:61]:Running Broadcom Mesa driver 21.2.5 on V3D 4.2 (0xffffffff)
INFO     2022-01-14 11:22:49,331 [startup.check_driver:73]:GPU:   (v3d drivers)
INFO     2022-01-14 11:22:49,331 [startup.check_driver:73]:GPU:   (vc4-drm drivers)
WARNING  2022-01-14 11:22:49,336 [startup.check_vulkan:132]:Vulkan is not available or your system isn't Vulkan capable
DEBUG    2022-01-14 11:22:50,828 [lutriswindow.update_store:451]:Showing 1 games
WARNING  2022-01-14 11:22:54,937 [linux.get_arch:198]:Unsupported architecture aarch64
[...]

Issue #3819, opened 2021-11-14
Error when installing Lutris on Raspberry Pi 4 and Pop Os


chest069:
Error while installing package: 95.2000

I got this error when trying to install Lutris on my Pi 4 and Pop OS. Anyone know how to fix this? Thanks for any help.

tannisroot:
We don't support ARM platforms at the moment.

PR #5707, opened 2024-11-05
Fix improper use of non-amd64/i386

theofficialgman:
see commit messages for details

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -13,7 +13,7 @@ Vcs-Browser: https://github.com/lutris/lutris
 Vcs-Git: https://github.com/lutris/lutris.git
 
 Package: lutris
-Architecture: all
+Architecture: amd64 i386
 Depends: ${misc:Depends},
          ${python3:Depends},
          python3-yaml,
--- a/lutris/util/linux.py
+++ b/lutris/util/linux.py
@@ -192,8 +192,6 @@ class LinuxSystem:  # pylint: disable=too-many-public-methods
             return "x86_64"
         if machine in ("i386", "i686"):
             return "i386"
-        if "armv7" in machine:
-            return "armv7"
         logger.warning("Unsupported architecture %s", machine)
 
     @staticmethod


strycore:
You gonna have to provide more details

theofficialgman:
@strycore please read the commit messages that go into detail why these changes are here

theofficialgman:
image

strycore:
You can run Lutris on Arm

theofficialgman:
Again, please read the commit messages

strycore:
I've read them

theofficialgman:
I see so you do not understand it. I will make it clear for you. You cannot run lutris (ie: to connect with existing services like Humble Bundle, GOG and Steam to make your game libraries easily available. Game downloads and installations are automated and can be modified through user made scripts) on a PURE arm system.

lutris can only be run (by run I mean actually used to run anything like described above through your own description) through translation/emulation layers like box64/fex/QEMU. On these systems, the entire amd64/i386 architecture is spoofed and amd64/i386 packages are installed either in a separate rootfs (ie: on FEX) or on the host system (ie: on Box64 with amd64/i386 package architectures enabled). In both cases, lutris will "think" its running on an amd64/i386 system. There is no need for allowing lutris to install on a PURE ARM system (like with `all`) since it will not function.

theofficialgman:
> Sometimes it requires to install the Arm version of a runner manually but that's about it.

What arm runners? You have no such systems in place for arm runners.

theofficialgman:
> Yes, our Arm support is incomplete and experimental.

Don't expose something that doesn't work to users. Developers looking to improve support for other architectures have no issues running lutris manually.

strycore:
I've received only positive feedback about having Lutris available on Arm so no it's not going away.

If you want to be helpful, contribute to the build system so we can have Arm runners.

theofficialgman:
> > i386 runner (due to bad programming, you can see that easily)
> 
> You're really trying to rub me the wrong way, aren't you?

No i am sick of users coming to me or posting on other support forums (on arm systems) saying along the lines of:
"I installed lutris, it doesn't work or doesn't work with 'insert thing here'. it works on my pc, why doesn't it work on 'insert system/os name here'? something must be wrong with 'insert system/os name here'." When in reality the problem lies solely with lutris which shouldn't be installable in the first place.

theofficialgman:
> I've received only positive feedback about having Lutris available on Arm

I have only received negative feedback from having lutris available to install on debian based systems on non-supported architectures.

theofficialgman:
> I've received only positive feedback about having Lutris available on Arm

I see no positivity from having it available, only confusion and issues that are not fixed or handled.
https://github.com/lutris/lutris/issues/418
https://github.com/lutris/lutris/issues/3367
https://github.com/lutris/lutris/issues/3819
https://github.com/lutris/lutris/issues/3988
https://github.com/lutris/lutris/issues/4621

theofficialgman:
and to reiterate again:

lutris can only be run through translation/emulation layers like box64/fex/QEMU. On these systems, the entire `amd64`/`i386` architecture is spoofed and `amd64`/`i386` packages are installed either in a separate `amd64`/`i386` rootfs (ie: on FEX) or on the host system (ie: on Box64 with amd64/i386 package architectures enabled). In both cases, lutris will "think" its running on an `amd64`/`i386` system.

before this PR:
- lutris can be installed on pure armhf/arm64 systems and NONE of the lutris runners will work
- lutris can be installed on a emulated amd64/i386 system (box64/fex/qemu) and the lutris runners have a chance of working

after this PR:
- lutris cannot be installed on pure armhf/arm64 systems 👍🏻 
- lutris can be installed on a emulated amd64/i386 system (box64/fex/qemu) and the lutris runners have a chance of working

strycore:
Lutris is a python UI which is architecture agnostic, using all/any as the architecture for python packages is very common and recommended. 

The issue you're trying to address is blocking Lutris from being installed on Arm because we don't have automated installation of runners. 

Aside from Wine and Linux native games, almost no runner require the use of Box64 or Fex as most of them are available on Arm. 

I get that you're trying to address a valid concern but your solution is not the correct one and I will never cripple Lutris because some users can't figure out how arm machines work.

You're also completely ignoring that there are other distros than Ubuntu and Debian and we might ship Lutris for Arm on those as well.

theofficialgman:
> Aside from Wine and Linux native games, almost no runner require the use of Box64 or Fex as most of them are available on Arm.

What runners are left? Flatpak, yeah not enough to justify all the broken/useless ones. https://github.com/lutris/lutris/tree/master/lutris/runners . Keep in mind that libretro (retroarch) doesn't produce arm64 Linux cores so everyone of your runners relying on that are useless (I've tried to volunteer to add support for automation for building them but the whole project is on life support and none of the maintainers respond).

strycore:
> What runners are left? 

 'web',
 'flatpak',
 'zdoom',
 'dosbox',
 'xemu',
 'easyrpg',
 'mame',
 'mednafen',
 'scummvm',
 'libretro',
 'fsuae',
 'vice',
 'atari800',
 'hatari',
 'snes9x',
 'mupen64plus',
 'dolphin',
 'ryujinx',
 'yuzu',
 'cemu',
 'pcsx2',
 'rpcs3',
 'vita3k',
 'osmose',
 'reicast',
 'redream',
 'pico8',
 'jzintv',
 'o2em'

theofficialgman:
you clearly do not understand the point. most of the runners have NO (official or even if you compile from source) arm support

no arm build: upstream developer does not provide an official arm build. it could be built from source
fundamentally x86_64: upstream developer does not provide an official arm build and it cannot be build from source
no linux build: upstream developer does not provide an official linux build. it could be built from source

~~'dosbox',~~ fundamentally x86_64 only
~`'xemu',~~ no arm build
~~'easyrpg',~~ no arm build
~~'mame',~~ no arm build
~~'mednafen',~~ no linux build
~~'scummvm',~~ no arm build
~~'libretro',~~ no arm build
~~'fsuae',~~ no arm build
~~'vice',~~ no arm build
~~'hatari',~~ no arm build
~~'snes9x',~~ no arm build
~~'mupen64plus',~~ no arm build
~~'dolphin',~~ no arm build
~~'ryujinx',~~ no arm build
~~'cemu',~~ fundamentally x86_64 only
~~'pcsx2',~~ fundamentally x86_64 only
~~'rpcs3',~~ no arm build
~~'vita3k',~~ no arm build
~~'osmose',~~ fundamentally x86_64 only (and dead)
~~'reicast',~~ no linux build (dead)
~~'jzintv',~~ fundamentally x86_64 only (and dead)
~~'o2em'~~ fundamentally x86_64 only (and dead)

so that leaves you with a small list of things that actually work and have builds for arm linux. This proves the invalidity of this statement "most of them are available on Arm" because the majority (over half) do not.

PR #5710, opened 2024-11-06
Runner download: properly handle download cancel action

B--B:
This PR fixes an exception raised when user start a runner download and then press the cancel button (read commit message for details). in addition to what I have already written I have tried to crash Lutris in every way I could think of (malformed urls, decompression of non existent archives, corrupted archives) but I have not been able to do so, Lutris handled all the other exceptions as before

### DIFF
--- a/lutris/gui/config/runner_box.py
+++ b/lutris/gui/config/runner_box.py
@@ -96,8 +96,6 @@ class RunnerBox(Gtk.Box):
 
         if self.runner.is_installed():
             self.emit("runner-installed")
-        else:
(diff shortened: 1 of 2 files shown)


danieljohnson2:
This seems right to me!

B--B:
Thanks for merging this Daniel! I forgot to mention that I left untouched the part related to bios/addons download, because if the user press cancel during BIOS download most likely will end up with a broken runner installation (runner already installed but BIOS missing), so an exception will be still raised in that case, tried with Atari 800 that download BIOS package and the error dialog appear as it should

PR #5713, opened 2024-11-08
searchable_combobox --> searchable_entrybox

B--B:
This commit switch from a combobox to a custom entry box used on MAME "machine" option, removing the dropdown button.
Reasons for this change:

1) Gtk.Bin has been removed in Gtk4, in case of future migration to Gtk4 this should work fine with few changes (pack_start/get_toplevel) [1] 
2) Dropdown button was terrible because:
  - It was causing lutris freeze sometimes even for a few seconds when pressed for the first time
  - The list had hundreds of entries, it was unusable
  - The list window was causing several Gtk warns when created

Additionally, this commit refactor a bit the code, slighty modifing some functions (mostly get_active_id because even if it was working fine i had some compatibility issues when upgrading or downgrading Lutris to a version with/without entry box) and adds status icon to the search box.

Test: tested on KDE plasma 6.1.5 and KDE framework 6.6.0, no issues and migration from old Lutris versions or downgrade to old versions works as expected. Code checked with pylint and flake8

[1] https://docs.gtk.org/gtk4/migrating-3to4.html#adapt-to-gtkbin-removal

PS: In case you want to keep searchable_combobox code just let me know and I will add it back, but it was used only by MAME runner


danieljohnson2:
Well, this may be a good idea, but @strycore will need to make the call.

But I would ask- is there *any* way for a user to get the complete list of options with this control?

B--B:
Not the complete list, but if you type even a single character all results that contain the character will appear. I've removed the dropdown button because it' unusable, at least for me on KDE, it takes seconds for loading the first time with Lutris stuck, have a ton of entries and you cannot scroll quickly, but I have an idea about the complete list, give me some time :)

danieljohnson2:
I appreciate the problem this UI had; but I do think it's desirable to know what the options are somehow.

B--B:
Now when user click on the entry box the full list will appear, as bonus we can also fast scroll the list, it wasn't possible previously

danieljohnson2:
So it seems like the list appears in full if you enter any character but then erase it again. Right?

What I do not understand is why this code does not block the UI for several seconds just like the old code did. It's populating just as large a list-model, isn't it?

I mean, it seems to work better in my testing, I just don't know why that is.

B--B:
No idea honestly, i tried to fix the old impl but I wasn't able to do so, everytime when you click for the first time the dropdown button gtk prints some warns about window size and UI stucks, I've tried to see if it was possible to set a fixed window size for the list at least to see if it was related to Lutris freeze but nothing worked, that's why I tried a different approach, especially when I saw that Gtk Bin has been removed from Gtk4

B--B:
@danieljohnson2 sorry I misunderstood, I had a saved entry and I was removing it before trying, so I didn't realize that you have to interact at least once or the list doesn't appear even when clicking on the input box. I solved it by restoring the previous behavior of the input box (1 character to make the list appear) and I added a popup list, just click on the search icon and it will appear as in the old implementation. It triggers the same warnings as the previous implementation, but fortunately it doesn't stuck Lutris

danieljohnson2:
I like this version better.

B--B:
Me too, don't like so much remove features, but initially i didn't think this solution. Let me know if you want to change some other thing.

danieljohnson2:
Yes, I had this problem too. The problem is that the mame runner code for populating the list is crap. I have not been able to make it work right after *several* tries. It's ``write_xml_list`` in ``mame.py``- it fails and writes nothing. You can create the ``mame.xml`` file manually to bypass this: ``mame -list-xml >~/.cache/lutris/mame/mame.xml`` I think it is.

Somehow ``mame -list-xml`` doesn't work when Lutris does it. It's weird.

B--B:
> When I tested this branch, I didn't get any popup. The magnifying glass turned into 3 dots, and I could type stuff but nothing would show up. I don't see anything in the logs either. Anything that would case that?

I can confirm that Daniel is right, not related to this PR but it's an issue. Lutris create mame.xml and systems.json the first time you open a MAME game config, but the search box is not populated, you need to close Lutris and open it again, then it will work.

danieljohnson2:
I can well believe that you have to restart Lutris- if ``mame -list-xml`` even works, that code is very dicey. It's trying to load up the file in the background, I believe.

But I am not even getting that far with it.

strycore:
Please rebase against latest master branch

danieljohnson2:
This PR is affected by the big refactor that I did- the changes made to ``boxes.py`` need to go to ``widget_generator.py``.

I could probably get this working, if you would like.

strycore:
yes, that would be nice

danieljohnson2:
OK, I've rebased it myself and merged the result; I believe it is what @B--B would have wanted, as it were.

GitHub does not recognizes that we're merging this PR- maybe there's some way I could have done this so that it did, but I am not the GitHub master. So I am closing this.

Issue #5714, opened 2024-11-11
Continuous re-installation of ge-Proton at startup


PrimaIVoid:
### Bug description

**Almost** every time i try to launch Lutris from terminal or directly from the app, in the lower-left corner it seem to try download/estrac/install the latest version of the **ge-proton** runner, even though in my system and in the config of lutris i already have the latest available ge-proton set and ready to use. Also tried used it with some games as it works without problems

The only output i receive about this behaviour when Lutris is trying to re-download the same runner is:
`2024-11-11 10:03:28,951: No local runtime path in /home/zodiac/.local/share/lutris/runtime/ge-proton`

This warning says the truth since in that directory i don't have any "ge-proton" folder, but for my knowledge all the relevant runners are stored in: 
`/home/zodiac/.local/share/lutris/runners/`
In here in fact i have the latest `wine` and `Ge-Proton` builds that works and are updated by Lutris automatically
I have the feeling this might be related to my setting the default choice of **UMU-Latest** as Wine runner, even though does not seem to be related to ge-proton that is anyhow already installed and available as needed

I have noticed this weird behaviour since about 2 weeks, as of now i'm using the latest v0.5.18

### How to Reproduce

This behaviour is kind of hard to reproduce even for me, since it does **not always** trigger.
After restarting my entire Computer-System and try to open Lutris for the first time it **always** force the ge-proton re-download

After this happens, i may close Lutris and wait a few minutes and try opening it again, and it is always random if it triggers or not the forced re-download of ge-proton

### Expected behavior

Lutris should be aware that the latest version of ge-proton runner is already installed/set and ready on my system, not try to redownload it over again

### Log output

```shell
`2024-11-11 10:03:28,951: No local runtime path in /home/zodiac/.local/share/lutris/runtime/ge-proton`
```

### System Information

```shell
Vulkan support: YES
Esync support: YES
Fsync support: YES
Wine installed: YES
Gamescope: NO
Mangohud: NO
Gamemode: YES
Steam: NO
In Flatpak: NO
[System]
OS: Linux Mint 21.2 Victoria
Arch: x86_64
Kernel: 5.15.0-125-generic
Lutris Version: 0.5.18
Desktop: XFCE
Display Server: x11
[CPU]
Vendor: GenuineIntel
Model: Intel(R) Core(TM) i5-6200U CPU @ 2.20GHz
Physical cores: 4
Logical cores: 8
[Memory]
RAM: 8 GB
Swap: 2.0 GB
[...]

PR #5715, opened 2024-11-11
Proton: fix proton download everytime Lutris check for updates

B--B:
Proton is a special case, it's treated like a runtime but it's not a runtime so we must set the runner dir. Additionally it use a custom dir that must be set or local_runtime_path will fail and Lutris will download it again the next time updates are checked.

Test: updated proton, restarted Lutris and manually checked for updates

Fixes lutris/lutris#5714

NOTE: probably not the cleanest thing to do but works fine with minimal changes

### DIFF
--- a/lutris/runtime.py
+++ b/lutris/runtime.py
@@ -311,6 +311,9 @@ class RuntimeComponentUpdater(ComponentUpdater):
     @property
     def local_runtime_path(self) -> str:
         """Return the local path for the runtime folder"""
+        if self.name == "ge-proton":
+            proton_dir = "proton/GE-Proton/"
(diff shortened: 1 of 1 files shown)


danieljohnson2:
This looks reasonable, but I need to test it and I won't be able to until this afternoon. I expect I'll merge it then.

B--B:
No worries Daniel, take your time! I've also tried to manually delete some runtime components and ran an update, everything works for me. I don't like it so much but should be 100% safe and can be used as temporary fix until Proton is properly handled by updater.

danieljohnson2:
OK, I think I understand. I'll take it, but I don't like having the checks for ge-proton all over the place, so I'll fix that afterwards.

PR #5717, opened 2024-11-12
Ubuntu: packaging: add AppArmor profile for bwrap

B--B:
Starting with Ubuntu 23.10 Canonical begin to restrict unprivileged user namespaces.[1] After this change AppArmor completely block bwrap, breaking Umu, which stops with the following error:

```
pressure-vessel-wrap[290705]: E: Child process exited with code 1: bwrap: setting up uid map: Permission denied
```

The solution is to add the missing AppArmor profile for bwrap. Ubuntu already has it on apparmor-profiles package, but it's still in testing phase. This commit adds the profile to Lutris package and add the rules during the deb installation.[2] Since it's an experimental profile it may not be perfect or cause some issues on some corner cases, but it's still much more better than leave bwrap completely unconfined or Umu broken for all Ubuntu users on 23.10+.

Note: The profile will break AppArmor with ABI version < 4, for this reason this patch include a postinst script[3] that will remove bwrap profile in case Lutris is installed on an old Ubuntu version or Debian (ABI 3).

Note for packaging: dh-apparmor now is a required build dependency.

Test:
1) Kubuntu 24.10 (Real HW) ABI 4: AppArmor OK (rules applied), Umu works
2) Ubuntu 23.04 (VM) ABI 3: AppArmor OK (rules removed), Umu works
3) Debian 12 (VM) ABI 3: AppArmor OK (rules removed), Umu works

[1] https://ubuntu.com/blog/ubuntu-23-10-restricted-unprivileged-user-namespaces
[2] https://wiki.debian.org/AppArmor/Contribute/FirstTimeProfileImport
[3] https://www.debian.org/doc/debian-policy/ch-maintainerscripts.html#summary-of-ways-maintainer-scripts-are-called

EDIT
v2: restart apparmor in postinst if ABI = 4 (load rules without the need of a reboot) and renamed apparmor profile (avoid dpkg conflicts when Ubuntu will move the profile from extra to apparmor.d)

### DIFF
--- a/debian/apparmor-profile
+++ b/debian/apparmor-profile
@@ -1,0 +1,68 @@
+# This profile allows almost everything and only exists to allow
+# bwrap to work on a system with user namespace restrictions
+# being enforced.
+# bwrap is allowed access to user namespaces and capabilities
+# within the user namespace, but its children do not have
+# capabilities, blocking bwrap from being able to be used to
+# arbitrarily by-pass the user namespace restrictions.
+#
+# Note: the bwrap child is stacked against the bwrap profile due to
+# bwraps use of no-new-privs
+
+# disabled by default as it can break some use cases on a system that
+# doesn't have or has disable user namespace restrictions for unconfined
+# use aa-enforce to enable it
+
+abi <abi/4.0>,
+
+include <tunables/global>
+
+profile bwrap /usr/bin/bwrap flags=(attach_disconnected,mediate_deleted) {
+  allow capability,
+  # not allow all, to allow for pix stack
+  # sadly we have to allow  m every where to allow children to work under
+  # stacking.
+  allow file rwlkm /{**,},
+  allow network,
+  allow unix,
+  allow ptrace,
+  allow signal,
+  allow mqueue,
+  allow io_uring,
+  allow userns,
+  allow mount,
+  allow umount,
+  allow pivot_root,
+  allow dbus,
+  allow px /** -> bwrap//&unpriv_bwrap,
+
+  # the local include should not be used without understanding the userns
+  # restriction.
+  # Site-specific additions and overrides. See local/README for details.
+  include if exists <local/bwrap-userns-restrict>
+}
+
+profile unpriv_bwrap flags=(attach_disconnected,mediate_deleted) {
+  # not allow all, to allow for pix stack
+  allow file rwlkm /{**,},
+  allow network,
+  allow unix,
+  allow ptrace,
+  allow signal,
+  allow mqueue,
+  allow io_uring,
+  allow userns,
+  allow mount,
+  allow umount,
+  allow pivot_root,
+  allow dbus,
+
+  allow pix /** -> &unpriv_bwrap,
+
+  audit deny capability,
+
+  # the local include should not be used without understanding the userns
+  # restriction.
+  # Site-specific additions and overrides. See local/README for details.
+  include if exists <local/unpriv_bwrap>
+}
--- a/debian/control
+++ b/debian/control
@@ -4,6 +4,7 @@ Priority: optional
 Maintainer: Mathieu Comandon <mathieucomandon@gmail.com>
 Build-Depends: debhelper-compat (= 12),
                appstream,
+               dh-apparmor,
                dh-sequence-python3,
                meson,
 Rules-Requires-Root: no
--- a/debian/postinst
+++ b/debian/postinst
@@ -1,0 +1,18 @@
+#!/bin/sh
+
+set -e
+
+# Remove apparmor profile if abi < 4.0
+if [ ! -e /etc/apparmor.d/abi/4.0 ]; then
+    rm -f /etc/apparmor.d/lutris-bwrap-userns-restrict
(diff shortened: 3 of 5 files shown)


RayJW:
This seems like a dangerous change. My understanding of bwrap and AppArmor is very surface level, but it looks like this just blatantly allows almost everything, which would be completely against the core idea of AppArmor. The umu people already fixed this problem [upstream](https://github.com/Open-Wine-Components/umu-launcher/pull/218), and their provided profile seems a lot more targeted. Is there any justification why Lutris would need all these extra permissions if the only error is with umu?

B--B:
Have I lost some changes or they actually are using the same rules i'm shipping here? https://github.com/Open-Wine-Components/umu-launcher/commit/418c5d2aab3a5ceb5d57661d50f780a2cf3960d7    https://github.com/Open-Wine-Components/umu-launcher/blob/main/packaging/deb/debian/apparmor/bwrap-userns-restrict-umu

RayJW:
> Have I lost some changes or they actually are using the same rules i'm shipping here?

No, you might be right then. Sorry for the confusion, I now see, that they changed the rules they're shipping since the PR I linked. In that case wouldn't it be easier for Lutris to also just symlink to these preinstalled rules? That way if anything changes it won't be missed here.

B--B:
No worries :) Sure, this can be done in the same way it has been added on UMU, but the file is 1:1 and if we want to copy it from apparmor extra profiles then we need to add another dep on lutris, because the profile is in apparmor-profiles packages which is not shipped by default on Ubuntu. But yes, if Lutris devs decide to depend on another package no problems for me, it's an easy edit!

B--B:
@danieljohnson2 updated commit description and ready for the review, I think I covered all the cases: our profile name is lutris-bwrap-userns-restrict, below ABI 4 it will be simply removed so nothing changes for Debian/old Ubuntu. On ABI 4 first we check if there are Ubuntu rules for bwrap already set (so we cover even future cases when the profile will be enabled again by Canonical) or other custom rules like bwrap-userns-restrict-umu, if found lutris profile will be removed because we don't need it, otherwise we install our profile. I've also added a postrm script because by default Ubuntu remove AppArmor profiles only with an apt purge. There's only a case that can still cause some trouble: if someone needs bwrap with root privileges this profile will break it if installed.  We try to catch even this case, but it's really user-dependent: for make bwrap work with root privileges you need to set bwrap as unconfined with a profile like the one linked by RayJW in umu PR, if user named that file bwrap-userns-restrict or bwrap-userns-restrict-something-else then our profile will be removed and everything will work (even Umu because bwrap is in unconfined mode), but if the file name is different our profile will be installed and AppArmor will apply the most restrictive rules, breaking "sudo bwrap".  But It's a minor case, and the same happens installing umu deb and will happen when Canonical will enable bwrap profile again.
In case of duplicates (we fail to detect custom rules or user install another deb that ship the same profile like Umu) we are lucky, AppArmor parse the duplicated rules and discard them, no problems at all.

PS: no need to hurry with this, but please consider it or something similar before 0.5.18 deb is out, or Ubuntu users on 23.10+ will have to fix Umu manually by themselves

danieljohnson2:
I'll review this when I get a chance, but I won't merge it. I really don't understand this new-fangled AppArmor jiggery-pokery. Somebody with a bit more competence should be looking at this.

danieljohnson2:
Well, I've read over it. It's full of punctuation and letters and things. This makes me nervous.

I think it requires a reboot on uninstall- but not on install. Why is this? Could we not restart the AppArmor service thing on uninstall instead?

strycore:
It will be easy to revert if this causes any kind of problem, let's try it out before the 0.5.18 release

B--B:
Sorry for the late reply, I saw it has been merged but i want to answer to Daniel questions. Yes, unfortunately a reboot is needed due to how AppArmor works: restarting the service will force AppArmor to load the rules again, even new ones if you add them, but does not unload them, this must be done before AppArmor restart with something like "apparmor_parser -R /etc/apparmor.d/profile-name". This may be problematic because dpkg scripts must exit with status 0, but if apparmor_parser fail for some reason it doesn't exit with status 0 and will break dpkg. If apparmor_parser fail we can force the script exit with status 0 and avoid dpkg errors, but if happens and you install Lutris after the removal without rebooting AppArmor rules will be treated as duplicated (because even if the profile file has been deleted the rules are still loaded in AppArmor) and our profile won't be copied so Umu will still work until a reboot, then AppArmor will block bwrap again. But I will take a look at it, I want to see if is possible to try to unload the rules and fallback to reboot only if really needed.

danieljohnson2:
I'm still not happy about reboot-on-uninstall, thought I trust it will only happen if you have AppArmor.

B--B:
> I'm still not happy about reboot-on-uninstall, thought I trust it will only happen if you have AppArmor.

I agree, that's why I want to see if there's another way to do this without the need of reboot. Fortunately only AppArmor user on Ubuntu will experience this, AppArmor 4 is a Canonical addition, there are a ton of patches (99 only on kernel side)  so I do not expect this feature on other distros soon

PR #5721, opened 2024-11-13
Return of the Saved Searches!

danieljohnson2:
Well, 0.5.18 is taking a while, so I figured I'd take the leftover code for saved searches and glue it all together. It really didn't take that much glue. Here's what the search UI looks like:
Screenshot from 2024-11-13 17-31-15
A linked button looks like part of the search box. Click it and there's a pop-over:
Screenshot from 2024-11-13 17-32-06
The search text is still in the same box, and you can edit it by using the pop-overs controls. It updates immediately and is reflected quickly in the view behind. But notice you can enter a name. That tag button is the save button- these are supposed to be 'dynamic categories' now so that seems appropriate. Click that and here's where you go:
Screenshot from 2024-11-13 17-32-18
We reset the search but switch to the newly saved category. The saved search has an edit button on that is pretty much identical to the work from the previous PR, #5536.

One concern you may have is the icon in the pop-over button. @strycore's effort used ``filter-symbolic`` here. My icon theme hasn't got this icon (and it's ugly and its mother dresses it funny) so I picked some fallbacks. You are seeing ``edit-find-replace-symbolic`` here.

But ``filter-symbolic`` is preferred if you do have it; here's what it looks like with the Breeze icon theme:
Screenshot from 2024-11-13 17-32-56
Not a fan, but I would never question maximum leader within earshot.

### DIFF
--- a/lutris/gui/config/edit_saved_search.py
+++ b/lutris/gui/config/edit_saved_search.py
@@ -3,7 +3,7 @@ from copy import copy
 from gettext import gettext as _
 from typing import Callable, Dict, List, Optional, Tuple
 
-from gi.repository import Gtk
+from gi.repository import GObject, Gtk
 
 from lutris import runners, services
 from lutris.database import categories as categories_db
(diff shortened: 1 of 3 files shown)


strycore:
> 0.5.18 is taking a while 
If you feel this is stable enough, this can be merged for 0.5.18. I'd like some changes to this UI but we can make changes in the next release.

> One concern you may have is the icon in the pop-over button. @strycore's effort used filter-symbolic here. My icon theme hasn't got this icon (and it's ugly and its mother dresses it funny) so I picked some fallbacks. You are seeing edit-find-replace-symbolic here.

The icon is going to change based on the user's theme and the only way to have something consistent is for us to ship our own icon.

Issue #1218, opened 2018-10-19
Add checksum check for runtime


tannisroot:
As runtime can get corrupted during download due to unstable internet connection or interference from the ISP, it may cause errors for the user.
Lutris should run a checksum check for the runtime packages that it downloads and automatically re-download them if the check fails.

manuelVo:
Maybe it would be a good idea to sign the runtime and check that signature after downloading while we're at it. The way it's built right now if anyone were to break into the servers distributing the runtime they'd be able to distribute malware through the runtime. That could be prevented by signing the runtime.

LeXofLeviafan:
> automatically redownload

And if the checksum is [always wrong](https://forums.lutris.net/t/wine-nofshack-runner/7856/3)?

mYnDstrEAm:
This is **very** important for security reasons, as outlined in my issue linked above (closed as duplicate).

People don't even need to break into the servers distributing the runtime, for example it could be the developers themselves distributing them on purpose or unintentionally because some of the software they use for coding & building got infected for example.

Moreover, it would need to check after the download has finished but also afterwards to make sure the launcher is not replaced by some malware version of it. I don't know how this could be best implemented but maybe via regular checks of the hash.

A difficulty is getting the correct hash for the launchers, if possible hashes signed by developers. Eventually the binaries should be [reproducible](https://reproducible-builds.org/) so that they are validated to match the source code.

I think for secure gaming on Linux this issue and #4556 would need to be solved.

d-damien:
Given the size of some files, and the human element (you cannot know whether the user will change to a bad internet connection, is in a hurry etc) I believe the error message should offer the choice to the user to retry download or cancel the install. But never allowing to run with a wrong checksum. The user could make their own script if they know what they are doing.

@LeXofLeviafan ISP messing with the content of their clients are a shame, and CDNs well... accidents happen I guess. But it is simply not normal the file differs. The error message could include a very brief explanation about CDNs and ISPs, and invite the user to try from another place.

Given how spread (and proprietary) the video game community is, it is indeed an important element for security.

PR #5723, opened 2024-11-14
check hash for download installers

duhow:
Implements #1218 - not sure if this is enough to close it.

Using [6 Days A Sacrifice](https://lutris.net/games/install/3246/view) as example, we add `checksum`:

```yaml
  files:
  - game:
      filename: 6days_se.zip
      url: http://www.g4g.it/files3/2009/FullyRamblomatic_SpecialEditions_www.g4g.it.zip
      checksum: md5:5a546c58dd0323a6dcbc31dfcd90c339
```

### DIFF
--- a/lutris/gui/installer/file_box.py
+++ b/lutris/gui/installer/file_box.py
@@ -221,6 +221,7 @@ class InstallerFileBox(Gtk.VBox):
     def on_download_complete(self, widget, _data=None):
         """Action called on a completed download."""
         logger.info("Download completed")
+        self.installer_file.check_hash()
         if isinstance(widget, SteamInstaller):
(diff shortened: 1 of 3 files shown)


danieljohnson2:
Does this work with InstallerFileCollection? I feel like this will crash for any GOG or Amazon game, since they use that instead.

duhow:
I don't expect any compatibility issues, as the function seems to run only if checksum is defined.

https://github.com/lutris/lutris/blob/1492c47749cdc41b435ca445ae08632ffd77bb18/lutris/installer/installer_file.py#L216-L217

danieljohnson2:
That looks better.

danieljohnson2:
Even if ruff doesn't think so!

PR #5725, opened 2024-11-14
Move fluidsynth as optional command

duhow:
Apparently this is used for ScummVM?

I don't have this program and keeps warning every time I start Lutris:

> 2024-11-14 20:22:13,904: Command 'fluidsynth' not found on your system


strycore:
Yes, this can be marked as optional.

PR #5736, opened 2024-11-21
Runners: Duckstation: fix wrong options

B--B:
Some command line options in Duckstation runner does not exist, fix them and add all options available:

- batch mode (Advanced, set to True by Default for better integration with Lutris)
- Fastboot mode
- Slowboot mode
- Fullscreen
- No Fullscreen
- Start with no controller detection
- Custom path for settings .ini file (Advanced)

Test: runner downloaded and installed, tried to install and play a couple of games (Xenogears and Parasite Eve), options now works fine.

Ref: https://github.com/stenzek/duckstation/wiki/Command-Line-Arguments

### DIFF
--- a/lutris/runners/duckstation.py
+++ b/lutris/runners/duckstation.py
@@ -1,10 +1,12 @@
-from gettext import gettext as _
-from os import path
+"""DuckStation Runner"""
 
-from lutris import settings
+import os.path
+from gettext import gettext as _
+
 from lutris.exceptions import MissingGameExecutableError
 from lutris.runners.runner import Runner
 from lutris.util import system
+from lutris.util.log import logger
 
 
 class duckstation(Runner):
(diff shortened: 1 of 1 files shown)


danieljohnson2:
Whoa there, this PR changes the names ("option" key) for many of the options. That will break any existing games installed for this runner, and in particular changing ``main_file`` to ``rom_file`` is not a good idea- other parts of Lutris are looking for the ``main_file`` option in particular. Don't change that one!

B--B:
Oh nice catch, I wanted to change the label because DuckStation support also other formats (ISO in this case is a little bit misleading) but I've changed the option name, will fix it. Big picture and nogui does not exist on Duckstation command line so that commands do nothing at the moment, but if you prefer we can set batch mode as nogui option so user choice won't be discarded.

danieljohnson2:
I would prefer that, but ``rom_file`` is the main thing that needs to be fixed.

danieljohnson2:
Well, this fixes obsolete CLI flags, but also adds a few, so it's partly fixes and partly enhancement. I don't merge features as a rule, since @strycore should make the call on feature work.

Buuuut perhaps I can stretch the point, since the features are minor and the fixes are needed for this runner to work. He can always force-choke me if he disapproves, Darth Vader style.

Issue #5729, opened 2024-11-18
App ID is not set


OlegAckbar:
### Bug description

Although there's code for setting app id for Lutris in reality it's not set. This prevents Lutris from having app icon in task switcher on KDE.

### How to Reproduce

1. Open Lutris on KDE
2. Try to switch tasks
3. See that Lutris window doesn't have icon in task switcher

### Expected behavior

Lutris should have app id properly set to `net.lutris.Lutris`

### Log output

```shell
INFO     2024-11-18 20:29:25,827 [application.do_command_line:479]:Starting Lutris 0.5.18
INFO     2024-11-18 20:29:25,888 [startup.run_all_checks:120]:"card0" is AMD Radeon RX 6700 XT (1002:73df 1462:3982 amdgpu) Driver 24.2.7
DEBUG    2024-11-18 20:29:26,004 [xrandr.get_resolutions:91]:Retrieving resolution list
DEBUG    2024-11-18 20:29:26,006 [xrandr._get_vidmodes:18]:Retrieving 101 video modes from XrandR
DEBUG    2024-11-18 20:29:26,111 [path_cache._update_missing_games:114]:Checking for missing games
WARNING  2024-11-18 20:29:26,137 [runtime.__init__:171]:Runtime updates are disabled. This configuration is not supported.
DEBUG    2024-11-18 20:29:26,137 [api.check_stale_runtime_versions:52]:Modified at Пн 18 ноя 2024 20:21:45, will update after Вт 19 ноя 2024 02:21:45
DEBUG    2024-11-18 20:29:26,138 [lutriswindow.create_runtime_updater_cb:1329]:Runtime up to date
WARNING  2024-11-18 20:29:26,141 [lutriswindow.update_missing_games_sidebar_row:465]:Path cache out of date? (1 IDs missing)
DEBUG    2024-11-18 20:29:27,113 [http._request:88]:POST https://lutris.net/api/users/library?since=1731950881
INFO     2024-11-18 20:29:30,950 [application.do_shutdown:1048]:Shutting down Lutris
```

### System Information

```shell
Vulkan support: YES
Esync support: YES
Fsync support: YES
Wine installed: YES
Gamescope: YES
Mangohud: YES
Gamemode: YES
Steam: YES
In Flatpak: NO
[System]
OS: Arch Linux rolling n/a
Arch: x86_64
Kernel: 6.11.8-2-cachyos
Lutris Version: 0.5.18
Desktop: KDE
Display Server: wayland
[...]

PR #5738, opened 2024-11-21
Fix Lutris icon not used in KDE switcher on Wayland

B--B:
Wayland require the name of the program to be set (must be the same as application_id) or a generic Wayland icon will be used in app switcher instead the Lutris specific one.

Test: tested on Wayland, Lutris icon is used. No changes on X11 (it was already working even without this)

Fixes: lutris/lutris#5729


danieljohnson2:
OK, I think I understand; the default GTK behavior is [to take the last component of ``argv[0]``](https://docs.gtk.org/glib/func.set_prgname.html) which would just be 'lutris', so this changes it to the right thing.

I'll merge this fix.

PR #5752, opened 2024-11-27
feat(util/linux.py): add AOSC OS optenv32 path (/opt/32) to multiarch_lib_folders

MingcongBai:
AOSC OS is a non-multiarch distribution which ships its 32-bit x86 runtime (under the name optenv32) in /opt/32, and the "native" 64-bit x86-64 runtime in /usr/lib.

Add this path combo to make Lutris work on AOSC OS.


danieljohnson2:
I think that adding little tweaks for all these little distros is probably not a great idea.

I wonder if there's a way we could support any of them with a single strategy. Like, maybe there could be a ``LUTRIS_MULTIARCH_LIB_DIRS`` environment variable and the distro's packaging could arrange to set that somehow?

I am not sure what would be convenient for the packaging team at AOSC OS in that way.

PR #5753, opened 2024-11-28
Added retry logic.

guylamar2006:


danieljohnson2:
Might want to give us a bit more on what his is about. Why is retrying a failed installation helpful? Won't it just fail again?

strycore:
This is unlikely to be helpful and there is very little explanation to why this was made. 
We don't want to provide an experience where the user will have repeated failures, this retry thing isn't useful.

PR #5774, opened 2024-12-07
Use real MangoHud with Gamescope but skip both for steam runner

jmarcet:
Launching `mangohud lutris` and within `Lutris` enabling `Gamescope` yields no errors in my system (tested with more than 70 games) unlike the current implementation of `MangoHud` in `Lutris`.

Hence, this PR changes how `MangoHud` is set in the launch arguments so it can be combined with `Gamescope` instead of using `gamescope --mangoapp`, which if it behaved exactly the same way as the real `MangoHud` it could be a non issue, but the fact is the output is not the same, AFAIK it cannot be configured and it misses very interesting options available with the real `MangoHud`.

At the same time it disables setting `Gamescope` with the `steam runner` which at least in my system does not work and needs to be set within `Steam`, unlike `MangoHud` which can be launched as `mangohud steam-native` with no issues. Anyhow, I think it is less confusing this way, having both `MangoHud` and `Gamescope` not settable for the `steam runner`.

```
Vulkan support: YES
Esync support: YES
Fsync support: YES
Wine installed: YES
Gamescope: YES
Mangohud: YES
Gamemode: NO
Steam: YES
In Flatpak: NO

[System]
OS: Arch Linux rolling n/a
Arch: x86_64
Kernel: 6.12.3-z790i-1-1
Lutris Version: 0.5.18
Desktop: KDE
Display Server: wayland

[CPU]
Vendor: GenuineIntel
Model: 13th Gen Intel(R) Core(TM) i9-13900K
Physical cores: 24
Logical cores: 32

[Memory]
RAM: 62.5 GB
Swap: 0.0 GB

[Graphics]
Vendor: NVIDIA Corporation
OpenGL Renderer: NVIDIA GeForce RTX 4080 SUPER/PCIe/SSE2
OpenGL Version: 4.6.0 NVIDIA 565.77
OpenGL Core: 4.6.0 NVIDIA 565.77
OpenGL ES: OpenGL ES 3.2 NVIDIA 565.77
Vulkan Version: 1.4.303
Vulkan Drivers: NVIDIA GeForce RTX 4080 SUPER (1.3.289)
```

### DIFF
--- a/lutris/runner_interpreter.py
+++ b/lutris/runner_interpreter.py
@@ -10,32 +10,10 @@ from lutris.util.linux import LINUX_SYSTEM
 from lutris.util.log import logger
 
 
-def get_mangohud_conf(system_config):
-    """Return correct launch arguments and environment variables for Mangohud."""
-    # The environment variable should be set to 0 on gamescope, otherwise the game will crash
-    if system_config.get("gamescope"):
-        return None, None
-    if system_config.get("mangohud") and system.can_find_executable("mangohud"):
-        return ["mangohud"], {"MANGOHUD": "1", "MANGOHUD_DLSYM": "1"}
-    return None, None
-
-
 def get_launch_parameters(runner, gameplay_info):
     system_config = runner.system_config
     launch_arguments = gameplay_info["command"]
     env = {}
-
-    # MangoHud
-    if runner.name == "steam":
-        logger.info(
-            "Do not enable Mangodhud for Steam games in Lutris. "
-            "Edit the launch options in Steam and set them to mangohud %%command%%"
-        )
-    else:
-        mango_args, mango_env = get_mangohud_conf(system_config)
-        if mango_args:
-            launch_arguments = mango_args + launch_arguments
-            env.update(mango_env)
 
     prefix_command = system_config.get("prefix_command") or ""
     if prefix_command:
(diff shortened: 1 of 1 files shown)


danieljohnson2:
I do not understand what this is about. What are these errors you are getting with the current software?

jmarcet:
> I do not understand what this is about. What are these errors you are getting with the current software?

In theory `gamescope` and `manngohud` cannot be mixed, instead it is recommended to use `gamescope --mangoapp`.

When using `gamescope --mangoapp` I miss some info I have with plain `mangohud` which can be configured to personal taste unlike `mangoapp`. At least I could not find how to tune it.

The way `mangohud` is launched within `lutris`, if patched just to allow using both `mangohud` and `gamescope` together, there are many games where there are `vulkan` errors reported and the game quits.

However launching `mangohud lutris` and enabling only `gamescope` wthin `lutris` solves those errors, thus having `mangohud` and `gamescope` working together.

Hence this PR. It changes how `mangohud` is launched and ditches the `--mangoapp` option.

I have carried tests only in my system but since I saw the errors I had read about in `reddit` (I cannot find the correct thread now) and just by changing the launch order and dropping the envs `{"MANGOHUD": "1", "MANGOHUD_DLSYM": "1"}` everything works, I thought it could be useful sharing it.

In `Steam` I achieve the same with `mangohud gamescope -W 3840 -H 2160 -ef --hdr-enabled --rt -- %command%`

danieljohnson2:
Some brief research suggests that if you try to use both like this, the one you run first works and the second one does not- so your PR (if I follow it) will make mangohud work, but gamescope won't.

I do not think that's an improvement over "-mangoapp".

jmarcet:
I know what the available info suggests but I am using BOTH at the same time successfully. Without `gamescope` I have no HDR or if I force it with the relevant env vars I get washed out colors.

Screenshot_20241030_223603

Screenshot_20241215_124418

jmarcet:
I found what was different about my setup which made `mangohud` and `gamescope` able to run at the same time. I did not have vanilla `gamescope` installed, but `gamescope-nvidia-git-3.15.9` from Arch's AUR, hence I am closing this PR.

Thanks for your time and the great `lutris` 😄

PR #5797, opened 2024-12-15
Stop pointing winetricks to wine in open_wine_terminal

vasi:
When the user runs `winetricks`, they should get winetricks, not wine.

Fixes bug introduced in 4a62275daa55b6c6ecc76592dd34f61e0e41523f


vasi:
cc @danieljohnson2

danieljohnson2:
Did you know I get a notification every time anyone posts a message on this repo? You don't need to cc me like that. I see all! If I'm ignoring you, it's just because I'm a jerk.

Anyway, the fix looks fair to me. I'll merge it. Thanks!

PR #5798, opened 2024-12-15
translation: pt_BR update

victorsouzaleal:
A translation update before christmas HO HO HO
image

### DIFF
--- a/po/pt_BR.po
+++ b/po/pt_BR.po
@@ -5,8 +5,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: lutris\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2024-11-21 09:17-0300\n"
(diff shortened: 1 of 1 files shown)


danieljohnson2:
Bah, farsa! But I'll merge it anyway.

Looks like we're translating the changelog now, for Flathub. Not sure anyone intended that.

But that makes wonder. Should we be doing translations of stuff like "installed:yes" in the new fancy search? My concern with it is that we can save these things and a saved search might break if you changed languages on it.

(I hope Google Translate can be trusted on Christmas cheer...)

victorsouzaleal:
It didn't make any sense but I understood your "Bah Humbug!"  :rofl: 

I don't think it's worth translating these things related to fancy search, the tooltips are translated and very clear. (and knowing Brazilian users, a minority will use this and I'm not in it :smile: )

PR #5807, opened 2024-12-19
Adds BIOS directory to user storage config 

### LINKED ISSUE #431: Provide a way to specify BIOS file for libretro/retroarch cores
Hello,

I'm trying to find where to place BIOS file for PCXS Rearmed and Mednafen PSX cores, I can't find the information neither on lutris wiki, nor retroarch wiki.

So please provide a way, via lutris GUI, to select BIOS file for libretro cores.

P.S : Meanwhile, If you could also tell me where to place the file that would help me :)

wrightgabriel0220:
Closes #431 

When the BIOS directory is modified, that directory is validated (to ensure scanning isn't run on directories that are too large or obviously incorrect) then scanned. The results of that scan are cached in `~/.cache/lutris/bios-files.json`, where each file's name, size, date created, date modified, and md5 hash are stored. When a game is run using the Retroarch runner, Lutris checks this cache against the checksums of the BIOS files that Retroarch requests. If any are available in the cache, they are automatically moved into Retroarch's system directory and the game starts right away without issue.

Here's a quick video demo of the feature working:
https://github.com/user-attachments/assets/d9cdedaf-b5d4-4d8b-9c7b-29356f2934a1

Let me know if anything needs adjusting here! :)


danieljohnson2:
Is this a setting in the Updates tab that is just for libretro? Shouldn't it just be a libretro runner setting if so? Or will other runners use this eventually?

wrightgabriel0220:
> Is this a setting in the Updates tab that is just for libretro? Shouldn't it just be a libretro runner setting if so? Or will other runners use this eventually?

My understanding is the goal is to use this for other emulator runners like PCSX2 or FS-UAE which similarly require a bios file to be deposited in some folder somewhere, but that it's more important to get this working for RetroArch first since that encompasses a lot of content. https://github.com/lutris/lutris/issues/431#issuecomment-2502960861

danieljohnson2:
Okay, that sounds fair. Though I wonder who's going to do the changes to the other runners.

wrightgabriel0220:
I'm happy to apply that to the other runners, although maybe that should be a secondary issue and PR? I'd need to gather a list of all the runners that need this (unless we have that somewhere) and modify the runner script for each one. It wouldn't be super hard if we happen to have an easy way to know which runners need a BIOS. Otherwise, it'll just take me some time.

danieljohnson2:
I wouldn't pack that into this PR; better to be more focused. It can come later.

wrightgabriel0220:
Perfect. Actually, looking at the issues for 0.6.0, it looks like there's a couple that directly correlate with this for different runners (seeing PS2/PCSX2 and FS-UAE just glancing but there might be more). I'll follow up on those next.

strycore (approved):

PR #5813, opened 2024-12-21
Rename the 'sandbox' options to never say 'sandbox'.

danieljohnson2:
A sandbox is a security thing, and this feature was never that. This is about placing folders like 'Documents' in particular places, instead of in your actual 'Documents' directory.

 'Sandbox' is a very misleading term for this.

So this PR changes it, at shown:
image

"Desktop Integration Folder Location" is not obvious, but it more suggests "confusion" than "security", so that's an improvement.

### DIFF
--- a/lutris/runners/wine.py
+++ b/lutris/runners/wine.py
@@ -604,21 +604,22 @@ class wine(Runner):
             {
                 "option": "sandbox",
                 "type": "bool",
-                "section": _("Sandbox"),
-                "label": _("Create a sandbox for Wine folders"),
+                "section": _("Desktop Integration Folder Isolation"),
+                "label": _("Place desktop folders specifically"),
                 "default": True,
                 "advanced": True,
                 "help": _(
                     "Explicitly specify the location for desktop integration folders.\n"
-                    "If turned off, these are placed in '%s'"
+                    "If turned off, these are placed in '%s'.\n"
+                    "These folders include 'Documents', 'Pictures' and others."
                 )
                 % os.path.expanduser("~"),
             },
             {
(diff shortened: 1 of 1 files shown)


strycore:
Please never write "Place desktop folders specifically" because it doesn't mean anything at all.

danieljohnson2:
I'm open to suggestions, boss. How *do* we describe this thing? Can it be done without word-wrapping here?

strycore:
After inspection, the ability to set a custom location for Windows folders is a bullshit feature and should be removed.

We should have a checkbox with "Integrate system file in the prefix", that's it.

danieljohnson2:
I can manage that. I'd like the default to be to *not* integrate, so each prefix has its own Documents, Pictures, etc. I find it most annoying when a Wine game creates directories in ~/Documents or something like that. Is that okay?

strycore:
I believe that should have been the default behavior already and if it wasn't we're just lamers.

strycore:
Also, there shouldn't be a custom section, isolation checkbox should go where the wine prefix is set

danieljohnson2:
OK, let's close this PR- I'll change this in master. "Integrate system file in the prefix" it is, in Prefix section.

danieljohnson2:
OK, I've adjusted the setting. Here's what it looks like:
Screenshot from 2024-12-24 18-54-19

It's with the prefix, no section, and that means it is on the Game options. You can now only set this on a per-game basis, not globally for the runner. Defaults to off, and it is still an Advanced setting- the only one on the Game options tab for a Wine game.

PR #5826, opened 2024-12-29
Migrate the "GE-Proton (Latest)" wine version to "ge-proton"

danieljohnson2:
I understand we're going with "ge-proton" as the name of the auto-updating Umu Wine version. We're going to use "GE-Proton (Latest)" as the displayed name, but internally it will be "ge-proton".

This PR is a migration to convert the runner settings and all games to that name.


danieljohnson2:
Thanks!

Issue #5761, opened 2024-12-03
Tray Icon not Properly Displaying for Flatpak Package


RayJW:
### Bug description

I already filed an issue with the Flathub repo, but as I elaborated [there](https://github.com/flathub/net.lutris.Lutris/issues/461), I think this would ideally be fixed upstream. The solution is rather simple, the Application ID for the AppIndicator would have to match the reverse domain style e.g. `net.lutris.Lutris`, as this is how the icons are being installed. I don't see a downside to this for upstream, but as I also pointed out in the Flathub issue, there would be workarounds for this.

### How to Reproduce

1. Install the Lutris Flatpak from Flathub
2. Enable tray icon
3. Observe the broken tray icon on most DEs

### Expected behavior

To properly display the tray icon for the Flatpak as well.

### Log output

```shell
❯ flatpak run net.lutris.Lutris -d                    
2024-12-03 20:37:27,215: Command 'fluidsynth' not found on your system
Error: unable to open display 
2024-12-03 20:37:27,225: ['glxinfo', '-B'] command failed: Command '['glxinfo', '-B']' returned non-zero exit status 255.
2024-12-03 20:37:27,225: No available glxinfo output
2024-12-03 20:37:27,225: Invalid glxinfo received
INFO     2024-12-03 20:37:27,563 [application.do_command_line:479]:Starting Lutris 0.5.17
'DISPLAY' environment variable not set... skipping surface info
MESA: error: Use of VkSurfacePresentModeCompatibilityEXT without a VkSurfacePresentModeEXT set. This is an application bug.
INFO     2024-12-03 20:37:27,609 [startup.run_all_checks:120]:AMD Radeon RX 6950 XT (1002:73a5 1849:5238 amdgpu) Driver None
DEBUG    2024-12-03 20:37:27,859 [path_cache._update_missing_games:114]:Checking for missing games
DEBUG    2024-12-03 20:37:27,895 [api.check_stale_runtime_versions:52]:Modified at Di 03 Dez 2024 20:27:49, will update after Mi 04 Dez 2024 02:27:49
DEBUG    2024-12-03 20:37:27,910 [lutriswindow.create_runtime_updater_cb:1244]:Runtime up to date
DEBUG    2024-12-03 20:37:28,861 [http._request:88]:POST https://lutris.net/api/users/library?since=1733254615
INFO     2024-12-03 20:37:33,889 [application.do_shutdown:1051]:Shutting down Lutris
```

### System Information

```shell
[...]

Issue #5657, opened 2024-10-16
No tray icon on KDE Wayland


Victor239:
### Bug description

On Kubuntu 24.10 on a X11 desktop the Lutris tray icon is visible. On a Wayland desktop it is not. 

I found two related issues with no tray icon on Wayland #4934 #5249 but these are on GNOME desktops, and from what I understand GNOME is deprecating the system tray. This is not the case for KDE, and all my other apps work fine (I've got 10+ apps with system tray icons).

### How to Reproduce

1. Launch Lutris
2. Observe no tray icon 

### Expected behavior

Tray icon should be visible in KDE Wayland environment, just like it is for X11. 

### Log output

```shell
INFO     2024-10-16 21:35:19,318 [application.do_command_line:479]:Starting Lutris 0.5.17
INFO     2024-10-16 21:35:19,403 [startup.run_all_checks:120]:AMD Radeon 780M (1002:15bf f111:0006 amdgpu) Driver 24.2.3
WARNING  2024-10-16 21:35:19,438 [libretro.get_platform:114]:Game don't have a core set
WARNING  2024-10-16 21:35:19,438 [game.set_platform_from_runner:442]:The Libretro runner didn't provide a platform for Castlevania: Circle of the Moon (libretro)
WARNING  2024-10-16 21:35:19,439 [libretro.get_platform:114]:Game don't have a core set
WARNING  2024-10-16 21:35:19,439 [game.set_platform_from_runner:442]:The Libretro runner didn't provide a platform for Dragon Ball Z: The Legacy of Goku (libretro)
WARNING  2024-10-16 21:35:19,439 [libretro.get_platform:114]:Game don't have a core set
WARNING  2024-10-16 21:35:19,439 [game.set_platform_from_runner:442]:The Libretro runner didn't provide a platform for Men in Black: The Series (libretro)
WARNING  2024-10-16 21:35:19,440 [libretro.get_platform:114]:Game don't have a core set
WARNING  2024-10-16 21:35:19,440 [game.set_platform_from_runner:442]:The Libretro runner didn't provide a platform for The Legend of Zelda: Oracle of Seasons (libretro)
F: Filesystem suffix "ro" is not applicable for --nofilesystem
F: Filesystem suffix "ro" is not applicable for --nofilesystem
DEBUG    2024-10-16 21:35:19,562 [xrandr.get_resolutions:91]:Retrieving resolution list
[...]

PR #5839, opened 2025-01-04
Refactor Icon Name and Config to Reverse Domain Name Structure

RayJW:
This PR should fix some issues with the tray icon, especially with the Flatpak. It doesn't seem to affect the regular package at all and the Flatpak works correctly now, once [this PR](https://github.com/flathub/net.lutris.Lutris/pull/477) over there is merged. I personally think, this is the easiest approach to solve these issues. However, for the sake of transparency, here are some other “fixes” I found across some projects.

- Apply a patch at build time to change what I changed here only for the Flatpak ([example](https://github.com/flathub/org.virt_manager.virt-manager/blob/master/virt-manager-systray-icon-use-reverse-dns-naming-convention.patch))
- Build logic into the functions I touched in this PR to detect a Flatpak environment and only change the variables according to that.

As I already mentioned, I think this is the least disruptive approach. It's basically a set and forget solution, whereas the other two options may cause more maintenance work down the road and make the code/build more convoluted.

The only opposition to this PR might be, that it makes the icon names more convoluted. But in my limited experience more and more applications adopt the reverse domain name structure whether they have a Flatpak or not as it's become standard over various platforms with time.

fixes: #5761, #5657 (maybe)


danieljohnson2:
I think you've missed a spot. How about ``about-dialog.ui``? That refers to the icon name 'lutris'. Also ``status_icon.py`` line 88?

This is why I didn't want to do this; I was sure I'd missing icon names somewhere.

RayJW:
> I think you've missed a spot. How about `about-dialog.ui`? That refers to the icon name 'lutris'. Also `status_icon.py` line 88?
> 
> This is why I didn't want to do this; I was sure I'd missing icon names somewhere.

You're right. That's a bad oversight on my part. Makes sense, there are obviously other parts using the icon. I will gladly check again for more than tray references, unless you want to go down any of the other routes?

danieljohnson2:
It's an oversight, but I see no other way than the hard way, if you want to make this work.

I think you need to search for ``"lutris"`` in the .py and .ui files, and check each one to see if it is an icon name. The ones inside ``_(   )`` can be ignored- those can't be icon names because they are localized.

Have fun! :smile:

RayJW:
> I think you need to search for `"lutris"` in the .py and .ui files, and check each one to see if it is an icon name. The ones inside `_( )` can be ignored- those can't be icon names because they are localized.

I checked all references to `"lutris"`, `"lutris.png"`, or `icon =` in some sort and I think I got (let's say most of them to be safe) now. Perhaps you could quickly build the deb and quickly try it out, so it doesn't break, say the desktop icon or anything? I'm not familiar with the build process and don't have an `apt` machine around.

danieljohnson2:
I'm running Nobara, which is a Fedora derivative.

RayJW:
> I'm running Nobara, which is a Fedora derivative.

Oh, I see. Maybe I can quickly throw together an Ubuntu VM. Do you know how I would go about building the `.deb` package? Couldn't find anything in the docs.

danieljohnson2:
I'm not really familiar with the .deb build process. I haven't built one in years, and it wasn't for Lutris when I did it.

RayJW:
> I'm not really familiar with the .deb build process. I haven't built one in years, and it wasn't for Lutris when I did it.

OK, no worries. In that case, I'll conclude my testing here, as I do not have time to look into that as well. However, I did build the Flatpak and test the master branch Lutris and both seem to work fine, and I couldn't find any missing icons or similar.

strycore:
> He generally (but not always) does PRs shortly before the release.

My own shortcomings should not be made into established rules that govern the project. I handle PRs before release because that's when I get off my ass and start looking at them. It would be much preferable to review PRs in a timely manner. Ideally, they should be review or merged within a matter of hours.

Doesn't look like this patch would have bad consequences for the native package so ok to merge.

PR #5852, opened 2025-01-12
🇮🇷 Fix Persian translation

alr86:
Do you have any plan to support Weblate/Crowdin?
And why I cannot try it out? When I run the translation test commands, Lutris launches in RTL style but with English language


alr86:
> And why I cannot try it out? When I run the translation test commands, app launches in RTL style but with english language

Reported in #5853

PR #5893, opened 2025-01-28
Full Polish translation

ToRRent1812:
Tested internally, seems good so far.


danieljohnson2:
Always good to get our Polish polished. Thanks!

Issue #5882, opened 2025-01-24
Cover art /banner/icons stored in cache dir


thomas-mc-work:
### Bug description

Hello team!
I've noticed that all images are stored in `$XDG_CACHE_HOME/lutris/{banners,coverart}`. That means if you migrate your setup to another machine then you usually don't move the cache folder. The result is all icons are missing on the launchers in the interface. Also it should not harm the setup if you wipe the cache folder.

Solution: Persist this data in the already established `$XDG_DATA_HOME/lutris` folder.

### How to Reproduce

1. Create any game including a cover art or banner or icon
2. Remove the cache folder of Lutris (usually `~/.cache/lutris`)
3. Restart Lutris
4. See all images are gone

### Expected behavior

All images should stay in place when the cache folder is removed.

### Log output

```shell
(Not a technical issue but rather regarding the software design)
```

### System Information

```shell
[System]
    OS:              Ubuntu 24.04 Noble Numbat
    Arch:            x86_64
    Kernel:          6.8.0-51-generic
    Desktop:         XFCE
    Display Server:  x11

    [CPU]
    Vendor:          GenuineIntel
    Model:           Intel(R) Core(TM) i7-4790 CPU @ 3.60GHz
    Physical cores:  4
    Logical cores:   8

    [Memory]
    RAM:             15.5 GB
    Swap:            31.7 GB

    [Graphics]
    Vendor:          AMD
    OpenGL Renderer: AMD Radeon RX 580 Series (radeonsi, polaris10, LLVM 17.0.6, DRM 3.57, 6.8.0-51-generic)
    OpenGL Version:  4.6 (Compatibility Profile) Mesa 24.0.9-0ubuntu0.3
    OpenGL Core:     4.6 (Core Profile) Mesa 24.0.9-0ubuntu0.3
    OpenGL ES:       OpenGL ES 3.2 Mesa 24.0.9-0ubuntu0.3
    Vulkan Version:  1.3.275
    Vulkan Drivers:  AMD Radeon RX 580 Series (RADV POLARIS10) (1.3.274)
```

danieljohnson2:
This change has already been made for 0.5.18, which is the current version. You are running something older.

But I should note that 0.5.18 will continue to use the old directories if they exist. This was to make my life easier- so I can switch between Lutris versions for troubleshooting without breaking anything.

So once you upgrade to 0.5.18, you'll want to move your ``banners`` and ``coverart`` over by hand, I guess.

[...]

PR #5897, opened 2025-01-28
Add a migration to move banners and cover-art from .cache to .local/share

danieljohnson2:
v0.5.18 has logic to use ``~/.local/share/lutris`` for banners and coverart instead of ``~/.cache/lutris``- but just continues to use ``~/.cache/lutris`` if those directories are found there. That was a convenience for troubleshooting, but I do not think we want it to be that way forever.

This PR removes that logic and uses only ``~/.local/share`` for these things, and also adds a migration to move your banners and covert art to the current location that 0.5.18 already prefers.

See #5882.

### DIFF
--- a/lutris/migrations/__init__.py
+++ b/lutris/migrations/__init__.py
@@ -3,7 +3,7 @@ import importlib
 from lutris import settings
 from lutris.util.log import logger
 
-MIGRATION_VERSION = 15  # Never decrease this number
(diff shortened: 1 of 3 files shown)


thomas-mc-work:
> v0.5.18 has logic to use `~/.local/share/cache` for banners

Did you mean `~/.local/share/lutris`?

danieljohnson2:
Yes, I did. But Python knows what I mean!

PR #5923, opened 2025-02-13
Update melonDS to the latest version

Noxellar:
melonDS has been updated to 1.0RC so it would be good to point to the new linux package. I *guessing???* that this is a permalink that always points to the newest package though I could be wrong.

There is a pretty major caveat, that being that melonDS now requires a bunch of Qt6 libraries. I'm currently using the latest stable flatpak release of Lutris which doesn't have those. Not sure how this affects the regular packaged version/s of Lutris. It would be nice if somebody could troubleshoot that for me since I'm not experienced enough to deal with it. Thanks!

### DIFF
--- a/share/lutris/json/melonds.json
+++ b/share/lutris/json/melonds.json
@@ -4,7 +4,7 @@
     "platforms": ["Nintendo DS"],
     "runner_executable": "melonds/melonDS",
     "flatpak_id": "net.kuribo64.melonDS",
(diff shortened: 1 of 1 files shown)


xenoxriar:
Alternatively, would it be acceptable to instead link to an appimage rather than a binary? Probably bad practice and might not work but it would mean you don't need to manage any libraries since they're all bundled in the executable..

PR #5940, opened 2025-02-23
Add logging ability to the MAME runner

telanus:
Add logging ability to the MAME runner. Makes finding an error easier.

### DIFF
--- a/lutris/runners/mame.py
+++ b/lutris/runners/mame.py
@@ -178,6 +178,30 @@ class mame(Runner):  # pylint: disable=invalid-name
             "default": False,
         },
         {
+            "option": "verbose",
+            "type": "bool",
+            "section": _("Debugging"),
+            "label": _("Verbose"),
+            "help": _("display additional diagnostic information."),
+            "default": False,
+        },
+        {
+            "option": "log",
+            "type": "bool",
+            "section": _("Debugging"),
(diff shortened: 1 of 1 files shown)


danieljohnson2:
Seems nice, but consider using the fancy new ``conditional_on`` key to disable some of these settings when logging is off, like this:
```python
        {
            "option": "verbose",
            "type": "bool",
            "conditional_on": "log",
            "section": _("Debugging"),
            "label": _("Verbose"),
            "help": _("display additional diagnostic information."),
            "default": False,
        },
```

This would disable "verbose" in the UI if "log" is off. If MAME ignores ``-verbose`` when ``-log`` is not present, this would be a good thing to do.

telanus:
> This would disable "verbose" in the UI if "log" is off. If MAME ignores `-verbose` when `-log` is not present, this would be a good thing to do.

Not really necessary, as the three logging options work independent of each other, -verbose gives a certain level of info, that the other 2 doesn't. oslog and log is similar, but oslog gives a little bit more than log.

strycore:
When you add such option, please set them as Advanced. No one want to be burdened by a bunch of options about logging.

strycore:
Since the oslog option was not even parsed correctly, I've merged all 3 options into one.

PR #5959, opened 2025-03-01
add optional support of balanced directories for games config files a…

eikeno:
…nd covers/banners

Add support via environment variable LUTRIS_BALANCE_ASSETS of an additional directory in the storage path of game config files (YAML) and for games coverarts and banners.

When this variable is set, the following happens:

Instead of:
~/.local/share/lutris/games/cyberpunk_2077_goodies_collection-1664997482.yml the following file is read (or created when relevant): ~/.local/share/lutris/games/c/cyberpunk_2077_goodies_collection-1664997482.yml

Instead of:
~/.local/share/lutris/coverart/cyberpunk_2077_goodies_collection.jpg the following file is read (or created when relevant): ~/.local/share/lutris/coverart/c/cyberpunk_2077_goodies_collection.jpg

Intent:
While storing all the YAML and games related covers/banner files in a single directory works perfectly, when reaching high number of games, longer access time and lower perceived reactiveness of the GUI can happen, for example when switching from "Library/Favorites" "Library/Games".

This if of course mitigated when using modern fast storage such as SDD, but however it is possible to obtain slowdowns when dealing with a few thousands+ entries. The problem is notably perceptible when using emulation runners and trying to manage large collections with many thousands of ROM entries.

The proposed changes tries to be as minimal as possible, and non intrusive as it's clearly intended for rather niche use case, and not intending to be a default option: when the LUTRIS_BALANCE_ASSETS env var is not set, behaviour is unchanged.

The modification doesn't fully solve the scaling problem, as it is still possible that a specific directory holds too many files after a certain time, notably due to the prevalence of certain first letters in games titles, depending on the language (i.e, many title starting with letter 'w' en english while much less common in french) BUT
has the advantage of remaining very simple, not adding a new dependancy that could possibly provide more advanced/tunable balancing logic, while increasing a lot the treshold of appearance of the slowness.

Been using this locally for a few years, now proposing it upstream in case this can be of any interest.

Available for any modifications, explanations, tests etc. if needed.

Cheers

### DIFF
--- a/lutris/config.py
+++ b/lutris/config.py
@@ -21,7 +21,10 @@ def write_game_config(game_slug: str, config: dict):
     """Writes a game config to disk"""
     configpath = make_game_config_id(game_slug)
     logger.debug("Writing game config to %s", configpath)
-    config_filename = os.path.join(settings.CONFIG_DIR, "games/%s.yml" % configpath)
+    if os.environ.get("LUTRIS_BALANCE_ASSETS"):
+        config_filename = os.path.join(settings.CONFIG_DIR, "games/%s/%s.yml" % (game_slug[0], configpath))
+    else:
+        config_filename = os.path.join(settings.CONFIG_DIR, "games/%s.yml" % configpath)
     write_yaml_to_file(config, config_filename)
     return configpath
 
@@ -30,8 +33,12 @@ def duplicate_game_config(game_slug: str, source_config_id: str):
     """Copies an existing configuration file, giving it a new id that this
     function returns."""
     new_config_id = make_game_config_id(game_slug)
-    src_path = os.path.join(settings.CONFIG_DIR, "games/%s.yml" % source_config_id)
-    dest_path = os.path.join(settings.CONFIG_DIR, "games/%s.yml" % new_config_id)
+    if os.environ.get("LUTRIS_BALANCE_ASSETS"):
+        src_path = os.path.join(settings.CONFIG_DIR, "games/%s/%s.yml" % (game_slug[0], source_config_id))
+        dest_path = os.path.join(settings.CONFIG_DIR, "games/%s/%s.yml" % (game_slug[0], new_config_id))
+    else:
+        src_path = os.path.join(settings.CONFIG_DIR, "games/%s.yml" % source_config_id)
+        dest_path = os.path.join(settings.CONFIG_DIR, "games/%s.yml" % new_config_id)
     copyfile(src_path, dest_path)
     return new_config_id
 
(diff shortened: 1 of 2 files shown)


danieljohnson2:
I'm not super thrilled; it's weird to me to have a environment variable control this, and I do not like the repeating code to check it and form the paths- surely that should be factored out into a function or something.

And just how many games do you have installed that you're crushing your file-system under the load? I'd not have expected the number of config files in the directory to be a bottleneck.

I imagine you have benchmarked this change. Can you share your numbers?

eikeno:
I opted for the environment variable as i understood from reading the website / FAQ etc.  that adding new options was not always welcome. No problem however, I'll re-think this and check where a potential option for this would fit best.

Similar train of thought regarding the repeating code. If using a function for factoring this is OK, I'll modify that as well.

Not performed state of the art benchmark, just noticed the expected improvement (i.e from 10+ secs to maybe 2 or 3 when changing section on "Library") I'll see what I can do to provide accurate figures with my setup.  My test  config  currently has around 10K entries.

I'll update when possible.

Thanks

danieljohnson2:
That's not what I meant; an environment variable is still an option. I don't think it should be an option.

What I'd prefer is that if we're going to go to this organization, we do it all the time. But somehow we need to be compatible with existing installs- or at least we need a migration for them. See ``lutris/migrations/__init__.py`` if you'd like to consider that.

eikeno:
Thanks for the clarification.
Makes sense. I'll check that and see if I can come up with something clean enough

eikeno:
I've taken the time to evaluate impact of this PR on performances, adding timestamps at on_sidebar_changed() and update_revealer(), with actual collection (circa 10k entries) and crafted a test sample with 100K entries and the verdict is that is has near to zero impact.

I'm not 100% sure, but I think what caused this: I first did this mod a long time ago, and I'm pretty sure the game-paths.json wasn't existing at that time (or not under the current form), and i remember noticing important reactivity improvements at that time, but this got out of my head and confused what caused what it seems :/

The results i found with my test setup for a transition between "Recent" to "Games" were around 3.9s in both cases (balanced dirs or flat) with 9824 entries,  and around 70s  with 100,000 fake entries.

Clearly, what takes time when entries count grows is somewhere else, probably around filters, their calculation and application in relation with Gtk stuff. I didn't go too much further in that direction as there are Async calls making the monitoring tasks quite complex.

One thing I can exclude with not too much surprise, is the SQL part. It accounts for very little (which is good).

So I'm closing this as there no proven gains, and hope to be able to propose other PR in the future .

Thanks for your time and guidance.

PR #5960, opened 2025-03-01
fix Frenh locale utf to utf8 to avoid error message

eikeno:
In get_locale_choices():
French locale setting is set to "fr_FR.utf" while utf8 is used for other languages. When using the French locale in the GUI error appears, despite /etc/locale.gen is OK. Using utf8 fixe the problem and no errors are reported.


danieljohnson2:
Oh, that's clearly a bug. Let me get ``git blame`` to see what idiot fat-fingered that...

...

... never mind. Let's just merge this fix.

Issue #5393, opened 2024-03-29
Games from Steam Families not appearing in Lutris


bjoern-tantau:
I don't know if it was the same with the old Steam family sharing. But in the new Steam Families the games you get through your family are not shown on the profile page and thus also not in Lutris. Even if you have them installed.

But the Families games do show up in the last played section of the profile. Maybe that could be a workaround to get them into Lutris.

Of course it would be nicer if all the games were showing in Lutris.

Gab0-here:
I have the same issue, most of the games I play are on the steam family. Even if I can open steam and play them from there, it would definitely be better if it was integrated within Lutris.

It does make sense why it can't, since it's an opt-in option that has to be configured on every single client, and I don't even know what API they use, or if it can be activated there.

With that said you can still add them manually, and they WILL get recognized and work, you just need to add them as "locally installed game", select STEAM as the launcher, and put the steam app ID (for example, 1091500 for cyberpunk 2077), if you don't know the ID or how to get it, you can search for it on SteamDB.

![Screenshot from 2024-07-12 23-31-47](https://github.com/user-attachments/assets/c266bf4f-1b02-43a8-929b-a546588c7b54)

![Screenshot from 2024-07-12 23-32-01](https://github.com/user-attachments/assets/a3489570-0da8-410d-8414-818f7e34422c)

![Screenshot from 2024-07-12 23-33-29](https://github.com/user-attachments/assets/6573ac93-550c-488b-a1c9-99bab8722b61)

![Screenshot from 2024-07-12 23-33-38](https://github.com/user-attachments/assets/df4ad4c2-50cd-4c55-a0b5-08af83647a27)

voidlesity:
Any updates on this?
It would be really cool to have installed family games shown in lutris

GamerWierdo100:
I too have this issue

Alphenex:
Same issue, this definitely should be fixed/added

AliensInc:
I don't use the latest release, I use the git-version and this work.
Yesterday I installed a couple of games from steam and they all was there.

Download https://github.com/lutris/lutris and then run the bin/lutris script, do not install.

If this works then it has been fixed in the code(not yet released), if not there gotta be something wrong with your installation.

Just my thoughts.

bjoern-tantau:
> I don't use the latest release, I use the git-version and this work. Yesterday I installed a couple of games from steam and they all were there.
> 
[...]

Issue #5840, opened 2025-01-05
Steam families games not showing up in Lutris


GamerWierdo100:
### Bug description

Games which are shared with you through the Steam Families feature which are available to download, install, and play in Steam are not visible in Lutris, my friend who is also having this problem said that he could manually add these games with their application IDs

### How to Reproduce

1. Join a Steam Family
2. Confirm that the games are installable and playable through the steam client
3. Open lutris and try to find those games shared with you

### Expected behavior

The shared games should be detected and appear the same as any other

### Log output

```shell
~  lutris -d
2025-01-05 15:47:32,303: Command 'vulkaninfo' not found on your system
2025-01-05 15:47:32,454: The Battle.net source is unavailable because Google protobuf could not be loaded: No module named 'google'
INFO     2025-01-05 15:47:32,526 [application.do_command_line:480]:Starting Lutris 0.5.18
DEBUG    2025-01-05 15:47:32,527 [system._execute:109]:Executing lspci
INFO     2025-01-05 15:47:32,536 [startup.run_all_checks:120]:"card1" is Advanced Micro Devices, Inc. [AMD/ATI] Raphael (1002:164e 1458:d000 amdgpu) Driver None
DEBUG    2025-01-05 15:47:32,536 [system._execute:109]:Executing lspci
INFO     2025-01-05 15:47:32,544 [startup.run_all_checks:120]:"card2" is NVIDIA Corporation AD103 [GeForce RTX 4070 Ti SUPER] (10de:2705 1458:413e nvidia) Driver 565.77
ERROR    2025-01-05 15:47:32,544 [startup.check_libs:62]:i386 libgnutls.so.30 missing (needed by gnutls)
DEBUG    2025-01-05 15:47:32,664 [path_cache._update_missing_games:114]:Checking for missing games
DEBUG    2025-01-05 15:47:32,680 [api.check_stale_runtime_versions:52]:Modified at Sun 05 Jan 2025 03:47:17 PM , will update after Sun 05 Jan 2025 09:47:17 PM
DEBUG    2025-01-05 15:47:32,680 [lutriswindow.create_runtime_updater_cb:1358]:Runtime up to date
DEBUG    2025-01-05 15:47:36,889 [base.wipe_game_cache:180]:Deleting games from service-games for steam
DEBUG    2025-01-05 15:47:36,889 [sidebar.on_service_games_loading:718]:Service steam games loading
[...]

PR #5968, opened 2025-03-03
Add support for Steam Family Sharing

rosxz:
Add support for Steam Family Sharing by introducing a "Steam Family" source that lists the user's steam family games.
Set "STEAM_FAMILY_INCLUDE_OWN = True" in lutris settings to include the user's own Steam games.

Closes issue #5393 and #5840.


rosxz:
I noticed the total play time of the user isn't being pulled from steam. This is because the service id is different from steam's service id (on the match_game/get_game_service_api_page methods). 
I don't know the full implications, but in this case it seems to me it could be easily fixed by handling the steamfamily id as an alias to steam service on the lutris api.
Otherwise, there's the following hack (which again, I'm not 100% confident fixes everything):
```diff --git a/lutris/services/steamfamily.py b/lutris/services/steamfamily.py
index f476dce73..c566855d2 100644
--- a/lutris/services/steamfamily.py
+++ b/lutris/services/steamfamily.py
@@ -133,6 +133,18 @@ class SteamFamilyService(SteamService, OnlineService):
             records.extend([game for game in own_games if game["appid"] not in ids])
         return records

+    def match_game(self, service_game, lutris_game):
+        """Correctly set the service id for storing game entry and replace it again"""
+        self.id = SteamFamilyService.id
+        super().match_game(service_game, lutris_game)
+        self.id = SteamService.id
+
+    def match_games(self):
+        """Swap id to Steam service id to get Steam metadata (playtime)"""
+        self.id = SteamService.id
+        super().match_games()
+        self.id = SteamFamilyService.id
+
     def load(self):
         """Load the list of games"""
         try:```

strycore:
Nice! Would be nice to have this integrated in the Steam service instead of having a separate source at some point

PR #5974, opened 2025-03-05
feat: add buttons to zoom log text

Merrit:
The log text can be quite small and present an accessibility issue. This change will implement buttons that allow users to zoom in and out for better readability.

| Before | After |
|--------|--------|
| lutris-logs-before | lutris-logs-after |

[lutris-logs.webm](https://github.com/user-attachments/assets/7b957d58-5d90-41d9-8565-deadc593717f)

### DIFF
--- a/lutris/gui/dialogs/log.py
+++ b/lutris/gui/dialogs/log.py
@@ -4,7 +4,7 @@ import os
 from datetime import datetime
 from gettext import gettext as _
 
-from gi.repository import Gdk, GObject, Gtk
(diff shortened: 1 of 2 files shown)


Zebra2711:
Maybe `Ctrl` + `scroll` should be used instead of adding more buttons.

Merrit:
There isn't already an excess of buttons on this interface, and a shortcut would likely go undiscovered by the majority of users.

Even if it included a shortcut (which sounds reasonable), purely for discoverability it should have a way for the average user to know this feature exists without having pre-existing knowledge of the shortcut.

PR #5980, opened 2025-03-10
add Chinese (Traditional) locale

Zebra2711:


danieljohnson2:
This looks fine, but I did test entering some random custom locale code into the locale box in Lutris, and it all seemed to go just fine. I am not sure what problem you have been having.

Zebra2711:
> This looks fine, but I did test entering some random custom locale code into the locale box in Lutris, and it all seemed to go just fine. I am not sure what problem you have been having.

idk, but it seems like the dependency installation is broken in the workflow action of Mypy

the51stater:
> This looks fine, but I did test entering some random custom locale code into the locale box in Lutris, and it all seemed to go just fine. I am not sure what problem you have been having.

with Lutris 5.17, you can enter the custom locale code
with Lutris 5.18, it's now allowed to enter a custom locale code. You can only choose one from the selection list.
Screenshot From 2025-03-09 00-03-07

the51stater:
BTW, we cannot declare that the code 'zh_CN.utf8' owns the 'Chinese' language option. The 'zh_CN.utf8' is only used by the people from China. It's not used by the people from Taiwan and some other countries. At least so far. Maybe there will be a new language option named 'Taiwanese Mandarin' in the future, but that has not happened officially.

Zebra2711:
> with Lutris 5.17, the custom locale code was zh_TW.utf8 after being updated to Lutris 5.18 through the deb file installation, the code zh_TW.utf8 was said (invalid)

That did not work. You should set the environment variable using LC_ALL=zh_TW.utf8 instead.

the51stater:
> That did not work. You should set the environment variable using LC_ALL=zh_TW.utf8 instead.

According to my memory, with the old Lutris 5.17, the environment variable did not work on the game Black Desert. Only the locale did.
I will test it again with the new Lutris 5.18.
However, that will result in confusion because the user does not know if the environment variable can override the locale.

Zebra2711:
> I will test it again with the new Lutris 5.18
Check my repo, i also made a deb package for testing

the51stater:
> Check my repo, i also made a deb package for testing

Thank you for that deb package. However, it's uninstallable due to uninstallable dependencies.
My OS is Ubuntu 24.10.

the51stater:
The workaround with the environment variable LC_ALL works. It can override my system default locale, English. All traditional Chinese characters in the game Black Desert are readable with this workaround.
Screenshot From 2025-03-12 16-20-17

danieljohnson2:
I see that on 0.5.18 it labels ``zh_TW.utf8`` as "(invalid)" which is wrong; this is fixed in 0.5.19 already, which is why I didn't see it. But I think it should work if you enter ``zh_TW.utf8`` in the box and ignore the "(invalid)" thing.

The bad news is that 0.5.19's release went wrong, and there's no .deb or Flatpak. The word from maximum leader is to hang on for 0.5.20.

Zebra2711:
> The following packages have unmet dependencies:
> lutris: 相依關係: glib2 but it is not installable
> 相依關係: gnome-desktop but it is not installable
> 相依關係: gtk3 but it is not installable
> 相依關係: python-cairo but it is not installable
>  ...
> 相依關係: xorg-xrandr but it is not installable
Those dependencies are from my Arch Linux system. I forgot to map that to the Ubuntu pkgs

PR #5984, opened 2025-03-12
update girepository dev for CI

Zebra2711:
Update dependency to libgintrospection2.0-dev for PyGObject 3.51.0 compatibility


danieljohnson2:
Seems like we need this; let's give it a try.

PR #5987, opened 2025-03-13
Fixes #5205

### LINKED ISSUE #5205: EA App login page showing in a foreign language compared to the system one
Hello,

New Lutris user here. I  installed the EA App launcher, but when connecting my EA account to add my game library to Lutris, the sign in page is in a foreign language (compared to my distro language which is English). Looks like Arabic or something alike, I am not sure, sorry ^^ !

I do not think this is an expected behaviour, at least on a system not configured in this language.

image

Thanks :) !

LoKolbasz:
Added English fallback locale to prevent EA login screen being displayed in a language that looks like Arabic, when `LANG` was set to an unsupported locale. Tested with `LANG=hu_HU.UTF-8`, EA and Epic Games Launcher logins. #5205

### DIFF
--- a/lutris/gui/dialogs/webconnect_dialog.py
+++ b/lutris/gui/dialogs/webconnect_dialog.py
@@ -41,7 +41,10 @@ class WebConnectDialog(ModalDialog):
         if lutris_locale:
             webview_locales = [lutris_locale.split(".")[0]] + webview_locales
         logger.debug(
-            f"Webview locale fallback order: {webview_locales[0]}{''.join(' -> ' + i for i in webview_locales[1:])}"
+            f"Webview locale fallback order: "
+            f"[Lutris locale]: '{lutris_locale}' -> "
+            f"[env: LANG]: '{environment_locale_lang}' -> "
+            f"[Default]: '{webview_locales[-1]}'"
         )
         self.context.set_preferred_languages(webview_locales)
 
--- a/lutris/gui/dialogs/webconnect_dialog.py
+++ b/lutris/gui/dialogs/webconnect_dialog.py
@@ -2,7 +2,7 @@
 
 import os
 from gettext import gettext as _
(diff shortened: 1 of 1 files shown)


danieljohnson2:
This seems like a good idea, but should we just use ``os.environ`` here? You can set the 'locale' in the Lutris configuration, and that seems like it should override this.

Also, I think code will just crash if ``LANG`` is not set. That's probably bad.

LoKolbasz:
> You can set the 'locale' in the Lutris configuration, and that seems like it should override this.

Where? I couldn't find it.

danieljohnson2:
Open "Preferences" then select "global options"- search for "locale".

To access this in the code, you do something like this:

```python
            lutris_config = LutrisConfig()
            locale = lutris_config.system_config.get("locale")
```

And then if that's ``None`` you could use ``os.environ`` instead.

LoKolbasz:
Thanks. I'll look into it tomorrow.

LoKolbasz:
> This seems like a good idea, but should we just use `os.environ` here? You can set the 'locale' in the Lutris configuration, and that seems like it should override this.
> 
> Also, I think this code will just crash if `LANG` is not set. That's probably bad.

Applied your suggestions.

I also added a debug message containing the locale fallback order, to make a locale misconfiguration easier to diagnose.

Tested with setting German as a the locale for lutris, and it was correctly reflected in the EA login screen.

danieljohnson2:
Getting better. Some more suggestions while we are at it:

1. We don't need all those local variable type notations. At least ``webview_locales`` and ``lutris_config`` should be deduced by mypy anyway.
2. I think it would be more pythonic to say ``if environment_lang:`` on line 38, and likewise on line 41. This works to 'not None and not blank' for a ``str``, which I think is appopriate- you don't want a blank language code.

danieljohnson2:
I'm happy with this. Merging.

PR #5995, opened 2025-03-15
Changed ROM import from adding single file, to adding a folder.

Zybyte85:
This allows people to bulk import ROMs. It also is recursive, so if you have a folder full of more folders with ROMs, it will scan all of them.

### DIFF
--- a/lutris/gui/addgameswindow.py
+++ b/lutris/gui/addgameswindow.py
@@ -42,8 +42,8 @@ class AddGamesWindow(ModelessDialog):  # pylint: disable=too-many-public-methods
         (
             "application-x-firmware-symbolic",
             "go-next-symbolic",
-            _("Import a ROM"),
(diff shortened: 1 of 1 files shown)


danieljohnson2:
I think this PR would mean you can't just pick one ROM to import via this dialog, right?

I should note, we could always import many ROMs at once via drag and drop. But this will be more discoverable, to be sure.

Zybyte85:
Yes, unfortunately it does. I tried, but there doesn't seem to be a way to
open a file dialog to open a file or a directory.

Zybyte85:
Couldn't we have it so you drag and drop single files and use the menu for folders? I feel like most of the time you are uploading folders instead of single files of ROMs anyway, so for the rare case of a single ROM you could just drag it.

danieljohnson2:
I think that's what this PR will do. I'll leave it to the boss to decide if he wants this.

PR #6003, opened 2025-03-18
add setting to allow override system locale with global options locale

eikeno:
Add an option in global options, visible in "advanced" mode, that allows to override system locale with the locale set in Lutris global options.

The interest is for people like me needing en_US (or anything else) system locale, but who still appreciate being able to have GOG download localized installer files in their native (or preferred) language, when available.

Not providing anything related to translations of added strings to help review.
Please let me know if any changes are required.

### DIFF
--- a/lutris/sysoptions.py
+++ b/lutris/sysoptions.py
@@ -426,16 +426,6 @@ system_options = [  # pylint: disable=invalid-name
     },
     {
         "section": _("Game execution"),
-        "option": "override_system_locale",
-        "type": "bool",
-        "default": False,
-        "advanced": True,
-        "conditional_on": "locale",
-        "label": _("Override system locale"),
-        "help": _("Use the locale defined above in priority to any system value, whenever possible."),
-    },
-    {
-        "section": _("Game execution"),
         "option": "prefix_command",
         "type": "string",
         "label": _("Command prefix"),
--- a/lutris/sysoptions.py
+++ b/lutris/sysoptions.py
@@ -426,6 +426,16 @@ system_options = [  # pylint: disable=invalid-name
     },
     {
         "section": _("Game execution"),
+        "option": "override_system_locale",
+        "type": "bool",
+        "default": False,
+        "advanced": True,
+        "conditional_on": "locale",
+        "label": _("Override system locale"),
+        "help": _("Use the locale defined above in priority to any system value, whenever possible."),
+    },
+    {
+        "section": _("Game execution"),
         "option": "prefix_command",
         "type": "string",
         "label": _("Command prefix"),
--- a/lutris/util/i18n.py
+++ b/lutris/util/i18n.py
@@ -5,12 +5,15 @@ import locale
 from lutris.config import LutrisConfig
 from lutris.util.log import logger
 
+
 def get_locale_override():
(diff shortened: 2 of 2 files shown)


danieljohnson2:
Hmm. Maybe we should *always* override these things with the Lutris locale setting, if it is set. Why would you ever want to have this new option turned off?

eikeno:
There might be a lot of reasons to want to avoid that: for example that might be that game A has nice a native language adaptation while game B's one is garbage.

Another one might be that using the native language version could cause another version of the installer to be used, maybe with different feature. 

I've used the game 'Flatout' on GOG for testing, and noticed the installer versions are different for the en_US and the fr_FR/es_ES/IT ones.
 I didn't checked this point any further, could be totally harmless in this instance, but shows that there can be corner cases with providers, and allowing choice either globally or per game can be handy.  Also, now (from what I've seen), only GOG service uses this code, but could be welcome for some future services (for the same kind of reasons).

My main idea was the option will be enjoyed by some but not wanted by others, at least as a default, and could cause user complaints that the behaviour changed 'without warning'.

I also understand your point of view, though. Maybe a middle ground would be to remove the option from "advanced" so that it gets some visibility for those who needed,  while keeping it turned by default for now, so that it can be communicated about in next release changelog, and maybe warn the feature could become enable by default in future (depending on eventual positive or negative feedback).

What do you think ?

danieljohnson2:
If the option is just there because someone *might* want this, or *maybe* someday we'll want it for something, then I don't think we should have an option. 

To me, this switch feels like a switch to turn off a bug. If that's what it is, we should just fix the bug. 

If only GOG uses it (as it appears to me) that would argue it's just a bug in the GOG service in particular.

Remember, users can change the Lutris global setting before installing, then change it back after, if they actually need to switch locales for installation for some reason.

I am now feeling more strongly that this should just be a bug fix, with no UI. The GOG service should use our locale settings, if that is set.

That's what I think.

eikeno:
ok. I've modified accordingly. UI part removed

danieljohnson2:
OK, I'll call this a bug fix and merge it.

AliensInc:
> If the option is just there because someone _might_ want this, or _maybe_ someday we'll want it for something, then I don't think we should have an option.
> 
> To me, this switch feels like a switch to turn off a bug. If that's what it is, we should just fix the bug.
> 
> If only GOG uses it (as it appears to me) that would argue it's just a bug in the GOG service in particular.
> 
> Remember, users can change the Lutris global setting before installing, then change it back after, if they actually need to switch locales for installation for some reason.
> 
> I am now feeling more strongly that this should just be a bug fix, with no UI. The GOG service should use our locale settings, if that is set.
> 
> That's what I think.

I also use en_US as system locale but I wonder why would I then need another for games, can't figure that out😜

I've used en_US since way back. I'm an old progammer and then there was a big problem with the error messages from ODBC. The error used both English and in my case Swedish in the same sentence. I really disliked that so from that on I started to use en_US for everything.

Also, sending an error log to support in local language is not a good idea as they probably can't read it, OK translate it but that's just more work for them and om some cases you might need the 'exact' layout of the error message.

...my 5 cents...

Issue #5977, opened 2025-03-06
Something wrong with the new Emulator BIOS feature for dc (Flycast)


Otieze:
### Bug description

Have the file in the right location and md5 checks (e10c53c2f8b90bab96ead2d368858623) (even though flycast did not use to need bios files before 0.5.19, games used to launch normally but Lutris brags about it now), Other cores work correctly (for example PSX Beetle works fine)

### How to Reproduce

launch dc game using libretro's flycast

### Expected behavior

game launches

### Log output

```shell
'dc/dc_boot.bin'

Traceback (most recent call last):

  File "/usr/lib/python3.13/site-packages/lutris/exception_backstops.py", line 29, in wrapper
    result = function(*args, **kwargs)

  File "/usr/lib/python3.13/site-packages/lutris/game.py", line 769, in configure_game
    raise error

  File "/usr/lib/python3.13/site-packages/lutris/util/jobs.py", line 30, in target
    result = self.function(*a, **kw)

  File "/usr/lib/python3.13/site-packages/lutris/runners/libretro.py", line 257, in prelaunch
    required_firmware_checksum = checksums[required_firmware_filename]
                                 ~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^

KeyError: 'dc/dc_boot.bin'

Lutris log:
[INFO:2025-03-06 18:18:50,760:application]: Starting Lutris 0.5.19
[INFO:2025-03-06 18:18:50,772:startup]: "card1" is Advanced Micro Devices, Inc. [AMD/ATI] Ellesmere [Radeon RX 470/480/570/570X/580/580X/590] (1002:67df 1462:3417 amdgpu) Driver 24.3.4
[WARNING:2025-03-06 18:18:50,940:lutriswindow]: Path cache out of date? (2 IDs missing)
[WARNING:2025-03-06 18:19:00,604:game]: Game still running (state: running)
[INFO:2025-03-06 18:19:00,604:game]: Stopping Banjo-Kazooie (libretro)
[INFO:2025-03-06 18:19:36,156:game_import]: Installing GoldenEye 007
[WARNING:2025-03-06 18:20:03,662:game]: Game still running (state: running)
[INFO:2025-03-06 18:20:03,662:game]: Stopping GoldenEye 007 (libretro)
[INFO:2025-03-06 18:21:50,482:game_common]: Creating new configuration with runner wine
[...]

PR #6007, opened 2025-03-20
Search BIOS files in Lutris RA dir instead of user level RA.

eikeno:
There are references to ~/.config/retroarch but when the runner is installed, it is created at ~/.local/share/lutris/runners/retroarch

The references cause correct BIOS files under
~/.local/share/lutris/runners/retroarch/system to be skipped.

Due to recent changes, missing firmware won't cause crash, but this still prevent expected behavior (I presume) to avoid interference between Lutris' RA runner setup and a user level setup,and discovery of bios files placed in the runner directory.

before change, with ~/.local/share/lutris/runners/retroarch/system populated (files manually added in system/dc/):

[libretro.prelaunch:269]:Firmware 'dc/dc_boot.bin' not found! [libretro.prelaunch:269]:Firmware 'dc/naomi.zip' not found! [libretro.prelaunch:269]:Firmware 'dc/hod2bios.zip' not found! [libretro.prelaunch:269]:Firmware 'dc/f355dlx.zip' not found! [libretro.prelaunch:269]:Firmware 'dc/f355bios.zip' not found! [libretro.prelaunch:269]:Firmware 'dc/airlbios.zip' not found! [libretro.prelaunch:269]:Firmware 'dc/awbios.zip' not found!

after change, system/ similarly populated:
[libretro.prelaunch:268]:Firmware 'dc/dc_boot.bin' found (No checksum info) [libretro.prelaunch:268]:Firmware 'dc/naomi.zip' found (No checksum info) [libretro.prelaunch:268]:Firmware 'dc/hod2bios.zip' found (No checksum info) [libretro.prelaunch:268]:Firmware 'dc/f355dlx.zip' found (No checksum info) [libretro.prelaunch:268]:Firmware 'dc/f355bios.zip' found (No checksum info) [libretro.prelaunch:268]:Firmware 'dc/airlbios.zip' found (No checksum info) [libretro.prelaunch:268]:Firmware 'dc/awbios.zip' found (No checksum info)

Now BIOS files are detected, if present.

probable fix for issue #5977

### DIFF
--- a/lutris/runners/libretro.py
+++ b/lutris/runners/libretro.py
@@ -182,7 +182,7 @@ class libretro(Runner):
         """Return the system directory used for storing BIOS and firmwares."""
         system_directory = retro_config["system_directory"]
         if not system_directory or system_directory == "default":
-            system_directory = "~/.config/retroarch/system"
+            system_directory = get_default_config_path("system")
         return os.path.expanduser(system_directory)
 
     def prelaunch(self):
@@ -212,7 +212,7 @@ class libretro(Runner):
             retro_config["rgui_config_directory"] = get_default_config_path("config")
             retro_config["overlay_directory"] = get_default_config_path("overlay")
             retro_config["assets_directory"] = get_default_config_path("assets")
-            retro_config["system_directory"] = "~/.config/retroarch/system"
+            retro_config["system_directory"] = get_default_config_path("system")
             retro_config.save()
         else:
             retro_config = RetroConfig(config_file)
--- a/lutris/runners/libretro.py
+++ b/lutris/runners/libretro.py
@@ -123,14 +123,11 @@ class libretro(Runner):
         return ""
 
     def get_core_path(self, core):
-        """Return the path of a core, prioritizing Retroarch cores"""
-        lutris_cores_folder = os.path.join(self.directory, "cores")
-        retroarch_core_folder = os.path.join(os.path.expanduser("~/.config/retroarch/cores"))
+        """Return the path of a core from libretro's runner only"""
+        lutris_cores_folder = get_default_config_path("cores")
         core_filename = "{}_libretro.so".format(core)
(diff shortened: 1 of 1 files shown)


danieljohnson2:
Looks good, but what about line 128, in ``def get_core_path``? Should that still be using ``~/.config/retroarch/cores``? It does not agree with the default config settings elsewhere in ``libretro.py``.

eikeno:
You're right.

A question: get_core_path() comment states "Return the path of a core, prioritizing Retroarch cores"

should I:

1) modify so that only cores in retro_config["libretro_directory"] are checked, and remove 
logic of searching "~/.config/retroarch/cores"  at all ?

2) modify so that "lutris" core is prioritzed over possible RA core ?

I'm personally more for 1) because:
- Since libretro runner will install its own retroarch files, it would in all case make more sense to a least prioritize Lutris version. 
- Also Retroarch versions used by Lutris and at user level might mismatch and create weird issues in some cases (at least, creates potential for that).
- As for BIOS files, a user wanting to share the same set of files in RA and Lutris still can do it with symlink, but at least it's intentional and under his responsibility.

Please let me know what's preferred, I'll update accordingly

danieljohnson2:
I agree we should at least prioritize Lutris's core. Since we don't use ``~/.config/retroarch`` for anything else, it's odd to use it just for this. Feels like a leftover.

I'd say: try removing ``~/.config/retroarch`` from the search and see if that causes any problems for you.

eikeno:
done. Tested ok here with a DC game. Tested under a few conditions (new install, delete retroarch.cfg etc) and all is fine.

If merged, Ii'll do a separate PR to docs repo for updating the libretro section doc (and add a bit more details, like symlink).

danieljohnson2:
'kay. This looks like a fix to me, so I'll merge it now.

PR #6009, opened 2025-03-22
modify refs from ~/.config/lutris to ~/.local/share/lutris

eikeno:
Also make utils/clean_config use CONFIG_DIR from settings.py instead of
hardcoded ref.

1 typo fix


danieljohnson2:
Looks legit!

PR #6010, opened 2025-03-24
Add support for downloading local files with file:// url

bnerickson:
This adds support for downloading LOCAL files specified in ymls using the requests-file FileAdapter: https://pypi.org/project/requests-file/ .  This allows:
1. A user to specify their own file location without running an http server.
2. A user to specify their own file location in their own custom ymls for installers that require a user manually download the file from a location online.  Ex: https://lutris.net/games/install/30460/view .  If Lutris supports a "silent" installation option in the future, this would allow users to install without manually having to specify the file location.

This was the sample yml I used for testing where the `t1_setup` variable is set differently in my actual environment:

```
description: 'Installs the "Retro Version" configuration for Starsiege: Tribes (Tribes
  1)'
game_slug: starsiege-tribes
gogslug: ''
humblestoreid: ''
installer_slug: starsiege-tribes-retro
name: 'Starsiege: Tribes'
notes: 'Please download manually the installer for "Starsiege: Tribes - Retro Version"
  from ModDB at this URL: https://www.moddb.com/games/tribes/downloads/starsiege-tribes-retro-version'
runner: wine
script:
  custom-name: 'Starsiege: Tribes - Retro'
  files:
  - t1_setup: /local/path/to/tribes/installer/retro_141_installerv2.exe
  game:
    arch: win32
    exe: $GAMEDIR/drive_c/t1retro/Tribes.exe
    prefix: $GAMEDIR
    working_dir: $GAMEDIR/drive_c/t1retro
  installer:
  - task:
      args: /S /D=C:\\t1retro
      executable: t1_setup
      name: wineexec
      prefix: $GAMEDIR
  - execute:
      command: sed -i 's/$pref::VideoFullScreen = "FALSE";/$pref::VideoFullScreen
        = "TRUE";/' $GAMEDIR/drive_c/t1retro/config/ClientPrefs.cs
slug: starsiege-tribes-retro
steamid: null
version: Retro
year: 1998
```


danieljohnson2:
I see that this PR adds a dependency (``python3-requests-file``). If I follow, it's to support 'file:' URLs in install scripts.

That seems like its for personal (per user) installer scripts, rather than using the "N/A:" prefix that would work for a second user by prompting for a file.

I can't say whether this is a good idea, but I'd feel better if we didn't need another dependency for it. I see that you don't get "file:" for free with the 'requests' method, which surprised me.

bnerickson:
> I see that this PR adds a dependency (`python3-requests-file`). If I follow, it's to support 'file:' URLs in install scripts.

Correct.  It "extends" (for lack of a better word) support for "file://" urls within the python "requests" library.

> That seems like its for personal (per user) installer scripts, rather than using the "N/A:" prefix that would work for a second user by prompting for a file.

Correct.  I wouldn't expect this to be used on lutris ymls that are uploaded to the online database.

> I can't say whether this is a good idea, but I'd feel better if we didn't need another dependency for it. I see that you don't get "file:" for free with the 'requests' method, which surprised me.

Yes, I wish it was native in "requests" as well.

This solution could be broken out such that "file://" urls are treated differently such as copying the file to the destination location instead of being "converted" (by pre-pending the "file://" to the file variable name) to a url.  However, natively treating local "file://" urls like any other url seemed like the cleanest option requiring the least amount of change.

eikeno:
I've drafted this:
https://github.com/eikeno/lutris/commit/803e38edf6f561eab72ad3a0a96410abfb6dafa5

as this local, i thought maybe we could bypass requests in this case, while keeping progress feedback. I'm not sure if the progress must be used or just for remote stuff ? If that could be avoided; seq_local_copy() could be replaced with usual copy
I've kept chunksize at 8192 for local copy (as is used with requests), increasing it should help speed up with larger files though.

tested with a test installer script with file:// URLs in it, works as expected/

I hope I'm not "intruding",  just wanted to share my draft to see hat you guys think, as this problem is interesting.

bnerickson:
> I've drafted this: [eikeno@803e38e](https://github.com/eikeno/lutris/commit/803e38edf6f561eab72ad3a0a96410abfb6dafa5)
> 
> as this local, i thought maybe we could bypass requests in this case, while keeping progress feedback. I'm not sure if the progress must be used or just for remote stuff ? If that could be avoided; seq_local_copy() could be replaced with usual copy I've kept chunksize at 8192 for local copy (as is used with requests), increasing it should help speed up with larger files though.
> 
> tested with a test installer script with file:// URLs in it, works as expected/
> 
> I hope I'm not "intruding", just wanted to share my draft to see hat you guys think, as this problem is interesting.

No you're not intruding, this is great, thanks!  Like I said above, either approach works for me; so it's up to the maintainers which (if any) approach is preferred.  I think my only comment/suggestion on your approach @eikeno would be splitting async_download into two defs: `def async_download` for normal downloading using the requests module and another def  like `def file_copy` that does the copying if the url is a file instead of having giant if/else statements.

PR #6016, opened 2025-03-25
Add support for downloading local files with file:// url

eikeno:
possible variant of PR #6010:  support file:// in scripts URL and download locally keeping the progression bar feature.
No added dependency.

### DIFF
--- a/lutris/util/downloader.py
+++ b/lutris/util/downloader.py
@@ -2,6 +2,7 @@ import bisect
 import os
 import threading
 import time
+import urllib
 from typing import Any
 
 import requests
@@ -35,7 +36,7 @@ class Downloader:
         self.referer = referer
         self.stop_request = None
         self.thread = None
-        self.url_is_file = url.startswith("file://")
+        self.url_is_file = urllib.parse.urlparse(self.url).scheme == 'file'
 
         # Read these after a check_progress()
         self.state = self.INIT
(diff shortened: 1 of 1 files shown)


danieljohnson2:
I do like this better on the whole.

Some detail questions. Why is this using a Session() object when that wasn't needed before?

And line 145 has an ``isinstance`` check that I don't understand. What is this guarding against? Can this can ever fail?

danieljohnson2:
Might be better to parse the file URL with urllib, like this:
```python
urllib.parse.unquote(urllib.parse.urlparse("file:///usr/lib%201").path) == '/usr/lib 1'
urllib.parse.urlparse("file:///usr/lib").scheme == 'file'
```

That should do a more reliable job of decoding file URLs.

eikeno:
First ones were leftovers of my testing phase.Removed. I also added your urllib statements, indeed much safer this way. Thanks

eikeno:
@bnerickson  gave an example in his variant of the same feature in #6010 
not needing this myself (just how to do it without adding extra dependancies), maybe @bnerickson  can give more details on this if possible ?

thanks

bnerickson:
> What is the goal of this feature? Installers using local files would be very limited in the amount of users they can reach

@strycore My take on the feature includes some justifications as-well-as references to instances where I've seen the feature requested in reddit/forums in the overview: https://github.com/lutris/lutris/pull/6010

In addition, and particular to my case, it moves Lutris closer to being able to install games without user input via zero-touch provisioning (ZTP).  The eventual goal is to be able to rebuild a PC and install all the games using the installer yml files without:

1. Internet access.
2. A user prompt to install a game
3. Fingers-crossing that the Dropbox/Google URL where the installer is hosted is gone

strycore:
I have several problems with this
* Users have to maintain their own local scripts
* This ignores the fact that we have a file cache that can already serve this purpose
* Fully automated installs are planned but we have no plan to implement them like that so that seems a bit counter productive and not aligned with the project goals.

Feels like a quick and dirty hack

Issue #5550, opened 2024-07-31
Feature request: expose 'gamescope_game_res', etc. options


QmwJlHuSg9pa:
I'm hoping to read options such as 'gamescope_game_res' (working in tandem with dgVoodoo2 resolution-forcing trickery) via a game's prelaunch_command Python script. Is there currently a sane way to expose these values to pre-launch scripts? Else, perhaps env vars could be generated for this purpose?

PR #6021, opened 2025-03-28
expose runner's env to prelaunch command script

eikeno:
By adding `env' dictionnary passed to the process a new key: LUTRIS_RUNNER_SYSTEM_CONFIG
containing json dump of the the dict: `self.runner.system_config'

Implements feature request: #5550

### DIFF
--- a/lutris/game.py
+++ b/lutris/game.py
@@ -569,6 +569,7 @@ class Game:
             logger.warning("Command %s not found", command_array[0])
             return
         env = self.game_runtime_config["env"]
+        env["LUTRIS_RUNNER_SYSTEM_CONFIG"] = json.dumps(self.runner.system_config)
         if wait_for_completion:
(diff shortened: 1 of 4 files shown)


danieljohnson2:
I'm of two minds on this feature. That's a possibly big chunk of JSON you are dumping into the environment. On the other hand, it's just the ``system_config`` and you might want the ``running_config`` or the ``game_config``.

So here's my idea. Can we have additional options right next to the environment list box that would enable ``LUTRIS_RUNNER_SYSTEM_CONFIG`` and two more for ``LUTRIS_RUNNER_RUNNER_CONFIG`` or ``LUTRIS_RUNNER_GAME_CONFIG``?

eikeno:
yes that sounds feasible  without too much complication. I'll check this and update

danieljohnson2:
I like it!

PR #6025, opened 2025-04-01
gamemode: avoid executing gamemoderun when gamemode is not installed

Zebra2711:
from lutris 0.5.8
    * The old versions of gamemode are no longer supported. Make sure you have the one that ships with a `gamemoderun` executable.
    see [`changelog`](https://github.com/lutris/lutris/blob/3a7b62febfe5eaceee2be9a1ae3d059aa055fdd5/debian/changelog#L381C1-L382C56)


Zebra2711:
> This does not seem right. This changes from checking for either old or new versions GameMode to checking for _both_ of them. Who is going to have both?

I'm not sure, but it seems like `self.is_feature_supported("GAMEMODE")` always return true in my system

danieljohnson2:
``self.is_feature_supported("GAMEMODE")`` is checking for the presence of a library that's part of game-mode, I think, so I wouldn't expect the result to change too much day-to-day. But it is a dependency on the implementation of GameMode, and it might be better to be rid of it.

PR #6027, opened 2025-04-02
prefix: avoid to create prefix in file system that not support linux symlink

Zebra2711:
This may avoid creating a wine prefix hang in Lutris. More testing is needed, so I'll leave it as a draft for now


danieljohnson2:
I would suggest that it might be better not to raise an exception when the test for file-system lameness fails; rather log a warning about it and return True, allowing the prefix creation to be attempted.

eikeno:
Why not change:
```
if os.path.islink(prefix):
	prefix = os.readlink(prefix)
```
to:
```
if os.path.exists(prefix) and os.path.islink(prefix):
	prefix = os.readlink(prefix)
```
as a link which exists but is broken is not usable and shouldn't imho go further

When going further, it reaches in_not_allow_file_system() and does:
```
 if not os.path.exists(prefix):
        prefix = os.path.dirname(prefix)
```
a broken symlink will match that condition, and wine files will possibly end up in the parent directory.
I guess this is a problem (for me it would).
Maybe I'm missing a point here, but hope this helps.

Zebra2711:
Fix broken symlink handles here: [lutris/runners/commands/wine.py#L158-L162](https://github.com/Zebra2711/lutris/blob/prefix/lutris/runners/commands/wine.py#L158-L162)

eikeno:
these lines check that the parent of prefix exists, not that prefix itself is a link, or broken link.  See:
```
eikeno@archmsi /tmp$ rm -f WINE  
eikeno@archmsi /tmp$ ln -sf /fake WINE                                        
eikeno@archmsi /tmp$ python          
Python 3.13.2 (main, Feb  5 2025, 08:05:21) [GCC 14.2.1 20250128] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> prefix="/tmp/WINE"
>>> if not os.path.exists(os.path.dirname(prefix)): print("error")
... 
>>> 
```

link is broken bur error not raised.

Zebra2711:
> these lines check that the parent of prefix exists, not that prefix itself is a link, or broken link. See:
> 
> ```
> eikeno@archmsi /tmp$ rm -f WINE  
> eikeno@archmsi /tmp$ ln -sf /fake WINE                                        
> eikeno@archmsi /tmp$ python          
> Python 3.13.2 (main, Feb  5 2025, 08:05:21) [GCC 14.2.1 20250128] on linux
> Type "help", "copyright", "credits" or "license" for more information.
> >>> import os
> >>> prefix="/tmp/WINE"
> >>> if not os.path.exists(os.path.dirname(prefix)): print("error")
> ... 
> >>> 
> ```
> 
> link is broken but error not raised.

where is:
```python
if os.path.islink(prefix):
        prefix = os.readlink(prefix)
```

eikeno:
it wasn't in lines you highlighted ;-)

anyway:
>>> if os.path.islink(prefix):
...     prefix = os.readlink(prefix)

prefix now equals '/fake'  while it could already be spotted as error.

Zebra2711:
For example:
symlink `$HOME/.wine -> /tmp/hdd/.wine`. `.wine` directory did not exist at the time, but I want to create a prefix with that symlink. But yes, that symlink is broken? So I can't create the prefix?

eikeno:
I see what you mean, but that equates to consider a broken link as an intent, which i don't personally think is a good idea:i think it's better to just report that what is currently configured as a prefix path is a broken link, then user can decide by himself what to do.  That's the problem with symlinks, it can quickly become a Pandora's box once "opened" - the fact it's broken could as well be intentional as a genuine configuration error, not mentioning if the target is also a symlink, or other weird stuff that can happen. Just my opinion anyway.

Let's see what maintainers think of it.

Zebra2711:
`df -Th .` seems unable to check the file system type correctly when a filesystem is mounted via FUSE

Zebra2711:
Do I need to remove the `in_not_allow_file_system` function or change something else? It looks clean now

eikeno:
Hey
mere suggestions:

slight rephrasing:
- in_not_allow_file_system() > is_disallowed_filesystem()
- Not found directory -> Directory not found
- Can't create prefix on file system that not support linux symlink -> Can't create prefix on file system that doesn't support linux symlink


L156
if _stat.st_uid != 1000 is not safe, nothing says user has UID 1000 (even if common), better use os.getuid() or something like that 
or
 imho even better would be try creation in all cases, and check if PermissionError is raised.

Reasoning: System could be multiusers, and intentionally sharing resource (like an external game drive or mountpoint) at group level.  If user is in a group that grants him necessary permissions (via std unix rights or ACL), maybe let it do it as long no exception is raised.

hope this helps

Twig6943:
Bump as this still happens and its been a year since last activity on this PR (sorry for thr notification)

danieljohnson2:
I left this one to the boss because I found the write-up confusing. Is this a new feature, or does it address a bug? Maybe both?

You should add "Resolves #<issue number>" in your PR description at the top to point me at the bug this fixes. Or you can try to just explain it.

The actual code looks okay. I would ask you not to raise Exception; raise RuntimeError. Some of the message have grammar issues, but I wouldn't refuse the PR for that. I'd merge it and fix them myself.

strycore:
The patch looks pretty harmless. I very much doubt that it will improve the speed of anything, but also it will not hurt to have it

Issue #5988, opened 2025-03-14
vdf parser stops at escaped double quote


jn64:
### Bug description

My Steam username starts with a double quote, let's say `"foo`

Steam stores it as `\"foo`

```
$ grep PersonaName $(locate loginusers.vdf)
		"PersonaName"		"\"foo"
```

Lutris shows my account as `\` in _Preferences > Accounts > Steam accounts_.

edit: wrong issue title, not sanitisation issue but just vdf parser stopping on `"`

### How to Reproduce

1. Set Steam username to a string containing a double quote `"`

### Expected behavior

Show correct Steam username

### Log output

```shell
NA
```

### System Information

```shell
Lutris Version: 0.5.19
```

strycore:
Fixed

PR #6028, opened 2025-04-03
fix : handle escaped double quote in steam account name

eikeno:
Simply remove the escaped double quotes if found in AccountName or PersonaName.                                                                                      

Trying to maintain the double quote in final display would need much  more modification in different places, for very little added value.
At least the name is displayed with this changed, instead of '\' as  before.                                                                                           

fixes issue: #5988


danieljohnson2:
Why are you looking for 'AccountName'? Lutris only looks for 'PersonaName' as far as I can tell.

PR #6029, opened 2025-04-03
raise error if wine prefix is a broken symlink or unusable

eikeno:
currently, in that case, this is not detected and regular startup is attempted; causing timeout message and lack of feedback via GUI.

This intentionally doesn't try to perform checks regarding the link target itself - could be subject to separate PR if really needed, but I think just the basic check done here would cover already enough situationx where it would be actually useful.

For example in my case I use shared wine prefix, and it sometimes happen the link is broken, because i created it wrongly, or external device it points to is not mounted - and notice that only due to abnormal startup time and message in console.

With this, the error is reported via GUI with no wait time nor risky attempts to start the game anyway.

### DIFF
--- a/lutris/exceptions.py
+++ b/lutris/exceptions.py
@@ -37,6 +37,16 @@ class DirectoryNotFoundError(MisconfigurationError):
         self.directory = directory
 
 
+class SymlinkNotUsableError(MisconfigurationError):
+    """Raise this error if a symlink that is required is not usable."""
+
+    def __init__(self, message=None, link=None, *args, **kwarg):
+        if not message and link:
+            message = message or _("The link {} could not be used.").format(link)
+        super().__init__(message, *args, **kwarg)
+        self.link = link
+
+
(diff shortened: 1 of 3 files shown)


Zebra2711:
I think it was fixed earlier in my merge request #6027

eikeno:
mmm actually not trying to catch the same problem as in your PR, but you got me thinking  and i moved the check to prelaunch, which makes much more sense, and avoids collision with your PR (i also tested merging both and it's ok).  

with  [3b58c07] now getting the correct (at least from what i expected) error (console + gui):
ERROR    2025-04-04 22:15:10,017 [jobs.target:32]:Error while completing task <bound method wine.prelaunch of <lutris.runners.wine.wine object at 0x7bd55637a410>>: <class 'lutris.exceptions.SymlinkNotUsableError'> Link /storage/nas/GAMES_5TB8/Games/LUTRIS/PC/F/FlatOut/WINE couldn't be used, please check permissions or broken link.

PR #6031, opened 2025-04-04
use icoextract intead of extract-icon, which is outdated

Zebra2711:
extract-icon is outdated and cannot handle large .exe files.
```
ERROR    2025-04-10 20:30:07,375 [wine.extract_icon:1351]:Unable to extract icon from /mnt/wind/tmp/HoYoPlay/games/Genshin Impact game/GenshinImpact.exe: 'Data length less than expected header length.'
  File "/home/zebra2711/git/lutris/lutris/runners/wine.py", line 1328, in extract_icon
    groups = extractor.get_group_icons()
  File "/home/zebra2711/git/lutris/lutris/util/wine/extract_icon.py", line 92, in get_group_icons
    grp_icon_dir.__unpack__(data)
    ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
  File "/usr/lib/python3.13/site-packages/pefile.py", line 1016, in __unpack__
    raise PEFormatError("Data length less than expected header length.")
pefile.PEFormatError: 'Data length less than expected header length.'
```
refs: https://github.com/jlu5/icoextract.git

### DIFF
--- a/lutris/runners/wine.py
+++ b/lutris/runners/wine.py
@@ -46,7 +46,7 @@ from lutris.util.wine.d3d_extras import D3DExtrasManager
 from lutris.util.wine.dgvoodoo2 import dgvoodoo2Manager
 from lutris.util.wine.dxvk import REQUIRED_VULKAN_API_VERSION, DXVKManager
 from lutris.util.wine.dxvk_nvapi import DXVKNVAPIManager
-from lutris.util.wine.extract_icon import PEFILE_AVAILABLE, ExtractIcon
+from lutris.util.wine.extract_icon import PEFILE_AVAILABLE, IconExtractor
 from lutris.util.wine.prefix import DEFAULT_DLL_OVERRIDES, WinePrefixManager, find_prefix
 from lutris.util.wine.vkd3d import VKD3DManager
 from lutris.util.wine.wine import (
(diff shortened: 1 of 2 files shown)


Zebra2711:
@danieljohnson2 can u have a closer look at this?

lutris/runners/wine.py
```diff
@@ -1324,29 +1324,15 @@ def extract_icon(self, game_slug):
...
-                elif icons[i].size == wantedsize:
-                    icons[i].save(pathtoicon)
-                    return True
-
-            if biggesticon >= 0:
-                resized = icons[biggesticon].resize(wantedsize)
-                resized.save(pathtoicon)
-                return True
-            return False
+            if not icon.size == wantedsize:
```
danieljohnson2, line 1332: The way I read this, you've changed it so that instead of searching for the right size (that loop you've removed) you just take the first .ico in the .exe and resize that.

That seems off. We're going to export as a .png, so the resulting file won't contain multiple images. Shouldn't we still be picking the best one?

Zebra2711: As far as I know, .exe files contain .ico files of the same size. The old code seems incorrect. I also looked into KDE's icoutils code, which displays thumbnails for .exe, .dll, and similar files in KDE dolphin. It also takes the first icon it gets from these files. Mostly, .exe and .dll files only have one icon file. If they have more than that, I don't think they store multiple different icons in different .ico files with different sizes

danieljohnson2: An .exe contains a bunch of embedded .ico files, but each .ico contains a bunch of images of different sizes. The idea is that you select the right size for what you need and you get a 'pixel perfect' image.

Typically there might be a 16x16 size that is actually drawn differently to look clearer at that small size. You don't want to grab that one and scale it up to 256x256 or something.

The old code took ``group[0]`` and scanned through the images inside. That's the 0th icon in the file, but it tries to pick the best image to make a PNG of.

PR #6033, opened 2025-04-04
skip load env have empty key

Zebra2711:
fixes lutris#6032


danieljohnson2:
I'll take it, but I think I need to add a guard in case "env" is just not in ``system_config``.

PR #6039, opened 2025-04-08
Exherbo distribution support

beholders-eye:
This adapts current implementation on detecting i{3,6}86 multilib dependencies but for Exherbo, I hope it's not too disruptive.


lutris/util/linux.py
```diff
@@ -432,7 +444,7 @@ def new_from_ldconfig(cls, ldconfig_line):
     @property
     def arch(self):
         """Return the architecture for a shared library"""
-        detected_arch = ["x86-64", "x32"]
+        detected_arch = ["x86-64", "x32", "libc6"]
```
beholders-eye, line 447: In Exherbo we always cross compile to the "emulated arch target", so it becomes "libc6".

danieljohnson2:
I am *highly* skeptical that @strycore will accept this one. It's very very "bodge Lutris to agree with our oddball distro", and I don't think we like that.

beholders-eye:
> I am _highly_ skeptical that @strycore will accept this one. It's very very "bodge Lutris to agree with our oddball distro", and I don't think we like that.

Thanks for your words, let's see, a no I already have. 😄

danieljohnson2:
Guess I was wrong.

PR #6043, opened 2025-04-10
add Azahar runner

### LINKED ISSUE #5854: [Feature Request] add Azahar runner
Currently The only runner of 3DS titles seems to be citra, but it has been discontinued for quite a while, it would be nice if Azahar was added as a runner

eikeno:
fixes: #5854


danieljohnson2:
Why does this have both a .py and a .json version of the runner? We should only need one, surely?

eikeno:
fixed. I proceeded by comparing a lot of other runners and files and did not spot the double use. 
Thanks !

TheRedCyclops:
Could you also use the flatpak when it's installed? the id is: org.azahar_emu.Azahar

eikeno:
added. Thanks for the heads up !

TheRedCyclops:
Is there the possibility of importing the preexisting games from Azahar?

eikeno:
if you mean that games already imported in Azahar itself will be listed as regular entries for the Azahar runner in Lutris, no.

Now, maybe you can try the top left "+" button and import ROMs, but I have no idea if this will work as expected 🤷‍♂️

As another option, In the time when i used Lutris for retro stuff (moved to ES-DE for that now), i came up with mass importers shell scripts. If interested to have a look it's on   https://github.com/eikeno/lutris-scripts/

Be sure that bin/string_to_slug.sh is in your PATH as it is needed by bin/importers/add_3ds_game2lutris.sh

Example use:
```
export PATH="$PATH:./bin"
find /my/gamedir -type f -name "*.3ds" -exec ./bin/importers/add_3ds_game2lutris.sh "{}" \;
```

haven't used it in a while, might need tweaking

PR #6058, opened 2025-04-18
Sort games by last modified date

erri120:
This PR adds a new field to the `games` DB table called `modified_at`. A new migration initializes this column with values taken from the `installed_at` column. Whenever the game gets modified, `modified_at` will be set to the current time. Finally, the values are exposed in the UI via a new column in the list view and a new sorting option in the grid view.


strycore:
from a usability standpoint, I don't really see why this is needed / wanted.

PR #6068, opened 2025-04-21
Add Norwegian bokmål translation

Martini-141:


danieljohnson2:
Can't argue with translations!

PR #6070, opened 2025-04-23
feat: Rename game config files when slug changes

Zebra2711:
* Renames game config files when a game's slug changes
* Use the new game name's slug when duplicating game configs


lutris/config.py
```diff
@@ -36,6 +36,18 @@ def duplicate_game_config(game_slug: str, source_config_id: str):
     return new_config_id
 
 
+def rename_config(old_config_id: str, new_slug: str):
```
danieljohnson2, line 39: I think you should specify the return type here.

~~But perhaps it would be better if you did not return ``None``, but instead ``old_config_id``, so the return type can be ``str``- and the one place that calls this can unconditionally assigned it, without bringing a walrus into it.~~ Never mind, this won't work- you still want the new config id, not the old, and you do not have it here. ``-> Optional[str]`` is correct here.

lutris/config.py
```diff
@@ -36,6 +36,18 @@ def duplicate_game_config(game_slug: str, source_config_id: str):
...
+    old_slug, timestamp = old_config_id.rsplit("-", 1)
+    if old_slug == new_slug:
+        return None
+    new_config_id = f"{new_slug}-{timestamp}"
+    config_games_dir = os.path.join(settings.CONFIG_DIR, "games")
```
danieljohnson2, line 44: Why not use ``GAME_CONFIG_DIR`` here?

danieljohnson2 (approved):
I can see some minor improvements, but this looks pretty good to me.

danieljohnson2 (approved):

PR #6071, opened 2025-04-23
Make it possible to build on bookworm and ubuntu

majick:
* PyGObject > 3.50.0 has a dependency on girepository-2.0 But it's only available on unstable OS versions. Pinned to 3.50.0 instead.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -62,7 +62,7 @@ setup(
         "evdev",
         "lxml",
         "pillow",
-        "PyGObject",
+        "PyGObject == 3.50.0",
         "pypresence",
         "PyYAML",
         "requests",


danieljohnson2:
Does this mean those distros are shipping PyGObject > 3.5, but not it's updated dependency? Aren't those distros broken, or do I not understand this?

majick:
In system python? Who knows what's being shipped. There's never any accounting for system python, it's not a general purpose runtime. Maybe that's where your confusion is coming in, that lutris is being installed from source against system python?

libgirepository — the build dependency — is 1.0 on shipping OSes, so any attempt to bring PyGObject latest (such as via the `setup.py`) into a clean runtime python (such as a `pyenv`) will fail. PyGObject > 3.50.0 cannot be installed without hand-building the library, which seems excessive.

Compared to just picking up a set of bindings that can actually build on the OS, that is. This is the only lutris dependency that will fail to build on current stable.

danieljohnson2:
You seem to know what's being shipped, but it sounds like whatever it is, it is broken, and this is a workaround.

But it seems strange to ship PyGObject 3.5 without the right libgirepository. I wonder how this came to pass.

majick:
Thanks for merging.

Just to be clear: I'm utterly uninterested in what Python bindings are shipped by the OS vendor because lutris git is not a vendor-provided piece of software; therefore whatever PyGObject Debian ships is irrelevant. System python isn't something anyone should be using for third party software. Especially software being installed from `pip install -e`

That'll go into whatever alternative a given user is running, either through `pyenv` or some other not-running-system-python means of the user's choice. At which point that's where the workaround comes in.

Only the native binary libraries are relevant here, and on Debian that's 1.0.

danieljohnson2:
Oh, I won't merge this. I don't understand what you are telling me well enough to judge it. I'll leave it to @strycore.

majick:
Thanks! If it helps @strycore here's even more detail:

```
% [clone the repo]
% pyenv virtualenv 3.12.2 lutris   # not system python because system python is cancer
% pyenv activate lutris
% pip install -e .  # into a stock Python venv as specified above
[... many lutris deps get installed ...]
[PyGObject latest (as specified in setup) downloads but fails to build]
[Because it's a binding to libgirepository-2.0, and that's not shipped.]

% ls -l /usr/lib/x86_64-linux-gnu/libgirepository*
lrwxrwxrwx 1 root root     24 Jan 19  2023 /usr/lib/x86_64-linux-gnu/libgirepository-1.0.so -> libgirepository-1.0.so.1
lrwxrwxrwx 1 root root     28 Jan 19  2023 /usr/lib/x86_64-linux-gnu/libgirepository-1.0.so.1 -> libgirepository-1.0.so.1.0.0
-rw-r--r-- 1 root root 223264 Jan 19  2023 /usr/lib/x86_64-linux-gnu/libgirepository-1.0.so.1.0.0

[We don't have 2.0 on Bullseye]

% sed -i -e 's|PyGObject|PyGObject==3.50.0|' setup.py
% pip install -e .
[ pip does its thing again...]
[ success, because 3.50.0 binds to libgirepository-1.0.so.1 , \
 which we have, rather than libgirepository-2.0 that current \
 PyGObject does. That we don't have. Because it's not shipped.]
```

Sometimes a transcript makes more sense than my prose description!

PR #6075, opened 2025-04-25
add export .yml script

Zebra2711:
Maybe useful, ehh. Draft


danieljohnson2:
My OCD demands that the 'S' in 'Script' in that menu command be in lowercase.

But also, this looks like it exports parts of the game's configuration as a .yml file. How is this a script? What does it do?

Zebra2711:
> My OCD demands that the 'S' in 'Script' in that menu command be in lowercase.
> 
> 
> 
> But also, this looks like it exports parts of the game's configuration as a .yml file. How is this a script? What does it do?

That yml file can be exported and shared, i cheated that yml config to make it installable by importing it into Lutris using the "Add game" option and selecting a local YAML script. Also, it should probably be named something like export "configuration" instead of "script". I might be approaching this in a wrong way, but it's actually useful.

dpfister22:
I think calling this a "script" is fine since the yaml files are called "scripts" in other places. For example, when installing a game there is an option that says "Install from a local install script" and "Run a YAML install script".

danieljohnson2:
Well, if these files can be imported by "Run a YAML install script" then I guess it is legit.

Those things are scripts because we can run them like a program, not because they are YAML.

PR #6084, opened 2025-04-27
Add option to change the Wine graphics driver

ventureoo:
Lutris already has option to select Wine audio driver, so in my opinion due to maturation of the winewayland driver, which has been available since Wine 9.0, it makes sense to add option to select Wine graphics backend.

P. S. I'm not sure if paths hardcoding for winewayland is the right solution, so suggestions for an alternative detection of winewayland in selected Wine/Proton version are welcome.


danieljohnson2:
I like it, but we usually use lowercase for option keys, so maybe ``"Graphics"`` should be ``"graphics"``. Or perhaps something more specific than that? Maybe ``"graphics_driver"``?

danieljohnson2:
I didn't appreciate that. Had I understood that, I'd have kept my trap shut. But I guess it can go either way.

ventureoo:
> I didn't appreciate that. Had I understood that, I'd have kept my trap shut. But I guess it can go either way.

No need to worry. I'll just return everything to the way it was.

lutris/runners/wine.py
```diff
@@ -411,6 +431,20 @@ class wine(Runner):
...
         },
+        {
+            "option": "proton_hdr",
+            "section": _("Graphics"),
+            "label": _("Enable HDR (Experimental)"),
+            "type": "bool",
+            "default": False,
+            "advanced": True,
+            "visible": _get_wine_wayland_warning,
+            "conditional_on": "Graphics",
```
danieljohnson2, line 443: This won't work- "conditional_on" works for boolean settings, but "Graphics" is not a boolean.

We need to write a fancy "conditional" here- it needs to check if Wayland is selected *and* that the Wine version is Proton.

lutris/runners/wine.py
```diff
@@ -554,6 +588,24 @@ class wine(Runner):
...
+                (_("Auto"), "auto"),
+                ("Wayland", "wayland"),
+                ("X11", "x11"),
+            ],
+            "default": "auto",
+            "warning": _get_wine_wayland_warning,
+            "help": _(
+                "Which graphics backend to use.\n"
+                "By default, Wine automatically picks the right one "
+                "for your system."
```
danieljohnson2, line 607: This help comment may be a little optimistic. Does Wine auto-detect here anything really?

lutris/runners/wine.py
```diff
@@ -594,6 +646,7 @@ class wine(Runner):
     reg_prefix = "HKEY_CURRENT_USER/Software/Wine"
     reg_keys = {
         "Audio": r"%s/Drivers" % reg_prefix,
+        "Graphics": r"%s/Drivers" % reg_prefix,
```
danieljohnson2, line 650: Can we avoid setting this when using Proton? Not obvious how to do this.

lutris/runners/wine.py
```diff
@@ -1178,6 +1236,11 @@ def get_env(self, os_env=False, disable_runtime=False):
         if not self.runner_config.get("dxvk") or not LINUX_SYSTEM.is_vulkan_supported():
             env["PROTON_USE_WINED3D"] = "1"
 
+        if self.runner_config.get("Graphics") == "wayland" \
```
danieljohnson2: We're not setting ``PROTON_ENABLE_WAYLAND``- we should be. That's how to tell Proton to use the Wayland driver - not a registry key.

lutris/util/wine/wine.py
```diff
@@ -96,6 +96,22 @@ def is_gstreamer_build(wine_path: str) -> bool:
     return system.path_exists(os.path.join(base_path, "lib64/gstreamer-1.0"))
 
 
+def is_winewayland_available(runner_version: str) -> bool:
```
danieljohnson2, line 99: Hmm. No sign of my fancy Proton version number check.

Will this check for Wayland .drv files work? It should be True for GE-Proton10-1 and later but false for GE-Proton9.

ventureoo: Yeah, I think it should be fine, because regular builds of Proton don't have winewayland files, and older Proton-GE versions don't either AFAIK.

lutris/runners/wine.py
ventureoo: Oops, yeah, totally forgot about that.

lutris/util/wine/wine.py
danieljohnson2: I checked Proton-Tkg-2074 and it does have those files.

lutris/runners/wine.py
ventureoo: It depends on the configuration. If DISPLAY is reset then winewayland is used by default anyway, similarly if the X11 backend fails to initialize (probably in environments without Xwayland).

danieljohnson2: I guess that's fair enough then.

lutris/runners/wine.py
danieljohnson2: I think I see the place- about line 1000 of this file, it already special cases ``Desktop`` and ``WineDesktop`` to be skipped sometimes. You could also do ``Graphics`` there.

lutris/runners/wine.py
ventureoo: I guess we can just drop the `conditional_on`, because `visible` should be enough already?

lutris/util/wine/wine.py
ventureoo: > I checked Proton-Tkg-2074 and it does have those files.

Yes, Proton-TKG also has winewayland. I think they were one of the first to attempt to use it, although I'm not sure if they use `PROTON_ENABLE_WAYLAND` variable.

lutris/runners/wine.py
ventureoo: Fixed, thanks

lutris/runners/wine.py
ventureoo: I saw this too and made the same solution, hopefully it will work properly.

lutris/util/wine/wine.py
ventureoo: Hmm, looks like Proton-TKG doesn't have `PROTON_ENABLE_WAYLAND` variable :\

Maybe we should also add a `DISPLAY=` variable as a safety precaution, but this needs to be tested.

danieljohnson2: If they do not, how do you activate the feature? With the registry key?

We need to figure out which method to use for which Wine build, somehow.

ventureoo: > If they do not, how do you activate the feature? With the registry key?
> 
> We need to figure out which method to use for which Wine build, somehow.

Nope, from what I read in this issue https://github.com/Frogging-Family/wine-tkg-git/issues/1425, the only way to activate it is to unset `DISPLAY` environment variable, as it works for the current upstream as well. Not sure if this will work for Proton-TKG versions based on Proton 9.

lutris/runners/wine.py
danieljohnson2: Well, make sure "visible" is correct- now that you mention it, it looks off. A warning were a boolean is expected? I think that'll do the opposite of what you'd want- visible if there is a warning, invisible if not.

lutris/util/wine/wine.py
danieljohnson2: This is getting complicated! We probably can't handle every custom Proton build; I'd rather handle what we ship (and that's GE-Proton and wine-ge).

lutris/runners/wine.py
ventureoo: > Well, make sure "visible" is correct- now that you mention it, it looks off. A warning were a boolean is expected? I think that'll do the opposite of what you'd want- visible if there is a warning, invisible if not.

Sorry, yes in the previous version I used same function as to get warning, but current version should be better I think since I returned your `_is_proton_wayland_available` function from https://github.com/lutris/lutris/pull/6127  in a simplified form for that.

lutris/util/wine/wine.py
danieljohnson2: And ``wine-ge`` does not support Wayland at all (it is too old) so hey presto now my PR is the way to go!

Maybe we should support vanilla Wine builds with Wayland too, though. We don't ship them, but we'll use them if installed.

danieljohnson2:
I've put some more review comments in; I don't think we're there yet.

danieljohnson2 (changes requested):
I guess I need to do another "review" in GitHub? We're not there yet- I think the registry setting and PROTON_ENABLE_WAYLAND thing should be mutually exclusive, and that the Proton HDR setting should be enabled only when Wayland is first selected, not just when it is available.

lutris/runners/wine.py
```diff
@@ -411,6 +440,19 @@ class wine(Runner):
...
             "conditional_on": "dgvoodoo2",
         },
+        {
+            "option": "proton_hdr",
+            "section": _("Graphics"),
+            "label": _("Enable HDR (Experimental)"),
+            "type": "bool",
+            "default": False,
+            "advanced": True,
+            "visible": _is_proton_wayland_available,
```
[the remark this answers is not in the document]
ventureoo: Indeed, I added a new condition and renamed function to `_is_proton_hdr_available`.

lutris/runners/wine.py
```diff
@@ -1004,6 +1065,16 @@ def set_regedit_keys(self):
...
             elif key in self.runner_config:
+                if (
+                    value
+                    and key == "Graphics"
+                    and value == "wayland"
+                    and not proton.is_proton_path(self.get_executable())
+                ):
+                    if not is_winewayland_available(self.read_version_from_config()):
+                        logger.warning("Your Wine version does not support winewayland graphics driver")
+                        continue
```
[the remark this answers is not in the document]
ventureoo: Yeah, I must have been a little bit last night to have missed that. Fixed, thanks!

danieljohnson2 (approved):
OK, I think this is good enough. I will close my PR.

lutris/util/wine/wine.py
ventureoo: Seems to be Proton-TKG just added support for the `PROTON_ENABLE_WAYLAND` variable, so I think this is solved now, probably for all known Proton builds. 

https://github.com/Frogging-Family/wine-tkg-git/commit/f4ee327a19cb06c9a7c7502c84a7e62b65e0fc14

PR #6088, opened 2025-04-29
Update `StartupWMClass` to the desktop file

fiftydinar:
According to the information I got, `StartupWMClass` is called `net.lutris.Lutris` now.

This should resolve the edge-cases where icon is not shown in Gnome's dash & maybe in some other desktop environments.

### DIFF
--- a/share/applications/net.lutris.Lutris.desktop
+++ b/share/applications/net.lutris.Lutris.desktop
@@ -1,6 +1,6 @@
 [Desktop Entry]
 Name=Lutris
-StartupWMClass=Lutris
+StartupWMClass=net.lutris.Lutris
 Comment=Video Game Preservation Platform
 Categories=Game;PackageManager;
 Keywords=gaming;wine;emulator;


Zebra2711:
> This should resolve the edge-cases where icon is not shown in Gnome's dash & maybe in some other desktop environments.

change it just do not thing and you will make issue when lutris run on x11? 
When Lutris is run with `GDK_BACKEND=x11`, its WM_CLASS is Lutris; when run with `GDK=wayland`, its WM_CLASS is net.lutris.Lutris. This change fixes the issue with GNOME's dash on Wayland but creates an issue when running on X11?

fiftydinar:
> When Lutris is run with `GDK_BACKEND=x11`, its WM_CLASS is Lutris; when run with `GDK=wayland`, its WM_CLASS is net.lutris.Lutris. This change fixes the issue with GNOME's dash on Wayland but creates an issue when running on X11?

I can't confirm this, at least in Xwayland, `GDK_BACKEND=x11` also sets wmclass as `net.lutris.Lutris`
So in both cases, it's `net.lutris.Lutris`

I just didn't test X11-only.

fiftydinar:
> > When Lutris is run with `GDK_BACKEND=x11`, its WM_CLASS is Lutris; when run with `GDK=wayland`, its WM_CLASS is net.lutris.Lutris. This change fixes the issue with GNOME's dash on Wayland but creates an issue when running on X11?
> 
> I can't confirm this, at least in Xwayland, `GDK_BACKEND=x11` also sets wmclass as `net.lutris.Lutris` So in both cases, it's `net.lutris.Lutris`
> 
> I just didn't test X11-only.

You are actually right, I tested it wrongly, so I'll close this, as this can't be dynamically set.

PR #6089, opened 2025-04-29
fix Lutris icon is not shown in Gnome's dash (wayland) 

Zebra2711:


Zebra2711:
@fiftydinar Could you try this? Should be fixed

danieljohnson2:
Is there a particular issue that this fixes? An issue number?

fiftydinar:
> @fiftydinar Could you try this? Should be fixed

I confirm it's fixed!

> Is there a particular issue that this fixes? An issue number?

https://github.com/lutris/lutris/pull/6088

It solves the edge-cases where compositor can't get the icon from the application, either through the name of the desktop file, id of the application or through parsing `StartupWMClass`. In my case, that reflected with the generic icon in Gnome's dash when using an AppImage. Gnome shell does this differently for flatpak desktop entries & believe, so the issue is not seen there. It's also recommended to set this so other desktop environments can easily parse this too.

`StartupWMClass` is different for X11 & Wayland currently, with desktop file defaulting to `Lutris`, which spawns when X11/Xwayland is used. While for Wayland, that's `net.lutris.Lutris`.

This PR solves this issue by making `StartupWMClass` uniquely named in all cases, `net.lutris.Lutris`.

danieljohnson2:
Okay, I'll merge this when I am back at home.

PR #6102, opened 2025-05-02
fix: add missing assets when creating Steam shortcut

lfont:
Add missing assets for the horizontal grid and icon.

### DIFF
--- a/lutris/util/steam/shortcut.py
+++ b/lutris/util/steam/shortcut.py
@@ -170,17 +170,17 @@ def set_artwork(game):
     shortcut_id = generate_appid(game)
     source_cover = resources.get_cover_path(game.slug)
     source_banner = resources.get_banner_path(game.slug)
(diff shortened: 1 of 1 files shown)


danieljohnson2:
Looks to me like we should take this but it's not clear what this fixes. Is there an issue on our repo about this? Can you tell me what the bug is?

lfont:
Hi,
Sorry, my description probably isn't as clear as I thought.
Without this PR, when you create a Steam shortcut via the context menu (right-click > Create Steam Shortcut), only two of the four (not including the logo for which we do not have a suitable fallback solution, but which is not mandatory) Steam images used to display games are created.

This PR adds the two missing images:
- The horizontal grid: This is displayed after a game has been played for a certain amount of time. The most recently played game is displayed with a larger, horizontal image. Without it, only a gray rectangle is visible.
- The icon: This is the image displayed to the left of the game name in the left pane of the Steam Library. Without it, a gray square is displayed.

If this isn't clear enough, please feel free to contact me and I'll try to add screenshots.

Issue #5544, opened 2024-07-24
Lutris can't access vulkaninfo: "vulkaninfo --summary" fails on Lutris but it works fine on my terminal


mattepm:
### Bug description

I am trying to run a non-Steam game through Lutris with Wine, and there is no way to make it work with my Vulkan drivers.

Looking at the error I get running Lutris in debug mode, it seems like it can't access the **vulkaninfo** command:
`ERROR    2024-07-24 13:09:09,354 [system.read_process_output:188]:['vulkaninfo', '--summary'] command failed: Command '['vulkaninfo', '--summary']' returned non-zero exit status 1.
`

Actually, if I run `vulkaninfo --summary` through another terminal, everything works.

I am 99.9% sure my GPU is compatible with Vulkan.
I read it could be something related to lib32 NVidia libraries being not set up properly, and verifying it that looks all fine too.

I understand there may be compatibility or driver issues, but it sounds really like Lutris can't access the vulkaninfo command.
Is there any way to fix that?

### How to Reproduce

1. Open terminal and type lutris -d
2. Open a game
3. Check the terminal and see the error

### Expected behavior

It should correctly forward the DirectX calls to the Vulkan driver.

### Log output

```shell
INFO     2024-07-24 13:09:09,342 [application.do_command_line:479]:Starting Lutris 0.5.17
ERROR: [Loader Message] Code 0 : vkCreateInstance: Found no drivers!
Cannot create Vulkan instance.
This problem is often caused by a faulty installation of the Vulkan driver or attempting to use a GPU that does not support Vulkan.
ERROR at ./vulkaninfo/vulkaninfo.h:674:vkCreateInstance failed with ERROR_INCOMPATIBLE_DRIVER
ERROR    2024-07-24 13:09:09,354 [system.read_process_output:188]:['vulkaninfo', '--summary'] command failed: Command '['vulkaninfo', '--summary']' returned non-zero exit status 1.
INFO     2024-07-24 13:09:09,354 [startup.run_all_checks:120]: (10de:28a0 1025:159c nvidia) Driver 530.30.02
DEBUG    2024-07-24 13:09:09,459 [path_cache._update_missing_games:114]:Checking for missing games
DEBUG    2024-07-24 13:09:09,478 [api.check_stale_runtime_versions:52]:Modified at Wed 24 Jul 2024 09:25:51 , will update after Wed 24 Jul 2024 15:25:51 
[...]

PR #6125, opened 2025-05-14
Update gpu.py

blackPantherOS:
Error handling resulting from unnecessary imports.  Referer:
https://github.com/lutris/lutris/issues/5544#issuecomment-2878724077

### DIFF
--- a/lutris/util/graphics/gpu.py
+++ b/lutris/util/graphics/gpu.py
@@ -120,11 +120,17 @@ class GPU:
     def get_vulkaninfo(self) -> Dict[str, Dict[str, str]]:
         """Runs vulkaninfo to find the GPU name"""
         subprocess_env = dict(os.environ)
-        subprocess_env["VK_DRIVER_FILES"] = self.icd_files  # Currently supported
-        subprocess_env["VK_ICD_FILENAMES"] = self.icd_files  # Deprecated
-        vulkaninfo_output = system.read_process_output(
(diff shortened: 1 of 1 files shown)


blackPantherOS:
Not clear for me, what is ruff failing because:

```
$ruff check gpu.py 
All checks passed!
```

eikeno:
on https://github.com/lutris/lutris/actions/runs/15013338352/job/42185940806?pr=6125
you can see details after clicking  on "Check Format" item in the list:


Run ruff format . --check
Would reformat: lutris/util/graphics/gpu.py
1 file would be reformatted, 296 files already formatted
Error: Process completed with exit code 1.

you can run `make format` to solve the problem (make sure you're in the correct env/venv if you're using one)
if using a venv, make sure to `make dev` before to install check tools (they need specific versions) via pip in the env.


hope this helps

blackPantherOS:
> Hang on, this PR changes how we invoke `vulkaninfo` so that, if possible we would not use the env-vars `VK_DRIVER_FILES` and `VK_ICD_FILENAMES`.
> 
> I don't think that's right; that means we're not selecting the ICD for the specific GPU we want to probe.
> 
> I don't know what this is meant to do, but I do not like it.

I would clarify that we select the ICD only if it has not been selected yet. Since Vulkan is already initialized in our system, re-specifying the environment variables results in incorrect behavior when running Lutris. See the quoted thread on top.

PR #6127, opened 2025-05-14
Support for PROTON_ENABLE_WAYLAND and PROTON_ENABLE_HDR.

danieljohnson2:
Just some options to enable some new env-vars that have been added in Proton 10.

image

These enable ``PROTON_ENABLE_WAYLAND`` and, if you enable that, you can have ``PROTON_ENABLE_HDR``. They are available only on Wayland. I do not know how to check if HDR support is available, so I just check for Wayland. I suppose HDR is even more experimental here, if it's experimental in gamescope.

These are disabled by default since there are compatibility issues; indeed, one way you can tell that this works is that it breaks the graphics of Doom 2016. They are also considered "Advanced" features.

### DIFF
--- a/lutris/runners/wine.py
+++ b/lutris/runners/wine.py
@@ -424,18 +424,20 @@ class wine(Runner):
             "default": False,
             "advanced": True,
             "visible": _is_proton_wayland_available,
-            "help": _("Enable Proton's support for the Wayland Display Server."),
+            "help": _(
+                "Enable Proton's support for the Wayland Display Server. Some Proton versions don't support this."
+            ),
         },
         {
             "option": "proton_hdr",
             "section": _("Graphics"),
-            "label": _("Enable HDR"),
+            "label": _("Enable HDR (Experimental)"),
             "type": "bool",
             "default": False,
             "advanced": True,
             "visible": _is_proton_wayland_available,
             "conditional_on": "proton_wayland",
-            "help": _("Enable Proton's support for the High Dynamic Range graphics. Requires Wayland."),
+            "help": _("Enable Proton's support for High Dynamic Range graphics. Requires Wayland."),
         },
         {
             "option": "esync",
--- a/lutris/runners/wine.py
+++ b/lutris/runners/wine.py
@@ -35,7 +35,7 @@ from lutris.runners.commands.wine import (  # noqa: F401 pylint: disable=unused-
 )
 from lutris.runners.runner import Runner
 from lutris.util import system
-from lutris.util.display import DISPLAY_MANAGER, get_default_dpi
+from lutris.util.display import DISPLAY_MANAGER, get_default_dpi, is_display_x11
 from lutris.util.graphics import drivers, vkquery
 from lutris.util.linux import LINUX_SYSTEM
 from lutris.util.log import logger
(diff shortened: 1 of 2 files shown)


eikeno:
tested ok (wayland + hdr) with plague a tale requiem. rtx 3070, hyprland, arch

Play-InTheWay:
Though, it's to keep in mind that [vanilla Proton may not support it soon](https://github.com/ValveSoftware/Proton/issues/4638#issuecomment-2845811100) but in other hand other runners of Wine or Proton's forks actually do. 

Ideally to evade confuse the users it should be limited to the versions that works, but I don't know if it's something feasible.

danieljohnson2:
I don't know how to tell what versions it'll work in, but in the one we actually ship, it should work. More or less.

Play-InTheWay:
Of course, but I'm thinking about people that think that by enabling it will work in vanilla Proton or even in older versions. Though, the worst could happen, are dealing with multiple bug reports. 
An alternative is to include some form of advice in the settings.

eikeno:
It's in advanced settings, which should prevent at least one portion of problems i guess.

Given the overall current status of HDR on Linux currently maybe just adding "(Experimental)"  as is done for Gamescope's HDR support in System Option could be useful and sufficient for now so people know what they're getting into ?

Would be really a pity not to add this anyway, as we don't use Vanilla proton as a default, and this adds a welcome feature (meaning wayland as well as HDR here)  addition without systematically having to use Gamescope.

IMHO, if there is caution to be had, it's not in the adoption of an optional feature like this, but rather  in related issue trackers / forums / etc  to simply establish (pinned comments / policy) what should be reported or not, and how, to avoid floods of useless/similar reports.  Providing such option, while giving a word of caution can only help spot real problems and moving forward better support over time.

There are reasons Vanilla Proton has to consider that can explain why some option are removed for now (or not enabled), due to the context (millions of accounts, having to deal with maximum diversity of OS/hardware  out there, in addition to their own steamdeck / steam OS etc.), cost of tech support  and so on. None of this apply to Lutris, and it's a chance for it to be on early adopters side, as long as Defaults settings are safe and conservative (which they currently are).

Just a few random morning thoughts.

Zebra2711:
We should check the proton version to grey it out if the version is older than 1747214205?

danieljohnson2:
I'll go ahead and add "(Experimental)" to the HDR options, like gamescope has, and add a sentence about limited support for Wayland.

I do not know about checking the proton version- I believe we need not only Proton 10 but *GE-Proton* with @GloriousEggroll's special sauce at this point. He's put in some extra patches for this, I believe.

danieljohnson2:
I've gone ahead and added a version number check (>=1747214205) for this feature.

The actual version text in the file is ``1747214205 GE-Proton10-1`` though. I could check for GE-Proton specifically, but I wonder if we should be so picky. Other Proton wranglers might support this stuff too.

ventureoo:
I think this should already be addressed to https://github.com/lutris/lutris/pull/6084?

danieljohnson2:
Hmm. That's a good point, I'd forget about #6084.

I don't think it makes a lot of sense to have two options, one for Proton and one for old-school Wine. But we should use the ``PROTON_...`` env-var to control Proton, and we'd like to have the HDR option.

I wonder if there's some way to combine these PRs?

danieljohnson2:
I guess I could fork your repo, re-implement my changes on top of yours, and submit a grand unified PR from there. I'd be kind of taking over this feature. What do you think?

ventureoo:
> I guess I could fork your repo, re-implement my changes on top of yours, and submit a grand unified PR from there. I'd be kind of taking over this feature. What do you think?

I have nothing against it and can add your changes on top of mine myself, so I'll leave it up to you what you're more comfortable with.

danieljohnson2:
I'll give it a try.

ventureoo:
> I'll give it a try.

I just tried to combine your changes with mine, please see https://github.com/lutris/lutris/pull/6084

danieljohnson2:
I'll check it out.

danieljohnson2:
I'm closing this PR in favor of #6084.

PR #6136, opened 2025-05-18
Fixed two minor ScummCM issues related with argument handling

neuromancer:
This small commit will fix two issues related with argument handling when using the ScummVM runner:

* If the `game_id` is not set, it will avoid adding a `None` in the command line. This is useful if you want to set the ScummVM runner to use `--auto-detect` in their arguments and you don't want to use the `game_id` directly.
* It will correctly set the arguments key name to `args` when the games are imported from a GOG installation. Previously, when using `arguments` as a key in the dictionary returned by `_get_scummvm_arguments` did not had any effect.


danieljohnson2:
Thanks!

PR #6149, opened 2025-05-26
Adds service that joins all other logged in services together.

Gonzako:
This is a draft PR to get some eyes on my code. 
Code right now calls the correct services "install_by_id"
but the service only feeds the steams media so all  other services and actually installed games don't highlight.
How'd I got about retrieving the media model for each service and feeding it to its views?


danieljohnson2:
I don't know what's upsetting you. I marked it a draft because it was clearly not ready to be merged, not as an endorsement.

Gonzako:
> this is not a draft, it is a flawed patch that ignores the direction lutris is taking. The only way this is getting merged is if it rewritten from scratch with with a **completely different logic**

Oh, it's perfectly fine. I've no mind giving this a re-spin. How'd I got about pushing this concept forward? I tried using both the client sync option and the website steam sync and neither worked for me using the latest releases.

PR #6153, opened 2025-05-28
winekill: Force-kill processes if waiting too long

Zebra2711:


danieljohnson2:
This does not look right- you have ``winekill`` call ``run_winekill`` which calls ``winekill`` again.

Except you don't because the loop exits after 20 cycles, and your code only kicks in after 30.

But the thing is, Lutris already has logic to force-kill naughty games that don't exit when we kill them. Look at ``Game.force_stop_game`` and ``Game.force_kill_delayed``. Isn't that working?

Zebra2711:
> This does not look right- you have `winekill` call `run_winekill` which calls `winekill` again.

yeah, I was wrong, lol. But we can also kill the current `wineserver` with `wineboot -k`, it's much faster

Issue #6155, opened 2025-05-29
working_dir for installer scripts


rienafairefr:
Hi to all,

I wanted to setup a few "local install scripts" so that I can offer reproducible "easy install scripts" for non-tech-literate people, on a few linux machines.

I already can mount a remote folder (from my NAS) containing installers, I can write the installation scripts yaml, mostly, the only nag I found is that when I run "lutris install -i script.yml", there is apparently no way to specify to a "wineexec" step an executable that is _right there_ in the same folder as the installer script yaml. I have to specify the full absolute path on the mounted folder, which is not super "portable".
Maybe there is a way that I didn't see, maybe I completely misunderstood the point of these local install scripts. 

A workaroung would probably by a "copy-then-install", where the setup files are copied to the game prefix path, then executed there. But I couldn't make it work either (the src is also not relative to the install script). 

I'm tempted to send a pull request for a variable a bit like $GAMEDIR that could provide that functionality (like, idk, $INSTALLSCRIPTDIR), but before that I'm interested in your point of views on that working_dir (it is actually by default set to the wine prefix, so has no notion of that install script directory)

danieljohnson2:
Your proposed ``$INSTALLSCRIPTDIR`` has really nothing to do with the ``working_dir``. The ``working_dir`` is the current directory we set before starting the game when you play it; it is not he current directory of Lutris itself; because we change that on the fly, we can't trust it.

I'd say go for the PR.

rienafairefr:
Right, thanks, there are two places with a working_dir, one for the game, and one for the winexec steps, here's an example/prototype
```
game_slug: nice-game
slug: nice-game
name: Nice Game
runner: wine
version: 1.0.0
script:
  game:
    arch: win32
    exe: $GAMEDIR/drive_c/Games/Game/Launcher.exe
    prefix: $GAMEDIR
    working_dir: $GAMEDIR/drive_c/Games/Game # I'm not talking about this one 
  installer:
    - task:
        name: create_prefix
        prefix: $GAMEDIR   
    - task:
        executable: setup.exe
        working_dir: /mnt/installers/nice-game # I'm talking about this one
        name: wineexec
        prefix: $GAMEDIR
```
If the /mnt/installers/nice-game could be `$INSTALLSCRIPTDIR` that'd be a "portable YML installer"

danieljohnson2:
I guess it makes sense when you illustrate it like that.

Zebra2711:
[...]

PR #6161, opened 2025-05-31
INSTALLERDIR variable

rienafairefr:
like discussed in issue #6155 : I think this variable would be useful to make installer scripts 'more portable', it makes Lutris aware of where the script is, so it can use it.
This is a prototype usage : 
```
game_slug: nice-game
slug: nice-game
name: Nice Game
runner: wine
version: 1.0.0
script:
  game:
    arch: win32
    exe: $GAMEDIR/drive_c/Games/Game/Launcher.exe
    prefix: $GAMEDIR
  installer:
    - task:
        name: create_prefix
        prefix: $GAMEDIR   
    - task:
        executable: setup.exe
        working_dir: $INSTALLERDIR
        name: wineexec
        prefix: $GAMEDIR
```

### DIFF
--- a/lutris/installer/__init__.py
+++ b/lutris/installer/__init__.py
@@ -33,6 +33,12 @@ def read_script(filename):
 
 def get_installers(game_slug=None, installer_file=None, revision=None):
     # check if installer is local or online
+
     if system.path_exists(installer_file):
-        return [normalize_installer(i) for i in read_script(installer_file)]
-    return get_game_installers(game_slug=game_slug, revision=revision)
+        installers = [normalize_installer(i) for i in read_script(installer_file)]
+        for installer in installers:
+            installer["_lutris_installer_file"] = installer_file
+    else:
+        installers = get_game_installers(game_slug=game_slug, revision=revision)
+
+    return installers
--- a/lutris/installer/installer.py
+++ b/lutris/installer/installer.py
@@ -1,6 +1,7 @@
 """Lutris installer class"""
 
 import json
+import os
 from gettext import gettext as _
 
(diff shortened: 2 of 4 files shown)


danieljohnson2:
Boy this seems like a lot of change to add one variable.

And ``script.script["script"]`` rubs me the wrong way every time. Think we have enough script? We might need more script!

rienafairefr:
I think capturing the installer_file in get_installers and then using it afterwards was the cleanest way (the other ways I thought would have even more changes), but maybe there is is a better way I didn't see.

I agree `script.script["script"]` is really really repetitive ^^ I just pushed a rename to `data` for that element, so it becomes `installer.data["script"]` instead, a bit less vomit-inducing

danieljohnson2:
I see what you are doing here: wrapping the dict we get from a YML or JSON file in a very basic data-class.

I'm not unsympathetic, and in Java I would absolutely do this. But this is Python where everything is very dynamic and rather casual compared to Java.

Passing around dicts for 'just data' is very conventional in dynamic languages like Python, and Lutris does it routinely.

Besides, your Installer class is rather a lot like the existing LutrisInstaller. I'm not comfortable having both.

I only merge bug-fixes, so this will have to wait for Maximum Leader. But I think this would be rejected in its current state.

I suggest what you could do instead is add a key-value pair to the installer dict yourself, with a distinguished name like "_lutris_installer_file" or something like that. It's not the kind of rigorous name-spacing you could have with separated classes, but it's good enough and should mean a *lot* less code churn.

rienafairefr:
You're right that a key-value pair for passing through that information is way easier on the code changes, I pushed that code up :+1:

PR #6163, opened 2025-06-01
deps: add missing python libraries for extract icon from .exe

Zebra2711:
Are we still hiding this feature?


Zebra2711:
[extract_icon.py#L5C1-L11C29](https://github.com/lutris/lutris/blob/aaea5ea6d4306b50c734114be3b57a1fd8b1f0a4/lutris/util/wine/extract_icon.py#L5C1-L11C29)

danieljohnson2:
It's a pretty marginal feature. The ruling from Maximum Leader I got was that if this dumb thing causes any more trouble, it needs to go.

It just provides icons for Wine programs only, only if we don't have an icon anyway, and only if we can get it out of the .EXE, and only when you save. Super jank.

Zebra2711:
> ... and only when you save. Super jank.

It is fixed by this #6031

danieljohnson2:
It is? I do not see how.

PR #6183, opened 2025-06-12
Add a special Wine version to represent the default Umu behavior

danieljohnson2:
This adds a new fake Wine version ``proton`` (presented as "Proton (Latest)") that has *no* Wine-path. Instead it executes ``umu-run`` with no ``PROTONPATH`` set. Umu will download and use a version of Proton of its choice.

This is *not* the same was "GE-Proton (Latest)"; that's a different version with @GloriousEggroll *extra* special sauce. I'm getting an older Proton version from Umu actually. This PR leaves "GE-Proton (Latest)" as it was.

This involves some bodges to redirect around the things that use the Wine-path so they use ``umu-run`` in some way instead. ~~Hope I got them all;~~ Nope, ``CabInstaller`` doesn't support ``umu-run`` at all and it isn't going to work. That may break some installers (on "Proton (Latest)").

It's all done with ``if`` statements, so there's definitely a call for a cleanup pass on this once we've sorted out how all this works.

### DIFF
--- a/lutris/exceptions.py
+++ b/lutris/exceptions.py
@@ -78,6 +78,10 @@ class MissingExecutableError(MisconfigurationError):
     """Raised when a program can't be located."""
 
 
+class UndefinedExecutableError(MissingExecutableError):
+    """Raised when a program isn't actually specified, and shouldn't be needed."""
+
+
 class MissingMediaError(LutrisError):
     """Raised when an image file could not be found."""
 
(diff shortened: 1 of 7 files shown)


strycore:
> umu-run with no PROTONPATH set. Umu will download and use a version of Proton of its choice.

> This is not the same was "GE-Proton (Latest)"

They should be the same.

PR #6220, opened 2025-11-08
Fix EA App Support

neurolag:
EA dropped the Origin API.
Due to these changes, the current EA integration script in Lutris does not work anymore.

Changes made in this PR will drop the use of the Origin API and instead use the new GraphQL API provided by EA.

Some of the changes have been adapted from https://github.com/BellezaEmporium/galaxy-integration-ead
A huge load of the changes have been inspired by @anadius' `Origin Helper.user.js` here: https://gist.github.com/anadius/e42c8da5d97e93c8a8602dce7adb6afd

GraphQL queries for fetching the boxarts and banners have been extracted from the EA Desktop application using `strings`.

With these changes, Lutris is able to fetch metadata of EA games once again.

### DIFF
--- a/lutris/services/ea_app.py
+++ b/lutris/services/ea_app.py
@@ -59,11 +59,30 @@ class EAAppGames:
 class EAAppMedia(ServiceMedia):
     service = "ea_app"
     file_patterns = ["%s.jpg"]
-    size = (0, 0)
+    name = NotImplemented
 
     @property
     def dest_path(self):
-        return os.path.join(settings.CACHE_DIR, self.service, "icon")
+        return os.path.join(settings.CACHE_DIR, self.service, self.name)
+
+    def get_media_url(self, details: Dict[str, Any]) -> Optional[str]:
+        image = details["baseItem"][self.name]["largestImage"]
+        return image.get("path", None) if image != None else None
+
+
+class EAAppKeyArt(EAAppMedia):
+    name = "keyArt"
+    size = (192, 108)
+
+
+class EAAppPackArt(EAAppMedia):
+    name = "packArt"
+    size = (135, 240)
+
+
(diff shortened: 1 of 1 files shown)


danieljohnson2:
Looks like a lot, but I think we need this to fix EA. Merging!

Thanks for all this work.

neurolag:
Awesome - thank you so much 😄

neurolag:
Ooh I just noticed that I'm still downloading game metadata in pages of 10
https://github.com/lutris/lutris/blob/910fe01e10080e1149450ac0042b6420c5aab4d8/lutris/services/ea_app.py#L345

I actually just had this included for testing purposes 🫠 
I don't think this will do any harm but maybe we want to increase this to 100 or so

neurolag:
@danieljohnson2 pinging just in case you want to change this
tysm

danieljohnson2:
Heard you the first time. I'm working on something else right now. I'll get back to you today.

neurolag:
oooh sorry - I wasn't sure whether or not GitHub would let you know since it's closed.
Thank you so much :smile:

danieljohnson2:
OK, that looks pretty safe, so if you say its okay I'll slip that in for you.

neurolag:
Nice, thank you so much 😄
I just double checked in the gist I linked and they use a limit of 100, too

schleusser:
I'm sorry but that fixes nothing for me, same error.

danieljohnson2:
If you are getting the same error *exactly*, you are not running the new EA App code- it's a rewrite.

Does your log still contain the code ``ea_account_info.find("user").find("personaId").text`` in the backtrace? That code's not in Lutris master, so if you see it, you aren't running the master.

PR #6234, opened 2025-11-15
Type Error in wine.py

CMDR-Shrine:
Noticed a bug when killing star citizen, traced it back to a small type error


danieljohnson2:
I don't understand; what's the bug here?

PR #6236, opened 2025-11-18
Update lutris to 3.14 python

ryzendew:
Lutris is not working on Fedora 43 due to python 3.14 
With this Pr now it works! not sure if it will be merged but if any edits are needed let me know!


danieljohnson2:
This does not look right to me.

What's going on with the zoom-in and zoom-out buttons? Why are you changing how the winetricks stuff works?

This seems to want to change from 3.12 to somehow both 3.12 and 3.14 at once. What's up with that?

PR #6239, opened 2025-11-19
vk icd files: add xe->intel dict entry to loader_map

sharkautarch:
fixes error message when using intel xe kernel driver (wherein `DRIVER=xe` instead of `DRIVER=i915`)


danieljohnson2:
This looks solid, I'll take it.

PR #6240, opened 2025-11-20
Fix and Update Persian translations

alr86:


danieljohnson2:
Looks good as far as I know!

PR #6244, opened 2025-11-20
Fix for sorting date columns

Cthululz:
I consistently ran into issues with sorting the date columns of "Last Played Time" and "Installed At".
The columns sorted by the displayed date which begins with the time of day.
image
image

This fixes that issue and sorts by the actual dates instead.
image

image

I ran it locally and it works as expected.


danieljohnson2:
I see what you mean; you can't combine a sort_func and a model like this. I'll take your PR.

PR #6266, opened 2025-12-01
Don't write empty TryExec to desktop files.

sarg:
Flatpak's lutris generates desktop files with empty `TryExec` (see https://github.com/lutris/lutris/pull/5634). Such files are ignored by application launchers, e.g. rofi. [Freedesktop spec](https://specifications.freedesktop.org/desktop-entry/latest/basic-format.html) doesn't really indicate that empty values should be treated as absent, so to avoid misinterpretation it is better to not add the key at all.


danieljohnson2:
Makes sense to me. I'll merge the fix.

PR #6286, opened 2025-12-12
When installing from local script, pass script directory location to script as $SCRIPTDIR variable

### LINKED ISSUE #6283: Allow to reference files relative to the YAML script
Lutris has been accepting installation from local script since v0.5.10. This is awesome because I can now keep a backup of my games on my NAS along with their installation script in the same directory. However, since the script and the location of the files relative to it is known in that case, it seems redondant having to manually locate them. It would be nice to have a $SCRIPTDIR variable that indicates the location of the currently exectuted script. Optionnally Lutris could also provide $SCRIPTNAME although I'm not sure if it's useful.

I can code that provided I'm indicated where to look in the code.

christophehenry:
Fixes #6283 

This probably needs tests. In which case, could you point me to tests I could use as model?

### DIFF
--- a/docs/installers.rst
+++ b/docs/installers.rst
@@ -72,6 +72,8 @@ variables get substituted for their actual value during the install process.
 Available variables are:
 
 * ``$GAMEDIR``: Absolute path where the game is installed.
+* ``$SCRIPTDIR``: Absolute path to the directory containing the YAML script, when installing from
+    local script; in other conditions, this variable is empty.
 * ``$CACHE``: Temporary cache used to manipulate game files and deleted at the
   end of the installation.
 * ``$RESOLUTION``: Full resolution of the user's main display (eg. ``1920x1080``)
@@ -333,7 +335,6 @@ stuttering. (``pulse_latency: true``)
 game is running.  Useful for games that handle other layouts poorly and don't
 have key remapping options. (``use_us_layou: true``)
 
-
 Fetching required files
 =======================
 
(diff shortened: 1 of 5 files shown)


christophehenry:
Now that I think of it, maybe a special way to list files would be more adapted? Something like:

```
files:
- scriptdir: setup.exe
- N/A: please locate extension
```

`scriptdir:` would then try to locate the file relative to the YAML script and ask to locate the faile like `N/A:` does if not found.

WDYT?

danieljohnson2:
I think that that is confusing; it suggests what you want is not the script-dir at all, but rather the directory that contains the setup files. Not the same thing at all!

christophehenry:
You suggest something more descriptive? like `script-relative: setup.exe`? Or another solution entirely?

danieljohnson2:
I think if you want to specify a path relative to the script, ``$SCRIPTDIR`` is better.

If that isn't what you want to do, I don't understand what you do want to do.

christophehenry:
This is what I want to do. I was just thinking this is unusable in the `files:` section because current implementation always interpret that as a file to download from internet and will fail. But maybe this isn't useful and specifying it in `task:` is sufficient. Then this PR can stay as it is.

Should I write tests? Maybe this also require documentation?

danieljohnson2:
Lutris isn't super heavy on automated testing. Additional documentation (I guess in ``installers.rst``) would be desirable though.

christophehenry:
I edited `docs/installers.rst` to document this feature. I tested this locally. I think it's ready for review.

danieljohnson2:
OK, I got around to reviewing this PR. You certainly documented the heck out of it!

I really have only on comment; I think it would have been better to smuggle the ``installer_file` inside the individual installer dictionaries in ``installers``- there'll be only one in the read-from-file case. This will mean quite a few less touches to the code where you've had to pass this along.

I don't insist, but that's really the one suggestion I can make here.

christophehenry:
I thought about doing that but it didn't seem appropriate to put that in the dictionary since it doesn't correspond to an actual declaration in the YAML script. I also thought about creating a new object that would contain the path and the dictionary but it felt too much refactoring and I couldn't ensure I would brake anything elsewhere. I'm fine putting this in the installer dictionary though. What should should be the corresponding key? `installer_file` is ok?

danieljohnson2:
``installer_file`` would be fine, it seems unused. We already tamper with the script you read in ``normalize_installer`` today; this doesn't seem so awful to me.

I'll leave in your hands.

christophehenry:
WTF… I don't why Github keeps closing this… Anyway, I integrated your suggestions.

PR #6293, opened 2025-12-17
Fix German translation for GC

s7a7ic:
Fixes spelling of "Nintenndo" in the German translation.
When using a German locale, the UI now shows the correct icon for the GameCube platform and games.

If GC games were already in the library and showing the generic icon, I had to remove and re-add them in lutris to show the correct system icon after applying the change.

lutris_gc_german


danieljohnson2:
Spelling fix merged!

PR #6294, opened 2025-12-18
Overhaul standalone Script-Generation

### LINKED ISSUE #4427: flatpaked lutris-wrapper seems to disown processes and exit, so steam can't detect non-steam game as running
So far, any game I've installed via flatpak lutris does this, and it is a problem because steam controller schemes do not apply in this case. And this is a huge problem on steam deck in desktop mode. Thankfully, as long as you do not use the multitasking features to open a browser or something, it does apply the controller scheme when launching from game mode.

newhinton:
script generation (-b) now prints the full absolute path of the target script.
This also adds an option to add a game to steam in big-picture mode, which generates a standalone script, and adds that to steam. This should fix the "disown" bug which doesnt allow steam to pass controller input to lutris games.

closes #4427


newhinton:
~~Todo:~~

~~Revert changes to original steam shortcut and create a new one for "big-picture"~~

Edit: Done

Issue #6300, opened 2025-12-22
Steam shortcut created by Lutris breaks after upgrades with nix package manager


harmadasg:
When adding a steam shortcut a few properties are set, including the path for the lutris binary. In the current implementation this path is always the full absolute path, which can be problematic when lutris is installed with the nix package manager. On nix, binaries in `$PATH` are symlinks pointing to immutable derivations in the Nix store. After a system or package upgrade, Lutris may be rebuilt, resulting in a different store path, while the Steam shortcut continues to reference the old one.

- Installed lutris via nix and lutris is added to path. On nix lutris is just a symlink pointing to one specific lutris derivation in nix store.
- Added a new steam shortcut from lutris. When checking the game properties on steam I can see the following path under target: `"/nix/store/dqps9041bah2v9c4s5ybdiz9zzk9llb3-lutris-unwrapped-0.5.19/bin/lutris"`
- Update nix packages. This time something changed in the lutris package so a new derivation is created in the nix store. Now the lutris on `$PATH` resolves to a different entry in the nix store
- When trying to use the existing shortcut in steam the game does not start in my case. Most likely reason is that the lutris binary path points to an older entry in the nix store, and not he latest one after the upgrade
- As a workaround I can delete/recreate the shortcut, or manually change the path to point to the currently active lutris derivation.

PR #6301, opened 2025-12-22
Nix steam shortcut fix

harmadasg:
Doing something similar as to what was done for flatpak. Additionally I had to change `start_dir` too, as it also used the lutris binary location. I simply use the path for the home dir, this is the same approach used by the heroic games launcher.

Resolves https://github.com/lutris/lutris/issues/6300

### DIFF
--- a/lutris/util/steam/shortcut.py
+++ b/lutris/util/steam/shortcut.py
@@ -133,26 +133,22 @@ def generate_shortcut_id(game):
 
 
 def generate_shortcut(game, launch_config_name):
-    lutris_binary = shutil.which("lutris")
+    lutris_binary = "lutris"
 
     launch_options = format_installer_url(
         {"action": "rungameid", "game_slug": game.id, "launch_config_name": launch_config_name}
     )
 
     launch_options = shlex.quote(launch_options)
-    start_dir = f'"{os.path.dirname(lutris_binary)}"'
 
-    if "/nix/store/" in lutris_binary:
-        lutris_binary = "lutris"
-        start_dir = f'"{Path.home()}"'
-    if lutris_binary == "/app/bin/lutris":
+    if is_flatpak_lutris():
         lutris_binary = "/usr/bin/flatpak"
         launch_options = "run net.lutris.Lutris " + launch_options
     return {
         "appid": generate_shortcut_id(game),
         "AppName": game.name,
         "Exe": f'"{lutris_binary}"',
-        "StartDir": start_dir,
+        "StartDir": f'"{Path.home()}"',
         "icon": resources.get_icon_path(game.slug),
         "LaunchOptions": launch_options,
         "IsHidden": 0,
@@ -163,6 +159,10 @@ def generate_shortcut(game, launch_config_name):
         "DevkitOverrideAppID": 0,
         "LastPlayTime": 0,
     }
+
+
+def is_flatpak_lutris():
+    return shutil.which("lutris") == "/app/bin/lutris"
 
 
 def set_artwork(game):


danieljohnson2:
I don't think we like distro specific stuff like this. Can this be done in a way that works on Nix and other distros but without an explicit "is this Nix?" check?

harmadasg:
@danieljohnson2, thanks for the feedback. As mentioned in https://github.com/lutris/lutris/issues/6300, for Nix it is important to use only `lutris` and not the actual absolute path to the lutris binary. I guess the reason for including the full path in the current implementation is that we can't be sure that lutris in included on `$PATH`.

Based on this, I think we can apply the following change:

1. Check if `lutris` is included on `$PATH`. If it is, use only `lutris` instead of the full absolute path.
2. If not included we should fall back to the old behavior and create the shortcut with the absolute lutris binary path

Afaik this should work safely on other distros and also for Nix. Ofc I would not change the [currently used workaround](https://github.com/lutris/lutris/blob/master/lutris/util/steam/shortcut.py#L143-L145) for flatpak lutris. What do you think?

danieljohnson2:
I think what you propose would be an improvement- but I'm not sure. If you give us a PR like that, I won't merge it though, I'd wait for the boss on this.

harmadasg:
@danieljohnson2, I have looked into this, it turned out that the shortcut creation logic already works similar how I described my improvement. https://github.com/lutris/lutris/blob/d69ef8e7c01907a80ce1e4e5cdd79868af6441da/lutris/util/steam/shortcut.py#L135
To retrieve the full lutris binary path `shutil.which` is used and it actually requires `lutris` to be included on `$PATH` otherwise `None` is returned. To simply test this out we can run Lutris from a local git repository and try to create a Steam shortcut (just make sure the lutris package is not installed). Lutris will report an error similar to this:
```
[INFO:2026-01-08 20:53:27,723:shortcut]: Creating Steam shortcut for Baldur's Gate 3 (wine)
[ERROR:2026-01-08 20:53:27,723:exception_backstops]: Error handling signal 'activate': expected str, bytes or os.PathLike object, not NoneType
  File "/tmp/lutris/lutris/exception_backstops.py", line 79, in error_wrapper
    return handler(*args, **kwargs)
  File "/tmp/lutris/lutris/game_actions.py", line 360, in on_create_steam_shortcut
    steam_shortcut.create_shortcut(game, launch_config_name)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/lutris/lutris/util/steam/shortcut.py", line 93, in create_shortcut
    shortcuts = list(shortcuts) + [generate_shortcut(game, launch_config_name)]
                                   ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/lutris/lutris/util/steam/shortcut.py", line 143, in generate_shortcut
    start_dir = f'"{os.path.dirname(lutris_binary)}"'
                    ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
  File "<frozen posixpath>", line 178, in dirname
TypeError: expected str, bytes or os.PathLike object, not NoneType
```

Based on this I think we can just simply use `lutris` in favor of the full binary path when creating Steam shortcuts (except for Flatpak Lutris). I have also updated the PR according to my findings, please let me know what you and the boss think. Thank you!

danieljohnson2:
I don't think the boss is looking right now, but you've sold me that this fixes a bug- crashing when ``lutris`` is not on the path. Making it work in Nix is a okay too, I guess.

I'll go ahead and merge it.

Issue #5986, opened 2025-03-13
Was #5808 implemented


kdliu86:
Was the fix for #5808 implemented in .5.19?

danieljohnson2:
It should be in there.

kdliu86:
Is there a link?  I am wary of upgrading to .5.19 after spending a bunch of time reverting due to .18 breaking all my installs.

danieljohnson2:
I don't think you should; 0.5.19 was a problematic release with no .deb or Flatpak. Can you wait for 0.5.20?

kdliu86:
Another user informed me that this was not fixed and is still broken in .19.

christophehenry:
Hi! Just jumping on this one to say that I just tested on master and winetricks is still impossible on pure-Proton prefixes on main [which these lines confirm](https://github.com/lutris/lutris/blob/master/lutris%2Frunners%2Fcommands%2Fwine.py#L439-L441). I don't understand this exclusion since I tested UMU winetricks on a pure-Proton prefix and it works as expected.

kdliu86:
So does this work in 0.20?

christophehenry:
I haven't tested 0.5.20 yet, but it should, yes.

strycore:
ok closing

kdliu86:
Still broken

```
------------------------------------------------------
------------------------------------------------------
warning: wineserver not found!
------------------------------------------------------
```

1. Select any game.
2. From the winetricks dropdown select open bash terminal
3. Try any winetricks command.
4. Get wineserver not found
5.

christophehenry:
Well then IDK. It was working on my machine during the tests but I did not expect it to always work which is why I kept the warning message. Maybe it's working only with a very recent version of winetricks or a very recent version of Proton. I can test again tomorrow.

christophehenry:
So I just tested here and everything is working. Here's what versions I find in the logs:

> Using winetricks 20250102-next - sha256sum: d00f5634463f5ed00e3af93160e827cb99ab8c647fe7b31848fed806dac50eb5 with wine-10.0 and WINEARCH=win64

But I can't find the warning message I left anywhere which is disconcerting 🤔

kdliu86:
> So I just tested here and everything is working. Here's what versions I find in the logs:
> 
[...]

PR #6309, opened 2025-12-27
fix(winetricks): Use regular winetricks for official Proton builds instead of silently failing

christophehenry:
I don't known why Lutris wasn't doing this up until then but it's working on my machine. Maybe it's due to a recent update? Using:

```
$HOME/.local/share/lutris/runtime/winetricks
Using winetricks 20250102-next - sha256sum: d00f5634463f5ed00e3af93160e827cb99ab8c647fe7b31848fed806dac50eb5 with wine-10.0 and WINEARCH=win64
```

UMU is also correctly executing winetricks verbs with:

```
WINEPREFIX=… $HOME/.local/share/lutris/runtime/umu/umu-run winetricks
```

However, executing UMU on official Proton builds prefixes within Lutris' code also fails with UMU trying to execute `protonfixes` which doesn't exist in official Proton builds. I suppose it's because UMU behaves differently when `PROTONPATH` environement variable is set.

Anyway, I thing the regular `winetricks` code path should be tried when dealing with official Proton builds. If it fails, it fails but it's still better than the current behavior of not trying at all.

This also shoud solve #5986.

### DIFF
--- a/lutris/runners/commands/wine.py
+++ b/lutris/runners/commands/wine.py
@@ -352,11 +352,10 @@ def wineexec(
     command_parameters = []
     if proton.is_proton_path(wine_path):
         command_parameters.append(proton.get_umu_path())
-        if winetricks_wine and wine_path not in winetricks_wine:
+        if winetricks_wine and wine_path not in winetricks_wine and "winetricks" not in args:
             command_parameters.append("winetricks")
     else:
         command_parameters.append(wine_path)
-
     if executable:
         command_parameters.append(executable)
     command_parameters += split_arguments(args)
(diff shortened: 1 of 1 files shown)


lutris/runners/commands/wine.py
```diff
@@ -352,11 +352,10 @@ def wineexec(
     command_parameters = []
     if proton.is_proton_path(wine_path):
         command_parameters.append(proton.get_umu_path())
-        if winetricks_wine and wine_path not in winetricks_wine:
+        if winetricks_wine and wine_path not in winetricks_wine and "winetricks" not in args:
```
christophehenry: Not exactly related with the current feature but when debugging, I observed a duplication of the `winetricks` argument. Didn't seem to affect the behavior of Winetricks itself but it may change in the future and trigger a bizarre bug.

PR #6314, opened 2025-12-29
Update Norwegian Bokmål translation

Martini-141:
Quick refresh, adding new translations and changed some old ones.


danieljohnson2:
I'll take it!

PR #6317, opened 2025-12-29
Add a .lutris file association that execute local file installation

### LINKED ISSUE #6284: Support opening local script by double-click on a `.lutris` file
Complementary to #6283, support local script by double-click on a `.lutris` file would greatly simplify installing games from local script. I'm not sure how this work but, if I'm correct, this would require:

1. modify the Lutris CLI to accept a script argument without an option
2. modify the installation packages (RPM, DEB, etc) to declare an `application/lutris-script` MIME type associated with the `.lutris` extension.

Is anything here blocking?

christophehenry:
Fixes #6284

Adds a new `application/x-lutris` MIME type and a `.desktop` file to associate with lutris local installation. This works in combination with #6286 to easier game installation by just double-clicking a `.lutris` file.


christophehenry:
Ok I think this works. RPM builds and installs correctly here.

PR #6326, opened 2026-01-04
ruff: remove ignore B009 (unnecessary getattr)

duthils:
More details at: https://docs.astral.sh/ruff/rules/get-attr-with-constant/

### DIFF
--- a/lutris/sysoptions.py
+++ b/lutris/sysoptions.py
@@ -545,7 +545,7 @@ def with_runner_overrides(runner_slug):
         runner = runners.import_runner(runner_slug)
     except runners.InvalidRunnerError:
         return options
-    if not getattr(runner, "system_options_override"):
+    if not runner.system_options_override:
         runner = runner()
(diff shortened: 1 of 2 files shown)


danieljohnson2:
I'm rejecting this fix and fixing it myself.

The mechanical fix you are doing is wrong; the problem is that this code is trying to ask "is system_options_override set on the runner class, or only on the instance?"

This matters for the JSON runner, where it is set on the instance, so we need to create a runner here in that case. This wants ``hasattr`` not ``getattr``- but for JSON, it is actually being set on both, so ``hasattr`` is always true anyway. What a mess.

My fix is to *always* create the runner. This fails with a stack overflow for the Pico8 runner, but it is broken in several ways- one is that it decides whether Pico8 can be run alone depending on the configuration, and does this check during runner init. This falls apart if you change the Pico8 configuration anyway.

Quite a mess. I have also removed the run-alone capability of the Pico8 runner as well.

PR #6350, opened 2026-01-19
Fix typo for update_advanced_switch_visibility

ItsAllAboutTheCode:
The `update_advanced_switch_visibility` call in game_common.py:build_header_bar would fail if it was ever called with a non-None `self.notebook` member (currently self.notebook just happens to be `None`) so it no exceptions occur.


danieljohnson2:
Gotcha!

PR #6355, opened 2026-01-20
Fix: Update Ubisoft configuration path to include AppData/Local

italoghost:
### Description

**Context:**
The Ubisoft Connect launcher has recently updated its directory structure, moving the local configuration cache from `Program Files (x86)` to the user's `AppData/Local` folder within the Wine prefix.

**The Problem:**
Currently, Lutris only looks for the configuration file in the legacy path. When this file is missing, Lutris falls back to the Ubisoft Web API. However, the Web API is currently returning incomplete metadata (missing the `thumbImage` field for many titles), which leads to games not being synced or displayed in the library at all. This change ensures that Lutris can always find the local configuration regardless of the launcher version.

**Changes in this PR:**
- Updated `get_configurations` in `lutris/services/ubisoft.py` to search for the configuration file in both the legacy and the new `AppData/Local` paths.
- Implemented `glob` to dynamically handle different Windows usernames (e.g., `steamuser`, `owner`, or system-specific names) inside the Wine prefix.
- Added `import glob` to the service module.

**Testing:**
Verified locally that with this change, Lutris correctly finds the `configurations` file in `drive_c/users/steamuser/AppData/Local/...` and successfully syncs the full library, even when the Web API metadata is failing.

---

**Note on Implementation:**
This is my first PR and it was developed with the assistance of an AI (Google Gemini).

### DIFF
--- a/lutris/services/ubisoft.py
+++ b/lutris/services/ubisoft.py
@@ -59,9 +59,7 @@ class UbisoftCover(ServiceMedia):
             return
         base_dir = ubi_game["directory"]
         asset_file = os.path.join(
-            base_dir,
-            "drive_c/Program Files (x86)/Ubisoft/Ubisoft Game Launcher/cache/assets",
-            url,
+            base_dir, "drive_c/Program Files (x86)/Ubisoft/Ubisoft Game Launcher/cache/assets", url
         )
         cache_path = os.path.join(self.dest_path, self.get_filename(slug))
         if os.path.exists(asset_file):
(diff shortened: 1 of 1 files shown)


lutris/services/ubisoft.py
```diff
@@ -138,16 +139,27 @@ def is_connected(self):
...
+            "drive_c/Program Files (x86)/Ubisoft/Ubisoft Game Launcher/cache/configuration/configurations",
+            "drive_c/users/*/AppData/Local/Ubisoft Game Launcher/cache/configuration/configurations",
+        ]
+
+        for path_pattern in possible_paths:
+            full_pattern = os.path.join(base_dir, path_pattern)
+            matching_paths = glob.glob(full_pattern)
+            if matching_paths:
+                config_path = matching_paths[0]
+                if os.path.exists(config_path):
```
danieljohnson2: I think you mean ``os.path.isfile`` here; a directory won't do.

italoghost:
I changed the code a little, following your comments, adding a for loop to check if the paths are real files and selecting the most recently modified file from a list. I also added some debug loggers.

I copied the configuration file (and the path!) from my old installation (the one where the configuration file is in `Program Files (x86)`) and tested the function. It identified the three paths (the two from `Local` and the one from `Program Files (x86)`), but loaded the one from `steamuser` — the most recent one.

danieljohnson2:
OK, this is looking pretty good to me. I'll go ahead and merge this, then reformat it for you.

Issue #5173, opened 2023-12-07
Amazon Prime integration not downloading all files


BrandowLucas:
When launching specific games with the Amazon Prime Gaming service such as Metal Slug i get this error after 10sec of launching:

![Screenshot_20231207_003728](https://github.com/lutris/lutris/assets/96845660/50d3c316-b7c5-4790-a9c3-8205b98e48ff)

or

![Screenshot_20231207_013648](https://github.com/lutris/lutris/assets/96845660/26cbe70a-2d47-4b9c-8314-fbd847d0d0c7)

This seems like a Lutris specific issue because when I'm using either the official Amazon .exe app or Heroic Games Launcher Metal Slug and other affected games work as expected.

Here is a list with amazon games i've been getting the same error so far:

- Twinkle Star Sprites
- The Last Blade

BrandowLucas:
Well, changing the downloaded Metal Slug Lutris version to the one provided by Heroic Games Launcher i was able to boot the game, it seems the Lutris downloaded game version is broken or something like that..

The only difference here i noticed from Heroic Games Launcher to Lutris is that Heroic creates a separate prefix in "~/Games/Heroic/Prefixes/default/Metal Slug/drive_c/" and download the game on a separate folder outside the wine prefix in "~/Games/heroic/Metal Slug", while on Lutris it creates a wineprefix on "~/Games/amazon/metal-slug/drive_c/" and install the game inside its wineprefix ("~/Games/amazon/metal-slug/drive_c/game/").

BrandowLucas:
Well I just found out that comparing both Heroic and Lutris Metal Slug installation folder, Heroic download extra files to the game (the same for every other game I was having issues with) while Lutris missed some files. I think that is the problem, Lutris is just not downloading all files and thus making the game incomplete and crash-able.

strycore:
I can confirm that 
![image](https://github.com/lutris/lutris/assets/280336/3411b25d-783d-44be-91e1-8fdcc6467e72)

After copying the 4 files from the Heroic install, the game launches fine.

legluondunet:
Is this issue still present?

BrandowLucas:
The issue is still reproducible, please re open it.

lutris 0.5.19-9
tested with Metal Slug (mslug1.exe)

legluondunet:
@BrandowLucas could you try to reproduce this issue with latest Lutris Github code?
Do you know how to do?

BrandowLucas:
[...]

PR #6393, opened 2026-02-08
Fix Amazon Games integration missing files due to hash collisions

BrandowLucas:
Previously, the Amazon service manifest parsing would drop files with identical content (same hash) because it used the hash as a dictionary key, keeping only the last path encountered. This caused games like Metal Slug (and some others) to crash due to missing files.

This change updates `structure_manifest_data` to store a list of paths for each hash. It also updates the installer command `autosetup_amazon` to copy the downloaded file to all required locations.

Fixes: #5173


BrandowLucas:
@danieljohnson2

danieljohnson2:
Looks good to me. Merging!

PR #6412, opened 2026-02-15
mypy: add strict typing to exception_backstops.py

duthils:


danieljohnson2:
Hang on there- this uses type parameter lists, which are a Python 3.12 feature I think. We support back to 3.8

duthils:
> Hang on there- this uses type parameter lists, which are a Python 3.12 feature I think. We support back to 3.8

Oh right, thanks for noticing! I updated the PR with a typing syntax compatible with Python 3.8.

I also see that running `mypy . --python-version 3.8` returns some errors, I'll open another PR for those, if you don't have an objection.

danieljohnson2:
No need for a new PR; I'l just fix the type annotations for Python 3.8 compatibility.

Thanks for the heads up on that.

duthils:
Thank you very much! I also rebased this PR and added `--python-version 3.8` to the mypy check in github action

PR #6421, opened 2026-02-17
Add categories to --list-games

Kate6:
(1) Add a categories list to the output of --list-games --json
(2) Allow filtering the output of --list-games to only a specific category.


JassonCordones:
Remember to tun mypy and ruff check on your source files

strycore:
> Remember to tun mypy and ruff check on your source files

We should add pre-commit hooks to the project to have those run automatically

PR #6433, opened 2026-02-17
remove unused file util/dolphin.py

duthils:
I could not find any reference to functions in this file, and 
it cannot be imported since `util/dolphin/__init__.py` is imported
instead.


ItsAllAboutTheCode:
@danieljohnson2 It looks like this PR can be closed now.

danieljohnson2:
Oh yeah, I had forgotten about it entirely or I would have just merged it. Ah well.

PR #6470, opened 2026-02-21
util.system: remove unused function kill_pid

duthils:
This function is reimplemented in lutris-wrapper

### DIFF
--- a/lutris/util/system.py
+++ b/lutris/util/system.py
@@ -4,7 +4,6 @@ import hashlib
 import os
 import re
 import shutil
-import signal
 import stat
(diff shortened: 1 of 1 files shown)


danieljohnson2:
Gotta be careful with this- certain module functions (and the members of one class) are exposed to installer scripts to run by name, so I had to do a lot of checking to make sure this is safe.

I think it is safe. I've also got a PR up to block access to everything outside of ``lutris.runners.commands.*`` and the ``ScriptInterpreter``'s methods, so in future we can be more sure about this!

PR #6478, opened 2026-02-21
feat: Add GOG cloud save synchronization support

e88z4:
Implement RFC-compliant OAuth2 with PKCE-based cloud save sync for GOG.com. Enables automatic upload/download of game saves before launch and after exit.

Key features:
- OAuth2 PKCE authentication flow (RFC 6749, RFC 7636)
- Automatic sync on game launch (download) and exit (upload)
- Conflict detection with local-wins protection
- Gzip compression/decompression for cloud storage
- Platform-specific save location resolution (Windows/Linux/Mac)
- Progress tracking with detailed logging

Implementation details:
- Game-scoped authentication using builds API credentials
- Platform mapping: lowercase for builds API, capitalized for remote-config
- MD5 checksum verification for download integrity
- Timestamp-based conflict detection
- GTK dialog for manual conflict resolution

Technical fixes:
- Added gzip decompression for downloaded saves (was writing compressed)
- Fixed platform parameter capitalization mismatch between APIs
- Implemented dual platform mapping for different GOG endpoints

Testing:
- 111 unit tests with 98-99% code coverage
- Verified upload/download integrity with MD5 checksums
- Tested conflict detection and data protection
- Validated OAuth2 token refresh flow

Files:
- lutris/services/gog_cloud.py: Core cloud storage protocol (~958 lines)
- lutris/services/gog_cloud_hooks.py: Game lifecycle hooks (~287 lines)
- lutris/gui/dialogs/cloud_sync.py: Conflict resolution dialog
- lutris/game.py: Integration with game launch/exit
- tests/test_gog_cloud.py: Comprehensive test suite
- tests/test_gog_cloud_hooks.py: Hook integration tests

### DIFF
--- a/lutris/game.py
+++ b/lutris/game.py
@@ -744,12 +744,21 @@ class Game:
             self.start_prelaunch_command(self.runner.system_config["prelaunch_wait"])
 
         # GOG cloud save sync (download from cloud before launch)
-        try:
-            from lutris.services.gog_cloud_hooks import sync_before_launch  # noqa: PLC0415
+        if getattr(self, "service", None) == "gog" and getattr(self, "appid", None):
+            try:
+                from lutris.gui.dialogs.cloud_sync_progress import CloudSyncProgressDialog  # noqa: PLC0415
+                from lutris.services.gog_cloud_hooks import sync_before_launch  # noqa: PLC0415
 
-            sync_before_launch(self)
-        except Exception as ex:
-            logger.warning("GOG cloud sync before launch failed: %s", ex)
+                dialog = CloudSyncProgressDialog(
+                    game=self,
+                    sync_func=sync_before_launch,
+                    direction="pre-launch",
+                )
+                dialog.connect("destroy", lambda _d: self.start_game())
+                dialog.run_sync()
+                return True
+            except Exception as ex:
+                logger.warning("GOG cloud sync before launch failed: %s", ex)
 
         self.start_game()
         return True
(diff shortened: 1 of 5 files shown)


strycore (approved):

danieljohnson2:
This is a cool feature, but it's super annoying to have a dialog pop up for a brief flash every time I start a game.

If there aren't any very loud screams about it, I'm going to call this a bug, and replace that dialog with a progress bar in our 'download queue' which appears under the sidebar. We use this for Lutris component downloads, but it was always intended for more.

@e88z4? @strycore?

I'm also thinking about a way to turn this off. Not sure about that, but it seems like something you might want.

danieljohnson2:
OK, I've got [a branch up](https://github.com/lutris/lutris/tree/dj/gog_save_sync_stuff) if anyone is interested.

1. Reports progress in the download queue (under the sidebar)
2. Is a bit less inclined to pop the 'conflict' dialog- so if you skip, you don't always get that on game stop.
3. Has a "Cloud save sync" option on GOG games and system preferences. Defaults to true.

This is a bit much for a "bug fix" but I'm inclined to stretch the rules and merge it myself- but I'll pause a bit to see if anyone calls me on my :poop: before I do that.

strycore:
Thanks, this was needed.

e88z4:
@danieljohnson2 , 

This is good idea. The main reason I had the dialog box because I have a few games that have so many files to be uploaded. It took a while to be synced to the cloud. My first intention is to give user notification that the cloud sync is in progress and the game is about to start as soon as the cloud sync occurs. I don't want user to cancel the process because they might suspect the game crashes.

I like the idea to make this optional as well.

danieljohnson2:
I'm my testing, cancelling the sync never caused a game to crash. Possibly it would if it were downloading files. 

But the initial sync can be quite slow, so I think we need to be able to cancel it.

e88z4:
@danieljohnson2 , @strycore , 

There is one use case that I didn't have a chance to test. GOG starts implementing size restriction for cloud storage for each game for 200MB. I don't think the PR has a handler to handle this situation. When I get a chance or if any of you have a chance to test and handle this use case, then it will be great.

Do we want to give user notification about their storage capacity as well? I will let you decide how to proceed from here. Whatever you decide, I can try to implement.

strycore:
> I'm my testing, cancelling the sync never caused a game to crash.

Cancelling a sync while it's ongoing is likely to cause the cloud save to be in an inconsistent state. It took a while to sync my Cyberpunk 2077 saves but I don't think being able to interrupt the sync process is a good idea. We just need good visual feedback

danieljohnson2:
I think that's not so bad- Lutris will try to sync again after game exit, and will resend all the files it didn't manage to send before, plus any that it did but which changed. And you aren't blocked while it does this, so it should be able to complete.

Two issues occur:
1. We probably need to keep Lutris from exiting while this happens, like we do during game execution.
2. If the user tries to start the same game again, we'd have two overlapping syncs- that's probably not a great idea.

I will take a look at ths.

danieljohnson2:
OK, I think I covered these cases. Me and my :robot:.

1. Lutris won't quit while syncing, just as it won't while a game is running. It'll hide the Lutris window only.
2. If a post-game sync is running for a game, the pre-game-sync for the game will wait for it. This prevents the game starting while a sync is in progress, as well.

Issue #6494, opened 2026-02-25
Flatpak build: NameError: threading not defined in util/http.py after latest update


moiseslopez1014:
### Bug description

Lutris flatpak version doesn't launch since latest commit

### How to Reproduce

1. Go to 'Terminal' 
2. type on 'flatpak run net.Lutris.lutris'

### Expected behavior

Run, as usual

### Log output

```shell
As I have flatpak, lutris -d doesnt work, instead i used:

moises@mentolao:~$ flatpak run net.lutris.Lutris
Gtk-Message: 13:10:02.380: Failed to load module "xapp-gtk3-module"
Traceback (most recent call last):
  File "/app/bin/lutris", line 68, in <module>
    from lutris.gui.application import LutrisApplication
  File "/app/lib/python3.13/site-packages/lutris/gui/application.py", line 37, in <module>
    from lutris.api import get_runners, parse_installer_url
  File "/app/lib/python3.13/site-packages/lutris/api.py", line 20, in <module>
    from lutris.util import http, system
  File "/app/lib/python3.13/site-packages/lutris/util/http.py", line 41, in <module>
    class Request:
    ...<160 lines>...
            return ""
  File "/app/lib/python3.13/site-packages/lutris/util/http.py", line 46, in Request
    stop_request: threading.Event = None,
                  ^^^^^^^^^
NameError: name 'threading' is not defined. Did you forget to import 'threading'?
```

### System Information

```shell
OS: Linux Mint 22.3 cinnamon
Lutris commit id: 4ee07112b96f43605bf81e936a93a6fc6f099ffd2710293ee3e13dc0935feb85
```

ItseMe-ger:
Bug description
Lutris flatpak version doesn't launch since latest commit

How to Reproduce
Steps to reproduce the behavior:

Go to 'Terminal'
type on 'flatpak run net.Lutris.lutris' / klick on the startbutton as usual
Expected behavior
Run, as usual

flatpak run net.lutris.Lutris  -d
Traceback (most recent call last):
  File "/app/bin/lutris", line 68, in <module>
    from lutris.gui.application import LutrisApplication
  File "/app/lib/python3.13/site-packages/lutris/gui/application.py", line 37, in <module>
    from lutris.api import get_runners, parse_installer_url
  File "/app/lib/python3.13/site-packages/lutris/api.py", line 20, in <module>
    from lutris.util import http, system
[...]

PR #6496, opened 2026-02-25
adding missing lib imports in lutris.util.http to fix some name errors

typirr:
a quick fix for issue https://github.com/lutris/lutris/issues/6494

runtime NameError crashes occur in release 5.22 when running the flatpak (observed on nixOS unstable with bundled Python 3.13). the module lutris/util/http.py references lib names (threading, CookieJar, HTTPResponse, Generator) used in type annotations and runtime code but does not import them.

What this change does
adds the missing imports to  lutris/util/http.py

import threading
from http.cookiejar import CookieJar
from http.client import HTTPResponse
from typing import Generator

Reproduction
run the flatpak build of Lutris 5.22 you will see tracebacks like:

NameError: name 'threading' is not defined
NameError: name 'CookieJar' is not defined
NameError: name 'HTTPResponse' is not defined

Fix verification
applied the patch to the flatpak bundle on nixOS unstable 26.05 and Lutris starts successfully afterwards.

Conclusion
iam not an expert so take it as a grain of salt with revision

### DIFF
--- a/lutris/util/http.py
+++ b/lutris/util/http.py
@@ -11,7 +11,6 @@ from ssl import CertificateError
 from typing import TYPE_CHECKING, Any, Collection, Dict, Generator, Optional
 from http.cookiejar import CookieJar
 from http.client import HTTPResponse
-from typing import Generator
 import certifi
 
 from lutris.settings import PROJECT, SITE_URL, VERSION, read_setting


danieljohnson2:
This should have been fixed in 5.22.

Are you sure you don't have 5.21?

typirr:
@danieljohnson2 oh.... the version that i have right now is indeed 5.21 and i thought its 5.22

typirr:
i still dont know why flathub installed 5.21 insted of 5.22 or that its not on flathub yet, sorry for disturbing though!

PR #6516, opened 2026-03-05
mypy: add typing annotations to api.py

duthils:
Note: I added type aliases for the different API payloads, to start adding some semantics to type annotations, though they could be much more elaborate. I can revert them to built-in types if you prefer.

### DIFF
--- a/lutris/api.py
+++ b/lutris/api.py
@@ -27,14 +27,14 @@ from lutris.util.strings import time_ago
 API_KEY_FILE_PATH = os.path.join(settings.CACHE_DIR, "auth-token")
 USER_INFO_FILE_PATH = os.path.join(settings.CACHE_DIR, "user.json")
 
-Game: TypeAlias = Dict[str, Any]
-GamesPage: TypeAlias = Dict[str, Any]
-Installer: TypeAlias = Dict[str, Any]
-InstallerInfo: TypeAlias = Dict[str, str]
-Runner: TypeAlias = Dict[str, Any]
-RunnerVersion: TypeAlias = Dict[str, Union[str, bool]]
-RuntimeVersions: TypeAlias = Dict[str, Any]
-UserInfo: TypeAlias = Dict[str, Any]
+DbGameDict: TypeAlias = Dict[str, Any]
+GamesPageDict: TypeAlias = Dict[str, Any]
+InstallerDict: TypeAlias = Dict[str, Any]
+InstallerInfoDict: TypeAlias = Dict[str, str]
+RunnerDict: TypeAlias = Dict[str, Any]
+RunnerVersionDict: TypeAlias = Dict[str, Union[str, bool]]
+RuntimeVersionsDict: TypeAlias = Dict[str, Any]
+UserInfoDict: TypeAlias = Dict[str, Any]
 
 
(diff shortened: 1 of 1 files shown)


danieljohnson2:
What  is ``Union[str | bool]``?

Also, I think maybe you should use typing.TypeAlias for your type aliases?

duthils:
> What is `Union[str | bool]`?

If I understand correctly, a runner version API response contains something like:
```
    {
        "version": "wine-ge-8-26",
        "architecture": "x86_64",
        "url": "https://...",
        "default": True,
    },
```
Since the `default` key has a boolean value, `get_default_runner_version_info` returns `Dict[str, Union[str, bool]]`


> Also, I think maybe you should use typing.TypeAlias for your type aliases?

Thank you, I added it to explicit the aliases.

duthils:
> What is `Union[str | bool]`?

Oh I realized you probably meant the `|` symbol... That was the mypy suggestion, but I have reverted it to a comma `Union[str, bool]` to keep the same format as other Union

danieljohnson2:
'Kay. I figure we'll switch from ``Union[A, B]`` to the sweeter ``A | B`` syntax that 3.10 supports after a decent interval, but I want to give 0.5.22 a little time to make sure there aren't any more ~~disastrous~~ critical bugs before taking the plunge.

But I don't think there's any point in doing both forms at once.

danieljohnson2:
I'm okay with aliases, but I am not okay with this much ambiguity. Game is really important class in Lutris, and you're using that an alias for Dict.

I think it would be helpful to add some sort of suffix like ``GameDict`` so we know it's not a Game object, and to make it more obvious it's a dict. We actually have a lot of dicts wandering around that represent different things (service-game version game-library game is a big point of confusion!) so these aliases could clarify things, but the fact that these are dicts is important.

danieljohnson2:
Just occurred to me to mention- I've been trying to name *variables* that contain game dicts ``db_game`` instead of ``game``. Possibly a name like ``DbGame`` would suit; though I've not been making enough of a distinction between library and service games (that is, dicts reprenting rows of the ``games`` table versus rows representing ``service_games``)

duthils:
@danieljohnson2 thanks a lot for the explanations and the suggestion. I have renamed the type aliases to `DbGameDict`, `InstallerDict`, etc.

Should I use those aliases in other modules (e.g. in `runner.py`), or keep them in `api.py` for now? Should I create a `types.py` module for this kind of type aliases?

danieljohnson2:
I'll merge this.

I think that these aliases might be placed in 'relevant' modules. I'd put ``DbGameDict`` in ``lutris/database/games.py``, where you find the functions that produce these things. That might be a good rule- the alias goes with the function that creates the dict.

But I may be thinking in C# here, where ``using Thingy`` brings all those declarations into scope at once, and circular references are not a problem. If you find that circular references are a problem, then a ``typealiases.py`` module might be the answer (these are not real types, just aliases).

I think Lutris would benefit from using these comprehensively. We have a million of these damn dicts, and it's far too easy to get confused. Python's type annotations aren't good for much as far as I can see, but they do provide documentation, and we badly want documentation on this.

PR #6519, opened 2026-03-07
Add full-version GOG update support

danieljohnson2:
GOG sometimes releases a new full-version installer rather than (or in addition to) incremental patches. This adds support for detecting and applying those as an in-place upgrade.

- Tracks the GOG installer version in the game config (`service_installer_version`) at install time, so future runs can detect when a newer version is available.
- Game installs that predate this feature won't have this, so we'll assume whatever's up on GOG is an upgrade and offer it.
- "Install updates" will check this against the API's current installer, and if they differ, appends a **Full version** entry to the update list alongside any available patches.
- Full-version installs install in-place: the game directory is reused, no new DB entry is created, and only the config is updated when done.
- Extras are skipped for in-place reinstalls to avoid re-downloading unrelated files.
- Itch.io already did something a lot like this, but I've updated it to work (and look) more like the GOG feature.

_Screenshot showing the installer window with patch options and the new "Full version" entry:_

Screenshot From 2026-03-06 19-47-58

A thought- we might assume that if there isn't a new 'full' installer, you are entirely up to date and give the "no updates" error. But I think GOG does not always post a new full version with its patches in a prompt way, so maybe this is not a good idea.

### DIFF
--- a/lutris/services/itchio.py
+++ b/lutris/services/itchio.py
@@ -543,11 +543,15 @@ class ItchIoService(OnlineService):
         logger.warning("No supported platforms found")
         return {}
 
-    def generate_installers(self, db_game: Dict[str, Any]) -> List[dict]:
+    def generate_installers(self, db_game: Dict[str, Any], runner_name: Optional[str] = None) -> List[dict]:
         """Auto generate installer for itch.io game"""
         details = json.loads(db_game["details"])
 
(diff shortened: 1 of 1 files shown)


danieljohnson2:
Closing: the GOG depot integration (gogdl) handles updates now, making the offline-installer full-version update path unnecessary.

PR #6531, opened 2026-03-10
mypy: add typing annotations to most of util/graphics

duthils:
I added `type: ignore` where code changes are needed to fix the typing issues. I plan to open separate PRs for those, unless you prefer otherwise.

### DIFF
--- a/lutris/util/graphics/glxinfo.py
+++ b/lutris/util/graphics/glxinfo.py
@@ -21,7 +21,6 @@ class GlxInfo:
                           of running the program, useful for testing.
         """
         self._output = output or self.get_glxinfo_output()
-        self._section = None
         self._attrs = set()  # Keep a reference of the created attributes
         self.parse()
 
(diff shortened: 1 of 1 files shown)


danieljohnson2:
I would accept minor fixes are part of a type-annotation PR; like turning ``return []`` to ``return None`` in this example.

I see a bunch of places that look like 'this can't be None, but mypy can't figure that out'. I don't know what fix you have in mind, but ``if not x: raise RuntimeError("OW!")`` or ``cast(Foo, x)`` would be acceptable, as long as you review the callers to make sure None isn't expected.

duthils:
Thanks, I added the fixing commits in this PR. The changes in `GlxInfo` do not make any changes to the logic, but the change in variable scope allows mypy to detect that it will not be `None` in the wrong place and removes the type-check error.

danieljohnson2:
OK; I think the types are what they should be, thought get_current_resolution() should probably actually return a tuple, not a list. I'll look at that after the merge.

PR #6534, opened 2026-03-11
Fix mistakes and improve JSON runner

JassonCordones:


JassonCordones:
Superseded by #6900 - same content, rebuilt as a single commit on top of current master. The old branch was based on my fork's master, which had accumulated ~220 unrelated commits through repeated merge-syncs, so this PR showed 220 commits for a one-file change.

PR #6535, opened 2026-03-11
mypy: add typing annotations in util/graphics

duthils:
There are a few logic changes in `LogicalMonitor`:
* `LogicalMonitor.monitors` now only returns not-`None` values
* `LogicalMonitor.get_config` now raises a RuntimeError if one monitor has no mode (previously raised an `AttributeError`)

### DIFF
--- a/lutris/util/graphics/displayconfig.py
+++ b/lutris/util/graphics/displayconfig.py
@@ -1,6 +1,7 @@
 """DBus backed display management for Mutter"""
 
 from collections import namedtuple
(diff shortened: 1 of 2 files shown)


danieljohnson2:
OK, the RuntimeError looks like it just provides a better error message for a crash that would have happened anyway. I think filtering ``monitors`` also avoids a crash. I see a lot of conversions that I'm not sure are to the right type, but they seem to be in unused properties anyway.

I guess I'll take it.

PR #6543, opened 2026-03-13
mypy: add typing annotations to util/display.py

duthils:


danieljohnson2:
I am not so sure about the ``Optional[Union[str, Iterable[Output]]]`` stuff- the functions handle None, but it's not really something you *should* pass in- you can tell from the warnings. Do we need the ``Optiona[...]`` here, really?

In Python, it's perfectly reasonable to have a non-optional parameter and still check for None- because the static type system is so very weak, you never know you won't get None.

duthils:
Sure, I replaced the `Optional` with an `if` condition in the code that `mypy` believed it could pass `None`. Is this OK?

danieljohnson2:
I think that's better than proliferating Optional[]. Merging ahoy!

PR #6551, opened 2026-03-16
Suppress externally-installed runners by default; remove 'Visible in Side Panel'

### LINKED ISSUE #6056: [Feature Request] Option to fully disable Runners
Currently runners will appear in the list of available runners when manually adding a game even if you have them hidden from the side panel. On a system with a lot of Flatpak emulators it will detect all of them and add them to the drop down menu. This is impractical if you are only using Lutris for PC games. The only way I have found to stop this functionality on the Flatpak version is to remove the "org.freedesktop.Flatpak" talk permission, however this also removes the ability to use the Flatpak version of Steam as a source. 

Image

One possible solution could be replacing the "Visible in Side Panel" option with a drop down including the option to disable the runner. Another one could be adding a disable button next to the install and options buttons within the "Runners" list.

danieljohnson2:
Resolves #6056.

## Summary

Runners installed outside of Lutris (Flatpak, Linux, system Wine, etc.) are now **suppressed by default**. Instead of a separate "Visible in Side Panel" option, users simply install and uninstall runners — this works even for things like Flatpak or Wine that come from the distribution.

- **Suppressed by default**: externally-detected runners (those with no Lutris-managed directory) are hidden unless the user explicitly reveals them.
- **Suppression hides runners**: suppressed runners disappear from the sidebar, game configuration runner drop-down, and saved search editor. But games that already use a suppressed runner still launch fine — suppression never blocks operational use.
- **Auto-unsuppress on install**: installing a game whose runner is suppressed silently unsuppresses it, just as a missing runner is automatically installed.
- **Flatpak-only runners get a popover menu**: when these can either have an internal Lutris version installed or be suppressed, the action button opens a menu with "Install Lutris version" and "Hide Flatpak version".
- **Removes "Visible in Side Panel"**: this option is replaced by suppression. A migration converts existing `visible_in_side_panel` settings to `suppressed` in runner YAML configs.
- **Refactors `is_installed()`**: now a template method — the base class handles `suppress_allowed` and `flatpak_allowed` flags, delegating to `check_installed(**kwargs)` for runner-specific detection. No subclass needs to override `is_installed()` anymore.

## Screenshot

Screenshot From 2026-03-18 17-26-06

*Basilisk II is installed via Flatpak — the hamburger menu offers to install the Lutris-managed version or hide the Flatpak.*

## Crossing the event horizon

While I think this is all nice and good, I think it really adds more complexity than it removes now and is probably getting closed as just too damn much fanciness.

But I may as well offer it. I think it's neat.

### DIFF
--- a/lutris/gui/application.py
+++ b/lutris/gui/application.py
@@ -1080,7 +1080,7 @@ Also, check that the version specified is in the correct format.
 
         try:
             runner = import_runner(runner_name)()
-            if runner.is_installed():
+            if runner.is_installed(suppress_allowed=False):
                 print(f"'{runner_name}' is already installed.")
             else:
                 runner.install(self.install_ui_delegate, version=None, callback=None)
@@ -1099,7 +1099,7 @@ Also, check that the version specified is in the correct format.
         except InvalidRunnerError:
             logger.error("Failed to import Runner: %s", runner_name)
             return
-        if not runner.is_installed():
+        if not runner.is_installed(suppress_allowed=False):
             print(f"Runner '{runner_name}' is not installed.")
             return
         if runner.can_uninstall():
--- a/lutris/gui/config/runner_box.py
+++ b/lutris/gui/config/runner_box.py
@@ -55,7 +55,7 @@ class RunnerBox(Gtk.Box):
         self.configure_button.set_margin_right(12)
         self.configure_button.connect("clicked", self.on_configure_clicked)
         self.pack_start(self.configure_button, False, False, 0)
-        if not self.runner.is_installed() or self.runner.is_suppressed():
+        if not self.runner.is_installed():
             self.runner_label_box.set_sensitive(False)
         self.configure_button.show()
         self.action_alignment = Gtk.Alignment.new(0.5, 0.5, 0, 0)
@@ -77,7 +77,7 @@ class RunnerBox(Gtk.Box):
             _button = Gtk.Button.new_from_icon_name("system-software-install-symbolic", Gtk.IconSize.BUTTON)
             _button.get_style_context().add_class("circular")
             _button.connect("clicked", self.on_unsuppress_clicked)
-        elif self.runner.is_installed():
+        elif self.runner.is_installed(suppress_allowed=False):
             _button = Gtk.Button.new_from_icon_name("edit-delete-symbolic", Gtk.IconSize.BUTTON)
             _button.get_style_context().add_class("circular")
             _button.connect("clicked", self.on_suppress_clicked)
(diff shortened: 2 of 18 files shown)


strycore:
> There is an existing runner option "Visible in Side Panel" that hides a runner from the sidebar

We can't keep adding those kind of options forever. This is just adding more stuff instead of solving the real issue, which is that Flatpaks are detected as installed whether a user wants them or not. Flatpak runners should be "installable" like any other runner, with maybe a message saying the runner is available as a Flatpak. 

I'm even considering runner backends: 

* OS provided binary
* Lutris build (against Sniper runtime)
* Flatpak
* AppImage
* Snap
* User provided binary

Right now we just have custom runner location in the options and some ugly hacks to handle Flatpak runners without good user feedback.

Seems like you got halfway there, for the Flatpak "installs", if you can deactivate them, then the correct move is to not have them active by default.

danieljohnson2:
"We can't keep adding those kind of options forever"? I don't know why not, everybody else does!

But if you read to the end, you will see that this is intend to replace, not add.

But I will adjust the default in the PR.

danieljohnson2:
On reflection, maybe you read the PR but are not impressed with vaguely promised future changes.

If we drop the "Show in sidebar" option, then the UI is simpler for users- there's only one UI, the install/uninstall of runners, and it works for all runners. It is more code, but some of that is an improvement on its own- factoring the Flatpak checks out from the runner-specific "external runner install" checks.

I'll add that to the PR, in the hopes it will be more palatable. It still won't be the big redesign you have in mind, though.

ItsAllAboutTheCode:
What occurs in the situation where a user has an externally installed runner they want to suppress, but also want to use the runner provided by Lutris?
i.e
A Situation where a distro version of a runner is older than what Lutris provides, but the user has the runner installed externally. The user wants to use the Lutris provided runner in that case.

danieljohnson2:
So, I've got some bugs to fix there- the just PR bugs out if you try to play a game with a 'suppressed' runner. I'll fix that tonight.

However, Lutris always has preferred an internally installed runner over a Flatpak one, if both exist, and that counts as 'installed' and 'not suppressed'.

This PR does not support "internally installed, but hidden from the sidebar only", like what it replaces does. This is why my first effort just kept both; but the boss doesn't want to pile on new layers of options, so I went with replacement.

danieljohnson2:
I'm sorted out a bunch of corner cases, so now you can run a game if the runner has been suppressed, installing a game will un-suppress its runner, and you can again install the Lutris runner 'on top of' the Flatpak.

But it has gotten rather fancier- and more complex- than I had originally envisioned. Ah, well.

ItsAllAboutTheCode:
From a usability perspective, I think it is good that the user can now choose to install the Lutris runner again over any system installed runner.

danieljohnson2:
Well, Lutris can already do that. I think this UI is clearer, but the complexity really got up there. I don't think the boss will like it.

PR #6554, opened 2026-03-17
mypy: add typing annotations to runners/__init__.py

duthils:


danieljohnson2:
I see you've started dropping in ``from __future__ import annotations``; We've being using ``"quoted.Identifiers"`` instead so far, and I'm not sure we want to piddle ``from __future__`` everywhere.

I'd be more comfortable if you stuck to the quotes, at least until the boss expresses a preference on this.

duthils:
Thank you for mentioning this, I replaced the import with quoted identifiers.

danieljohnson2:
I see this has already gotten into the code-base a few times- include one inserted by Claude on the bosses behalf. Perhaps he does not care about this.

Anyway, I'll merge it.

PR #6555, opened 2026-03-17
mypy: add typing annotations to database directory

duthils:
* I renamed `DbGameDict` to `ApiGameDict`, so that `DbGameDict` actually comes
  from the SQLite DB. Both types are currently equivalent, but I'm not sure if
  they are semantically the same or if they should be kept as separate types?
* I'm not too sure how to reconcile game IDs that are a strings in the object
  `Game.id`, but integers in the DB `games.id`. Any tips?

### DIFF
--- a/lutris/database/categories.py
+++ b/lutris/database/categories.py
@@ -71,7 +71,7 @@ def get_categories() -> List[Dict[str, Union[int, str]]]:
 def get_all_games_categories() -> Dict[str, List[int]]:
     games_categories = defaultdict(list)
     for row in sql.db_select(settings.DB_PATH, "games_categories"):
-        games_categories[row["game_id"]].append(row["category_id"])
+        games_categories[str(row["game_id"])].append(row["category_id"])
     return games_categories
 
 
@@ -144,7 +144,7 @@ def get_game_ids_for_categories(
     if filters:
         query += " WHERE %s" % " AND ".join(filters)
 
-    result = set(game["id"] for game in sql.db_query(settings.DB_PATH, query, tuple(parameters)))
+    result = set(str(game["id"]) for game in sql.db_query(settings.DB_PATH, query, tuple(parameters)))
     for smart_cat in _SMART_CATEGORIES:
         if excluded_category_names is not None and smart_cat.get_name() in excluded_category_names:
             continue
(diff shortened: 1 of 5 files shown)


danieljohnson2:
Lutris game IDs are strs; they need to be converted from int to str when they come off the database- see ``game.py`` line 89.

This predates my involvement with Lutris but my gut says they started out as ints, then 'widened' to strs over the course of developement. Note that *service game* ``appid``s have to be strings- we don't control them, they come from an external service. Given how sloppy Python typing is, keeping this consistent (and therefore str) is probably wise.

danieljohnson2:
I believe that API games an 'game.py'/'games.py' games are not the same at all- API games are entries from the on-line Lutris.net database, and 'games.py' games are from your PGA.db file.

But there are also 'service games'- these are from an API, but not *our* API; they are from GOG or Steam or whatnot. These are even more different, and there's variation between services. We store these in PGA.db also, but in a different table.

I am not convinced trying to separate ``ServiceGameDict[GOGService]`` from ``ServiceGameDict[ItchIoService]`` is worth the considerable trouble - I think ``ServiceGameDict`` would be as much as is reasonable.

So I think ``DbGameDict``/``DbServiceGameDict``/``ApiGameDict``. Note the latter is not kept on the PGA.db file, so I would not prefix it with Db.

duthils:
Thank you for your explanations!

I changed the type for game IDs to become `str` when extracted from the DB. Every time I checked where those `int`s were used, they were either discarded, converted to `str`, or sent back to the DB.

As for `DbServiceGameDict`, I'll introduce it in a subsequent PR when typing the module `database/services.py`, to avoid this one getting too large, if you have no objection.

danieljohnson2:
I'm vibe-coding like a maniac at the moment, so I thought I'd vibe review this. Claude found 4 issues, but only one is real. I'll let it explain.

duthils:
Thanks, I have fixed this one and added another `str` conversion in `get_game_ids_for_categories`, along with the relevant conversions when consuming its result.

I'm not sure if this PR is the right place to do this kind of conversion changes. I did the best I could to ensure they won't cause trouble, but I can also revert the conversions and reintroduce them when there is more typing coverage, in order to detect potential errors more easily. Any preference?

danieljohnson2:
I'm okay with a certain about of this sort of conversion, swept in with the type annotations. It's often needed to get type annotations be reasonable. I'll merge this.

Issue #6499, opened 2026-02-25
Lutris 0.5.22 Dark Theme in Flatpak Version is not working


A7md70242602GH:
### Bug description

When I set Theme as "System Default (default)" it stills Light theme while my system theme is dark, and when I set it Dark theme do the same thing.

### How to Reproduce

1. Go to 'Three Lines in the top right corner'
2. Click on 'The box in front of Theme'
3. Choose 'Dark'

### Expected behavior

Dark theme is not working.

### Log output

```shell
[INFO:2026-02-26 02:35:58,507:application]: Starting Lutris 0.5.22
[ERROR:2026-02-26 02:35:58,510:system]: ['/usr/bin/vulkaninfo', '--summary'] command failed: [Errno 2] No such file or directory: '/usr/bin/vulkaninfo'
[INFO:2026-02-26 02:35:58,733:startup]: "card1" is No GPU (10de:25a0 103c:8a22 nvidia) Driver 580.126.09
[ERROR:2026-02-26 02:35:58,734:system]: ['/usr/bin/vulkaninfo', '--summary'] command failed: [Errno 2] No such file or directory: '/usr/bin/vulkaninfo'
[INFO:2026-02-26 02:35:58,922:startup]: "card2" is No GPU (1002:1681 103c:8a22 amdgpu) Driver 25.3.3
```

### System Information

```shell
Vulkan support: YES
Esync support: YES
Fsync support: YES
Wine installed: YES
Gamescope: YES
Mangohud: YES
Gamemode: YES
Steam: YES
In Flatpak: YES
[System]
OS:   
Arch: x86_64
Kernel: 6.17.0-14-generic
Lutris Version: 0.5.22
Python Version: 3.13.11 (main, Nov 10 2011, 15:00:00) [GCC 15.2.0]
Desktop: KDE
Display Server: x11
[CPU]
Vendor: AuthenticAMD
Model: AMD Ryzen 7 6800H with Radeon Graphics
Physical cores: 8
Logical cores: 16
[Memory]
RAM: 30.5 GB
Swap: 23.3 GB
[Graphics]
Vendor: AMD
OpenGL Renderer: AMD Radeon 680M (radeonsi, rembrandt, LLVM 21.1.8, DRM 3.64, 6.17.0-14-generic)
OpenGL Version: 4.6 (Compatibility Profile) Mesa 25.3.3 (git-a9c6420d8c)
[...]

PR #6557, opened 2026-03-18
Fix dark theme not applied on startup

strycore:
## Summary
- Fix `_preferred_theme` class default matching the init value `"default"`, which caused the setter's guard clause to skip `_update_is_dark()` entirely on startup — `gtk-application-prefer-dark-theme` was never set
- Fix portal variant unpacking: `call_finish()` and `SettingChanged` signal params return `GLib.Variant` objects, but `_read_value()` compared them against Python ints (`GLib.Variant.__eq__` returns `NotImplemented` for non-Variant types, so comparisons always failed)

## Details

**Bug 1 — init skips theme application:** The class-level `_preferred_theme = "default"` matches the common init value from `settings.read_setting("preferred_theme") or "default"`. The setter's `if self._preferred_theme == preferred_theme: return` fires, so `_update_is_dark()` is never called. The app stays in light mode until the async portal response arrives — and if the portal fails (common on KDE/XFCE), it stays light permanently.

**Bug 2 — variant comparison always fails:** The portal `Read` method returns `(v)` and `SettingChanged` params are `(ssv)`. Indexing or destructuring these yields `GLib.Variant` objects, not native Python types. `_read_value()` compares against `int` literals `1` and `2`, but `GLib.Variant.__eq__(int)` returns `NotImplemented`, so the function always falls through to `return "default"`.

## Test plan
- [ ] Launch Lutris with system dark theme — verify dark mode applied immediately (no flash of light)
- [ ] Launch Lutris with system light theme — verify light mode
- [ ] Change system theme while Lutris is running — verify it follows
- [ ] Set "Dark" / "Light" / "System Default" in Lutris preferences — verify each works
- [ ] Test on GNOME and KDE

Fixes https://github.com/lutris/lutris/issues/6499

### DIFF
--- a/lutris/style_manager.py
+++ b/lutris/style_manager.py
@@ -1,3 +1,5 @@
+from typing import Optional
+
 from gi.repository import Gio, GLib, GObject, Gtk
 
 from lutris import settings
@@ -20,8 +22,8 @@ class StyleManager(GObject.Object):
     """
 
     _dbus_proxy = None
-    _preferred_theme = None
-    _system_theme = None
+    _preferred_theme: Optional[str] = None
+    _system_theme: Optional[str] = None
     _is_dark = False
 
     def __init__(self):
@@ -117,7 +119,7 @@ class StyleManager(GObject.Object):
         self._update_is_dark()
 
     @property
-    def preferred_theme(self) -> str:
+    def preferred_theme(self) -> Optional[str]:
         """Can be 'light' or 'dark' to override the theme, or 'default' to go with
         the system's default theme."""
         return self._preferred_theme


danieljohnson2:
I tested this; I find that is works just fine, but the unpack() doesn't do anything in my debugger. That is to say, ``values`` is a variant, but ``values[0]`` is an int according to type(values[0]). The same thing happens with the tuple destructuring later. 'unpack()' converts the variant to a tuple, but the final result does not seem to be affected.

I think Claude was wrong about what's happening there. I asked it about this, and it said it was right and the change necessary- and then I said by debugger disagreed, and it then it said it had been wrong after all. So it's an AI, take that for what you will.

But changing the initial ``_preferred_theme`` may fix this- I don't have the bug to test if it does. That does seem to force an additional theme update, I think.

PR #6561, opened 2026-03-20
Guard installer commands against paths outside the game directory

danieljohnson2:
## Summary
- Installer script commands (`extract`, `mkdir`, `merge`, `move`, `rename`, `write_file`, `write_json`, `write_config`, `chmodx`, `substitute_vars`) now verify that destination paths resolve within the game's target directory. An erroneous script that accidentally includes `..` in a path or otherwise resolves outside `$GAMEDIR` will get a clear `ScriptingError` instead of silently writing to the wrong place.
- Archive extraction (`tarfile`) now uses the `data` filter (Python 3.12+), which rejects tar members with `..` path components or absolute paths. Execute bits and other normal permissions are preserved. `tarfile.FilterError` is caught and surfaced as a clean `ExtractError`.

## Motivation
Installer scripts specify paths via variable substitution (`$GAMEDIR/foo`), and a typo or misconfigured variable can resolve outside the intended game directory. These guardrails catch that early with a helpful error message rather than letting files land in unexpected locations.

### DIFF
--- a/lutris/gui/installerwindow.py
+++ b/lutris/gui/installerwindow.py
@@ -31,6 +31,7 @@ from lutris.installer import InstallationKind, interpreter
 from lutris.installer.errors import MissingGameDependencyError, ScriptingError
 from lutris.installer.interpreter import ScriptInterpreter
 from lutris.util import xdgshortcuts
+from lutris.util.files import get_display_path
 from lutris.util.jobs import AsyncCall
 from lutris.util.linux import LINUX_SYSTEM
 from lutris.util.log import get_log_contents, logger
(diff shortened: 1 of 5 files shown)


strycore:
Thanks for the security hardening work here. I like the intent but I'm concerned that raising a `ScriptingError` will break a significant number of existing installers. Many scripts legitimately write outside `$GAMEDIR` — for example `write_config` to `~/.config/`, `extract` to `~/.local/share/`, `substitute_vars` on config files in various locations, etc.

Instead of hard-blocking, could we prompt the user for confirmation when a command targets a path outside the game directory? Something like:

> "The installer is trying to write to ~/.loki/quake3. Allow / Deny"

That way users are informed about out-of-gamedir writes and can make the call, but existing installers keep working when the user clicks Allow.

danieljohnson2:
I can do that, but I have to do work and stuff. Hang on.

danieljohnson2:
OK, @strycore, I got 'round to it. There will now be a prompt dialog for for writes outside of the game directory.

CLI installs will still just fail; advise if you want that to change.

danieljohnson2:
PS. I think #6474 is pretty safe from a compatibility point of view, too.

PR #6568, opened 2026-03-23
mypy: add typing annotations to database/*

duthils:


danieljohnson2:
A lot of files, but it looks pretty solid.

danieljohnson2:
However, I've taken a pass over it (and sic'ed Claude on it) and done some revisions. The ``str``/``int`` thing for game ids is a mess with a bunch of mismatches and bugs.

I've put in logic to try to convert all ``int`` game-ids to ``str`` in ``database/games.py``. It's gotta be ``str`` in parts of the UI, so it'll be ``str`` everywhere.

PR #6572, opened 2026-03-24
Updated references to `~/.local/share`, `~/.cache,` `~/.config` to use the XDG spec

ItsAllAboutTheCode:
References to those directories have been updated to use settings.DATA_DIR, settings.CACHE_DIR and settings.CONFIG_DIR respectively. Those refer to $XDG_DATA_HOME, $XDG_CACHE_HOME and $XDG_CONFIG_HOME values otherwise they fallback to the `$HOME/.local/share`, `$HOME/.cache` and `$HOME/.config`

References within to $HOME directories within any Flakpak related code has been unchanged and still references to `~` or `Path.home()` home values.

### DIFF
--- a/lutris/game_actions.py
+++ b/lutris/game_actions.py
@@ -4,11 +4,11 @@
 # pylint: disable=too-many-public-methods
 import os
 from gettext import gettext as _
-from pathlib import Path
 from typing import List
 
 from gi.repository import Gio, Gtk
 
+from lutris import settings
 from lutris.config import duplicate_game_config
 from lutris.database import games
 from lutris.database.games import add_game, get_game_by_field
@@ -402,7 +402,7 @@ class SingleGameActions(GameActions):
             steam_shortcut.create_shortcut(game, launch_config_name, True)
             db_game = games.get_game_by_field(game.id, "id")
             if db_game:
-                gamepath = f"{Path.home()!s}/.local/share/applications/lutris-{game.slug}.sh"
+                gamepath = f"{settings.XDG_DATA_DIR}/applications/lutris-{game.slug}.sh"
                 generate_script(logger, LaunchUIDelegate(), db_game, gamepath)
 
     def on_create_desktop_shortcut(self, *_args):
--- a/lutris/runners/ryujinx.py
+++ b/lutris/runners/ryujinx.py
@@ -3,6 +3,7 @@ import os
 from gettext import gettext as _
 from shutil import copyfile
 
+from lutris import settings
 from lutris.exceptions import MissingGameExecutableError
 from lutris.runners.runner import Runner
 from lutris.util import system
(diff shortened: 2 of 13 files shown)


danieljohnson2:
This is not correct- ``DATA_DIR`` is ``~/.local/share/lutris`` not ``~/.local/share``.

If you want to change these fixed paths to use ``GLib.get_user_data_dir()`` and ``GLib.get_user_cache_dir()`` and so on, you need to do a lot more.

I suggest adding new members to ``settings.py`` with ``XDG_`` prefixes: ``XDG_DATA_DIR = GLib.get_user_data_dir()``, then use that. Don't forget to define ``DATA_DIR`` in terms of ``XDG_DATA_DIR`` if you go this way.

ItsAllAboutTheCode:
> This is not correct- `DATA_DIR` is `~/.local/share/lutris` not `~/.local/share`.
> 
> If you want to change these fixed paths to use `GLib.get_user_data_dir()` and `GLib.get_user_cache_dir()` and so on, you need to do a lot more.
> 
> I suggest adding new members to `settings.py` with `XDG_` prefixes: `XDG_DATA_DIR = GLib.get_user_data_dir()`, then use that. Don't forget to define `DATA_DIR` in terms of `XDG_DATA_DIR` if you go this way.

Ah sorry, I'll missed the `lutris` part of the CONFIG_DIR and DATA_DIR, I'll add the XDG_DATA_DIR, XDG_CONFIG_DIR and XDG_CACHE_DIR constants.

danieljohnson2:
I have a further concern- inside a Flatpak those GLib calls resolve to paths inside the Flatpak, where ``~`` resolves to your real home directory. That's great for our own config and such, that's what we do.

But for runners, should we be using a Flatpak specific config directory? I think it depends on the actual behavior of the specific runner. This seems like it needs a lot of testing in a Flatpak test install.

ItsAllAboutTheCode:
> I have a further concern- inside a Flatpak those GLib calls resolve to paths inside the Flatpak, where `~` resolves to your real home directory. That's great for our own config and such, that's what we do.
> 
> But for runners, should we be using a Flatpak specific config directory? I think it depends on the actual behavior of the specific runner. This seems like it needs a lot of testing in a Flatpak test install.

So this is tricky, because using `~` would use the user home directory regardless of whether they had `XDG_*` set.
Even in a Flatpak environment, if a user ran lutris with XDG_DATA_HOME and XDG_CONFIG_HOME set, it shouldn't write to that user's real home directory.

danieljohnson2:
The problem is that we need to do whatever the 3rd party software does- be it a runner, GOG, EGS, scummvm, But I don't know which use ~ and which XDG_ stuff.

The ones that are for Lutris (there are some in egs.py that refer to ~/.cache/lutris) should use stuff like CACHE_DIR instead.

Most of these changes are for 3rd party software. I'm inclined to not touch those without a bug report.

ItsAllAboutTheCode:
Arlig

> The problem is that we need to do whatever the 3rd party software does- be it a runner, GOG, EGS, scummvm, But I don't know which use ~ and which XDG_ stuff.
> 
> The ones that are for Lutris (there are some in egs.py that refer to ~/.cache/lutris) should use stuff like CACHE_DIR instead.
> 
> Most of these changes are for 3rd party software. I'm inclined to not touch those without a bug report.

Alright, I'll close this PR for now and do some testing around the 3rd Party software especially Flatpak.

PR #6576, opened 2026-03-24
Change 7zip executable path from '7za' to '7z'

TheMaxus:
7za fails to open Nullsoft Installer Exes. 7z is already included in Lutris runtime and this change would help writing automated Lutris installers for older games that use Nullsoft Installers.

### DIFF
--- a/lutris/util/extract.py
+++ b/lutris/util/extract.py
@@ -210,6 +210,7 @@ def _extract_7zip(path: str, dest: str, archive_type: str = None) -> None:
         command.append("-t{}".format(archive_type))
     subprocess.call(command)
 
+
 def _get_7zip_path() -> str:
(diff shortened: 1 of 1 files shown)


danieljohnson2:
Hmmm. I see the runtime contains '7z', but maybe we should be able to fall back to '7za' if it's not found. You can have scenarios where the runtime hasn't been downloaded.

Also, is there an issue in the issue-tracker that this fixes? Some installer that doesn't work with '7za'?

TheMaxus:
It's a good idea to have a fallback and it would be trivial to implement.

There's no issue currently opened, but I provided more context related to this problem in #6577. In the installer I have to call 7z from exec task, but it doesn't feel right as extract exists.

PR #6579, opened 2026-03-24
mypy: add typing annotations to game_actions.py

duthils:


danieljohnson2:
Looks clean, but I'm not sure I should merge a type-annotation PR that fails mypy. It's the principle of the thing, you know?

danieljohnson2:
Okay. It's a bit casty but I don't have a good answer for that.

danieljohnson2:
I've gone in a trimmed it down to one cast, with a check ahead of it. The main idea is that the only window we ever have here is LutrisWindow, so we can just commit to that until proper intersection types arrive (Python 3.12).

duthils:
Very nice, thanks a lot!

PR #6583, opened 2026-03-25
feat: multi-profile support — per-user Wine prefixes, saves, playtime stats

scoofz:
Capture d’écran du 2026-03-27 21-32-40
Capture d’écran du 2026-03-27 21-32-55

[https://youtu.be/g9j7DJAZeBE](url)

## Motivation

When multiple users share the same machine and Lutris installation, game saves, Wine prefixes and playtime stats are currently all stored under the same paths. Switching between user profiles (e.g. parent/child on the same OS account) causes saves to be shared or overwritten.

This PR introduces a **profile system** that isolates per-user data while keeping game installations, runner binaries and global config shared. It also integrates transparently with **Steam Big Picture**: when a Steam user launches a Lutris game via a shortcut, their linked Lutris profile is selected automatically — no manual steps.

---

### 1. Multi-profile system (data layer)

**`lutris/database/schema.py`** — Two new tables:
- `profiles` — `(id TEXT PK, name TEXT, icon TEXT, created_at INTEGER, steam_id TEXT)`
- `profile_game_stats` — `(profile_id, game_id, playtime REAL, lastplayed INTEGER)`

**`lutris/database/profiles.py`** *(new)* — CRUD helpers: `add_profile`, `get_profile`, `get_all_profiles`, `update_profile`, `delete_profile`, `ensure_default_profile`, `get_profile_game_stats`, `update_profile_game_stats`, `get_profile_by_steam_id`, `set_profile_steam_id`. A `default` profile is seeded automatically on first run.

**`lutris/settings.py`** — Adds `PROFILES_DIR = ~/.local/share/lutris/profiles/`.

### 2. Profile manager (`lutris/profile.py`)

`ProfileManager` singleton. Reads the active profile from `lutris.conf` on startup. Exposes directory helpers: `get_wine_prefix_path(slug)`, `get_saves_path(slug)`, `get_profile_game_config_path(configpath)`.

### 3. Config overlay (`lutris/config.py`)

`LutrisConfig` accepts an optional `profile_id`. When set, a per-profile YAML overlay is loaded on top of the global game config and written back separately, leaving the shared config untouched.

### 4. Per-profile game stats (`lutris/game.py`)

`playtime` and `lastplayed` are read from / written to `profile_game_stats` for the active profile. The shared `games` table keeps aggregate playtime for backwards compatibility.

### 5. Wine prefix isolation (`lutris/runners/wine.py`)

`prefix_path` resolution order:
1. Explicit `_prefix` (constructor override)
2. **Per-profile path** (`profiles/{id}/wine-prefixes/{slug}/`) ← new priority
3. Game config YAML `prefix` key
4. `WINEPREFIX` env var
5. Auto-detection via `find_prefix()`

### 6. Wine prefix cloning on first launch

When a profile's Wine prefix doesn't exist yet, Lutris clones the source prefix (original game install) rather than starting blank — so existing saves are preserved on first use under a new profile.

- **`lutris/runners/wine.py`** — `get_prefix_clone_source()` and `_find_original_prefix()` locate the base prefix to clone.
- **`lutris/gui/dialogs/prefix_clone.py`** *(new)* — `PrefixCloneDialog`: progress dialog while `rsync` copies the prefix in the background.
- **`lutris/gui/dialogs/delegates.py`** — `clone_wine_prefix()` in launch delegates triggers the dialog when a clone is needed.
- **`lutris/game.py`** — Calls `delegate.clone_wine_prefix()` before prelaunch when the runner reports a clone source is available.

### 7. Startup profile selector (`lutris/gui/dialogs/profile_selector.py`)

Card-based dialog shown automatically when 2+ profiles exist. Clicking a card switches to that profile before the main window opens. Includes a "Don't show at startup" toggle. Suppressed automatically when Steam auto-detection succeeds (see below).

### 8. Profile manager UI (`lutris/gui/dialogs/profile_dialog.py`)

Create, rename, switch and delete profiles. Shows disk usage per profile. New: **Steam account linking** — each profile can be linked to a Steam account detected from `loginusers.vdf`. The association is stored as `steam_id` in the `profiles` table.

### 9. Steam account ↔ Lutris profile linking + automatic switching

**The goal:** when a Steam user launches a Lutris game via Big Picture (locally or via Moonlight/remote play), their Lutris profile and saves are used automatically.

**How it works:**

- Each profile stores a `steam_id` (steamid64). Link is set via the profile manager UI.
- On launch via `lutris:rungameid/…`, `application.py` reads `loginusers.vdf` to identify the active Steam account, looks up the linked profile via `get_profile_by_steam_id()`, and silently switches to it before the game starts.
- The startup `ProfileSelectorDialog` is suppressed when auto-detection succeeds, so Big Picture launches are seamless.
- Standalone launches (no game URL) always show the selector normally.

Profile selection priority: `--profile` CLI flag > `?profile=` URL param > Steam auto-detect > startup dialog.

**`lutris/api.py`** — Parses `?profile=<id>` from `lutris:rungameid/` URLs.

### 10. Dynamic WINEPREFIX in Steam shortcut scripts (`lutris/runner_interpreter.py`)

Steam runs exported `.sh` launcher files directly — bypassing Lutris's Python stack entirely. Previously these scripts had a hardcoded `WINEPREFIX` pointing to one profile's directory.

`_make_dynamic_wineprefix()` detects when the prefix path is inside `profiles/` and replaces the hardcoded value with a shell expression that runs an inline Python one-liner at script runtime:

```bash
export WINEPREFIX="$(python3 -c "
from lutris.util.steam.config import get_steam_users
from lutris.database.profiles import get_profile_by_steam_id
import os
users = get_steam_users()
profile_id = 'default'
if users:
    p = get_profile_by_steam_id(users[0].get('steamid64', ''))
    if p:
        profile_id = p['id']
print(os.path.expanduser(f'~/.local/share/lutris/profiles/{profile_id}/wine-prefixes/my-game'))
" 2>/dev/null || echo "/fallback/path")"
```

All future **"Create Steam shortcut"** exports are automatically profile-aware — no manual patching needed after adding a new game.

### 11. Steam shortcuts for all users (`lutris/util/steam/shortcut.py`)

`create_shortcut()` and `remove_shortcut()` previously only operated on the currently active Steam account's `shortcuts.vdf`. In a multi-user setup, adding a game while one account was active left the other accounts without the shortcut in their Big Picture library.

Both functions now iterate over all Steam user data directories and write/remove the shortcut for each account. A duplicate-detection guard prevents re-adding entries that already exist.

### 12. CLI flags (`lutris/gui/application.py`)

- `--profile <name|id>` — select a profile at launch
- `--list-profiles` — print available profiles

---

## GUI changes

- `lutris/gui/lutriswindow.py` — "Manage Profiles" entry in the hamburger menu
- `share/lutris/ui/lutris-window.ui` — "Manage Profiles" menu item
- `share/lutris/ui/lutris.css` — `.profile-card` / `.profile-card-active` styles

---

## Backwards compatibility

- A `default` profile is created automatically on first run — no migration needed.
- Existing Wine prefixes and saves remain untouched; they are picked up via the game-config / `find_prefix` fallback.
- New DB columns and tables are added non-destructively.
- Users who never create a second profile see no UI change (selector only appears with 2+ profiles).
- Steam shortcut exports without profile-based prefixes are unaffected by `_make_dynamic_wineprefix()`.

---

## Testing

Tested on Arch Linux / CachyOS with two profiles ("Papa" / "Vincent"), each linked to a separate Steam account, launching games via Moonlight/Steam Remote Play:

- Wine (native prefix) — saves isolated between profiles ✓
- Proton/UMU — Unix ISO path correctly converted to `Z:\…` ✓
- Profile selector at startup ✓ / suppressed when launched from Steam ✓
- Steam auto-switch: connecting as Vincent → Lutris switches to Vincent's profile automatically ✓
- Dynamic WINEPREFIX in `.sh` scripts resolves to the correct profile at runtime ✓
- Wine prefix cloned from original on first profile launch ✓
- `create_shortcut()` writes to all Steam user libraries ✓
- Create / rename / delete / link Steam account via profile dialog ✓
- Playtime tracked independently per profile ✓

---

## Open questions for maintainers

- Should `$PROFILEUSERDIR` be added to the savesync documentation?
- Profile IDs are UUIDs — is a human-readable slug preferred?
- Should the startup selector be opt-in (current: auto-shows when ≥2 profiles) or always opt-in?
- The dynamic WINEPREFIX shell expression requires `python3` in `PATH` at script runtime — acceptable trade-off?

🤖 Co-authored with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lutris/game_actions.py
+++ b/lutris/game_actions.py
@@ -387,6 +387,7 @@ class SingleGameActions(GameActions):
             return False
         from lutris.database.profiles import DEFAULT_PROFILE_ID
         from lutris.profile import get_profile_manager
+
         return get_profile_manager().current_profile_id != DEFAULT_PROFILE_ID
 
     def on_verify_prefix(self, _widget: Gtk.Widget) -> None:
@@ -418,6 +419,7 @@ class SingleGameActions(GameActions):
                 source_prefix = game.runner._find_original_prefix()  # type: ignore[attr-defined]
 
             if source_prefix and os.path.isdir(source_prefix) and source_prefix != prefix_path:
+
                 def _dir_size(path: str) -> int:
                     total = 0
                     for dirpath, _dirnames, filenames in os.walk(path):
@@ -432,9 +434,11 @@ class SingleGameActions(GameActions):
                 source_mb = _dir_size(source_prefix) // (1024 * 1024)
                 ratio = profile_mb / source_mb if source_mb else 1.0
                 if ratio < 0.90:
-                    status = _(
-                        "⚠ Prefix may be incomplete: %d MB (source: %d MB, %.0f%% present)."
-                    ) % (profile_mb, source_mb, ratio * 100)
+                    status = _("⚠ Prefix may be incomplete: %d MB (source: %d MB, %.0f%% present).") % (
+                        profile_mb,
+                        source_mb,
+                        ratio * 100,
+                    )
                 else:
                     status = _("✓ Prefix OK: %d MB (source: %d MB).") % (profile_mb, source_mb)
             else:


ItsAllAboutTheCode:
Do you have any screenshots or videos  of using the profile features in the GUI?

scoofz:
> Do you have any screenshots or videos of using the profile features in the GUI?

I don't, but I definitely can do a video to show the feature in action.

scoofz:
> Do you have any screenshots or videos of using the profile features in the GUI?

https://youtu.be/M8n_F4J0r_0

Here is a quick demo I shot this morning using the cartridges integration I made for it. It shows the profile selector that appears when you start lutris. It also shows that each user has a different game save.

ItsAllAboutTheCode:
> > Do you have any screenshots or videos of using the profile features in the GUI?
> 
> https://youtu.be/M8n_F4J0r_0
> 
> Here is a quick demo I shot this morning using the cartridges integration I made for it. It shows the profile selector that appears when you start lutris. It also shows that each user has a different game save.

Oh, that is pretty cool.
So this is for multiple people sharing the same user account on the system?

In a multi-user scenario where user have their own `/home/<username>` directory, I am assuming each user would just have one profile and take advantage of the sandboxing and security benefits that having a second OS account provides.

scoofz:
> > > Do you have any screenshots or videos of using the profile features in the GUI?
> > 
> > 
> > https://youtu.be/M8n_F4J0r_0
> > Here is a quick demo I shot this morning using the cartridges integration I made for it. It shows the profile selector that appears when you start lutris. It also shows that each user has a different game save.
> 
> Oh, that is pretty cool. So this is for multiple people sharing the same user account on the system?
> 
> In a multi-user scenario where user have their own `/home/<username>` directory, I am assuming each user would just have one profile and take advantage of the sandboxing and security benefits that having a second OS account provides.

Yeah, my goal is to have all the screens in the house being able to use moonlight / sunshine to my gaming VM and quickly select a game with their user profile and their game saves for a smooth user experience. I didn't plan for the use of a secondary user account though, that introduces a whole lot of permission issues and sunshine / moonlight issues too. In the long term, I'm looking for something as smooth as steam's big picture mode. Remote into, select user and play.

scoofz:
I'm actually rethinking my approach, since I can link my games to steam, I'm implementing a way to link the lutris user profiles to their respective steam account, allowing to automatically pick the right user profile with their game saves when logged in with steam big picture.

scoofz:
Updated, you can now link your lutris profile to a steam user account. Which makes the lutris shortcuts launched from steam / big picture, load the related profile's game save for the game.

scoofz:
## Follow-up improvements to prefix cloning (from my personal fork)

While testing this PR on my household setup I hit two issues that led to two small follow-up changes, sharing here for visibility in case reviewers find them useful.

### Bug fix: atomic Wine prefix cloning (`lutris/runners/wine.py`)

**Problem:** `shutil.copytree(original_prefix, prefix_path, symlinks=True)` writes directly to the final destination. If it fails mid-copy (disk full, permission error, SIGKILL), the destination directory is left partially written. On the next launch, Lutris sees the directory exists (and `user.reg` may already be there), skips the clone entirely, and tries to run the game with an incomplete prefix — resulting in a silent exit with return code 0 and no useful error message.

**Reproducer:** Horizon Zero Dawn on a non-default profile — 640 MB cloned vs 716 MB source, `user.reg` present, game exits in 4 s.

**Fix:** copy to a `.tmp` sibling, then `os.rename()` into place (atomic on Linux, same filesystem):

```python
tmp_path = prefix_path + ".tmp"
if os.path.exists(tmp_path):
    shutil.rmtree(tmp_path)
try:
    shutil.copytree(original_prefix, tmp_path, symlinks=True)
    os.rename(tmp_path, prefix_path)
except Exception:
    shutil.rmtree(tmp_path, ignore_errors=True)
    raise
```

### New: "Verify Wine prefix" right-click action (`lutris/game_actions.py`)

Added to `SingleGameActions`, shown only for Wine games on a non-default profile.

Checks:
1. Structural integrity (`user.reg`, `system.reg`, `drive_c/windows/`)
2. Size ratio against the source prefix (warns if < 90%)

Offers to delete and re-clone if either check fails. Useful as a manual escape hatch for prefixes corrupted after the initial clone (disk errors, interrupted copy, etc.).

Not proposing to add this to the upstream PR as-is — the profile system itself is the new contribution — but flagging the atomic clone fix as something worth incorporating if the clone logic lands.

Twig6943:
Wouldn't it make more sense to just use xdg user dirs instead of this whole profile system? e.g.

Playtimes:
~/.config/lutris/playtimes.json
~/.config/lutris/save_dirs.json > make saves default to ~ as well
for prefixes straight up ~/.local/share/wineprefixes

scoofz:
> Wouldn't it make more sense to just use xdg user dirs instead of this whole profile system? e.g.
> 
> Playtimes: ~/.config/lutris/playtimes.json ~/.config/lutris/save_dirs.json > make saves default to ~ as well for prefixes straight up ~/.local/share/wineprefixes

Since the goal is to have a remote, one user machine. We share the same account, the machine is always on, and sunshine brings us directly in steam big picture. The goal is to avoid using any kind of mouse or keyboard, like a remote game console if you will. The user system has to be integrated in the app. XDG makes it impossible to get that seamless experience since it involves painful user switching at the system level, using a mouse and keyboard, typing a password to log in... etc... 

I want to fire up moonlight, remote into my gaming machine, select my user and just play. And it's been working seamlessly so far for me with this.

Twig6943:
> > Wouldn't it make more sense to just use xdg user dirs instead of this whole profile system? e.g.
> > Playtimes: ~/.config/lutris/playtimes.json ~/.config/lutris/save_dirs.json > make saves default to ~ as well for prefixes straight up ~/.local/share/wineprefixes
> 
> Since the goal is to have a remote, one user machine. We share the same account, the machine is always on, and sunshine brings us directly in steam big picture. The goal is to avoid using any kind of mouse or keyboard, like a remote game console if you will. The user system has to be integrated in the app. XDG makes it impossible to get that seamless experience since it involves painful user switching at the system level, using a mouse and keyboard, typing a password to log in... etc...
> 
> I want to fire up moonlight, remote into my gaming machine, select my user and just play. And it's been working seamlessly so far for me with this.

Isn't RDP supposed to have good multi user support?

scoofz:
> Isn't RDP supposed to have good multi user support?

RDP is a remote desktop, it's no where near the performance you get from sunshine / moonlight in terms of latency / frame rate. 

RDP would create a new separate instance, so not really the console experience. 

RDP also introduces friction that I'm trying to avoid, login, password, the ability to use it on a TV or an xbox like I do right now with moonlight....  

Also, this, right there already works and it's stupid simple to use as it is.

PR #6587, opened 2026-03-26
mypy: add typing annotations to game.py

duthils:


danieljohnson2:
Looks good to me.

While you're still here, though, I'm thinking that 0.5.22 has been pretty solid, and we can start introducing the sweeter ``A | B`` and ``A | None``instead of ``Union[A, B]`` and ``Optional[A]``, going forward.

I don't think we're going to have to go back to Python 3.9 or anything.

duthils:
> While you're still here, though, I'm thinking that 0.5.22 has been pretty solid, and we can start introducing the sweeter `A | B` and `A | None`instead of `Union[A, B]` and `Optional[A]`, going forward.

Thanks, will do!

PR #6591, opened 2026-03-26
feat: Added support to edit Launch Configs from the Game Option Tab

### LINKED ISSUE #4648: Request: launchers
Apologies upfront if this is something that's been proposed before or is a milestone target, I could not find anything similar to it.

**tl;dr:**
The idea is to be able to configure tools, programs or plain custom shortcuts for a game, besides the runner configuration itself.
The simplest use case is mod managers or similar tools, where you want to work on the exact same prefix the game uses.

Currently there are two ways I see to do it in Lutris:
1. Changing the executable/command line temporarily to the tool's, running the game, then going back and reverting it. This guarantees it uses the same configuration, but is pretty obviously annoying. Maybe a launcher could be seen as a way to automate this process.
1. For Wine runners, the option for running an EXE in the prefix. Convenient for one-time programs like installers and alike, but annoying for something you might to use often. A big miss from this is that certain configurations seem to be ignored, like environment variables.
1. Creating a separate entry for each tool you need. This can be more easily accomplished by duplicating the original entry and then changing what the runner executes, particularly useful for reusing the same Wine prefix. While it is a good solution, it quickly pollutes the game list with entries for game-specific tools, besides lacking the ability to automatically retrieve thumbnails.

I'd like to propose a third option, the ability to define custom "launchers" for games.
The way it could work is by having a list of programs and arguments you can add to a game's options, which would then appear in the context menu for that game (or the menu from the play button's arrow). These would then execute the command line for the launcher clicked, inheriting all or most of the game's Lutris configuration.

Examples:
- Bethesda games are notoriously moddable, so the use of a mod manager is very, very common. Although the more popular mod managers can launch the games directly, there are tools which won't (like LOOT or other more game-specific tools), so having the ability to quickly launch said tools through Lutris would be very useful and time-saving.
- Often you may want to skip launchers when playing games, but you might need to still use them from time to time to change a setting or two.
- Maybe a game has a different executable for multiplayer, or you use a mod that works over the game's original data but with a different executable. Maybe the game simply has different command line options for things like benchmarks.

There is a very clear cut analog to this in Steam, many games offer different executables/tools/modes when right clicked. Sometimes they are launchers, a special version of the game, maybe a multiplayer mode/server etc. I think that's a very good design to start from, except we'd need to figure out how to best approach letting users define launchers.

The way I see it, something similar to how game-specific settings inherit global settings could work well. Have each launcher start out copying the game's settings, then you can go ahead and change the executable, command line arguments, maybe disable things like gamescope or virtual desktop. But the idea is that we can minimize the amount of "reconfiguration" necessary, the intent is to run the launcher with something as close to the game as possible.

Like how a Windows installation may have many programs installed with quick and easy access, we could benefit from having this for Wine prefixes that may need different programs installed.

I understand this would probably be a pretty big task, so I'm just throwing the idea out there.
I could even start work on it myself if there isn't anything similar already being worked on, but I'd like to make sure I'm well aligned in relation to UI/UX, user expectations and the goals of the project before starting work on a PR.

ItsAllAboutTheCode:
The GameBox class has been moved to it's own separate file and additional functions have been added to edit the launch configs.

To facilitate the changes, the WidgetGenerator class has been updated to include an `option_setters` member, that can be used to set the UI value of the widget from code.

This is needed as when launch configs are switched in the UI, the same Widget Generator object is used to view them.

resolves #4648 

Launch Config Editing demo
----
https://github.com/user-attachments/assets/d5e2ce22-3825-4107-8906-abda9eaba607

### DIFF
--- a/lutris/gui/config/boxes.py
+++ b/lutris/gui/config/boxes.py
@@ -36,7 +36,7 @@ def set_option_wrapper_style_class(wrapper: Gtk.Widget, class_name: str | None):
 
 
 class AdvancedSettingsBox(VBox):
-    """Intermediate vbox class for expsoing the Advanced Visibility options"""
+    """Intermediate vbox class for exposing the Advanced Visibility options"""
 
     def __init__(self, **kwargs) -> None:
         super().__init__(**kwargs)
@@ -197,8 +197,7 @@ class ConfigBox(AdvancedSettingsBox):
                 logger.exception("Failed to generate option widget for '%s': %s", option.get("option"), ex)
 
         show_advanced = settings.read_setting("show_advanced_options") == "True"
-        self._advanced_visibility = show_advanced
-        gen.update_widgets()
+        self.advanced_visibility = show_advanced
 
     def update_widgets(self):
         if self._widget_generator:
--- a/lutris/gui/config/game_common.py
+++ b/lutris/gui/config/game_common.py
@@ -181,7 +181,8 @@ class GameDialogCommon(SavableModelessDialog, DialogInstallUIDelegate):
             self.game_box = self._build_options_tab(
                 _("Game options"),
                 lambda: GameBox(self.config_level, self.lutris_config, self.game),
-                advanced=has_advanced(self.game),
+                # Override advanced visibility to True, since the Launch Configs will be editable by this tab
+                advanced=True,
                 searchable=is_searchable(self.game),
             )
         elif self.runner_name:
--- a/lutris/gui/config/game_config_box.py
+++ b/lutris/gui/config/game_config_box.py
@@ -45,7 +45,7 @@ LAUNCH_CONFIG_GAME_OPTIONS: list[dict[str, Any]] = [
         "option": "args",
         "type": "string",
         "label": _("Arguments"),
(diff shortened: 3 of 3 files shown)


danieljohnson2:
Seems neat, but surely this should be an advanced option (visible only when Advanced is switched on in the header-bar).

ItsAllAboutTheCode:
> Seems neat, but surely this should be an advanced option (visible only when Advanced is switched on in the header-bar).

You're right this should be an advanced option.
I'll update the Launch Config editing to only be available in that scenario.

ItsAllAboutTheCode:
> > Seems neat, but surely this should be an advanced option (visible only when Advanced is switched on in the header-bar).
> 
> You're right this should be an advanced option. I'll update the Launch Config editing to only be available in that scenario.



> > Seems neat, but surely this should be an advanced option (visible only when Advanced is switched on in the header-bar).
> 
> You're right this should be an advanced option. I'll update the Launch Config editing to only be available in that scenario.

I have moved the Launch config editing to now only be available when the Advanced switch is checked.

https://github.com/user-attachments/assets/3689ab6c-5b15-4e67-9fe3-c365d7f9407e

PR #6593, opened 2026-03-27
Feat: Add optional regex filtering to merge_folders function

TheMaxus:
As the title says. In some cases it's more convenient to match source paths with regex rather than putting multiple merges in the script. The change shouldn't break any existing installers. This also needs documentation should the PR be accepted and if that's the case I'll get on it as soon as I can.

### DIFF
--- a/lutris/util/system.py
+++ b/lutris/util/system.py
@@ -353,6 +353,7 @@ def substitute(string_template: str, variables: dict[str, str]) -> str:
         return variables[string_template]
     return template.safe_substitute(variables)
 
+
 def _match_regex(regex: re.Pattern[str] | None, path: str) -> bool:
     if not regex:
         return True
@@ -360,6 +361,7 @@ def _match_regex(regex: re.Pattern[str] | None, path: str) -> bool:
     if match:
         logger.debug("Path %s matches regex, copying", path)
     return bool(match)
+
 
 def merge_folders(source: str, destination: str, regex: re.Pattern[str] | None = None) -> None:
     """Merges the content of source to destination"""


TheMaxus:
Some reasoning behind the change is that in my installer I want to move all files ending with .REZ to game directory.
Merge doesn't seem to support wildcards, so it requires multiple merge tasks to do this.

```yaml
- merge:
    dst: $GAMEDIR/drive_c/Program Files (x86)/Fox/Aliens vs. Predator 2/
    src: $CACHE/disc/PREDATOR.REZ
- merge:
    dst: $GAMEDIR/drive_c/Program Files (x86)/Fox/Aliens vs. Predator 2/
    src: $CACHE/disc/ALIEN.REZ
- merge:
    dst: $GAMEDIR/drive_c/Program Files (x86)/Fox/Aliens vs. Predator 2/
    src: $CACHE/disc/MARINE.REZ
- merge:
    dst: $GAMEDIR/drive_c/Program Files (x86)/Fox/Aliens vs. Predator 2/
    src: $CACHE/disc/MULTI.REZ
```

With regex it becomes:

```yaml
- merge:
    dst: $GAMEDIR/drive_c/Program Files (x86)/Fox/Aliens vs. Predator 2/
    src: $CACHE/disc/
    regex: ".*\\.REZ$"
```

Which is more ergonomic to implement and in addition it seems like executing each task comes
with some overhead, so this seems to make the installer slightly faster. When writing other installers
regex allows to precisely match or unmatch any files, which is in my opinion a very useful tool to have.

I also tested the PR to be working with the example code.

PR #6595, opened 2026-03-27
mypy: add typing annotations to gui/widgets/utils.py

duthils:


danieljohnson2:
You have a unit test failure to fix; but I also wonder why you introduced ``GLib.Bytes``. ``array.array`` not good enough for you?

danieljohnson2:
I was going to ask Claude why it always says ``"LutrisAppilication | None"`` instead of ``"LutrisAppilication" | None`` but I guess I know now.

duthils:
Thanks, I'm working on fixing the test.
About `array.array`, `mypy` is complaining on the first argument that should be `GLib.Bytes`. I can restore the original and ignore the error if you prefer.

ItsAllAboutTheCode:
@duthils 

FYI: You can run the [`make install-hooks`](https://github.com/lutris/lutris/blob/d718903c4d3aa7954b048b1dd01a3d51896951e2/Makefile#L96-L97) command to install local pre-commit hooks that will enforce that the UnitTest, mypy, ruff and the check_annotations.py script have have been run before committing.

danieljohnson2:
It's okay if you can test this change.

duthils:
> @duthils
> 
> FYI: You can run the [`make install-hooks`](https://github.com/lutris/lutris/blob/d718903c4d3aa7954b048b1dd01a3d51896951e2/Makefile#L96-L97) command to install local pre-commit hooks that will enforce that the UnitTest, mypy, ruff and the check_annotations.py script have have been run before committing.

Thanks for the suggestion, for some reason the tests were successful on my machine. I'll investigate why.

danieljohnson2:
I see that ``image2pixbuf`` is never called anywhere. You can't test it; I think you can just delete it. We're using Cairo surfaces now anyway.

duthils:
Ah, thank you, I have removed the function `image2pixbuf`.

duthils:
For the record, the tests were not failing on my machine because I was running Python 3.14 instead of Python 3.12 like the CI.

danieljohnson2:
I looked into this and I have vibe-coded an additional check for our automatic tests to detect this bug; the CI should include it, as does my customary ``make sc``.

I'll merge this PR too.

danieljohnson2:
And I think I've got the upgrade to 3.10 type annotation done. Claude is very good at this - I reviewed it all, and found just one mistake which would have been harmless.

I've re-enabled the ruff check against implicit optionals in parameters; we've got ``| None`` everywhere now.

PR #6600, opened 2026-03-29
mypy: add typing annotations on gui/application.py

duthils:
* I had to enable `allow_redefinition_new` to avoid an error from mypy:
  ```python
    def do_command_line(...):
    ...
        if options.contains("list-service-games"):
        ...
            for game in service_game_list:
                game["installed"] = any(("service_id", game["appid"]) in item.items() for item in game_list)
        ...
        elif action in ("rungame", "rungameid"):
            game = Game(db_game["id"])
            game.game_error.register(on_error)
  ```
  * Mypy raised the following error, because it could not see that `game` changed type:
    ```
    error: "dict[str, str | int]" has no attribute "game_error"
    ```
* If you have a better solution instead of: `option: GLib.Variant = options.lookup_value(...)`, I'm all ears :)

### DIFF
--- a/lutris/gui/application.py
+++ b/lutris/gui/application.py
@@ -493,14 +493,12 @@ class LutrisApplication(Gtk.Application):
         migrate()
 
         run_all_checks()
-        if options.contains("dest"):
-            option: GLib.Variant = options.lookup_value("dest")
+        if option := options.lookup_value("dest"):
             dest_dir = option.get_string()
         else:
             dest_dir = None
 
-        if options.contains("output-script"):
-            option: GLib.Variant = options.lookup_value("output-script")
+        if option := options.lookup_value("output-script"):
             searchstring = option.get_string()
             export_script_game = (
                 games_db.get_game_by_field(searchstring, "id")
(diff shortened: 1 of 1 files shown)


danieljohnson2:
I don't really like using an experimental option like that one.

I suggest separating the variables- use 'db_game' for all uses where it is a dict, 'game' where it is a Game object.

I think ``option: GLib.Variant`` is okay if MyPy can't deduce that. Not sure why it can't.

duthils:
Thanks, I removed the option `allow_redefinition_new`.

As for `option: GLib.Variant`, it tells mypy that the value can not be `None`. Without it, mypy considers that `lookup_value` can return `None` and complains that `None` has no attribute `get_string`.

danieljohnson2:
I couldn't figure out the documentation for GLib, but Claude is a great search tool so here it is. It thinks we can get rid of those type annotations.

danieljohnson2:
Claude is great, but it''s a crying shame I have to ask for :robot: beep boop, though.

duthils:
Thanks, I applied all suggestions, and added a comment on the signal handlers return value.

For the record, the GLib API reference for `VariantDict` is here: https://lazka.github.io/pgi-docs/index.html#GLib-2.0/classes/VariantDict.html#GLib.VariantDict

danieljohnson2:
OK, enough quibbling. Merge!

danieljohnson2:
Had to tweak ``get_lutris_action`` a bit- it was crashing on startup because of a nesting error (that is, nesting is now needed, now that ``url`` and ``urls`` are distinct).

But that's why we're doing this now, hopefully well before the release. These sorts of things are bound to slip in.

PR #6606, opened 2026-03-30
mypy: add typing annotations to util/steam

duthils:
I removed some unused code in `vdf.binary_loads`, as `elif bytes is not str` is always True. I could not understand what the original meaning was, and this code is still present in the upstream VDF library.

### DIFF
--- a/lutris/util/steam/config.py
+++ b/lutris/util/steam/config.py
@@ -2,6 +2,7 @@
 
 import os
 from collections import OrderedDict
+from typing import Any, Iterable
 
 import requests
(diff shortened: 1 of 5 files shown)


danieljohnson2:
I agree with you. Claude identifies that as a Python 2 compatibility shim: ``bytes is not str`` is True on Python 3 but False on Python 2. We require Python 3.10 so we do not need it.

I'll merge this.

Actually, Claude will. I didn't want it to, but it's a bit over-eager. Oh well.

PR #6608, opened 2026-03-30
Add Ares multi-system runner

ItsAllAboutTheCode:
Added the a runner for the Ares Multi System Emulator from https://ares-emu.net/
The Ares Multi system emulator was created using Runner Creation Dialog to prove out the that feature as a POC.
That runner was completely created using that new dialog menu.


appoloin:
"human_name": "ares"   should be  "human_name": "Ares"

ItsAllAboutTheCode:
> "human_name": "ares" should be "human_name": "Ares"

Actually according to how the emulator is titled on the official wesbite lowercase "ares" is correct https://ares-emu.net/

PR #6439, opened 2026-02-18
Add the ability to create new runners in the Lutris UI

ItsAllAboutTheCode:
## Main Feature
A new "Add runner config" section had been added to the bottom of Runners stack in the Preferences menu which allows the user to open a dialog to create a runner: 

Added the "Create New Runner Config" dialog which allows users to create a new Lutris runner and configure each section of it (human_name, platform, game_options. runner_options, etc...).
The dialog validates that the values entered by the user can be used as valid JSON format for the runner.
Once the dialog the dialog is saved, the runner configuration is then immediately available within with the "Runners" stack in the preferences dialog for use.
Furthermore any runner configuration that resides in the "$XDG_DATA_HOME/lutris/runners/json/" directory can now be edited in the "Runners" stack section of the Preferences menu.

Refactored the JSON Runner to be built upon a ModelRunner base class which abstracts the definition of a runner to a dictionary structure.
The JSON runner now just marshals to/from dictionary stored in the ModelRunner base class.
Validation methods for the runner definition has been added to the ModelRunner which validates whether a supplied dict can be used for runner configuration.

e76d0bf5bacc1f51d51382c2c023481ec4b6a095

### Add Support to use YAML based runners
Runners created using YAML in addition to JSON. It uses the same "ModelRunner" backend that the JSON runner uses to marshal to/from a python dict.
A dropdown option has been added to Runner Creation Dialog to choose between JSON or YAML format when saving a newly created runner.
Once a runner is created, it's file format can't be changed to prevent users from creating multiple of the same runner by inadvertently changing the File Format option.
However it may be desirable in the future to allow changing the file format as that would allow conversion of runners between formats.

YAML runners are saved in a directory parallel to JSON runners.
i.e `$XDG_CONFIG_HOME/lutris/runners/yaml` for user created runners
Shared runners that come with lutris should be put in the source code directory of `share/lutris/yaml`

c8ee30051d3dcd4c3bc5e24d116868f456db6650

### Added POC yaml runner using the UI runner creation dialog
Used the UI runner creation dialog to create the eden runner.
83b6bbf0d3ef8bb7591a0c794673493fe526f60d

[...]

PR #6609, opened 2026-03-30
Fixed and issue with Vita3k reporting installed titles as missing and converted the runner to  hybrid python json runner

ItsAllAboutTheCode:
## Missing Title Vita 3K UI fix
Fix Lutris GUI reporting Vita installed titles as missing
The Vita3K emulator launches Vita titles using their Title ID and not a file path.
Therefore the Vita3K runner is setup to accept the title ID field.
However that title ID field was set as the 'main_file' option in the config yaml that gets generated when a Vita entry is added.
The missing installed title check in Lutris invokes the game.py [get_path_from_config](https://github.com/lutris/lutris/blob/4620e1ba2fafb90ac1eef99dad4e5f65017d8554/lutris/game.py#L643) which treats the `main_file` key as a file path.

As the Vita title ID is not a valid file path, the Missing Games check would flag all Vita installed titles as missing.

The fix here is to change the key from 'main_file' to 'title_id' so that the field is not treated as a file path.

## Vita3K hybrid python-json runner
Converted Vita3K runner to a hybrid JSON/Python runner
The Vita3K runner field definition now resides in the vita3k.json file.
The Python file contains a custom play() method which maps runner old entry_point_option field of "main_file" to the new "title_id" entry point since the Vita3K doesn't use file paths to run.

This commit is to be an example of how to define a runner using the Runner Creation Dialog UI(#6439) in JSON and then using Python to extend that runners functionality.

### DIFF
--- a/lutris/runners/json.py
+++ b/lutris/runners/json.py
@@ -65,6 +65,14 @@ class JsonRunner(Runner):
                 arguments += shlex.split(self.runner_config.get(option["option"]))
             else:
                 raise RuntimeError("Unhandled type %s" % option["type"])
+
+        # Prepend the option flag before for entry_point_option value
+        for option in self.game_options:
+            if self.entry_point_option != option["option"]:
+                continue
+            if "argument" in option:
+                arguments.append(option["argument"])
+
         main_file = self.game_config.get(self.entry_point_option)
         if not system.path_exists(main_file):
             raise MissingGameExecutableError(filename=main_file)
--- a/lutris/runners/vita3k.py
+++ b/lutris/runners/vita3k.py
@@ -1,7 +1,8 @@
 from gettext import gettext as _
+from pathlib import Path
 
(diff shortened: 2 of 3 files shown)


danieljohnson2:
This seems like a lot of change to take as a bugfix- but it sounds like a real bug (that all games are "missing").

I'm torn. I could probably be convinced.

ItsAllAboutTheCode:
> This seems like a lot of change to take as a bugfix- but it sounds like a real bug (that all games are "missing").
> 
> I'm torn. I could probably be convinced.

Well at first I made the change to fix the missing title issue with the Vita3K runner in the lutris UI.

However I thought since I was working on the Runner Creation Dialog at the time, it would be a good litmus test to convert over to JSON using the Runner Creation Dialog and show how future maintenance of a runner could be reduced by moving all of it's data to JSON and only having specialized functionality in Python.

PR #6611, opened 2026-03-31
Updated Multi-Platform runners to use the `Runner.platform_dict`

ItsAllAboutTheCode:
The `Runner.platform_dict provides a mapping of "Lutris platform name" -> "emulator platform name"
The "emulator platform name" is normally what is supplied to the emulator as a command line parameter to run that platform. 
The "Lutris platform name" is used by Lutris to store information about a platform in its pga.db and to provide a UI string.

Fixed the  `Runner.get_platform` to return the active Lutris platform name when a dictionary is used.

The Widget Generator changes are to allow the the choice box widget for the platform selection drop down to actually provide the choice of (key, value) when provided with a dictionary. Previously the choice widget would map (key, key) and therefore the game options would get serialized with the selected `key` and not the `value`. Now that a dictionary is passed as the set of choice option, the `value` needs to be serialized.

## Testing Done

* Verified that existing runner configurations for the multi-platform runners that have not been updated from their old keys (`model`, `machine`, `bios`)  for specifying a platform runs without needing user intervention.
* Verified that re-saving the existing runner configuration for the multi-platform runners update their old keys to use the `platform` key 

### Multi-platform runner existing configurations work with no modifications

https://github.com/user-attachments/assets/95d618fe-c406-4bf4-bfaf-44804bb54a78

* Verified all of  the Widget Generator choice types options:
### `list[ list[str] with length == 2]` 
list_of_list_len_2

### `list[tuple[str, str]]`
list_of_tuples_len_2

### `tuple[tuple[str, str]]`
tuple_of_tuples_len_2

### `list[str]`
list_of_strings

### `dict[str, str]`
dictionary_of_string_to_string

### DIFF
--- a/lutris/gui/config/widget_generator.py
+++ b/lutris/gui/config/widget_generator.py
@@ -2,7 +2,7 @@
 
 import os
 from abc import ABC, abstractmethod
-from collections.abc import Callable
+from collections.abc import Callable, Mapping
 from gettext import gettext as _
 from inspect import Parameter, signature
 from typing import TYPE_CHECKING, Any
@@ -420,11 +420,9 @@ class WidgetGenerator(ABC):
             # list[tuple[str, str]]
             # tuple[tuple[str, str]]
             # list[str]
-            # tuple[str]
-            # list[dict[str]]
-            # dict[str, str] - New code should use a dictionary
+            # Mapping[str, str]
             choice_iterable = None
-            if isinstance(choices, dict):
+            if isinstance(choices, Mapping):
                 choice_iterable = choices.items()
             if not choice_iterable:
                 if isinstance(choices, (list, tuple)) and choices:
--- a/lutris/migrations/__init__.py
+++ b/lutris/migrations/__init__.py
@@ -3,7 +3,7 @@ import importlib
 from lutris import settings
 from lutris.util.log import logger
 
-MIGRATION_VERSION = 18  # Never decrease this number
+MIGRATION_VERSION = 19  # Never decrease this number
 
 # Replace deprecated migrations with empty lists
 MIGRATIONS = [
@@ -25,6 +25,7 @@ MIGRATIONS = [
     ["migrate_banners_back"],
     ["migrate_proton_to_wine_dir"],
     ["retrieve_discord_appids"],
+    ["migrate_multi_system_runners_to_use_platform_dict"],
 ]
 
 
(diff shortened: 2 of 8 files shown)


danieljohnson2:
I need to review this.

Still, the first thing I see is that list of types for the choices, and I don't understand why; I don't really like ``list[list[str] with length == 2`` because we can't express that in the type system. ``list[tuple[str, str]]`` should be preferred, I think. I don't see why we should prefer ``dict[str, str]`` here.

Maybe the list should just be ``Iterable[tuple[str, str]]`` and ``Mapping[str, str]``?

ItsAllAboutTheCode:
> I need to review this.
> 
> Still, the first thing I see is that list of types for the choices, and I don't understand why; I don't really like `list[list[str] with length == 2` because we can't express that in the type system. `list[tuple[str, str]]` should be preferred, I think. I don't see why we should prefer `dict[str, str]` here.
> 
> Maybe the list should just be `Iterable[tuple[str, str]]` and `Mapping[str, str]`?

To clarify this is documenting how the current process works with the Choice widget. The only new case here is to support a `dict[str, str]`.
The reason a list to list of strings that has length of 2 is needed to be checked is that this widget can be configured in JSON, which doesn't have Tuple type.
For example here is a link to the ruffle runner that uses a list of list strings approach https://github.com/lutris/lutris/blob/139df2f31be82eec3e32f963311417253a1730f7/share/lutris/json/ruffle.json#L79-L84

The first four screenshots in the PR description references choice widget configuration from the existing JSON and Python files. 
The `list [ list[str] == len 2 ]` case is straight from ruffle.json
The `list[tuple[str,str]]`, `tuple[tuple[str,str]]` and `list[str]` are directly from sysoptions.py.
https://github.com/lutris/lutris/blob/139df2f31be82eec3e32f963311417253a1730f7/lutris/sysoptions.py#L165
https://github.com/lutris/lutris/blob/139df2f31be82eec3e32f963311417253a1730f7/lutris/sysoptions.py#L322-L326
https://github.com/lutris/lutris/blob/139df2f31be82eec3e32f963311417253a1730f7/lutris/sysoptions.py#L296

ItsAllAboutTheCode:
> I don't insist that migration be handled in the 'normal' way, but I suspect it would be cleaner to do so.
> 
> If you are concerned about 'forward compatibility', you can keep the old 'machine' or 'model' key in place, and just add 'platform'.

I added a migration script and removed all of the `def __init__` methods override now that the migration script takes care of updating the existing keys.

PR #6615, opened 2026-04-01
wine: get fstype correctly

Zebra2711:
fix 
```
Can't create the prefix on a file system that does not support Linux symbolic links.

  File "/app/lib/python3.13/site-packages/lutris/exception_backstops.py", line 35, in wrapper
    result = function(*args, **kwargs)

  File "/app/lib/python3.13/site-packages/lutris/game.py", line 822, in configure_game
    raise error

  File "/app/lib/python3.13/site-packages/lutris/util/jobs.py", line 62, in target
    result = self.function(*a, **kw)

  File "/app/lib/python3.13/site-packages/lutris/runners/wine.py", line 1128, in prelaunch
    create_prefix(prefix_path, wine_path=self.get_executable(), arch=self.wine_arch, runner=self)
    ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

  File "/app/lib/python3.13/site-packages/lutris/runners/commands/wine.py", line 163, in create_prefix
    raise RuntimeError(_("Can't create the prefix on a file system that does not support Linux symbolic links."))

RuntimeError: Can't create the prefix on a file system that does not support Linux symbolic links.
```
I fixed my patch by not reusing reuse`linux.LinuxSystem().get_fs_type_for_path(prefix)` because it returns the wrong value

### DIFF
--- a/lutris/util/linux.py
+++ b/lutris/util/linux.py
@@ -194,10 +194,10 @@ class LinuxSystem:  # pylint: disable=too-many-public-methods
     @staticmethod
     def get_drives() -> list[dict[str, Any]]:
         """Return a list of drives with their filesystems"""
-        lsblk_output = system.read_process_output(["lsblk", "-f", "--json"])
-        if not lsblk_output:
+        findmnt_output = system.read_process_output(["findmnt", "-J", "--list"])
+        if not findmnt_output:
             return []
-        return [drive for drive in json.loads(lsblk_output)["blockdevices"] if drive["fstype"] != "squashfs"]
+        return [drive for drive in json.loads(findmnt_output)["filesystems"] if drive["fstype"] != "squashfs"]
 
     @staticmethod
     def get_ram_info() -> dict[str, str]:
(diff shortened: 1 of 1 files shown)


danieljohnson2:
Well, I'd like to merge your bugfix, but I don't really understand it - or the bug it fixes. Can you explain a bit more?

Zebra2711:
@danieljohnson2 should be correct now. On my system, `sudo` is required to run `lsblk -f` to get the `FSTYPE`. Otherwise, the `FSTYPE` column is None, causing `is_disallowed_fs` to return True

jvalecillos:
I can confirm this bug is present in the current version ([c5020975cdfb](https://github.com/flathub/net.lutris.Lutris/commit/c5020975cdfb72f29ba74adec01842adc28430ad)) of the Flatpak on the beta channel.

danieljohnson2:
Well, this was merged on April 1st (ho ho) and that c5020975cdfb commit is from March 18th, so I think I can see the problem you are having there.

PR #6616, opened 2026-04-01
Add support for setting an explict runner name

ItsAllAboutTheCode:
The explicitly set runner name is prefixed in front of a runners `runner_executable` property if it exist.

This removes the need to manually prefix the `runner_executable` property with the name of the runner.

Furthermore, the Python class name or JSON filename containing the runner can be renamed without breaking the runner functionality in any way(.
* The installed runner would use the `name` within the file and therefore changing the name wouldn't change the directory that runners is looked up at
* Because the name field in runner definition hasn't changed, the runner installs to the same location on disk.

## Testing Done

* Verified that the `xenia` runner is able to be installed and launches without error
* Verified each JSON runner can be installed and launch without error.

### DIFF
--- a/share/lutris/json/86box.json
+++ b/share/lutris/json/86box.json
@@ -1,9 +1,10 @@
 {
+    "name": "86box",
     "human_name": "86Box",
     "description": "86Box is an open-source, low-level PC emulator.",
     "platforms": ["MS-DOS"],
     "runnable_alone": "True",
-    "runner_executable": "86box/86Box-Linux-x86_64-b7600.AppImage",
+    "runner_executable": "86Box-Linux-x86_64-b7600.AppImage",
     "flatpak_id": "net._86box._86Box",
     "download_url": "https://github.com/86Box/86Box/releases/download/v5.0/86Box-Linux-x86_64-b7600.AppImage",
     "game_options": [
--- a/share/lutris/json/ags.json
+++ b/share/lutris/json/ags.json
@@ -1,8 +1,9 @@
 {
+    "name": "ags",
     "human_name": "Adventure Game Studio",
     "description": "Graphics adventure engine",
     "platforms": ["Linux"],
-    "runner_executable": "ags/ags.sh",
+    "runner_executable": "ags.sh",
     "game_options": [
         {
             "option": "main_file",
--- a/share/lutris/json/basiliskii.json
+++ b/share/lutris/json/basiliskii.json
@@ -1,9 +1,10 @@
 {
+    "name": "basiliskii",
     "human_name": "Basilisk II",
     "description": "Basilisk II is an Open Source 68k Macintosh emulator",
     "platforms": ["Macintosh"],
     "runnable_alone": "True",
-    "runner_executable": "basiliskii/BasiliskII-x86_64.AppImage",
+    "runner_executable": "BasiliskII-x86_64.AppImage",
     "flatpak_id": "net.cebix.basilisk",
     "download_url": "https://github.com/Korkman/macemu-appimage-builder/releases/download/continuous/BasiliskII-x86_64.AppImage",
     "game_options": [
--- a/share/lutris/json/citra.json
+++ b/share/lutris/json/citra.json
@@ -1,9 +1,10 @@
 {
+    "name": "citra",
     "human_name": "Citra",
     "description": "Nintendo 3DS emulator",
     "platforms": ["Nintendo 3DS"],
     "require_libs": [],
-    "runner_executable": "citra/citra-qt",
+    "runner_executable": "citra-qt",
     "flatpak_id": "org.citra_emu.citra",
     "runnable_alone": true,
     "game_options": [
--- a/share/lutris/json/colem.json
+++ b/share/lutris/json/colem.json
@@ -1,8 +1,9 @@
 {
+    "name": "colem",
     "human_name": "Colem",
     "description": "ColecoVision / Coleco Adam emulator",
     "platforms": ["ColecoVision"],
-    "runner_executable": "colem/colem",
+    "runner_executable": "colem",
     "download_url": "https://fms.komkon.org/ColEm/ColEm56-Ubuntu-x86-bin.tgz",
     "game_options": [
         {
--- a/share/lutris/json/desmume.json
+++ b/share/lutris/json/desmume.json
@@ -1,8 +1,9 @@
 {
+    "name": "desmume",
     "human_name": "Desmume",
     "description": "Nintendo DS emulator",
     "platforms": ["Nintendo DS"],
-    "runner_executable": "desmume/bin/desmume",
+    "runner_executable": "bin/desmume",
     "flatpak_id": "org.desmume.DeSmuME",
     "game_options": [
         {
--- a/share/lutris/json/dgen.json
+++ b/share/lutris/json/dgen.json
@@ -1,8 +1,9 @@
 {
+    "name": "dgen",
     "human_name": "DGen",
     "description": "Sega Genesis emulator",
     "platforms": ["Sega Genesis"],
-    "runner_executable": "dgen/bin/dgen",
+    "runner_executable": "bin/dgen",
     "game_options": [
         {
             "option": "main_file",
--- a/share/lutris/json/frotz.json
+++ b/share/lutris/json/frotz.json
@@ -1,8 +1,9 @@
 {
+    "name": "frotz",
     "human_name": "Frotz",
     "description": "Z-code emulator (Zork, etc.)",
     "platforms": ["Z-Machine"],
-    "runner_executable": "frotz/frotz",
+    "runner_executable": "frotz",
     "entry_point_option": "story",
     "game_options": [
         {
--- a/share/lutris/json/melonds.json
+++ b/share/lutris/json/melonds.json
@@ -1,8 +1,9 @@
 {
+    "name": "melonds",
     "human_name": "melonDS",
     "description": "Nintendo DS emulator",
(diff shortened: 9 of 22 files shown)


ItsAllAboutTheCode:
Here is how the issue looks in practice.

Without the change in this PR:

https://github.com/user-attachments/assets/fc904a31-3684-4411-bbd0-fd320492e3f1


With the change in this PR:

https://github.com/user-attachments/assets/48571e78-9820-48d2-8b34-e5a3fffbfcbe

danieljohnson2:
This seems familiar. We have a nasty compatibility problem for runners like FS-UAE, where the runner has to hardcode ``runner_executable = "fs-uae/fs-uae"`` because it's directory name does not match its runner class name (``fsuae``)

That directory is actually part of the archive that is being extracted; we extract it into the ``runner`` directory, and what the archive contains is the ``fs-uae`` directory with the executable inside.

This is all rather fragile, so be careful. Make sure you've tested installing FS-UAE!

ItsAllAboutTheCode:
> This seems familiar. We have a nasty compatibility problem for runners like FS-UAE, where the runner has to hardcode `runner_executable = "fs-uae/fs-uae"` because it's directory name does not match its runner class name (`fsuae`)
> 
> That directory is actually part of the archive that is being extracted; we extract it into the `runner` directory, and what the archive contains is the `fs-uae` directory with the executable inside.
> 
> This is all rather fragile, so be careful. Make sure you've tested installing FS-UAE!

Yes that works.
This change doesn't modify how any current runners look up their path when extracting.
It just adds an additional search path of `<lutris-runner-dir>/<runner-name>/<runner-executable>` to the `get_executable` method.

https://github.com/user-attachments/assets/8468ce22-f21a-44ec-826b-63668b1c3d70

ItsAllAboutTheCode:
> > This seems familiar. We have a nasty compatibility problem for runners like FS-UAE, where the runner has to hardcode `runner_executable = "fs-uae/fs-uae"` because it's directory name does not match its runner class name (`fsuae`)
> > That directory is actually part of the archive that is being extracted; we extract it into the `runner` directory, and what the archive contains is the `fs-uae` directory with the executable inside.
> > This is all rather fragile, so be careful. Make sure you've tested installing FS-UAE!
> 
> Yes that works. This change doesn't modify how any current runners look up their path when extracting. It just adds an additional search path of `<lutris-runner-dir>/<runner-name>/<runner-executable>` to the `get_executable` method.
> 2026-04-01.17-41-41.mp4

It also works if the `fs-uae` prefix is removed from the runner_executable variable as well

https://github.com/user-attachments/assets/df8226b8-72ec-4f92-802e-cc3b61ce7025


The reason it works is because the FS-UAE runner overrides the directory property to return "fs-uae"
https://github.com/lutris/lutris/blob/406ee88b8137b37cc8772b351a814c19fa08abd1/lutris/runners/fsuae.py#L369-L371

danieljohnson2:
I see. But the JSON runner have to extract as the right subdirectory to work- that directory has to agree with what we expect (from the .json file) but it need not match the runner's name exactly.

So what's the scenario where we need this? Is not the real fix to correct the .json or the archive?

ItsAllAboutTheCode:
> I see. But the JSON runner have to extract as the right subdirectory to work- that directory has to agree with what we expect (from the .json file) but it need not match the runner's name exactly.
> 
> So what's the scenario where we need this? Is not the real fix to correct the .json or the archive?

The scenario is when the user authors a new runner.
They would have to make sure that the when they set the `runner_executable` field inside of the JSON they prefix the name of the executable with the name of the JSON file.

runner_creation_2

This issue in the above picture  is that the user sets the name of the runner to `ares-emu`, but didn't prefix the Runner Executable field with "ares-emu/".

So this changes prevents an extra point of human error.

danieljohnson2:
I haven't looked too close, but I had expected your fancy UI to do that for them.

What would happen if they set the ``runner_executable`` to ``hello_Im_a_dumb_name/ares-emu`` or some other invalid setting?

Maybe validation should be addressing this. I don't feel real great about this solution.

ItsAllAboutTheCode:
> I haven't looked too close, but I had expected your fancy UI to do that for them.
> 
> What would happen if they set the `runner_executable` to `hello_Im_a_dumb_name/ares-emu` or some other invalid setting?
> 
> Maybe validation should be addressing this. I don't feel real great about this solution.

Well I am going to push back on this here, because there is no indication for an author of a runner of the **contract** that the name of the file the runner configuration resides needs to be prefixed to the `runner_executable` property.

So to make it simpler for the author, the author doesn't needed to prefix the `runner_executable` property with the name of the runner file. The `get_executable` function would do that for them.

What the Creation UI does is validate that the `runner_executable` is  filename, however it can't know the `runner_executable` path is valid as the runner would need to be downloaded and installed, which happens when the user installs it via the menu.

danieljohnson2:
Well, yes, and that's a flaw. We should, in my option, have required every runner to be in a directory whose name = the runner's official name. The archive should contain the executable directory and not a directory.

But I can't change this - we would have to repack all the runners and I don't even have access to that. Even if I did, they would then be broken for previous Lutris versions. We're stuck.

I was expected the UI to ensure that the runner-executable path was in the form ``<runner-name>/<executable>``- that is, to disallow things like the ``fs-uae`` nonsense. But this would require the corresponding archive to be in that exact form, and there's no obvious way to check that.

ItsAllAboutTheCode:
> Well, yes, and that's a flaw. We should, in my option, have required every runner to be in a directory whose name = the runner's official name. The archive should contain the executable directory and not a directory.
> 
> But I can't change this - we would have to repack all the runners and I don't even have access to that. Even if I did, they would then be broken for previous Lutris versions. We're stuck.
> 
> I was expected the UI to ensure that the runner-executable path was in the form `<runner-name>/<executable>`- that is, to disallow things like the `fs-uae` nonsense. But this would require the corresponding archive to be in that exact form, and there's no obvious way to check that.

Actually I have a proposal in that regard.
Since the Create Runner UI is new, what if we added an  `name` field that would get used as the "runner name" file if set.
The name field would be optional.
If not set for JSON files, they would use their current behavior of using the stem of the filename as the "runner name".
For Python it would use the name of the Python class as it does currently.

That way even if the user renames the JSON file or changes the name of the Python class, the runner would still work.

So new runners going forward would get this functionality, old runners would work without issue.

danieljohnson2:
I do like your idea.

I asked Claude and it says I'm right and I'm so smart because it always says that. Customer is always right. Buuut, it suggests an alternative:

Detect if runner_executable is a bare-filename upon loading the .JSON (and later .YAML) and prepend the runner-name at that point (that is, in __init__).

This ``runner_executable`` correct for all users, instead of just in ``get_executable()``. But it means that any runner that likes to have further sub-directories (and some do) must include that first segment explicitly still, which is a confusing irregularity.

Maybe we can combine these ideas. Add "name" and and it is present, prepend it on the runner_executable in __init__? Your YAML UI editor would always provide name, so ``runner_executable`` becauses relative to the individual runner's  directory

ItsAllAboutTheCode:
> I do like your idea.
> 
> I asked Claude and it says I'm right and I'm so smart because it always says that. Customer is always right. Buuut, it suggests an alternative:
> 
> Detect if runner_executable is a bare-filename upon loading the .JSON (and later .YAML) and prepend the runner-name at that point (that is, in **init**).
> 
> This `runner_executable` correct for all users, instead of just in `get_executable()`. But it means that any runner that likes to have further sub-directories (and some do) must include that first segment explicitly still, which is a confusing irregularity.
> 
> Maybe we can combine these ideas. Add "name" and and it is present, prepend it on the runner_executable in **init**? Your YAML UI editor would always provide name, so `runner_executable` becauses relative to the individual runner's directory

That sounds like a plan! I'll get working on that.

ItsAllAboutTheCode:
> > I do like your idea.
> > I asked Claude and it says I'm right and I'm so smart because it always says that. Customer is always right. Buuut, it suggests an alternative:
> > Detect if runner_executable is a bare-filename upon loading the .JSON (and later .YAML) and prepend the runner-name at that point (that is, in **init**).
> > This `runner_executable` correct for all users, instead of just in `get_executable()`. But it means that any runner that likes to have further sub-directories (and some do) must include that first segment explicitly still, which is a confusing irregularity.
> > Maybe we can combine these ideas. Add "name" and and it is present, prepend it on the runner_executable in **init**? Your YAML UI editor would always provide name, so `runner_executable` becauses relative to the individual runner's directory
> 
> That sounds like a plan! I'll get working on that.

I made the changes. 
It works perfectly.

I was able to rename the runner file with no issue.
Furthermore the existing runners work with no modifications as well.

### Testing Environment
The runner creation dialog resulted in the following YAML content:
```yaml
description: ''
download_url: https://git.eden-emu.dev/eden-emu/eden/releases/download/v0.2.0-rc2/Eden-Linux-v0.2.0-rc2-amd64-gcc-standard.AppImage
entry_point_option: main_file
env: {}
flatpak_id: ''
game_options:
- default: ''
  label: ROM/ISO/exe file
  option: main_file
  type: file
human_name: TestRunner
name: test-runner
platforms:
  Nintendo Switch: Nintendo Switch
runnable_alone: true
runner_executable: Eden-Linux-v0.2.0-rc2-amd64-gcc-standard.AppImage
runner_options: []
system_options_override: []
working_dir: ''

```

https://github.com/user-attachments/assets/c35f56c3-bef2-4f58-9d50-b67d8cbd0157

danieljohnson2:
Looking good. I might want to make changes later- maybe make the name property settable instead of a separate 'runner_name'? But it's solid as it is.

PR #6617, opened 2026-04-01
mypy: add typing annotations to util/steam

duthils:
* I removed the method `AppManifest.get_platform()` because it is not used and it can't work as is, since it expects `get_steamapps_dirs` to return a `dict`, but it actually returns a `list[str]`, AFAICS. Is that OK?
* I didn't find any usage of `util.steam.vdf.vdict.VDFDict`. I added the typing annotations with a few ignores, but maybe we can delete it altogether, if it's not used?


danieljohnson2:
Dropping dead, broken code (rather than annotating it) is fine. If that removes the ``# type: ignore``s all the better!

But I do see some ``__iter__`` methods returning ``Any`` not ``Iterator[something]`` and that does not seem right.

duthils:
Thanks, I have removed the module `vdict.py`.

danieljohnson2:
Looking solid.

PR #6628, opened 2026-04-04
mypy: add typing annotations to monitored_command.py

duthils:


danieljohnson2:
Looks pretty good to me.

PR #6630, opened 2026-04-04
Security: Path traversal via unsanitized slug in filesystem path helpers

tuanaiseo:
## Problem

Path builders directly concatenate `game_slug` into filesystem paths. If `game_slug` can contain `../` or absolute path segments from untrusted metadata, callers may read/write outside intended media directories.

**Severity**: `medium`
**File**: `lutris/util/resources.py`

## Solution

Normalize and validate slugs before path construction (allowlist characters like `[a-z0-9-]`), and enforce that resolved paths stay under the intended base directory using `os.path.realpath` prefix checks.

## Changes

- `lutris/util/resources.py` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

### DIFF
--- a/lutris/util/resources.py
+++ b/lutris/util/resources.py
@@ -1,20 +1,41 @@
 """Utility module to handle media resources"""
 
 import os
+import re
 
 from lutris import settings
 
 
+_SLUG_RE = re.compile(r"^[a-z0-9-]+$")
+
+
+def _sanitize_game_slug(game_slug: str) -> str:
+    if not _SLUG_RE.fullmatch(game_slug):
+        raise ValueError("Invalid game slug")
+    return game_slug
+
+
+def _safe_join(base_path: str, filename: str) -> str:
+    base_real = os.path.realpath(base_path)
+    candidate_path = os.path.realpath(os.path.join(base_real, filename))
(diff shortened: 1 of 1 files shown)


danieljohnson2:
This seems alright, or will be after you satisfy ruff, but could you explain things a bit more?

Why do we need ``realpath``?

Why are we checking these specific paths? I'd have thought we sanitize slugs going in and out of the database or something like that so all uses get protection.

ItsAllAboutTheCode:
Looking at the Pull Request history of the author, this looks like this change was automated using some kind of AI tooling. The author has opened 96 PRs across 62 repos since the beginning of April

danieljohnson2:
Hmm. We hate AI around here, don't we? :smiling_imp: Maybe I should do something like this myself?

ItsAllAboutTheCode:
> Hmm. We hate AI around here, don't we? 😈 Maybe I should do something like this myself?

I have nothing against AI, it is just that this PR triggers some flags for nefarious behavior for me.

strycore:
PR  can't even do the bare minimum and pass the ruff check

PR #6640, opened 2026-04-07
Security: Path traversal vulnerability in firmware copy operation

barttran2k:
## Problem

The `get_firmware()` function uses `target_firmware_name` directly in `os.path.join(runner_system_path, target_firmware_name)` without sanitizing it for path traversal characters (e.g., `../`). If `target_firmware_name` contains directory traversal sequences, an attacker could write files to arbitrary locations outside the intended `runner_system_path`. Additionally, `cached_firmware_record['name']` is read from a JSON cache file and used as the source path for `shutil.copyfile()` without validation, allowing reading arbitrary files if the cache is tampered with.

**Severity**: `medium`
**File**: `lutris/util/retroarch/firmware.py`

## Solution

Sanitize `target_firmware_name` by stripping path separators (e.g., `os.path.basename(target_firmware_name)`) before using it in the destination path. Validate that `cached_firmware_record['name']` points to a file within the expected BIOS directory, and that the resolved destination path is within `runner_system_path` using `os.path.realpath()`.

## Changes

- `lutris/util/retroarch/firmware.py` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

### DIFF
--- a/lutris/util/retroarch/firmware.py
+++ b/lutris/util/retroarch/firmware.py
@@ -33,9 +33,19 @@ def get_firmware(target_firmware_name: str, target_firmware_checksum: str, runne
     for cached_firmware_record in bios_cache:
         # The checksum of the installed firmware we're looking at matches our target
         if cached_firmware_record.get("md5_hash") == target_firmware_checksum:
+            safe_firmware_name = os.path.basename(target_firmware_name)
+            if not safe_firmware_name:
+                logger.error("Invalid firmware name: %s", target_firmware_name)
+                return
+            source_path = os.path.realpath(cached_firmware_record["name"])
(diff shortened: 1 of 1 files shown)


danieljohnson2:
This is wrong is a few ways.

The ``target_firmware_name`` is generated via ``firmware_name = firmware_filename.split("/")[-1]`` elsewhere and I don't think that can be other than a single filename- though you'd think an ``os.path`` function would be more appropriate there.

The ``bios_cache`` contains paths generated by Lutris and saved there, and are under our control.

If you do want to do a check like this anyway, you should use ``system.path_contains`` instead of rolling your own. If that's not reliable enough, we should fix it.

I'll close this item without merging if you do not respond in a few days. And I will think you are a bot.

PR #6642, opened 2026-04-09
mypy: add typing annotations to gui/dialogs/__init__.py

duthils:


lutris/gui/dialogs/__init__.py
```diff
@@ -251,6 +262,9 @@ def __init__(self, message_markup: str, secondary: str | None = None, parent: Gt
         self.run()
         self.destroy()
 
+    # workaround missing typing stubs in Gtk
+    def run(self) -> Gtk.ResponseType: ...  # type: ignore
```
ItsAllAboutTheCode: Since this isn't a type stubs doesn't this need to call `return super().run()`

Doing a quick local test in the Python REPL the ellipsis statement does not call the base class
```python
>>> class Base:
...     def run(self):
...         return 7
...
>>> class Derived(Base):
...     def run(self) -> int: ...
...
>>> d = Derived()
>>> d.run()
>>> print(d.run())
None
>>>
```

lutris/gui/dialogs/__init__.py
```diff
@@ -556,36 +585,36 @@ class ClientLoginDialog(GtkBuilderDialog):
...
         connect_button.connect("clicked", self.on_connect)
 
-    def get_credentials(self):
+    def get_credentials(self) -> tuple[str, str]:
         username = self.username_entry.get_text()
         password = self.password_entry.get_text()
         return username, password
 
-    def on_username_entry_activate(self, widget):  # pylint: disable=unused-argument
+    def on_username_entry_activate(self, widget: Gtk.Entry) -> None:  # pylint: disable=unused-argument
```
ItsAllAboutTheCode: I believe if the argument name starts with an underscore, it will not be marked as unused.
So these widget arguments could be updated to `_widget` which should allow the annotation to be removed.

lutris/gui/dialogs/__init__.py
danieljohnson2: I'm pretty sure it does need that; otherwise all calls to this method just do nothing and return None everything.

lutris/gui/dialogs/__init__.py
danieljohnson2: That's the preferred idiom.

lutris/gui/dialogs/__init__.py
duthils: You're absolutely right, thank you. I ended up removing those methods, since calling `super().run()` causes `mypy` to throw the same error that the method was supposed to solve:
```
lutris/gui/dialogs/__init__.py:102: error: Call to untyped function "run" in typed context  [no-untyped-call]
```

lutris/gui/dialogs/__init__.py
duthils: Thanks, I applied this on the whole file.

duthils:
Sorry, I can't reproduce the failing tests, I'll investigate later.

ItsAllAboutTheCode:
> Sorry, I can't reproduce the failing tests, I'll investigate later.

The unit test failure isn't due to your changes.
There is an hard to reproduce import issue with the test code when it comes to importing from `lutris/services/__init__.py`

I don't know why it fails because the imports look all valid, but it is probably an issue with some `sys.path` machinery inside `nose2`

ItsAllAboutTheCode:
> > Sorry, I can't reproduce the failing tests, I'll investigate later.
> 
> The unit test failure isn't due to your changes. There is an hard to reproduce import issue with the test code when it comes to importing from `lutris/services/__init__.py`
> 
> I don't know why it fails because the imports look all valid, but it is probably an issue with some `sys.path` machinery inside `nose2`

It looks like the `test_gog_cloud` hit this issue as well and used importlib to skip loading `services/__init__.py`

https://github.com/lutris/lutris/blob/f3c6e98d5e662d1771fc911c2fea08610236036a/tests/_test_gog_cloud.py#L27-L31

~~Maybe there is a circular import loop involving `services/__init__.py` that appears when import order is fixed?
I know a lot filesystems doesn't guarantee a strict order when walking through files in a directory (`iterdir` or `listdir`).
Perhaps the reason this is happening on the unit test machine is due to how the inodes are traversed on that node.
Chances are in most traversals the circular loop with `services/__init__.py` is avoided by one of the modules that causes the cycle being imported early?
But then in a small subset of directory traversal orders, that cycle occurs.~~

Edit: It turns out the issue is with all the importlib chicanery that [test_gog_cloud.py](https://github.com/lutris/lutris/blob/master/tests/_test_gog_cloud.py#L27-L51) and [test_gog_cloud_hooks.py](https://github.com/lutris/lutris/blob/master/tests/_test_gog_cloud_hooks.py#L13-L48) is doing.

I posted a fix at https://github.com/lutris/lutris/pull/6644

PR #6644, opened 2026-04-10
Fixed Lutris UnitTest when run on github

ItsAllAboutTheCode:
The primary causes of the issue is the manipulation of how the gog_cloud.py and gog_cloud_hooks.py was getting imported in UnitTest via the importlib mechanism. That seemed to have broken the resolution of paths to `lutris.services.*`. As the GTK issues that it was trying to side step has been fixed at the `tests` root level, the workaround is no longer needed.

Additional changes have been added to help harden any potentially UnitTest environment and  circular import issues.
* Updated the github unit test run to use a virtual env to install the package dependencies and run the test

### DIFF
--- a/.github/workflows/static.yml
+++ b/.github/workflows/static.yml
@@ -107,7 +107,11 @@ jobs:
       - name: Install Python dependencies
         run: |
           python -m pip install --upgrade pip
+          python -m venv .venv
+          . .venv/bin/activate
           make req-python
           make dev
       - name: Run Unit Test
-        run: nose2 --exclude-ci
+        run: |
+          . .venv/bin/activate
+          nose2 --exclude-ci
--- a/lutris/gui/views/store_item.py
+++ b/lutris/gui/views/store_item.py
@@ -5,7 +5,8 @@ import time
 from lutris.database import games
 from lutris.database.services import ServiceGameCollection
 from lutris.runners import get_runner_human_name
-from lutris.services import SERVICES, LutrisService
+from lutris.services import SERVICES
+from lutris.services.lutris import LutrisService
 from lutris.services.service_media import MediaPath
 from lutris.util.log import logger
 from lutris.util.strings import get_formatted_playtime, gtk_safe
--- a/lutris/services/__init__.py
+++ b/lutris/services/__init__.py
@@ -67,9 +67,11 @@ SERVICES = get_services()
 
 
 # Those services are not yet ready to be used
-WIP_SERVICES = {
-    "mame": MAMEService,
-}
+def get_wip_services() -> dict[str, "type[BaseService]"]:
+    return {"mame": MAMEService}
+
+
+WIP_SERVICES = get_wip_services()
 
 if os.environ.get("LUTRIS_ENABLE_ALL_SERVICES"):
     SERVICES.update(WIP_SERVICES)
--- a/lutris/util/test_config.py
+++ b/lutris/util/test_config.py
@@ -5,12 +5,12 @@ import gi
 gi.require_version("Gdk", "3.0")
 gi.require_version("Gtk", "3.0")
 
-from lutris import startup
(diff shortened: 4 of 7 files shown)


danieljohnson2:
I think I get the idea- remove circular dependencies via more local imports, then the hackery in the GOG Cloud test isn't needed and we get rid of it.

But ``get_services()`` doesn't make sense to me- that method is called to initialize ``SERVICES`` so all those modules are getting imported immediately anyway. This needs to be a lazy-initialized thing, doesn't it? Or how does this work?

ItsAllAboutTheCode:
> I think I get the idea- remove circular dependencies via more local imports, then the hackery in the GOG Cloud test isn't needed and we get rid of it.
> 
> But `get_services()` doesn't make sense to me- that method is called to initialize `SERVICES` so all those modules are getting imported immediately anyway. This needs to be a lazy-initialized thing, doesn't it? Or how does this work?

Technically it isn't needed in the case of just importing `import lutris.services`, since there is a call to `get_services()` inside of the file.

The case when a specific name such as `from lutris.services import DEFAULT_SERVICES`, the import mechanism can technically avoid initializing the `SERVICES` constant.
Though I think in the case of the cPython implementation it may just initialize all the names from the file.

ItsAllAboutTheCode:
Actually Python 3.15 will be adding the lazy import feature.
So the need for function imports will go away in that version.
https://peps.python.org/pep-0810/

danieljohnson2:
I'm be amazed if it could do that. I think this code will result in the same circular references at runtime- but maybe mypy will not 'see' it.

I think we need to ditch ``SERVICES`` and replace it with calls to ``get_services``; then adorn that with @cache_single which will cache the result after the first call.

After that, the ``startup.py`` changes are perhaps no longer needed.

And we need commenting about this - it's not idiomatic. But I do see the need for something like this.

danieljohnson2:
Lazy imports sounds neat but bear in mind we need to support older Python versions- back to 3.10 at the moment. It will be a while before we can require Python 3.15.

ItsAllAboutTheCode:
> I'm be amazed if it could do that. I think this code will result in the same circular references at runtime- but maybe mypy will not 'see' it.
> 
> I think we need to ditch `SERVICES` and replace it with calls to `get_services`; then adorn that with @cache_single which will cache the result after the first call.
> 
> After that, the `startup.py` changes are perhaps no longer needed.
> 
> And we need commenting about this - it's not idiomatic. But I do see the need for something like this.

A slimy way to do it without changing imports is to change `SERVICES` to be a getter property.

danieljohnson2:
That would be a great way to do it- but I did not think you could have properties at module level. How can you do that? I've often wanted to do that.

ItsAllAboutTheCode:
> That would be a great way to do it- but I did not think you could have properties at module level. How can you do that? I've often wanted to do that.

The` __getattr__` function can be overrideden for the module itself, to do it. 
(The module is actually a class instance)
ex.
```python
def __getattr__(name):
    if name == 'SERVICES':
        @cache_single
        def get_services() -> dict[str, "type[BaseService]"]:
            ...
        return get_services()
    raise AttributeError(f"No attribute {name}")
```

However, as I believe the current services `__init__.py` code doesn't have a circular import at this point, I will revert the function local import changes and avoid any complicated solutions.

danieljohnson2:
That ``__getattr__`` trick is kinda inky.

OK, just pull that if it isn't needed. I had thought it was essential.

I might just do it myself separately though, via the @cache_single. Importing everything everywhere isn't a great thing in itsef.

ItsAllAboutTheCode:
I just pushed the reversion to those import changes.

danieljohnson2:
OK, now it's ripping out a bunch of awful module molestation in the tests, tweak the CIk, and a few other minor fixes.

I'll take it.

PR #6525, opened 2026-03-08
feat: Optimize GOG download speeds with parallel connections and resume support

e88z4:
## Summary

This PR introduces three phases of download speed optimization for GOG game downloads, with improvements that also benefit all other services.

---

## Phase 1: Chunk size & cache protection (`feat: Optimize download speed with larger chunks and cache protection`)

- **64× larger download chunks**: Increased from 8 KB to 512 KB, matching heroic-gogdl's proven chunk size. Reduces syscall overhead significantly.
- **Connection pooling**: New optional `session` and `chunk_size` parameters on `Downloader` for per-download customization. Fully backward-compatible.
- **Download cache protection** (`lutris/util/download_cache.py`): Prevents deletion of in-progress or recently downloaded files before installation completes. Tracks lifecycle states: `DOWNLOADING → DOWNLOADED → INSTALLING → INSTALLED/FAILED`. Failed installations preserve files for 7 days for retry — critical for 40 GB+ GOG downloads.
- Cache protection is integrated into `interpreter.py` cleanup, download progress boxes, and install commands.

---

## Phase 2: Multi-connection parallel downloader (`feat: Add GOG multi-connection parallel downloader`)

- New `GOGDownloader` class (`lutris/util/gog_downloader.py`) uses multiple simultaneous HTTP Range requests (4 workers by default) to download large GOG installer files in parallel.
- HTTP Range support detection with automatic fallback to single-stream for small files or servers without Range support.
- Connection pooling via `requests.Session` with `HTTPAdapter`.
- Per-worker retry with exponential backoff (3 attempts).
- Thread-safe progress tracking for UI integration.
- `InstallerFile` gains a `downloader_class` property for service-specific downloader injection. `DownloadCollectionProgressBox` uses it automatically.

---

## Phase 3: Download resume with persistent byte-range tracking (`feat: Phase 3 - download resume with persistent byte range tracking`)

- New `lutris/util/download_progress.py`: Persistent progress tracker using atomic JSON sidecar files (`.progress`) alongside downloads.
- Interrupted GOG downloads (hibernate, crash, network error, Lutris restart) resume from where they left off instead of starting over.
- Atomic write (`os.replace`) for crash-safe progress file updates.
- On start: checks for `.progress` file; resumes remaining ranges only if `file_size` matches; starts fresh if mismatched or file missing; skips entirely if all ranges complete.

---

## Quality gates

- `ruff check`: all checks passed
- `ruff format`: all files formatted
- `mypy`: success, no issues
[...]

PR #6655, opened 2026-04-20
fix(download_cache): correct CacheState typo in is_safe_to_delete

upperpalaeolithic:
## Summary

`is_safe_to_delete` in `lutris/util/download_cache.py` lists `CacheState.DOWNLOADING` twice in the protection tuple, omitting `CacheState.DOWNLOADED`:

```python
# Before (line 148)
if state in (CacheState.DOWNLOADING, CacheState.DOWNLOADING, CacheState.INSTALLING):
```

This means files in the `DOWNLOADED` state (download complete, awaiting installation) are not protected and can be deleted before installation finishes.

## Fix

Replace the duplicate `DOWNLOADING` with `DOWNLOADED`:

```python
if state in (CacheState.DOWNLOADING, CacheState.DOWNLOADED, CacheState.INSTALLING):
```

## Notes

- Regression introduced in #6525
- One-line change, no behaviour change for `DOWNLOADING` or `INSTALLING` states

## Test plan

- [x] Download a GOG game to completion, then cancel/interrupt before installation — confirm the downloaded file is preserved
- [x] Verify `make test` passes

### DIFF
--- a/lutris/util/download_cache.py
+++ b/lutris/util/download_cache.py
@@ -145,7 +145,7 @@ def is_safe_to_delete(file_path: str) -> bool:
         return True
 
     # Active states - never delete
-    if state in (CacheState.DOWNLOADING, CacheState.DOWNLOADING, CacheState.INSTALLING):
+    if state in (CacheState.DOWNLOADING, CacheState.DOWNLOADED, CacheState.INSTALLING):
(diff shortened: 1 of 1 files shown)


danieljohnson2:
Good catch!

PR #6666, opened 2026-04-27
Add new installer commands: input_text and input_dir

### LINKED ISSUE #6652: Suggestion: new script commands
After making a bunch of installer scripts I have some suggestions for new commands.

1. input_dir: inputs a directory and saves it as a variable, same as input_disc w/o the "Detect disc" button.
```yml
input_dir:
  id: OPTIONAL_ID # saves as $INPUT_{id}
  requires: file # optional, same as insert_disc
  description: Select directory
```

2. input_text: inputs a single line of text and saves it as a variable, this would be useful for requesting a CD-KEY when bypassing uncooperative installers.

Optionally, the transform and validade fields can ensure the text is in a correct format.
```yml
input_text:
  id: OPTIONAL_ID # saves as $INPUT_{id}
  transform:
    - toupper
    - remove: -, :, ;
  validate:
    len: 25
    isnumeric: true
```

3. Multiple file sources: sometimes patches for old games aren't a single file but vary depending on the region/language. They could be added in that drop down menu in the file selection screen.
```yml
files:
- patch:
    - english: www.exemple.com/patches/patch-123-english.exe
    - url:  www.exemple.com/patches/patch-123-brazil.exe
      name: patch.exe
```

(examples are just a draft)

Would love to hear feedback on this.

red-rj:
Greetings! This PR adds a couple of new installer commands: `input_text` and `input_dir`.

- `input_dir` is the same as `insert_disc` minus the disc detection stuff, it's less error prone then using the later and asking "pls don't use autodetect 🥺". It also has a `requires` arg, but it's optional.
- `input_text` asks the user for a line of text, will be useful for asking for CD-KEYs when bypassing installers or for questions that the script author can't know the answer ahead of time.
 
both are modeled after `input_menu`, they take an `id` argument and sets the result as `$INPUT_{id}`

any feedback is welcomed!

Fixes #6652

### DIFF
--- a/lutris/gui/installerwindow.py
+++ b/lutris/gui/installerwindow.py
@@ -997,7 +997,7 @@ class InstallerWindow(ModelessDialog, DialogInstallUIDelegate, ScriptInterpreter
             browse_button.set_halign(Gtk.Align.CENTER)
             browse_button.set_valign(Gtk.Align.START)
             callback_data = {"callback": wrapped_callback, "requires": requires, "alias": alias}
-            browse_button.connect("clicked", self.on_browse_clicked2, callback_data)
+            browse_button.connect("clicked", self.on_browse_dir_clicked, callback_data)
             vbox.pack_start(browse_button, False, False, 40)
 
             self.stack.present_replacement_page("ask_for_dir", vbox)
@@ -1008,7 +1008,7 @@ class InstallerWindow(ModelessDialog, DialogInstallUIDelegate, ScriptInterpreter
         previous_page = self.stack.save_current_page()
         self.stack.jump_to_page(present_ask_for_dir_page)
 
-    def on_browse_clicked2(self, btn, callback_data):
+    def on_browse_dir_clicked(self, btn, callback_data):
         dialog = DirectoryDialog(_("Select directory"), parent=self)
         folder = dialog.folder
         callback = callback_data["callback"]
--- a/lutris/gui/installerwindow.py
+++ b/lutris/gui/installerwindow.py
@@ -976,7 +976,7 @@ class InstallerWindow(ModelessDialog, DialogInstallUIDelegate, ScriptInterpreter
     # This page asks the user for a directory.
 
     def load_ask_for_dir_page(self, alias, message, requires, callback):
-        def present_ask_for_disc_page():
+        def present_ask_for_dir_page():
             """Ask the user to do insert a CD-ROM."""
 
             def wrapped_callback(*args, **kwargs):
@@ -993,7 +993,7 @@ class InstallerWindow(ModelessDialog, DialogInstallUIDelegate, ScriptInterpreter
             vbox.pack_start(label, False, False, 0)
 
             browse_button = Gtk.Button(label=_("Browse…"))
-            browse_button.set_size_request(200,35)
+            browse_button.set_size_request(200, 35)
             browse_button.set_halign(Gtk.Align.CENTER)
             browse_button.set_valign(Gtk.Align.START)
             callback_data = {"callback": wrapped_callback, "requires": requires, "alias": alias}
@@ -1006,7 +1006,7 @@ class InstallerWindow(ModelessDialog, DialogInstallUIDelegate, ScriptInterpreter
             self.display_cancel_button()
 
         previous_page = self.stack.save_current_page()
-        self.stack.jump_to_page(present_ask_for_disc_page)
+        self.stack.jump_to_page(present_ask_for_dir_page)
 
     def on_browse_clicked2(self, btn, callback_data):
         dialog = DirectoryDialog(_("Select directory"), parent=self)
--- a/lutris/gui/installerwindow.py
+++ b/lutris/gui/installerwindow.py
@@ -879,16 +879,21 @@ class InstallerWindow(ModelessDialog, DialogInstallUIDelegate, ScriptInterpreter
                     # to run, so we'll go to error page.
                     self.load_error_page(err)
 
-            vbox = Gtk.Box(orientation=Gtk.Orientation.VERTICAL, spacing=3)
+            vbox = Gtk.Box(orientation=Gtk.Orientation.VERTICAL, spacing=6)
+            vbox.set_valign(Gtk.Align.START)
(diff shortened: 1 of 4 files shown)


red-rj:
draft for now cause I need to update `docs/installer.rst`

ItsAllAboutTheCode:
FYI: You are going to need to run `make ruff` to format the files to fix the Static Analysis checks.

lutris/installer/commands.py
```diff
@@ -31,6 +31,7 @@ class CommandsMixin:
 
     # pylint: disable=no-member
     installer: LutrisInstaller = NotImplemented
+    user_inputs: list[dict]
```
ItsAllAboutTheCode: This should be more explicit.
It can be a dictionary of string to string

```suggestion
    user_inputs: list[dict[str, str]]
```

lutris/installer/commands.py
```diff
@@ -230,6 +231,40 @@ def _on_input_menu_validated(self, alias, menu):
         else:
             raise RuntimeError("A required input option was not selected, so the installation can't continue.")
 
+    def input_text(self, data):
+        alias = f"INPUT_{data.get('id', 'TEXT')}"
+        message = data.get("message", _("Type bellow:"))
```
ItsAllAboutTheCode: Typo: bellow -> below
```suggestion
        message = data.get("message", _("Type below:"))
```

lutris/installer/commands.py
```diff
@@ -230,6 +231,40 @@ def _on_input_menu_validated(self, alias, menu):
...
+        text = entry.get_text()
+        self.user_inputs.append({"alias": alias, "value": text.strip()})
+        self._iter_commands()
+
+    def input_dir(self, data):
+        """Request user to provide a path to a directory"""
+        requires = data.get("requires")
+        alias = f"INPUT_{data.get('id', 'DIR')}"
+        message = data.get("message", _("Click Browse to select a directory."))
+        if requires:
```
ItsAllAboutTheCode, line 251: It may be beneficial to perform a sanitation check here of the "requires" file to make sure it is just a basename.
I imagine it would not make sense for the "requires" string to be a path with multiple components (i.e "foo/bar"), nor a relative path(i.e "../baz").
If the "requires" string isn't the correct format, then an error should be raised.

You could use the [os.path.basename](https://docs.python.org/3/library/os.path.html#os.path.basename) method to check if "requires" is only a basename.
````python
if '..' in requires:
     raise RuntimeError(('"requires" path "%s" cannot contain ".." as that could allow it to be outside of the selected directory') % requires)

basepath_requires = os.path.basename(requires)
if basepath_requires != requires:
    raise RuntimeError(('"requires" path must be a basename path "%s"') % basename_path)
```

lutris/gui/installerwindow.py
```diff
@@ -925,6 +971,51 @@ def on_browse_clicked(self, widget, callback_data):
...
+
+            self.stack.present_replacement_page("ask_for_dir", vbox)
+
+            vbox.show_all()
+            self.display_cancel_button()
+
+        previous_page = self.stack.save_current_page()
+        self.stack.jump_to_page(present_ask_for_disc_page)
+
+    def on_browse_clicked2(self, btn, callback_data):
```
ItsAllAboutTheCode: The name `on_browse_clicked2` is not a good descriptive name. Adding a number to end of function name can lead to confusion later on.
How about naming this function `on_browse_input_dir_clicked`?

That way it is distinguished from the existing `on_browse_clicked` method.

red-rj:
> FYI: You are going to need to run `make ruff` to format the files to fix the Static Analysis checks.

there's no `ruff` target, do you mean `make sc`?

ItsAllAboutTheCode:
> > FYI: You are going to need to run `make ruff` to format the files to fix the Static Analysis checks.
> 
> there's no `ruff` target, do you mean `make sc`?

Sorry, yes, `make sc` or `make format`.

lutris/installer/commands.py
red-rj: if that's the case then `_find_matching_disc()`  would also need that fix, cause I copied the logic from there

Edit: "requires" should absolutely allow checks in sub-folders IMO, the point is to find a file that ensures the dir/cdrom is correct [like so](https://lutris.net/games/install/39828/view).

as of `..`, I dunno, I could check if you insist

ItsAllAboutTheCode: Ok, then the basename check wouldn't be needed, however the `..` check should be added to ensure the installer isn't attempting to check for files that are outside of the directory the user provided for it.

red-rj:
I left `_check_requires()` as a function just in case you all want to use it for `_find_matching_disc()` in a future PR

lutris/installer/commands.py
```diff
@@ -265,6 +265,18 @@ def _on_dir_selected(self, _btn, requires, dir, alias):
...
 
+    @staticmethod
+    def _check_requires(dir, requires):
+        """Check requires clause so it doesn't reach outside the target directory w/ relative paths,
+        returns the validaded path"""
+        target = Path(dir)
+        assert target.is_dir(), "requires must be checked against a directory!"
+        requires = Path(requires)
+        abspath = str((target / requires).resolve())
+        if target.name not in abspath:
```
ItsAllAboutTheCode: You can use the pathlib.Path is_relative_to function function to make sure the path is within the directory
https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.is_relative_to

ItsAllAboutTheCode: ex.
```python
target = Path(dir).resolve()
requires_path = (target / requires).resolve()
if not requires_path.is_relative_to(target):
   raise RuntimeError("Illegal 'requires' reaches outside the target directory", requires)
```

red-rj:
hey any updates on this? it's been a while

ItsAllAboutTheCode:
> hey any updates on this? it's been a while

I am not affiliated with the Lutris project, so all I can do is review your code and provide (hopefully) good feedback on the changes.
It would be up to the project maintainer (`strycore`) to determine whether to merge the PR.

PR #6671, opened 2026-05-02
feat: firstly show installed runners in preferences -> runners list box 

cebem1nt:


danieljohnson2:
I like it- thought I won't merge it myself since it is a feature.

But two things- I think it's better to say ``runner_name`` when the parameter is a ``str`` rather than a ``Runner``- particularly without type annotations, as here. So I don't like that aspect of this change.

Also, it might be simpler to add a ``key`` argument to ``sorted``. That can be a lambda or function, and could return a tuple. But it's not trivial- you would have to first resolve all the names to runner objects before sorting by not-installed, then name, if you see what I mean.

If I were the boss, though, I'd be willing to take it as is.

cebem1nt:
> But two things- I think it's better to say `runner_name` when the parameter is a `str` rather than a `Runner`- particularly without type annotations, as here. So I don't like that aspect of this change.

Do you refer to this part of code?

```py
for runner in not_installed_runners: # <---
   self._add_runner_list_box_row(runner)
   runner_count += 1
```

It is a `Runner` instance. I modified the constructor of `RunnerBox` so it accepts the instance directly, instead of string name. Otherwise there would be to much of unnecessary instance creation.

Yeah, annotations really miss here. I can add them in this PR or another one


> Also, it might be simpler to add a key argument to sorted. That can be a lambda or function, and could return a tuple. But it's not trivial- you would have to first resolve all the names to runner objects before sorting by not-installed, then name, if you see what I mean.

Sure, I can rewrite it that way.

danieljohnson2:
I actually meant ``RunnerBox.__init__``, but I see that's taking a Runner object as well. So fair enough, I guess.

javi0guerrero4-cloud:
RunnerBox.__init__,

PR #6687, opened 2026-05-12
fix(security): 2 improvements across 2 files

SlncTrZ:
## Summary

fix(security): 2 improvements across 2 files

## Problem

**Severity**: `Medium` | **File**: `lutris/util/yaml.py:L30`

The function `write_yaml_to_file` creates a temporary file with a predictable name (original path + '.tmp') in the same directory, then renames it. This is vulnerable to symlink attacks: an attacker with write access to the directory could create a symlink to an arbitrary system file, causing the application to overwrite it when the rename occurs. No use of secure temporary file APIs (e.g., `tempfile.mkstemp`) is present.

## Solution

Replace the manual '.tmp' naming with `tempfile.mkstemp()` to create an unpredictable temporary file, then use `shutil.move()` or `os.replace()` for atomic replacement.

## Changes

- `lutris/util/yaml.py` (modified)
- `lutris/runners/commands/dosbox.py` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced"

### DIFF
--- a/lutris/runners/commands/dosbox.py
+++ b/lutris/runners/commands/dosbox.py
@@ -47,4 +47,5 @@ def makeconfig(path, drives, commands):
         for drive in drives:
             config_file.write('mount {} "{}"\n'.format(drive, drives[drive]))
         for command in commands:
-            config_file.write("{}\n".format(command))
+            sanitized = command.replace('\n', ' ')
+            config_file.write("{}\n".format(sanitized))
--- a/lutris/util/yaml.py
+++ b/lutris/util/yaml.py
@@ -1,6 +1,7 @@
 """Utility functions for YAML handling"""
 
 import os
+import tempfile
 from typing import Any
 
 import yaml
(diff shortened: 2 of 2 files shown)


ItsAllAboutTheCode:
Looks like another bot.
It created 20 repos and 69 commits in about a week which is not something a human would do.
All the commits have the same pattern in PR description layout (Summary, Problem, Solution, Changes, Testing).
Also the bot never does any kind of testing of the changes.

That being said, the `write_yaml_to_file` function should use the the Python tempfile API, so the change isn't wrong.

The `dosbox.py` change doesn't make any sense though.

danieljohnson2:
I think it's AI slop, and I don't think the symlink complaint makes sense- renaming the symlink doesn't damage it's target. It would just mean we would read the target and try to install it. An attacker could just go over that actual installer file instead of the .tmp anyway.

I am not buying it. Closing.

PR #6692, opened 2026-05-16
fix: correct transactional behavior and resource leak in db_cursor

Abzaek:
## Problem

`db_cursor.__exit__` had two issues:

1. `commit()` was called unconditionally — even when the with-block exited due to an exception. Partial operations were committed, breaking transactional semantics.

2. No `try/finally` — if `commit()` itself raised, `close()` was never called, leaking database connections.

## Fix

Only commit on normal exit (`exc_type is None`), rollback on exception, and always close in a `finally` block.

### DIFF
--- a/lutris/database/sql.py
+++ b/lutris/database/sql.py
@@ -25,9 +25,14 @@ class db_cursor(object):
         cursor = self.db_conn.cursor()
         return cursor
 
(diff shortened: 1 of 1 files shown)


danieljohnson2:
Looks legit!

PR #6696, opened 2026-05-19
steam: use appmanifest StateFlags to track game running state

### LINKED ISSUE #2926: Post-exit script runs prematurely if Steam is already open
**Describe the bug**

If I set up a post-exit script for a Steam game, it is executed almost immediately after the game starts, not after it finishes.

Funnily, the opposite happens if Steam is _not_ already running. The post-run command won't run until Steam itself is closed.

**Expected behavior**

Lutris should wait for the game to actually end before running the post-exit command.

**Steps to reproduce**

- Open Steam and leave the client running

- Open lutris, choose a Steam game, set a post-exit command for that game (something that opens a poop-up window or generates some log)

- Run the game, see the script being executed before the game ends

**Lutris debugging output**

DEBUG    2020-06-06 19:22:19,085 [application.launch:409]:Launching HuniePop (steam)
INFO     2020-06-06 19:22:19,087 [runner.use_runtime:241]:Runtime disabled by system configuration
DEBUG    2020-06-06 19:22:19,099 [xrandr._get_vidmodes:15]:Retrieving video modes from XrandR
DEBUG    2020-06-06 19:22:19,109 [system.execute:40]:Executing pgrep steam$
INFO     2020-06-06 19:22:19,193 [runner.use_runtime:241]:Runtime disabled by system configuration
DEBUG    2020-06-06 19:22:19,193 [game.configure_game:316]:Launching HuniePop
DEBUG    2020-06-06 19:22:19,193 [game.configure_game:317]:{
  "command": [
    "/usr/games/steam",
    "steam://rungameid/339800"
  ],
  "env": {}
}
INFO     2020-06-06 19:22:19,194 [runner.use_runtime:241]:Runtime disabled by system configuration
DEBUG    2020-06-06 19:22:19,202 [command.start:135]:SDL_VIDEO_FULLSCREEN_DISPLAY="off"
DEBUG    2020-06-06 19:22:19,202 [command.start:135]:game_name="HuniePop"
DEBUG    2020-06-06 19:22:19,202 [command.start:135]:PYTHONPATH="/home/bento/devel/lutris:/home/bento/devel/lutris/bin:/usr/lib/python38.zip:/usr/lib/python3.8:/usr/lib/python3.8/lib-dynload:/home/bento/.local/lib/python3.8/site-packages:/usr/local/lib/python3.8/dist-packages:/usr/lib/python3/dist-packages:/usr/lib/python3.8/dist-packages"
DEBUG    2020-06-06 19:22:19,202 [command.start:136]:/home/bento/devel/lutris/share/lutris/bin/lutris-wrapper HuniePop 0 0 /usr/games/steam steam://rungameid/339800
Running /usr/games/steam steam://rungameid/339800
Initial process has started with pid 122644
Game is considered started.
/home/bento/.local/share/Steam/steam.sh: line 114: VERSION_ID: unbound variable
/home/bento/.local/share/Steam/steam.sh: line 114: VERSION_ID: unbound variable
Running Steam on debian  64-bit
/home/bento/.local/share/Steam/steam.sh: line 114: VERSION_ID: unbound variable
STEAM_RUNTIME is enabled automatically
Pins up-to-date!
Steam client's requirements are satisfied
/home/bento/.local/share/Steam/ubuntu12_32/steam -nominidumps -nobreakpad steam://rungameid/339800
WARNING: setlocale('en_US.UTF-8') failed, using locale: 'C'. International characters may not work.
Game is considered exited.
Initial process has exited.
All children have exited.
Exit with returncode 0
DEBUG    2020-06-06 19:22:19,587 [command.on_stop:179]:The process has terminated with code 0
DEBUG    2020-06-06 19:22:19,587 [command.stop:261]:Management process looks dead already.
INFO     2020-06-06 19:22:19,587 [command.on_stop:185]:Full shutdown prevented
DEBUG    2020-06-06 19:22:21,212 [game.beat:563]:Game thread stopped
WARNING  2020-06-06 19:22:21,213 [game.on_game_quit:593]:Game still running (state: running)
INFO     2020-06-06 19:22:21,213 [game.stop:578]:Stopping HuniePop (steam)
DEBUG    2020-06-06 19:22:21,213 [command.stop:261]:Management process looks dead already.
INFO     2020-06-06 19:22:21,221 [game.on_game_quit:601]:Running post-exit command: /tmp/lutris-post-run.sh
DEBUG    2020-06-06 19:22:21,221 [command.start:135]:SDL_VIDEO_FULLSCREEN_DISPLAY="off"
DEBUG    2020-06-06 19:22:21,221 [command.start:135]:game_name="HuniePop"
DEBUG    2020-06-06 19:22:21,222 [command.start:135]:PYTHONPATH="/home/bento/devel/lutris:/home/bento/devel/lutris/bin:/usr/lib/python38.zip:/usr/lib/python3.8:/usr/lib/python3.8/lib-dynload:/home/bento/.local/lib/python3.8/site-packages:/usr/local/lib/python3.8/dist-packages:/usr/lib/python3/dist-packages:/usr/lib/python3.8/dist-packages"
DEBUG    2020-06-06 19:22:21,222 [command.start:136]:/home/bento/devel/lutris/share/lutris/bin/lutris-wrapper /tmp/lutris-post-run.sh 1 0 lutris-post-run.sh /tmp/lutris-post-run.sh
DEBUG    2020-06-06 19:22:21,227 [game.on_game_quit:614]:HuniePop stopped at Sat, 06 Jun 2020 19:22:21
DEBUG    2020-06-06 19:22:21,227 [game.save:236]:Saving HuniePop (steam)
DEBUG    2020-06-06 19:22:21,237 [lutriswindow.on_game_updated:762]:Updating game HuniePop (steam)
Running /tmp/lutris-post-run.sh
Initial process has started with pid 122862
Game is considered started.

rawrmonster17:
Steam games launch via non-blocking URI handlers (`steam://rungameid/` or `steam -applaunch`), which exit immediately after telling Steam to launch the game. The actual game process runs under Steam's PID tree and does not inherit `LUTRIS_GAME_UUID`, so the default PID tracking cannot detect it.

This causes `beat()` to see no running processes and call `on_game_quit()` within ~2 seconds, triggering post-exit scripts while the game is still launching.

**Fix:** detect Steam's per-game **reaper** process, whose command line contains `SteamLaunch AppId=<appid>`. It exists for the whole game lifetime (through shader pre-cache and play) and exits when the game stops — for native, Proton/Wine and Flatpak Steam alike.

- `get_steam_game_pids(appid)` — find the reaper PID(s) by scanning `/proc` cmdlines.
- `steam.filter_game_pids()` — track the reaper PID (primary), with the appmanifest `AppRunning` flag kept as a secondary signal where Steam updates it.
- `Runner.keep_game_alive()` hook (default `False`, no change for other runners) — lets the Steam runner tell `Game.beat()` to keep monitoring even though Lutris's launch thread has exited. Bridges the launch window (bounded `LAUNCH_GRACE_SECONDS`) and the whole game lifetime; without it, `beat()` treated the live game's PIDs as orphans and SIGTERM'd them.
- `steam.force_stop_game()` — SIGTERM the game's reaper (stops just this game), never the Steam client.
- Cache the resolved appmanifest path per appid (avoid re-walking steamapps dirs every heartbeat).
- `STEAM_DATA_DIRS` — add the current Flatpak Steam layout so Flatpak libraries are found.
- Unit tests for all of the above.

> **Why not the appmanifest `AppRunning` flag (the original approach)?** I measured it live at 5 Hz through full play sessions on **both** native and Flatpak Steam: `StateFlags` stays `Fully Installed` the entire time the game is running — `AppRunning` is never set on either. So the appmanifest mechanism was inert on both platforms; the reaper process is the reliable signal.

### Verification (end-to-end through the Lutris GUI, Team Fortress 2)
- ✅ **Flatpak Steam** — ~4 min session monitored correctly (`has run for 244 seconds`, clean quit).
- ✅ **Native Steam** — ~82 s session monitored correctly (`has run for 82 seconds`, clean quit), launched via `/usr/bin/steam`.

In both cases: no premature `on_game_quit`, no orphan-SIGTERM of the running game, correct quit detection at the real exit. Previously both produced `ran for 2 seconds, did it crash?`.

CI: green (ruff, mypy on 3.10, syntax, annotations, translations, unit tests).

Fixes: #2926

### DIFF
--- a/lutris/game.py
+++ b/lutris/game.py
@@ -996,6 +996,16 @@ class Game:
         runs_only_prelaunch = False
         if self.prelaunch_executor and self.prelaunch_executor.is_running and self.prelaunch_executor.game_process:
             runs_only_prelaunch = game_pids == {self.prelaunch_executor.game_process.pid}
+
+        # Some runners (e.g. Steam) launch the game out-of-process via a
+        # non-blocking handoff: game_thread exits within seconds and there is a
+        # brief window before the game's own process tree appears. Let the
+        # runner keep the monitor alive across that window and for the whole
+        # game lifetime, so beat() doesn't quit (or kill the game as "orphans")
+        # prematurely. Returns False for normal runners (unchanged behaviour).
+        if not runs_only_prelaunch and self.runner.keep_game_alive(game_pids, self.game_thread.is_running):
+            return True
+
         if runs_only_prelaunch or (not self.game_thread.is_running and not game_pids):
             logger.debug("Game thread stopped")
             self.on_game_quit()
--- a/lutris/runners/runner.py
+++ b/lutris/runners/runner.py
@@ -721,6 +721,19 @@ class Runner:  # pylint: disable=too-many-public-methods
         the caller will SIGKILL them (after a delay)."""
         kill_processes(signal.SIGTERM, game_pids)
 
+    def keep_game_alive(self, game_pids: Iterable[int], game_thread_running: bool) -> bool:
+        """Whether beat() should keep monitoring even though Lutris's own launch
+        process (game_thread) may have exited.
+
+        Most runners keep the game under game_thread, so this returns False and
+        beat() uses its normal logic. Runners that launch the game out-of-process
+        (e.g. Steam, which hands off to a Steam-managed process tree via a
+        non-blocking URI) override this to (a) bridge the brief window between
+        the launcher exiting and the game's process tree appearing, and (b) keep
+        monitoring for the whole game lifetime.
+        """
+        return False
+
     def extract_icon(self, game_slug: str) -> bool | None:
         """The config UI calls this to extract the game icon. Most runners do not
         support this and do nothing. This is not called if a custom icon is installed
--- a/lutris/runners/steam.py
+++ b/lutris/runners/steam.py
@@ -1,23 +1,97 @@
 """Steam for Linux runner"""
 
 import os
+import signal
+import time
+from collections.abc import Iterable
 from gettext import gettext as _
+from typing import Optional
 
 from lutris.exceptions import MissingGameExecutableError, UnavailableRunnerError
 from lutris.monitored_command import MonitoredCommand
 from lutris.runners import NonInstallableRunnerError
-from lutris.runners.runner import Runner
+from lutris.runners.runner import Runner, kill_processes
 from lutris.util import linux, system
 from lutris.util.log import logger
-from lutris.util.steam.appmanifest import get_appmanifest_from_appid, get_path_from_appmanifest
+from lutris.util.steam.appmanifest import AppManifest, get_appmanifest_from_appid, get_path_from_appmanifest
 from lutris.util.steam.config import get_default_acf, get_steam_dir, get_steamapps_dirs
 from lutris.util.steam.vdfutils import to_vdf
 from lutris.util.strings import split_arguments
 
+# Steam appmanifest StateFlags that mean the game should be treated as "still
+# alive" so beat() keeps polling. "AppRunning" is the normal running case. The
+# others cover the transient window between the Play click and AppRunning being
+# set, when Steam runs a launch-time update/validate ("Update Running",
+# "Validating", "Update Started") and the launcher process may already have
+# exited. Persistent library states that are NOT tied to an active launch
+# (e.g. "Update Required", "Update Paused", "Downloading") are deliberately
+# excluded, so a game with a pending background update isn't seen as running.
+STEAM_ALIVE_STATES = {
+    "AppRunning",
+    "Update Running",
+    "Update Started",
+    "Validating",
+}
 
-def get_steam_pid():
-    """Return pid of Steam process."""
-    return system.get_pid("steam$")
+# Default time (seconds) to keep monitoring after the launcher exits while
+# waiting for Steam's per-game reaper to appear, before concluding the launch
+# failed. Covers slow hand-off, shader pre-caching and first-launch
+# prerequisites. Overridable per-game via the "launch_grace_seconds" option.
+DEFAULT_LAUNCH_GRACE_SECONDS = 120
+
+
+def get_steam_pid() -> Optional[str]:
+    """Return pid of Steam process.
+
+    Steam is single-instance, so this resolves to at most one PID (the call
+    uses pgrep without ``multiple``).
+    """
+    pid = system.get_pid("steam$")
+    # get_pid() can return a list only with multiple=True, which we never pass;
+    # narrow the type so callers get a plain Optional[str].
+    if isinstance(pid, list):
+        return pid[0] if pid else None
+    return pid
+
+
+def get_steam_game_pids(appid: str) -> list[int]:
+    """Return the PIDs of the Steam "reaper" processes launched for a given appid.
+
+    When Steam launches a game it spawns a reaper process whose command line
+    contains "SteamLaunch AppId=<appid>". This reaper exists for the entire
+    lifetime of the game (through shader pre-caching and play) and exits when
+    the game stops, for native, Proton/Wine and Flatpak Steam alike.
+
+    This is far more reliable than the appmanifest "AppRunning" StateFlag, which
+    is not updated for Flatpak Steam installs, and it covers the launch window
+    (the reaper appears before the game window does), so beat() does not call
+    on_game_quit() prematurely.
+    """
+    if not appid:
+        return []
+    # The reaper cmdline looks like: "reaper SteamLaunch AppId=440 -- ...".
+    # Match "AppId=<appid>" followed by a space or end of string so that appid
+    # 440 does not also match a reaper for 4400 / 44012 (substring collision).
+    prefix = "SteamLaunch AppId=%s" % appid
+    pids = []
+    for pid in os.listdir("/proc"):
+        if not pid.isdigit():
+            continue
+        try:
+            with open(os.path.join("/proc", pid, "cmdline"), "rb") as cmdline_file:
+                cmdline = cmdline_file.read().replace(b"\x00", b" ").decode("utf-8", "replace")
+        except OSError:
+            continue
+        if "reaper" not in cmdline:
+            continue
+        idx = cmdline.find(prefix)
+        if idx == -1:
+            continue
+        # Character immediately after the appid must be a space or end-of-string.
+        after = cmdline[idx + len(prefix) : idx + len(prefix) + 1]
+        if after in ("", " "):
+            pids.append(int(pid))
+    return pids
 
 
 def is_running():
@@ -87,11 +161,70 @@ class steam(Runner):
             "advanced": True,
             "help": _("Extra command line arguments used when launching Steam"),
         },
+        {
+            "option": "launch_grace_seconds",
+            "type": "string",
+            "label": _("Launch grace period (seconds)"),
+            "advanced": True,
+            "help": _(
+                "How long Lutris keeps monitoring after the Steam launcher exits "
+                "while waiting for the game to appear, before assuming the launch "
+                "failed. Covers shader pre-caching and first-launch prerequisites. "
+                "Leave blank to use the default (%s seconds)."
+            )
+            % DEFAULT_LAUNCH_GRACE_SECONDS,
+        },
     ]
     system_options_override = [
         {"option": "disable_runtime", "default": True},
         {"option": "gamemode", "default": False},
     ]
+
+    # Cache the reaper-PID scan for a fraction of a heartbeat so the two calls
+    # in a single Game.beat() (filter_game_pids + keep_game_alive) only walk
+    # /proc once. Kept well below HEARTBEAT_DELAY so each beat re-scans.
+    _REAPER_CACHE_TTL = 1.0
+
+    def __init__(self, config=None):
+        super().__init__(config)
+        # Cache for the resolved appmanifest path so beat() doesn't re-walk
+        # every steamapps dir on each heartbeat (see _get_cached_appmanifest).
+        self._cached_appmanifest_appid: Optional[str] = None
+        self._cached_appmanifest_path: Optional[str] = None
+        # Launch-window tracking for keep_game_alive().
+        self._launch_monitor_start: Optional[float] = None
+        self._steam_reaper_seen: bool = False
+        # Per-beat cache of the reaper PID scan (see _get_reaper_pids).
+        self._reaper_pids_cache: list[int] = []
+        self._reaper_pids_cache_at: float = 0.0
+
+    @property
+    def launch_grace_seconds(self) -> float:
+        """Launch grace period, from the per-game option if set and valid,
+        otherwise DEFAULT_LAUNCH_GRACE_SECONDS."""
+        raw = self.runner_config.get("launch_grace_seconds")
+        if raw not in (None, ""):
+            try:
+                value = float(raw)  # type: ignore[arg-type]
+            except (TypeError, ValueError):
+                value = None
+            if value is not None and value > 0:
+                return value
+            logger.warning("Invalid launch_grace_seconds %r; using default", raw)
+        return DEFAULT_LAUNCH_GRACE_SECONDS
+
+    def _get_reaper_pids(self) -> list[int]:
+        """Reaper PIDs for this appid, cached briefly so the two calls within a
+        single Game.beat() (filter_game_pids + keep_game_alive) share one
+        /proc scan instead of walking it twice."""
+        if not self.appid:
+            return []
+        now = time.monotonic()
+        if now - self._reaper_pids_cache_at < self._REAPER_CACHE_TTL:
+            return self._reaper_pids_cache
+        self._reaper_pids_cache = get_steam_game_pids(self.appid)
+        self._reaper_pids_cache_at = now
+        return self._reaper_pids_cache
 
     @property
     def runnable_alone(self):
@@ -201,6 +334,12 @@ class steam(Runner):
         return {"command": self.launch_args, "env": self.get_env()}
 
     def play(self):
+        # Reset launch-window tracking for this run (the runner instance may be
+        # reused across launches); see keep_game_alive().
+        self._launch_monitor_start = None
+        self._steam_reaper_seen = False
+        self._reaper_pids_cache = []
+        self._reaper_pids_cache_at = 0.0
         game_args = self.game_config.get("args") or ""
 
         binary_path = self.game_config.get("steamless_binary")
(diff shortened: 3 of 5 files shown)


danieljohnson2:
I looked at this and it looks right to me, but I don't know Steam well. I asked Claude, and it points out that this won't interact well with Steam if it is kept running in your tray. I'll have it post its review.

danieljohnson2:
*Beep boop!* 🤖 Robot reviewer reporting in. Two notable concerns plus some smaller things.

### Significant concern: the fix likely doesn't trigger in the common case

The new branch is gated on `steam_pid in candidate_pids`, but `candidate_pids` is computed in `Game.get_new_pids()` as `current_pids - prelaunch_pids`, and `prelaunch_pids` is snapshotted right before launch. So:

- If Steam was already running when the user clicked Play (Steam tray, autostart, etc. — the very common case), `steam_pid ∈ prelaunch_pids` and is therefore never in `candidate_pids`. The new branch is a no-op and `beat()` still fires `on_game_quit()` early.
- If Steam was cold-started by Lutris, `game_thread` *is* the Steam process, so `game_thread.is_running` already stays `True` for Steam's lifetime — the premature-exit bug wasn't really hitting in that path to begin with.

Did you test the "Steam already running in tray before Play" scenario? That's the path I'd expect to be reproducing #2926 against. I think the `steam_pid in candidate_pids` guard probably needs to be dropped (or replaced with something less restrictive) for this to actually help users.

### Risk: orphan-cleanup branch can SIGTERM Steam

In `Game.beat()`:

```python
if not self.game_thread.is_running and game_pids:
    self.runner.force_stop_game(game_pids)   # SIGTERM by default
    self.on_game_quit()
```

If you remove the `candidate_pids` guard so the fix actually kicks in for the "Steam already running" case, here's what happens next: `steam steam://rungameid/X` IPCs into the running Steam and exits in ~1 s, so `game_thread.is_running` flips to `False`. The very next beat sees `not is_running and game_pids == {steam_pid}` → `force_stop_game({steam_pid})` → SIGTERM to the user's Steam client. Bad.

The default `Runner.force_stop_game()` blindly `kill_processes(SIGTERM, ...)`. You'd probably want to override `force_stop_game()` on the Steam runner too, so it either no-ops or quits via `steam steam://...` / appid rather than signalling the Steam PID. (Or return a sentinel from `filter_game_pids` that keeps `beat()` polling without binding to a real PID.)

### Smaller stuff

- **I/O every heartbeat.** `get_appmanifest()` walks every steamapps dir and re-parses VDF on every beat (every 2 s per `HEARTBEAT_DELAY`). Worth caching the resolved appmanifest path once at `play()` and only re-reading `StateFlags` on subsequent beats.
- **Pre-`AppRunning` window.** Between Play click and `AppRunning` being set (Steam starting, "Update Running", "Validating", first-launch prereqs), the new branch returns `pids` unchanged. If the user hits this window while `game_thread` has already exited, premature `on_game_quit` still fires. Treating `Update Running`/`Validating`/`Update Started` as "still alive" would close that window.
- **Flatpak Steam.** `get_steam_pid()` uses `system.get_pid("steam$")` against host `/proc`. Under Flatpak the cmdline may not match — worth verifying given `play()` has a Flatpak branch.
- **No tests** for `filter_game_pids` on the Steam runner. A small unit test with a stubbed appmanifest and `get_steam_pid` would guard against regressions.
- **Types.** The base `filter_game_pids` is annotated `(candidate_pids: Iterable[int], game_uuid: str, game_folder: str) -> set[int]`; the override drops the annotations. Adding them keeps mypy happy.
- **Comment.** The inline `# Steam reports the game is still running. Add the Steam process itself as a tracked PID so beat() keeps polling.` is mostly *what*; the *why* (heartbeat needs non-empty PIDs to keep polling) is the load-bearing half.

### Style / conventions

- Pattern matches `wine.py:filter_game_pids` (override + `super()`). ✅
- 17/0 diff, no new imports — small blast radius. ✅

### TL;DR

Concept is right, but I think the `steam_pid in candidate_pids` gate makes this inert for the most common reproduction of #2926, and removing the gate exposes a SIGTERM-Steam footgun via the orphan-cleanup branch. Worth a second pass before merging.

*Beep boop, out.* 🤖

🤖 _Comment generated with [Claude Code](https://claude.ai/code)_

rawrmonster17:
Thanks for the detailed review — the bot was right on both counts, and digging in revealed the approach needed to change. I've reworked the PR.

**Key finding:** the appmanifest `"AppRunning"` StateFlag this PR originally relied on is **never set for Flatpak Steam**. I verified this live by polling `appmanifest_440.acf` at 5 Hz through a full Team Fortress 2 launch→play→quit cycle — the flag stayed `Fully Installed` the entire time while the game was clearly running (`LastPlayed` updated, processes alive). So the original mechanism was inert on Flatpak.

**New approach — detect Steam's per-game reaper process.** When Steam launches a game it spawns a `reaper` process whose command line contains `SteamLaunch AppId=<appid>`. It lives for the whole game lifetime (through shader pre-cache and play) and exits when the game stops — for native, Proton/Wine and Flatpak alike. This is what `get_steam_game_pids()` now keys on.

**Addressing the review's two main concerns:**
- *"Fix doesn't trigger in the common case (Steam already running)"* — the new detection doesn't depend on `candidate_pids`/the prelaunch snapshot at all; it independently looks for the reaper, so the "Steam already in tray" case works.
- *"Orphan-cleanup branch can SIGTERM Steam"* — `Game.beat()` treated the still-running game's PIDs as orphans and killed them ~2s in (because the `steam://rungameid` launcher exits immediately, so `game_thread.is_running` goes False). Fixed with a small `Runner.keep_game_alive()` hook (default `False`, no behaviour change for other runners) that lets the Steam runner tell `beat()` to keep monitoring. It bridges the launch window (bounded `LAUNCH_GRACE_SECONDS` so a genuinely failed launch still ends) and the whole game lifetime. `force_stop_game()` now SIGTERMs the game's reaper, never the Steam client.

Also added the current Flatpak Steam library path to `STEAM_DATA_DIRS` (`~/.var/app/com.valvesoftware.Steam/.local/share/Steam`), unit tests, and the appmanifest-path caching the bot suggested.

**Verification status:**
- ✅ **Flatpak Steam — verified end-to-end through the Lutris GUI.** A ~4-minute TF2 session is now monitored correctly: `has run for 244 seconds` with a clean quit, vs. the previous `ran for 2 seconds, did it crash?`. No premature `on_game_quit`, no SIGTERM of the game/Steam.
- 🔄 **Native (non-Flatpak) Steam — in progress.** I'm setting up a native Steam install to verify that path too before considering this done; I'll follow up here once it's confirmed.

Happy to adjust the `game.py` hook if you'd prefer a different shape — I kept it minimal and opt-in.

rawrmonster17:
Follow-up: **native (non-Flatpak) Steam is now verified too.**

I installed native Steam alongside the Flatpak one and ran the same end-to-end test through the Lutris GUI with Team Fortress 2:

- Launched via `/usr/bin/steam steam://rungameid/440`, the launcher exited immediately as expected.
- Lutris monitored correctly for the whole session and only called `on_game_quit()` when I actually quit: `Team Fortress 2 (steam) has run for 82 seconds` with a clean stop — no premature quit, no orphan-SIGTERM.

Notably, I confirmed by polling the appmanifest at 5 Hz that **native Steam also never sets the `AppRunning` StateFlag** during a running game (it stays `Fully Installed`), same as Flatpak — which is why the reaper-process approach is needed rather than the appmanifest flag.

So both code paths are now verified live:
- ✅ Flatpak Steam — 244 s session, clean quit
- ✅ Native Steam — 82 s session, clean quit

No code changes were needed between the two — the same `SteamLaunch AppId=<appid>` reaper detection covers both. I think this is ready for another look whenever you have time.

danieljohnson2:
*Beep boop!* 🤖 Robot reviewer back for round two. I went through the rework against my earlier concerns — short version: substantively resolved, not papered over. Longer version below.

## Original concerns vs. the new diff

- **"Fix doesn't trigger when Steam is already running"** — ✅ resolved. Detection no longer depends on `candidate_pids`/the prelaunch snapshot at all. `get_steam_game_pids()` scans `/proc` for the per-game reaper (`SteamLaunch AppId=<appid>` + `reaper`), so the "Steam already in tray" case works.
- **"Orphan-cleanup branch can SIGTERM Steam"** — ✅ resolved on two layers:
  - The new `Runner.keep_game_alive()` hook (default `False`, no behaviour change for other runners) lets the Steam runner tell `beat()` to keep monitoring while the reaper is alive or inside the launch grace window, so the orphan-kill branch isn't reached in the normal path.
  - `steam.force_stop_game()` explicitly filters Steam's own PID out of the kill list and SIGTERMs the reaper instead. Even if `force_stop_game` is reached, Steam can't be killed.
- **"I/O every heartbeat"** — ✅ `_get_cached_appmanifest()` resolves the path once per appid and re-reads only `StateFlags` thereafter.
- **"Pre-`AppRunning` window"** — ✅ `STEAM_ALIVE_STATES` includes `AppRunning`, `Update Running`, `Update Started`, `Validating`, with persistent-but-not-running states (`Update Required`, `Downloading`, etc.) deliberately excluded — exactly the split worth making.
- **"Flatpak Steam"** — ✅ and the digging here is the most valuable part of the rework. The author verified live (5 Hz polling through a full TF2 session) that **Flatpak Steam never updates `AppRunning` at all** — which is what made v1 inert there. That finding drove the pivot to reaper-process detection. Flatpak library path also added to `STEAM_DATA_DIRS`.
- **"No tests"** — ✅ 21 tests across 7 classes, covering reaper detection, all four `keep_game_alive` paths (launcher-alive / reaper-alive / launch grace / post-quit), `force_stop_game` sparing Steam, and the appmanifest path caching.
- **"Types / comments"** — ✅ override has full annotations; docstrings lead with *why* (failure mode of base behaviour, why the reaper is reliable, why appmanifest is a secondary signal).

End-to-end verification on both Flatpak (244 s) and native Steam (82 s) through the actual Lutris GUI, replacing the previous "ran for 2 seconds, did it crash?" with clean monitored sessions, is the right kind of evidence for this bug.

## Small issues introduced by the rework

All minor — nothing I'd call blocking — but worth flagging:

1. **Prefix-match false positive in `get_steam_game_pids`.** The needle is `\"SteamLaunch AppId=%s\" % appid` matched as a plain substring. If the user's appid is `440`, this also substring-matches a reaper for appid `4400` or `44012`. Real-world collision risk is low but real. Easy fix: append a trailing space to the needle (real reaper cmdlines look like `SteamLaunch AppId=440 --`), or match with a word boundary. This is the one item I'd actually ask for before merge.
2. **`/proc` walked twice per beat.** `filter_game_pids()` and `keep_game_alive()` both call `get_steam_game_pids()`. Walking `/proc` for a few hundred PIDs twice every 2 s isn't terrible, but caching the result for the duration of one beat (or passing it through from `filter_game_pids` to `keep_game_alive`) would halve it cheaply.
3. **`LAUNCH_GRACE_SECONDS = 120` is generous.** Two minutes is fine for shader pre-cache or first-launch prereqs, but on a failed launch the user waits 2 min for Lutris to realize. Not wrong, just worth flagging — could become a runner option later if anyone complains.
4. **`_get_steam_pid_int` has a dead `isinstance(..., list)` branch.** `get_steam_pid()` returns `system.get_pid(\"steam$\")` which is single-valued, never a list. Defensive but unreachable — fine to leave, fine to drop.
5. **Test fixture bypasses `__init__`.** `steam.steam.__new__(steam.steam)` then manually sets cache attributes. If `__init__` grows new state, every test fixture silently misses it. Constructing with a mock `LutrisConfig` would be sturdier.
6. **`Optional[str]` vs `str | None`.** The codebase mixes both. The new code uses `Optional[T]` throughout. Cosmetic.

## TL;DR

This is a much stronger PR than v1. The two substantive concerns I raised are genuinely resolved (and the pivot was driven by the author's own Flatpak verification — good engineering). The reaper-process approach is the right primitive here. I'd love to see the prefix-match guard tightened before merge; everything else is comfortably maintainer-discretion follow-up territory.

*Beep boop, out.* 🤖

🤖 _Comment generated with [Claude Code](https://claude.ai/code)_

rawrmonster17:
Pushed a follow-up (`8acf3b20f`) handling the remaining items from your round-two review:

**3. `LAUNCH_GRACE_SECONDS` is now configurable.** It's `DEFAULT_LAUNCH_GRACE_SECONDS = 120` (module constant) and overridable per-game via a new advanced runner option, `launch_grace_seconds`. Invalid/blank values fall back to the default. So anyone who hits a slow failed-launch wait can shorten it without a code change.

**4. Dead `isinstance(..., list)` branch removed — at the source.** Rather than just dropping the guard in `_get_steam_pid_int` (which would re-introduce a mypy union error, since `system.get_pid()` is typed `str | list[str] | None`), I annotated `get_steam_pid()` to return `Optional[str]` and do the single-value narrowing there. Callers now get a plain `Optional[str]` and `_get_steam_pid_int` is clean.

**5. Test fixture no longer bypasses `__init__`.** The `make_runner` fixture now constructs the runner through its real `__init__` (with config access stubbed so no DB/disk is touched), so any new instance state added to `__init__` can't be silently missed by tests.

Test count is now 36, including coverage for the configurable grace period (default / configured / invalid-fallback) and the `get_steam_pid` narrowing. CI is green (ruff, mypy 3.10, syntax, annotations, translations, unit tests).

I also re-ran the end-to-end GUI test on this commit before pushing — TF2 monitored correctly for a 66-second session with a clean quit (`has run for 66 seconds`), so the config/typing changes didn't regress the live behaviour.

danieljohnson2:
*Beep boop!* 🤖 Robot reviewer back for round three. Quick one this time — the rework addresses every item from my v2 review with a fix matched to its severity. Walking through each:

## v2 concerns vs. the new diff

1. **Prefix-match false positive in `get_steam_game_pids`** *(the one I'd "ask for before merge")* — ✅ Fixed properly. `cmdline.find(prefix)` plus a check that the next character is `\" \"` or end-of-string. Even better: there's a dedicated `test_does_not_match_appid_prefix_collision` covering appid `440` vs reapers for `4400` and `44012`, plus a `test_requires_reaper_in_cmdline` for the "non-reaper process mentioning appid" case.
2. **`/proc` walked twice per beat** — ✅ Fixed cleanly. New `_get_reaper_pids()` with `_REAPER_CACHE_TTL = 1.0` s, cache reset in `play()` so a re-launch starts fresh. `filter_game_pids` and `keep_game_alive` both go through the cached version; `force_stop_game` correctly bypasses (it needs fresh data and runs rarely).
3. **`LAUNCH_GRACE_SECONDS` should be configurable** — ✅ Now an advanced runner option (`launch_grace_seconds`) with `DEFAULT_LAUNCH_GRACE_SECONDS = 120` and validation: empty / None / non-numeric / negative all fall back to the default. Tests cover the default, configured value, invalid input, and negative input.
4. **`_get_steam_pid_int` dead list-handling branch** — ✅ Refactored the right way: the list narrowing was hoisted up into `get_steam_pid()` itself (which now narrows `system.get_pid()`'s `str | list | None` return to `Optional[str]` at the source), with a dedicated `GetSteamPidTest` class covering str-passthrough, None, list-narrowed-to-first and empty-list. `_get_steam_pid_int` is now a clean small `int()` cast.
5. **Test fixture bypassed `__init__`** — ✅ Fixed. `make_runner()` now does `steam.steam()` (real `__init__`) and uses `PropertyMock` for `game_config` and `runner_config`, so future `__init__` state additions can't be silently missed by the fixture.
6. **`Optional[str]` vs `str | None`** — Not changed; I'd marked this cosmetic, no objection.

End-to-end verification on both Flatpak (~4 min TF2 session, clean quit) and native Steam (multiple sessions, e.g. "has run for 66 seconds", clean quit) is in the commit message — same solid approach as before.

## Tiny new things I noticed

Nothing I'd call out as a blocker:

- The negative-`launch_grace_seconds` path silently falls back without a warning, while the parse-error path warns. Minor inconsistency — easy to align by logging in both branches if you want.
- `test_quit_after_reaper_seen_then_gone` manually expires the per-beat cache via `runner._reaper_pids_cache_at = 0.0` to simulate a fresh beat. Slightly leaky abstraction in the test, but pragmatic.
- The option uses `\"type\": \"string\"` for a numeric value, which is just the Lutris runner-option style.

## TL;DR

This was the cleanest "address all review feedback" rework I've seen in a while. Every concern got a fix matched to its severity (anchored substring check for the bug, cached scan for the perf nit, configurable option for the value-judgement concern, type narrowing at the source for the refactor opportunity), and test coverage was expanded to lock each one in.

If approving were my call, I'd approve. It isn't, so I'll leave that to the maintainer — but from this robot's seat, this looks ready.

*Beep boop, out.* 🤖

🤖 _Comment generated with [Claude Code](https://claude.ai/code)_

rawrmonster17:
Pushed a small follow-up (`a2cc28b`) for the one remaining item from your round-three review:

**Invalid `launch_grace_seconds` now warns consistently.** Previously a non-numeric value (`"abc"`) logged a warning, but a non-positive numeric one (`-5`, `0`) parsed fine, failed the `> 0` check, and fell back to the default *silently* — same user-facing outcome ("my value was ignored"), inconsistent feedback. The property is restructured so both invalid cases emit the same warning; valid positive values still return early with no warning. Extended `test_grace_default_and_invalid_fallback` to assert the warning fires for both the non-numeric and the negative case.

Squashed into the single commit. ruff, mypy (3.10) and the unit tests are green locally. This is a logic-only change to the warning path with no behavioural change to grace handling, so I didn't re-run the live GUI session for it — it's covered by the unit tests.

The other two round-three notes I left as-is: the per-beat cache expiry in `test_quit_after_reaper_seen_then_gone` (you flagged it as pragmatic), and `"type": "string"` on the option (that's the Lutris runner-option convention — there's no numeric option type). Happy to change either if you'd prefer.

danieljohnson2:
*Beep boop!* 🤖 Round four, brief one.

Confirmed `a2cc28b` addresses the one remaining minor item from my v3 review: both invalid `launch_grace_seconds` paths (non-numeric and non-positive) now converge on a single `logger.warning(...)` call, and the tests wrap both cases in `assertLogs(..., level=\"WARNING\")` to lock the behavior in. Path coverage checks out:

- `None` / `""` → default, no warning (legitimate "use default") ✅
- `\"120\"` → returns 120, no warning ✅
- `\"notanumber\"` → warns, returns default ✅
- `\"-5\"`, `\"0\"` → warns, returns default ✅

The `value is not None` guard before `value > 0` is correctly necessary (would otherwise raise `TypeError` on the non-numeric path). No new issues introduced.

Nothing else from me — LGTM from this robot's seat. Over to the maintainer.

*Beep boop, out.* 🤖

🤖 _Comment generated with [Claude Code](https://claude.ai/code)_

rawrmonster17:
Any update on this one?

danieljohnson2:
The update: I forgot about it.

But I'll merge it now, anyway.

PR #6699, opened 2026-05-19
feat: add GitHub Actions CI workflow

dashitongzhi:
Add Python CI workflow for automated testing and linting


danieljohnson2:
This is garbage. If you were paying attention you might notice we don't support Python 3.9, since that is EOL. And we don't need two PRs for this when we already do it, @dashitongzhi.

PR #6720, opened 2026-06-03
Add virtual all sources game view

ControlDrift:
Adds a virtual "All Sources" sidebar entry that aggregates games from enabled game sources into one searchable view.

image

This keeps existing source-specific views unchanged and does not merge metadata. Local installed service entries are skipped when the same service app id is already present in the service library, avoiding duplicate entries for the same source.

Tested with local Lutris data including Steam, Epic, GOG, Humble Bundle, EA App, and Ubisoft service games. Verified search, service media fallback, and existing launch/install action routing for mixed local and service entries.

### DIFF
--- a/lutris/gui/config/services_box.py
+++ b/lutris/gui/config/services_box.py
@@ -5,7 +5,7 @@ from gi.repository import GObject, Gtk
 from lutris import settings
 from lutris.gui.config.base_config_box import BaseConfigBox
 from lutris.gui.widgets.scaled_image import ScaledImage
-from lutris.services import SERVICES
+from lutris.services import IN_GAMES_VIEW_SETTING_SUFFIX, SERVICES, is_service_in_games_view
 
 
 class ServicesBox(BaseConfigBox):
@@ -44,10 +44,10 @@ class ServicesBox(BaseConfigBox):
             height_request=32,
             visible=True,
         )
-        in_games_view_key = service_key + "_in_games_view"
+        in_games_view_key = service_key + IN_GAMES_VIEW_SETTING_SUFFIX
         service = SERVICES[service_key]
         is_active = settings.read_bool_setting(service_key, section="services")
-        is_in_games_view = settings.read_bool_setting(in_games_view_key, default=True, section="services")
+        is_in_games_view = is_service_in_games_view(service_key)
 
         icon = ScaledImage.get_runtime_icon_image(
             service.icon, service.id, scale_factor=self.get_scale_factor(), visible=True
--- a/lutris/gui/lutriswindow.py
+++ b/lutris/gui/lutriswindow.py
@@ -141,12 +141,13 @@ class LutrisWindow(Gtk.ApplicationWindow, DialogLaunchUIDelegate, DialogInstallU
         self.views = {}
 
         self.dynamic_categories_game_factories: dict[str, Callable[[], list]] = {
+            "all_sources": self.get_all_source_games,
             "recent": self.get_recent_games,
             "missing": self.get_missing_games,
             "running": self.get_running_games,
             ".uncategorized": self.get_uncategorized_games,
         }
-        self.sortable_dynamic_categories = {".uncategorized", "missing", "running"}
+        self.sortable_dynamic_categories = {"all_sources", ".uncategorized", "missing", "running"}
 
         self.accelerators = Gtk.AccelGroup()
         self.add_accel_group(self.accelerators)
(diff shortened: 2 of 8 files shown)


danieljohnson2:
OK, I've been over this and I've set Claude on it. For once, we don't find the same things.

My concern here is architectural; that is, it's a lot of spot changes needed to handle your synthetic 'view ids' in the form ``service:<<service id>>:<<appid>``. That has got to be a bug factory. You are treating this new thing as a dynamic category, but also it is like a service with its own ``appids``. So what could we do about it?

1. The minimal answer is to do what you are doing, but with type annotations and helper functions. You do have some helpers (``parse_service_view_id``) but they aren't used consistently. Type annotations could make it clearer where magic 'view ids' are intended. By mypy is *very* poor at enforcing type safety for us, so I don't favor this.

2. Instead, we could go dynamic- that would fit Python better. Have a GameID *class* with subclasses maybe, which could represent different kinds of ID. Misuse would result in runtime errors. This is a more wide-ranging change, but would clarify Lutris's existing system of game-ids and appids (or so I would hope). Merging this close to release would be risky though.

3. Alternately, forget dynamic categories and make a new "all services" service. This new service would delegate everything it does to the other services, but it would then use your new ids as its own appids (ie, adjust the db_games dictionaries). This is a rewrite of the feature, but could offer more functionality- like reloading all services together.

Claude has much more detail-oriented stuff instead. I'll send him over next.

danieljohnson2:
Beep boop! 🤖 Took a pass through this — the feature is well scoped and the wiring (sidebar row → factory → activation/selection → media fallback) looks correct end-to-end. A few things worth tightening before merge:

### Correctness / efficiency
- **`ServiceGameCollection.get_for_service(service_id)` is called twice per service** in `get_all_source_games` — once to build the appid set, again to materialize the dicts. On large Steam/GOG libraries that's an avoidable hit. Cache the list and derive the set from it.
- **`filter_games` runs twice on local games**: once mid-build, then again on the combined list at the end. Just filter at the end.
- **`services.get_enabled_services()` is iterated twice** in the same function — it walks `SERVICES.items()` and reads a setting per entry. Bind it once.
- The disabled-services computation now lives in three places (here, `get_games_from_filters`, `services_box.py`). Worth extracting a small helper.

### Maintainability
- **Parsing of `service:<id>:<appid>` is duplicated** between `lutriswindow.parse_service_view_id` and `views/base.py._get_games_by_ids`. Import the helper so the wire format only lives in one place.
- `_view_id` short-circuits `StoreItem.id` for any dict carrying that key. Works, but a one-line comment in `store_item.py` pointing back to where it's set would help future readers.
- The `service_id or appid` fallback in `get_media_paths` is a genuine fix for service-game dicts that lack `service_id`, but it's a quiet behavior change unrelated to all_sources — worth calling out in the commit message.
- No type hints on the new public methods, unlike newer code in this file.

### Minor
- `split(":", 2)` is robust to colons in appids, but if anything ever stores a non-service id starting with `service:`, the selection/activation paths will misroute. Tiny risk in practice given numeric ids, but a rarer separator or a prefix-check guard would be more defensive.
- Broad `except Exception` around per-service loops is fine for UI resilience, but if `get_for_service` raises, the dedup pass keeps the local rows while the second loop silently drops the service rows — net "fail soft", just worth a comment.
- No tests for `get_service_view_id`/`parse_service_view_id` round-trip or the dedup filter — both would be cheap to cover.

Nice addition overall — looking forward to seeing it land. 🤖

---
*Reviewed by Claude (Opus 4.7) via Claude Code*

ControlDrift:
> OK, I've been over this and I've set Claude on it. For once, we don't find the same things.
> 
> My concern here is architectural; that is, it's a lot of spot changes needed to handle your synthetic 'view ids' in the form `service:<<service id>>:<<appid>`. That has got to be a bug factory. You are treating this new thing as a dynamic category, but also it is like a service with its own `appids`. So what could we do about it?
> 
> 1. The minimal answer is to do what you are doing, but with type annotations and helper functions. You do have some helpers (`parse_service_view_id`) but they aren't used consistently. Type annotations could make it clearer where magic 'view ids' are intended. By mypy is _very_ poor at enforcing type safety for us, so I don't favor this.
> 2. Instead, we could go dynamic- that would fit Python better. Have a GameID _class_ with subclasses maybe, which could represent different kinds of ID. Misuse would result in runtime errors. This is a more wide-ranging change, but would clarify Lutris's existing system of game-ids and appids (or so I would hope). Merging this close to release would be risky though.
> 3. Alternately, forget dynamic categories and make a new "all services" service. This new service would delegate everything it does to the other services, but it would then use your new ids as its own appids (ie, adjust the db_games dictionaries). This is a rewrite of the feature, but could offer more functionality- like reloading all services together.
> 
> Claude has much more detail-oriented stuff instead. I'll send him over next.

For now I have resolved the remarks in claude's review.

3. seems more elegant. You also prefer it?

danieljohnson2:
> 3\. seems more elegant. You also prefer it?

I do. But I have near-terminal case of [object-oriententitis](https://en.wikipedia.org/wiki/Object-oriented_programming)

For the record, I think the boss will be less picky than I am about these sorts of things, and you can get away without doing any of these options.

Issue #6724, opened 2026-06-05
You might want to think of another place for the flatpak version


AliensInc:
Seems that flathub will/has banned all projects that in clude any type of AI in the code.

Hope this is OK to link here!
https://www.youtube.com/watch?v=36vbquKEMyA

strycore:
I will remove Lutris from Flathub when they ask me to. As far as I know they ban projects that are entirely made by AI which is not the case for Lutris.

Do not post links to Lunduke videos on the Lutris bugtracker please.

danieljohnson2:
Some quick research online suggests that we're grandfathered in, but of course in the long run this will be fatal for Flathub if actually enforced.

I imagine it won't be; it'll just be another excuse to ban whoever they don't like, and no more.

AliensInc:
Maybe create AppImage instead if that's even doable.

@strycore, sorry, it won't happen again but in there it's pointed out that even use of it will be banned.
We''ll see what happens.

strycore:
Don't worry about it, no harm done and I recorded a fun video as a reaction to the whole thing.

If your source of information is the Lunduke Journal, I encourage you to source your information from elsewhere as well to form an opinion that is more rooted in reality.

The way I read the AI policy, this appears to apply to new submissions only, I see nothing about existing application. 
Also, they have this line, which I would hope applies to Lutris,

> Exceptions may be granted for mature, well-maintained projects.

https://docs.flathub.org/docs/for-app-authors/requirements#generative-ai-policy

An AppImage is on the roadmap but I doubt it will offer the same level of reliability as Flatpak,

PR #6726, opened 2026-06-06
Add AppImage packaging and release workflow

### LINKED ISSUE #5085: [Feature request] AppImage release
Hello,

have you considered an AppImage release?

danieljohnson2:
## Summary

- Add `make appimage` — builds a self-contained x86_64 AppImage in an Ubuntu 22.04 container (docker or podman) via linuxdeploy + linuxdeploy-plugin-gtk + appimagetool. Output is ~78 MB and bundles Python 3.10, GTK3, WebKit2GTK 4.1, GnomeDesktop, AyatanaAppIndicator, and Lutris itself.
- Add `.github/workflows/publish-appimage.yml` — on `release.released` / `release.prereleased`, builds and attaches the AppImage as a release asset via the bundled `gh` CLI; `workflow_dispatch` is also supported for ad-hoc builds that stop at the workflow-artifact upload.
- Filed against #6724: Flathub's recent AI-project ban makes Flatpak an unreliable distribution channel, and AppImage is arguably a better fit for Lutris anyway since the client deliberately needs unsandboxed access to host `xrandr`, `7z`, `wine`, `flatpak`, mounted drives, and the user's whole `$HOME`.

A few design points worth reviewing (also covered in the relevant commit messages):

- **Bundled Python:** the build image's `/usr/bin/python3.10` plus its stdlib, with tkinter/test/idle and the link-time-only `config-3.10-*/` directory stripped to keep size down.
- **PyGObject and dbus-python are pip-installed into the AppDir** (`'PyGObject<3.50'`, `dbus-python`, plus `pycairo` for the GTK→Cairo bridge). Both have switched to meson-python build backends, so the build image carries `meson` + `pycairo` (pip) and `libdbus-glib-1-dev` (apt) to satisfy the source build. PyGObject is pinned `<3.50` because 3.50+ requires the `girepository-2.0` ABI introduced in GLib 2.80; Jammy ships 2.72.
- **`AppRun` deliberately does *not* export `LD_LIBRARY_PATH`.** linuxdeploy already sets `$ORIGIN`-relative RPATH on every binary it deploys, and exporting `LD_LIBRARY_PATH` would leak into every host subprocess — which surfaced immediately when `flatpak` crashed trying to load our bundled libssl 3.0 instead of the host's 3.4. The build script `patchelf`s the RPATH on the pip-installed `_gi.so` / `_dbus*.so` after the linuxdeploy pass, since linuxdeploy only patches the copies it deploys into `usr/lib`, not the originals in `dist-packages/` where Python actually imports them.
- **`PATH` is left untouched** — we deliberately don't prepend `$APPDIR/usr/bin`, so host tools (`xrandr`, `7z`, `wine`, `flatpak`) keep winning, and host subprocesses like `umu-launcher` find host Python (with its host site-packages) via their `#!/usr/bin/env python3` shebangs. The AppImage requires host Python 3.10+ for UMU-based runners; modern distros satisfy this by default.
- **`linuxdeploy-plugin-gtk` is vendored at a pinned commit** with its upstream MIT licence alongside, addressing reviewer feedback that the plugin's upstream has been dormant since 2023. As the first practical use of the vendoring, the plugin's hardcoded `GTK_THEME="Adwaita:light/dark"` is patched out so the AppImage respects the user's actual GTK theme (upstream #39).

## Test plan

- [x] `make appimage` produces `dist/Lutris-0.5.23-x86_64.AppImage` locally on Fedora/podman, ~78 MB
- [x] AppImage boots GUI on Fedora 43 Nobara: window opens, GPU detected, DB loads, GOG/ZOOM service auth works, host GTK theme is respected
- [x] Host subprocess invocations work — Lutris correctly shells out to host `flatpak`, `xrandr`, etc. without `LD_LIBRARY_PATH` or `PYTHONPATH` poisoning
- [x] `workflow_dispatch` run on a personal fork builds + uploads the workflow artifact
- [x] A release event on a personal fork triggers the workflow and successfully attaches the AppImage to the release page
- [x] Maintainer review of design choices listed above
- [ ] First real release post-merge produces the attached AppImage as expected

Refs #6724.
Resolves #5085.

### DIFF
--- a/.github/workflows/publish-appimage.yml
+++ b/.github/workflows/publish-appimage.yml
@@ -1,0 +1,55 @@
+# Builds the Lutris AppImage and attaches it to the triggering GitHub
+# release. The build runs in the same Ubuntu 22.04 container the
+# `make appimage` recipe uses locally, so the artifact is bit-for-bit
+# the same as what developers produce on their own machines.
+#
+# Triggers:
+#   * release.released / release.prereleased — attaches to that release
+#   * workflow_dispatch                     — produces a workflow artifact
+#                                             only; does not touch a release
+name: Publish AppImage
+
+on:
+  release:
+    types: [released, prereleased]
+  workflow_dispatch:
+
+jobs:
+  appimage:
+    runs-on: ubuntu-22.04
+    permissions:
+      # Required so action-gh-release can upload the asset.
+      contents: write
+    steps:
+      - name: Checkout the repository
+        uses: actions/checkout@v6
+
+      - name: Build the AppImage
+        env:
+          # On release events, force the version to the tag (strip a
+          # leading 'v' if present). On workflow_dispatch this is empty,
+          # and build.sh falls back to awk'ing lutris/__init__.py.
+          RELEASE_TAG: ${{ github.event.release.tag_name }}
+        run: |
+          if [ -n "$RELEASE_TAG" ]; then
+              export LUTRIS_VERSION="${RELEASE_TAG#v}"
+          fi
+          make appimage
+
+      - name: Upload as workflow artifact
+        uses: actions/upload-artifact@v7
+        with:
+          name: lutris-appimage
+          path: dist/Lutris-*-x86_64.AppImage
+          if-no-files-found: error
+
+      # Only attach the asset when the workflow was actually triggered
+      # by a release event. workflow_dispatch runs stop at the artifact.
+      - name: Attach AppImage to release
+        if: github.event_name == 'release'
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          gh release upload "${{ github.event.release.tag_name }}" \
+              dist/Lutris-*-x86_64.AppImage \
+              --repo "${{ github.repository }}"
--- a/utils/appimage/build-in-container.sh
+++ b/utils/appimage/build-in-container.sh
@@ -1,0 +1,134 @@
+#!/usr/bin/env bash
+# Assemble the Lutris AppDir and produce an AppImage.
+#
+# This runs inside the Docker image defined by Dockerfile. It expects the
+# Lutris source tree to be mounted at /src, and writes the finished
+# AppImage to /src/dist/.
+set -euo pipefail
+
+SRC=/src
+OUT=$SRC/dist
+APPDIR=/tmp/Lutris.AppDir
+
+rm -rf "$APPDIR"
+mkdir -p "$APPDIR/usr/bin" "$APPDIR/usr/lib" "$OUT"
+
+# Bundle the Python interpreter + standard library so the AppImage doesn't
+# depend on the host having python3.10 installed. We use whatever python3
+# the build image ships (Ubuntu 22.04 → 3.10).
+PY_VER="$(python3 -c 'import sys;print("%d.%d"%sys.version_info[:2])')"
+cp -a "$(readlink -f /usr/bin/python3)" "$APPDIR/usr/bin/python${PY_VER}"
+ln -sf "python${PY_VER}" "$APPDIR/usr/bin/python3"
+cp -a "/usr/lib/python${PY_VER}" "$APPDIR/usr/lib/"
+# Drop test/idle bloat that no end user needs from the AppImage.
+rm -rf "$APPDIR/usr/lib/python${PY_VER}/test" \
+       "$APPDIR/usr/lib/python${PY_VER}/idlelib" \
+       "$APPDIR/usr/lib/python${PY_VER}/turtledemo" \
+       "$APPDIR/usr/lib/python${PY_VER}/tkinter" \
+       "$APPDIR/usr/lib/python${PY_VER}/lib2to3"
+
+# Install Lutris (Python package + bin/lutris + data files) into the AppDir.
+cd "$SRC"
+python3 setup.py install --root="$APPDIR" --prefix=/usr --no-compile
+
+# Install runtime Python deps directly into the AppDir's site-packages.
+SITE_PACKAGES=$(python3 -c 'import sysconfig; print(sysconfig.get_paths()["purelib"])')
+SITE_PACKAGES=${SITE_PACKAGES#/usr/}
+TARGET="$APPDIR/usr/$SITE_PACKAGES"
+mkdir -p "$TARGET"
+
+# PyGObject is satisfied by the system python3-gi we bundle from apt; pip
+# rebuilds against the host gobject-introspection and we want the version
+# wired up to the bundled typelibs, so we leave it to apt + linuxdeploy.
+# dbus-python is intentionally excluded here — it switched to a meson-python
+# build backend that pulls in a heavy toolchain, and the apt-shipped
+# python3-dbus is a perfectly good prebuilt copy we already have.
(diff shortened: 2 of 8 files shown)


.github/workflows/publish-appimage.yml
```diff
@@ -0,0 +1,55 @@
...
+          # On release events, force the version to the tag (strip a
+          # leading 'v' if present). On workflow_dispatch this is empty,
+          # and build.sh falls back to awk'ing lutris/__init__.py.
+          RELEASE_TAG: ${{ github.event.release.tag_name }}
+        run: |
+          if [ -n "$RELEASE_TAG" ]; then
+              export LUTRIS_VERSION="${RELEASE_TAG#v}"
+          fi
+          make appimage
+
```
ItsAllAboutTheCode, line 38: Would it be good to run the unit test pipeline against the built AppImage?
That way any potential missing dependencies can be caught before publishing.

danieljohnson2: I don't know how I would do that.

I also don't know how much it would really help; if the container where we run the test has the dependencies we want, but the .appimage has not, it will pass the test: The .appimage is not like a sandbox or a VM, it's more like you statically linked some of your dependencies. Dynamically linked stuff will still work, as long as those dependencies are available at runtime.

ItsAllAboutTheCode: Ok, this seems fine then.
I was thinking it would be helpful to have some kind of way to validate that the AppImage is running in the same environment as a normal loose file release.
i.e, running the AppImage should access the same $HOME/.config/lutris, $HOME/.local/share/lutris files of the normal release in an interchangable way.

Pretty much that user should be able to swtich between a AppImage releass and a normal release built at the same revision and not notice a difference.

danieljohnson2: If I understand it correctly (and perhaps I don't) there's no AppImage private .config or .local; it'll have to use the same ones as a .deb release would.

It seems to do that in my testing, anyway.

ItsAllAboutTheCode: That sounds good to me.
Since you tested it sounds like there are no edge cases or hiccups.

Twig6943:
Wouldn't it be better to make an appimage that's sharun based like this?

https://github.com/pkgforge-dev/Lutris-AppImage

Also for flatpak; a flatpak repo could be hosted / .flatpak files could be shipped (I'm all for appimage but saying this for flatpak folks)

danieljohnson2:
That seems to be a tool to run down our dynamic dependencies. I don't think we want to do that.

A lot of them are things you download in our UI; the main exception tot hat in Wine, which you are supposed to install from your distribution. We could make it a dependency of our .deb package too, but the boss says not to- Lutris is not a Wine front end but something broader, and we don't want Wine to be a requirement.

Or that is how I understand it.

utils/appimage/build-in-container.sh
```diff
@@ -0,0 +1,141 @@
...
+done < <(find "$TARGET" -maxdepth 3 \( -name '_gi*.so' -o -name '_dbus*.so' -o -name 'gi.repository*.so' \) -print0 2>/dev/null)
+echo "Passing ${#LIB_ARGS[@]} native extensions to linuxdeploy: ${LIB_ARGS[*]}"
+
+linuxdeploy \
+    --appdir "$APPDIR" \
+    --plugin gtk \
+    --executable "$APPDIR/usr/bin/python${PY_VER}" \
+    "${LIB_ARGS[@]}" \
+    --desktop-file "$DESKTOP_FILE" \
+    --icon-file "$ICON_FILE"
```
Samueru-sama, line 116: This is such a bad idea to use linuxdeploy, specially the gtk plugin.

This is abandoned and has a lot of known issues, even CPU-X now uses sharun instead. (made by the same person that made the GTK plugin btw)

Samueru-sama:
> Wouldn't it be better to make an appimage that's sharun based like this?

That appimage does not use sharun.

It uses runimage which is a container and I don't recommend. It was the only way to get Lutris to work properly. (Details why below).


> A lot of them are things you download in our UI; the main exception tot hat in Wine, which you are supposed to install from your distribution.

Note that since Lutris has the ability to download binaries **in the user's system** you will run into bugs related to the fact those binaries may not be able to work at all on the user system.

You could in theory spin a container, but that means access to userns which distros like Ubuntu put restrictions on.

So the only way for this to work right via horrible hack we do with LD_PRELOAD and sharun, see https://github.com/pkgforge-dev/Anylinux-AppImages/issues/486#issuecomment-4229902609

utils/appimage/build-in-container.sh
danieljohnson2: That *is* a concern. ``linuxdeploy`` is still active but the GTK plugin is old; ``sharun`` does not seem right for us as I have said; perhaps ``appimage-builder`` might be a better choice - it's looking for a maintainer but has been more recently touched than ``sharun`` at least. Perhaps there are other options.

What I will do right now is "vendor" the plugin in case it goes away. And we'll fix the locked-theme bug, which does affect Lutris.

Maybe when we make the GTK4 switch we'll also rework this. Need to think about what to do in the long term.

danieljohnson2:
This PR also had to deal with runners (and games) failing because they get a mix of AppImage and system stuff; there may still be bugs around that. Lutris itself uses the AppImage stuff, and (hopefully) the runners and games cleanly do not.

It's like you installed it from  your repo, except you didn't have to.

It's not containerized; if you want Flatpak, we already have that.

utils/appimage/build-in-container.sh
Samueru-sama: > `sharun` does not seem right for us as I have said

Why not? sharun can do every single thing that linuxdeploy or appimage-buidler do and even more.

The repo you were linked before actually does not use sharun btw.

Just so you are aware, when you use sharun you are not limited to building on the oldest system possible, **you can build on archlinux and the appimage will run on ubuntu14 even**, it also runs on non glibc systems.

Samueru-sama:
> because they get a mix of AppImage and system stuff; there may still be bugs around tha

You mean environment variables? 

We have a fix for that in [anylinux.so](https://github.com/pkgforge-dev/Anylinux-AppImages/blob/main/useful-tools/lib/anylinux.c)

This library was explicitly made for [Ghostty](https://github.com/pkgforge-dev/ghostty-appimage) because we had many issues with the appimage env leaking into child processes.

Overtime it has been expanded to include a lot of other fixes. One is the locales in `/usr/share` which most GTK apps have hardcoded and cannot be relocated easily (not a problem in python apps since that can be changed without recompiling).

strycore:
For the runners requirements, we are headed to having the Steam runtime 3 or 4 as the main way to run any executable, maybe coupled with additional folders of .so files for games and emulators that require them. This covers all cases: native, Flatpak, Snap and AppImage.

strycore:
Is there a reason the AppImage uses Ubuntu 22.04 as a base rather than 24.04 or 26.04? 

Besides that, this looks good and ready to merge

danieljohnson2:
That is to maximize compatibility - that's the oldest Ubuntu that has a compatible Python version.

As I understand it, if we use a later base version here, we can fail to link at runtime against an older glib.

strycore:
Ok that makes sense. Good to merge then

PR #6733, opened 2026-06-14
fix: include NULL-service games when excluding a service from the games view

### LINKED ISSUE #6716: Library shows no games when games from any source are hidden
### Bug description

The *Games* tab shows no games from other sources when there is any enabled game source like Steam or GOG that has *Show games from this source in the games view* unticked.

### How to Reproduce

1. Go to burger menu (top-right) > Preferences > Sources
2. Enable a source. Untick *Show games from this source in the games view* for that source.

Image

3. Close preferences to return to the library.
4. See that all games — even those from other platforms — are hidden.

Image

Note that installed games from other sources are still accessible from the *Recent* tab.

### Expected behavior

Games from other sources should still be present, as they are when there is no *Show games from this source in the games view* option left unticked.

*None of these games were installed via the hidden source:*
Image

### Log output

```shell
➜  ~ lutris -d
2026-05-30 16:57:15,344: The Battle.net source is unavailable because Google protobuf could not be loaded: No module named 'google'
INFO     2026-05-30 16:57:15,422 [application.do_command_line:468]:Starting Lutris 0.5.22
INFO     2026-05-30 16:57:15,481 [startup.run_all_checks:120]:"card1" is AMD Radeon RX 6700 XT (1002:73df 1da2:e445 amdgpu) Driver 26.1.1
DEBUG    2026-05-30 16:57:15,519 [xrandr.get_resolutions:91]:Retrieving resolution list
WARNING: running xrandr against an Xwayland server. See the xrandr man page for details.
DEBUG    2026-05-30 16:57:15,521 [xrandr._get_vidmodes:18]:Retrieving 30 video modes from XrandR
DEBUG    2026-05-30 16:57:15,615 [path_cache._update_missing_games:114]:Checking for missing games
DEBUG    2026-05-30 16:57:15,703 [api.check_stale_runtime_versions:52]:Modified at Sat 30 May 2026 04:35:02 PM , will update after Sat 30 May 2026 10:35:02 PM 
DEBUG    2026-05-30 16:57:15,704 [lutriswindow.create_runtime_updater_cb:1460]:Runtime up to date
INFO     2026-05-30 16:57:29,271 [application.do_shutdown:1076]:Shutting down Lutris
```

### System Information

```shell
Vulkan support: YES
Esync support: YES
Fsync support: YES
Wine installed: YES
Gamescope: NO
Mangohud: YES
Gamemode: YES
Steam: YES
In Flatpak: NO
[System]
OS: Arch Linux rolling n/a
Arch: x86_64
Kernel: 7.0.10-arch1-1
Lutris Version: 0.5.22
Python Version: 3.14.5 (main, May 10 2026, 18:26:20) [GCC 16.1.1 20260430]
Desktop: KDE
Display Server: wayland
[CPU]
Vendor: AuthenticAMD
Model: AMD Ryzen 5 5600 6-Core Processor
Physical cores: 6
Logical cores: 12
[Memory]
RAM: 31.3 GB
Swap: 32.0 GB
[Graphics]
Vendor: AMD
OpenGL Renderer: AMD Radeon RX 6700 XT (radeonsi, navi22, ACO, DRM 3.64, 7.0.10-arch1-1)
OpenGL Version: 4.6 (Compatibility Profile) Mesa 26.1.1-arch1.2
OpenGL Core: 4.6 (Core Profile) Mesa 26.1.1-arch1.2
OpenGL ES: OpenGL ES 3.2 Mesa 26.1.1-arch1.2
Vulkan Version: 1.4.350
Vulkan Drivers: AMD Radeon RX 6700 XT (RADV NAVI22) (1.4.348)
```

### Media (optional)

_No response_

rawrmonster17:
## Problem

Fixes #6716.

When a user unchecks **"Show games from this source in the games view"** for any service (e.g. Steam) in Preferences → Sources, *all* games disappeared from the library — including games from other services and manually-added games that have nothing to do with Steam.

## Root cause

SQLite three-valued logic: `NULL != 'steam'` evaluates to `NULL`, not `TRUE`. The `_create_filter` helper in `database/sql.py` was generating bare `service != ?` (or `service NOT IN (...)`) for an exclude filter. Any row with `service IS NULL` (manually-added games) was silently dropped because the WHERE clause treats NULL as falsy.

## Fix

When `negate=True` and the value set contains no explicit `None`, wrap the generated SQL fragment with `(field IS NULL OR ...)` so that NULL rows survive the exclusion:

```python
else:
    if negate:
        return f"({field} IS NULL OR {sql})"
    return sql
```

This preserves the existing behaviour of the `also_null` branch (when `None` is *explicitly* in the exclude set, NULL rows are still excluded via `IS NOT NULL AND ...`).

## Testing

- Added two regression tests to `tests/_test_pga.py` covering single-service and multi-service exclusion.
- All 238 existing tests continue to pass.

## Verified on

- Arch Linux, Lutris 0.5.22, NVIDIA RTX 4070 Ti, driver 610.43

### DIFF
--- a/lutris/database/sql.py
+++ b/lutris/database/sql.py
@@ -194,6 +194,8 @@ def _create_filter(field: str, value: Any, params: list[Any], negate: bool = Fal
         else:
             return f"({field} IS NULL OR {sql})"
     else:
+        if negate:
+            return f"({field} IS NULL OR {sql})"
         return sql
 
 
--- a/tests/_test_pga.py
+++ b/tests/_test_pga.py
@@ -52,6 +52,26 @@ class TestPersonnalGameArchive(DatabaseTester):
         self.assertEqual(len(game_list), 1)
         self.assertEqual(game_list[0]["name"], "installed_game")
 
+    def test_exclude_service_includes_null_service_games(self):
+        games_db.add_game(name="manual_game", runner="Linux")
+        games_db.add_game(name="steam_game", runner="Linux", service="steam")
+        games_db.add_game(name="gog_game", runner="Linux", service="gog")
+        game_list = games_db.get_games(excludes={"service": {"steam"}})
+        game_names = {g["name"] for g in game_list}
+        self.assertIn("manual_game", game_names)
+        self.assertIn("gog_game", game_names)
+        self.assertNotIn("steam_game", game_names)
+
+    def test_exclude_multiple_services_includes_null_service_games(self):
+        games_db.add_game(name="manual_game", runner="Linux")
+        games_db.add_game(name="steam_game", runner="Linux", service="steam")
+        games_db.add_game(name="gog_game", runner="Linux", service="gog")
+        game_list = games_db.get_games(excludes={"service": {"steam", "gog"}})
+        game_names = {g["name"] for g in game_list}
+        self.assertIn("manual_game", game_names)
+        self.assertNotIn("steam_game", game_names)
+        self.assertNotIn("gog_game", game_names)
+
     def test_game_with_same_slug_is_updated(self):
         games_db.add_game(name="some game", runner="linux")
         game = games_db.get_game_by_field("some-game", "slug")


danieljohnson2:
Good catch!

Duratlas:
As someone who doesn't know all the terminology of Github, this being merged doesn't mean that an update has been released with this fix yet, right? Since the current version, 0.5.22 is still from February. How would I go about getting this fix if that's possible, or do I have to wait for a Lutris update?

danieljohnson2:
No, being merged just puts it in the master branch. We have to build a release from there.

Some repositories offer a ``lutris-git`` packages which is not exactly the master branch, but should be much closer than the last proper release. You might look for that.

PR #6734, opened 2026-06-14
fix: refresh view after service sync regardless of active service

### LINKED ISSUE #4553: Global game view is not updated, when new game is detected in sources
- Activate one of `Epic Games Store`, `Ubisoft Connect`, (`Steam` if #4527 is merged).
  The behavior is the same, no matter which source view is selected. However, the Lutris view does not provide a context menu to completely remove the file from the library and other sources actually delete game data.
- Remove an installed game (no data is actually deleted).
- Remove the uninstalled game from the library.
  This is not actually necessary, the behavior is the same, if it is kept in the library.
- Activate global `Games` view.
  The game is no longer shown.
- Refresh game source **without** activating the source view.
- The deleted game is detected as installed, however the global game view is not updated.
  After changing the view to e.g. recent and then back to the global view, the game is shown.

rawrmonster17:
Fixes #4553

When a service finishes syncing, `on_service_games_loaded` only called
`update_store()` if the user was already viewing that specific service's
sidebar row. On the global "Games" view (`self.service is None`) or any
other service's view, the signal was a no-op — newly synced games
wouldn't appear until the user manually switched views and back.

The fix removes the stale guard. `update_store()` is safe to call
unconditionally: it uses a generation counter to coalesce rapid
reloads and cancel stale callbacks. This matches the existing pattern
in `on_categories_updated` and `on_local_library_updated`, which have
never had this guard.

### DIFF
--- a/lutris/gui/lutriswindow.py
+++ b/lutris/gui/lutriswindow.py
@@ -1129,8 +1129,7 @@ class LutrisWindow(Gtk.ApplicationWindow, DialogLaunchUIDelegate, DialogInstallU
 
     def on_service_games_loaded(self, service):
         """Request a view update when service games are loaded"""
-        if self.service and service.id == self.service.id:
-            self.update_store()
+        self.update_store()
 
     def on_categories_updated(self):
         self.update_store()


danieljohnson2:
That's an obscure bug there, but I'll take it.

PR #6735, opened 2026-06-14
fix: give each GOG parallel download worker its own stall monitor

rawrmonster17:
## Summary

Fixes a thread-safety bug in the GOG parallel downloader that can abort healthy downloads with a spurious "stall" error and can also hide genuine stalls.

Stall detection tracks its throughput window in two instance fields on the base `Downloader` (`_stall_start`, `_stall_bytes_at_start`). That is fine for the single-stream path, but `GOGDownloader` runs several worker threads, and each worker calls `_check_stall(stream_bytes)` with its **own** per-worker byte counter while mutating that one shared window — with no lock.

### The bug

When a fast worker opens the window at a high byte count and a slower worker then checks in with its smaller counter:

```
bytes_in_window = slower_worker_bytes - fast_worker_start   # negative
throughput = bytes_in_window / elapsed                       # negative
```

A negative throughput is always below `LOW_SPEED_LIMIT`, so the window never resets. After `LOW_SPEED_TIME` (30s) a perfectly healthy download is aborted with `DownloadStallError`, retried 3×, then fails. The reverse also happens: a fast worker can reset a genuinely stalled worker's window, so real stalls go undetected.

This affects every GOG game download, since `GOGDownloader` is the default downloader for GOG installer files.

### The fix

Extract the (unchanged) algorithm into a small `StallMonitor` class so each stream owns its own window:

- The base `Downloader` keeps a single monitor — the single-threaded path is behaviourally identical.
- Each GOG worker creates a **local** `StallMonitor` in `_download_range`, so worker counters never collide.
- `_stall_start` / `_stall_bytes_at_start` remain as backwards-compatible proxy properties.

## Testing

- Added `TestStallMonitorIsolation` regression tests: interleaved fast/slow streams stay healthy, a genuine stall still raises, and `_reset_stall_state` installs a fresh window.
- Adapted the existing pipelining stall test to drive the new per-worker mechanism through a deterministic fake clock instead of poking removed internals.
- All download-related tests pass (`_test_stall_detection.py`, `_test_pipelining.py`, `_test_gog_downloader.py`, `_test_download_progress.py`, `_test_collection_progress.py`).
- Verified end-to-end with a real 4-worker threaded download against a local Range-capable HTTP server with one deliberately throttled worker: download completes without a spurious stall and the output file is byte-for-byte correct (MD5 verified).

### DIFF
--- a/lutris/gui/dialogs/delegates.py
+++ b/lutris/gui/dialogs/delegates.py
@@ -9,7 +9,7 @@ from lutris.game import Game
 from lutris.gui import dialogs
 from lutris.gui.dialogs.download import DownloadDialog
 from lutris.services import SERVICES
-from lutris.util.downloader import Downloader
+from lutris.util.downloader import BaseDownloader, SimpleDownloader
 
 if TYPE_CHECKING:
     from lutris.config import LaunchConfigDict
@@ -104,7 +104,7 @@ class InstallUIDelegate(Delegate):
 
         The default is to download with no UI, and no option to cancel.
         """
-        downloader = Downloader(url, destination, overwrite=True)
+        downloader = SimpleDownloader(url, destination, overwrite=True)
         downloader.start()
         return downloader.join()
 
@@ -158,7 +158,7 @@ class DialogInstallUIDelegate(InstallUIDelegate):
     def download_install_file(self, url: str, destination: str) -> bool:
         dialog = DownloadDialog(url, destination, parent=self)
         dialog.run()
-        return dialog.downloader.state == Downloader.COMPLETED
+        return dialog.downloader.state == BaseDownloader.COMPLETED
 
 
 class DialogLaunchUIDelegate(LaunchUIDelegate):
--- a/lutris/gui/dialogs/download.py
+++ b/lutris/gui/dialogs/download.py
@@ -4,7 +4,7 @@ from gi.repository import Gtk
 
 from lutris.gui.dialogs import ModalDialog
 from lutris.gui.widgets.download_progress_box import DownloadProgressBox
-from lutris.util.downloader import Downloader
+from lutris.util.downloader import BaseDownloader
 
 
 class DownloadDialog(ModalDialog):
@@ -25,7 +25,7 @@ class DownloadDialog(ModalDialog):
         self.dialog_progress_box.start()
 
     @property
-    def downloader(self) -> Downloader:
+    def downloader(self) -> BaseDownloader:
         return self.dialog_progress_box.downloader
 
     def download_complete(self, _widget, _data):
--- a/lutris/gui/dialogs/runner_install.py
+++ b/lutris/gui/dialogs/runner_install.py
@@ -16,7 +16,7 @@ from lutris.game import Game
 from lutris.gui.dialogs import ErrorDialog, ModelessDialog, display_error
 from lutris.gui.widgets.stock_icon_image import StockIconImage
 from lutris.util import jobs, system
-from lutris.util.downloader import Downloader
+from lutris.util.downloader import SimpleDownloader
 from lutris.util.extract import extract_archive
 from lutris.util.jobs import schedule_repeating_at_idle
 from lutris.util.log import logger
@@ -369,7 +369,7 @@ class RunnerInstallDialog(ModelessDialog):
         if not url:
             ErrorDialog(_("Version %s is no longer available") % version, parent=self)
             return
-        downloader = Downloader(url, dest_path, overwrite=True)
+        downloader = SimpleDownloader(url, dest_path, overwrite=True)
         schedule_repeating_at_idle(self.get_progress, downloader, row, interval_seconds=0.1)
         self.installing[version] = downloader
         downloader.start()
--- a/lutris/gui/widgets/download_collection_progress_box.py
+++ b/lutris/gui/widgets/download_collection_progress_box.py
@@ -7,7 +7,7 @@ from gi.repository import GObject, Gtk, Pango
 
 from lutris.gui.dialogs import display_error
 from lutris.util.download_cache import CacheState, create_cache_lock, update_cache_lock
-from lutris.util.downloader import Downloader
+from lutris.util.downloader import BaseDownloader, SimpleDownloader
 from lutris.util.jobs import schedule_repeating_at_idle
 from lutris.util.log import logger
 from lutris.util.strings import gtk_safe, human_size
@@ -51,7 +51,7 @@ class DownloadCollectionProgressBox(Gtk.Box):
         self,
         file_collection: "InstallerFileCollection",
         cancelable: bool = True,
-        downloader: Downloader | None = None,
+        downloader: BaseDownloader | None = None,
     ) -> None:
         super().__init__(orientation=Gtk.Orientation.VERTICAL)
 
@@ -135,7 +135,7 @@ class DownloadCollectionProgressBox(Gtk.Box):
             os.remove(tmp_path)
 
         try:
-            downloader_cls = getattr(file, "downloader_class", None) or Downloader
+            downloader_cls = getattr(file, "downloader_class", None) or SimpleDownloader
             dl = downloader_cls(file.url, file.tmp_file, referer=file.referer, overwrite=True)
         except RuntimeError as ex:
             display_error(ex, parent=self.get_toplevel())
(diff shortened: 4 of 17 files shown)


danieljohnson2:
I'm not thrilled with this, but I am not thrilled with the original GOGDownloader either- it sort of body-snatches Downloader, leaving various disused bits behind. That's the source of this bug, but your version still does this. This mixing of implementation inheritance and polymorphism is icky. But it's not your fault either. I might accept your fix and clean up this business afterwards.

But Claude isn't thrilled easier, though for other reasons. I think its objection to your testing really should be looked at. I'll have it post here.

danieljohnson2:
Beep boop 🤖 — Daniel asked me to weigh in.

Your bug analysis is solid: the shared `_stall_start` / `_stall_bytes_at_start` really do get clobbered across workers, and a fast worker's high `bytes_at_start` minus a slow worker's small `stream_bytes` does go negative, which short-circuits the "speed is good — reset the window" branch and guarantees a spurious stall after `LOW_SPEED_TIME`. Worth noting it's a bit worse than the PR body says: the `self._reset_stall_state()` call at the top of `_download_range` means every worker is also stomping the window on entry, not just during chunk loops.

Direction of the fix — per-stream state owned by the stream — is right. A couple of pieces of the surface concern me though, and I think they're worth tightening before merge rather than after:

**1. The proxy `_stall_start` / `_stall_bytes_at_start` properties are not actually backwards-compatible.** They only reflect the *base* `Downloader`'s monitor — they have no relationship to any GOG worker's monitor (which lives as a local in `_download_range`). So any caller or test that was reading/writing them in the context of `GOGDownloader` would already have been getting nonsense, both before and after the patch. Their real purpose is to keep `_test_stall_detection.py`'s `dl._stall_start = get_time() - 31.0` pattern compiling. That's a test-shape problem leaking into production code as a misleading shim. I'd drop them and update those tests to construct a `StallMonitor` directly (which is what `TestStallMonitorIsolation` already does).

**2. The fake clock in the adapted `_test_pipelining.py` is fragile.** Returning `0.0` once then `100.0` forever is tightly coupled to the exact sequence of `get_time()` calls inside `StallMonitor.check()`. If the monitor ever takes the timestamp twice in one check, or callers add a `get_time()` elsewhere in `_download_range`, the test silently changes meaning. Hard to assert what; easy to break later. Injecting elapsed time into the monitor (or passing `now` into `check()`) would let the test express intent without faking a global.

**3. Is `StallMonitor` carrying its weight?** It's two ints, a float-or-None, and one method. A `@staticmethod _check_stall(stream_bytes, window_start, bytes_at_window_start) -> tuple[float | None, int]` plus two locals in `_download_range` would fix the same bug without introducing a new class or the temptation to grow proxy properties. Not a blocker — readability vs. minimalism is a judgment call — but if you keep the class, please drop the proxies along with it.

None of this changes the conclusion that the bug is real and the per-stream fix is the right shape. I'd just rather ship the leaner version that doesn't carry a wart explicitly labelled as a backwards-compatibility shim for code that wasn't working in the first place.

🤖 _Comment generated with [Claude Code](https://claude.ai/code)_

rawrmonster17:
Don't merge it yet. I want to take a stab at really just making this right all around instead of making you come back and refactor later. Might as well do it right the first time.

danieljohnson2:
OK. So you know, what I am thinking of is an abstract Downloader base class for the public interface, and a SimpleDownloader (or something) subclass that contains the implementation that is now in Downloader.

rawrmonster17:
Went ahead and did the structural cleanup you described rather than leaving it for later — an abstract base for the public interface, with the single-stream implementation moved into its own concrete subclass. Pushed in b606dbc.

- **`BaseDownloader`** is now the abstract public interface: state machine, progress/speed stats, retry + stall config, the shared `cancel` / `on_download_failed` / `on_download_completed` lifecycle, and `start()`. It declares the transfer hooks (`_prepare_destination`, `async_download`) plus two cleanup hooks — `_release_resources()` (every terminal path) and `_discard_persistent_state()` (completion and cancel only, **never** failure, so resume-after-crash is preserved).
- **`SimpleDownloader`** holds the original single-stream implementation unchanged — file handling, the retry loop, `_do_download`, `_prepare_retry`.
- **`GOGDownloader`** now extends `BaseDownloader` directly, so it no longer carries any of the single-stream members it used to inherit and ignore. Its resume detection became `_prepare_destination`; its progress-file cleanup became `_discard_persistent_state`.

That kills the body-snatching: `GOGDownloader` no longer has `file_pointer`, `_do_download`, or `_prepare_retry` hanging off it. Callers that build a single-stream downloader now construct `SimpleDownloader`; annotations that can hold either point at `BaseDownloader`.

No behavioural change intended — the single-stream path is byte-for-byte the same, and the GOG progress-file lifecycle (kept on failure, removed on complete/cancel) is reproduced exactly. Verified end-to-end for both single-stream and 4-worker parallel downloads (byte-correct output, no spurious stall), and mypy/ruff/tests are green.

The earlier review points are in too: the `_stall_start`/`_stall_bytes_at_start` proxy properties are gone (tests drive `StallMonitor` directly), and the pipelining test's fake clock is no longer coupled to how often `check()` samples the clock.

danieljohnson2:
Beep boop 🤖 — back for a second pass.

The structural refactor Daniel suggested looks well-executed and lands all three of my earlier nits as a side effect:

- The `_stall_start` / `_stall_bytes_at_start` proxy properties are gone. Stall tests now construct a `StallMonitor` directly, which is the right shape.
- The pipelining test's fake clock is decoupled from call count — it just stores `now` in a list and the chunk generator advances it explicitly between yields. That can't silently change meaning if `StallMonitor.check()` ever samples `get_time()` a different number of times.
- `BaseDownloader` / `SimpleDownloader` / `GOGDownloader` is exactly the split Daniel sketched. GOG no longer body-snatches a `file_pointer` it never used, and the shared lifecycle (`start` / `cancel` / `on_download_failed` / `on_download_completed`) lives in one place. The `_release_resources` (always) vs `_discard_persistent_state` (completion + cancel, never failure) distinction is what preserves resume-after-crash for GOG, and it's documented inline at the hook definitions, which is the right place for it.

I walked the cancel / completed / failed paths for both downloaders and the net effect matches the old code byte-for-byte where it should: SimpleDownloader still closes its `file_pointer` and removes `dest` on cancel; GOGDownloader still keeps the progress sidecar on failure and clears it on cancel/complete. One small accidental improvement: the old `on_download_completed` did an unguarded `self.file_pointer.close()` — the new `_release_resources` guards on `if self.file_pointer`, so the contract is a hair tighter.

Two small things I'd note but wouldn't block on:

1. **`_write_from_full_response` still has no stall detection.** Pre-existing gap (the old code didn't check there either), so it's not a regression from this PR — but if you're tightening stall semantics anyway, that fallback path stays uncovered. A worker that gets a 200 instead of 206 from a non-compliant server is the *most* likely place to encounter a slow drip. Easy to add: `stall_monitor = self._new_stall_monitor()` at the top, `stall_monitor.check(enqueued_bytes)` at the bottom of each chunk's branch where data was enqueued.

2. **`BaseDownloader._prepare_destination` / `async_download` raise `NotImplementedError`** but the class isn't formally `abc.ABC` with `@abstractmethod`. Doesn't matter functionally — nothing instantiates `BaseDownloader` directly — but if you want the abstractness to be enforced at construction rather than at `start()`, that's a one-import change.

LGTM from me on the bug fix and on Daniel's structural ask. The first one above is worth a follow-up issue regardless of when this lands.

🤖 _Comment generated with [Claude Code](https://claude.ai/code)_

danieljohnson2:
I think I'd have tried to move more out of ``BaseDownloader``, but it'll do fine as it is. I'll merge it if you feel its ready- let me know- but you can address Claude's remaining nits if you want to, first.

rawrmonster17:
Addressed both of Claude's remaining nits in the latest push (ac8a461):

1. **`_write_from_full_response` stall detection** — added `stall_monitor = self._new_stall_monitor()` before the loop and `stall_monitor.check(enqueued_bytes)` after each enqueued slice. The 200-fallback path now has the same protection as the normal 206 path.

2. **`BaseDownloader` formally abstract** — added `abc.ABC` as the base class and `@abc.abstractmethod` decorators on `_prepare_destination` and `async_download`. Missing implementations are now caught at instantiation rather than at `start()`.

Branch is squashed to one commit. All 155 download-related tests pass locally. Ready to merge whenever you are.

rawrmonster17:
fixing ruff failure now

danieljohnson2:
Ruff passed. Ready to merge?

rawrmonster17:
Yes this is good for me :D

PR #6736, opened 2026-06-15
style: modernize string formatting and simplify returns (ruff autofix)

rawrmonster17:
## Summary

Mechanical-only modernization sweep applied with `ruff 0.12.1`, scoped strictly to **safe** autofixes. No behavioral change — pure style/readability cleanup.

Adds `UP` and `RET` to the `select` list in `ruff.toml` (with unsafe/opinionated sub-rules suppressed and commented) so the project enforces these rules going forward — addressing the feedback that rules shouldn't be applied without being enabled in config.

### Rules enabled in ruff.toml

- **UP032** — convert `str.format()` calls to f-strings (101 sites)
- **UP028** — replace `yield`-over-`for` loop with `yield from` (4 sites)
- **UP008** — use `super()` instead of `super(__class__, self)` (1 site)
- **RET501 / RET502** — drop redundant `return None` / implicit `None` returns
- **RET505 / RET506 / RET507** — remove `else`/`elif` after `return` / `raise` / `continue`

### Deliberately suppressed in ruff.toml (with inline comments)

- **UP031** — `%`-format → f-string: gated behind `--unsafe-fixes` in ruff
- **UP006 / UP007 / UP035** — typing annotation modernization: requires Python 3.9+/3.10+ or drops still-needed `typing` members
- **RET503** — missing explicit `return`: can change behaviour
- **RET504** — assign-before-return: opinionated; intermediate variable often intentional

98 files, +322 / −351 (net −29 lines).

## Verification

All run against the project's pinned tooling (`ruff 0.12.1`, `mypy 1.16.1`, `--python-version 3.10`):

- `ruff check` — clean
- `ruff format --check` — clean
- `mypy` — Success, no issues in 327 source files

### DIFF
--- a/lutris/api.py
+++ b/lutris/api.py
@@ -93,7 +93,7 @@ def get_runtime_versions() -> RuntimeVersionsDict:
     if it is missing or stale."""
     if not system.path_exists(settings.RUNTIME_VERSIONS_PATH):
         return {}
-    with open(settings.RUNTIME_VERSIONS_PATH, mode="r", encoding="utf-8") as runtime_file:
+    with open(settings.RUNTIME_VERSIONS_PATH, encoding="utf-8") as runtime_file:
         return cast(RuntimeVersionsDict, json.load(runtime_file))
 
 
@@ -101,7 +101,7 @@ def read_api_key() -> dict[str, str] | None:
     """Read the API token from disk"""
     if not system.path_exists(API_KEY_FILE_PATH):
         return None
-    with open(API_KEY_FILE_PATH, "r", encoding="utf-8") as token_file:
+    with open(API_KEY_FILE_PATH, encoding="utf-8") as token_file:
         api_string = token_file.read()
     try:
         username, token = api_string.split(":")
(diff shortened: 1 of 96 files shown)


danieljohnson2:
I don't think I'll merge this, as it seems to me to be just a bunch of stylistic changes. You are enforcing Ruff rules that we don't have enabled in our ``ruff.toml``, and you're not enabling them either.

Maybe the boss will like it better, so I won't close it.

strycore:
> I don't think I'll merge this, as it seems to me to be just a bunch of stylistic changes. You are enforcing Ruff rules that we don't have enabled in our ruff.toml, and you're not enabling them either.

Agreed. If we are applying new ruff rules, they should be part of our ruff.toml

rawrmonster17:
Ok let me ask this. I am perfectly ok with just closing the pr. Would you want me to fix this to use the ruff.toml or just close this pr outright?

javi0guerrero4-cloud (changes requested):
Ho

rawrmonster17:
Rebased on master and updated to address the feedback — the rules are now enabled in `ruff.toml` so the codebase enforces them going forward.

Added `"UP"` and `"RET"` to the `select` list, with the unsafe/opinionated sub-rules (UP031, UP006/007/035, RET503/504) suppressed in the `ignore` list with inline comments explaining why each was excluded.

`ruff check` and `ruff format --check` both pass clean against the project's pinned `ruff 0.12.1`.

danieljohnson2:
I'm inclined not to really want this. Ruff beats me up and steals my lunch money often enough!

javi0guerrero4-cloud:
How

javi0guerrero4-cloud (commented):
Update

PR #6742, opened 2026-06-18
create galician language gl_ES

ninjum:
created the Galician language in case it could be added [gl_ES]


strycore:
thanks a lot!

PR #6746, opened 2026-06-19
Add GOG installer download-only cache flow

### LINKED ISSUE #6656: feat: add download-only mode to installer dialog (fetch without installing)
## Summary

Add an option to download GOG (and other) installer files to the local cache without proceeding to installation. The user would trigger the normal install flow but choose to stop after the download completes, leaving the files on disk for later offline use.

## Motivation

Currently the only way to cache installer files locally is to run through the full install dialog and cancel or fail mid-install — which is fragile and risks triggering the post-failure cleanup that was the subject of #6449.

Users with large GOG libraries on metered or unreliable connections need to be able to:
- Pre-download installers while connectivity is available
- Install later, fully offline
- Archive installer files against future GOG CDN unavailability

There is no supported way to do this today without going through each game's install dialog manually and hoping the download completes before something goes wrong.

## Proposed behaviour

- In the installer dialog, alongside the existing **Install** action, provide a **Download only** button (or a checkbox/toggle before proceeding)
- The download runs normally (using the existing parallel download infrastructure from #6525), including resume support and progress display
- On completion the installer files are left in the download cache in `DOWNLOADED` state; installation does not proceed
- A subsequent **Install from cache** path (could be a follow-up) would detect the cached files and skip the download step

## Scope

This is intentionally narrow — the download pipeline already exists and handles resume, progress, and cache state. The change would primarily be:
1. A UI affordance in the installer dialog to opt out of the install step
2. Leaving cache state as `DOWNLOADED` rather than advancing to `INSTALLING`

No new download logic required.

## Related

- #6449 — don't delete large downloads on failed install (addressed the accidental-deletion pain point)
- #6525 — parallel GOG downloads with resume support (the download infrastructure this would expose)

blakkguard:
## Summary

  Adds a GOG-focused download-only installer flow so users can download
  and keep installer files in the Lutris installer cache without
  immediately installing the game.

  This helps with offline or later installs by reusing cached GOG
  installer files instead of requiring the user to manually find the
  downloaded executable.

  Fixes #6656.

  ## Changes

  - Add a `Download only` action for supported GOG installers.
  - Store downloaded installer files in the installer cache.
  - Show the cached installer location after a download-only run
  completes.
  - Detect valid cached GOG installer files when starting an install.
  - Prompt the user to install from the cached installer when available.
  - Validate cached files by existence, size, and checksum where
  metadata is available.
  - Add cached installer management to Settings > Storage.
  - Add incomplete installer download cleanup in Settings > Storage.
  - Add tests for installer cache listing, deletion, and cached file
  validation.

  ## Scope

  This is intentionally limited to GOG installers. Other sources such as
  Epic, EA, and live-service launcher installs do not expose the same
  installer-file download flow.

  ## Testing

  Manually tested with a GOG game:

  - Download-only flow completed successfully.
  - Completion page showed the cached installer path.
  - Restarted Lutris and started the install again.
  - Lutris detected the cached installer and prompted to use it.
  - Installed successfully from the cached installer.
  - Verified cached installers appear in Settings > Storage.
  - Verified cached installer cleanup UI is present.

### DIFF
--- a/lutris/cache.py
+++ b/lutris/cache.py
@@ -25,13 +25,13 @@ def get_cache_path(create: bool = False) -> str:
     return settings.INSTALLER_CACHE_DIR
 
 
-def get_installer_cache_entries() -> list[dict]:
+def get_installer_cache_entries() -> list[dict[str, object]]:
     """Return cached installer folders with their size and file count."""
     cache_path = get_cache_path()
     if not os.path.isdir(cache_path):
         return []
 
-    entries = []
+    entries: list[dict[str, object]] = []
     for name in sorted(os.listdir(cache_path)):
         path = os.path.join(cache_path, name)
         if not os.path.isdir(path):
@@ -68,13 +68,13 @@ def delete_installer_cache_entry(path: str) -> None:
         shutil.rmtree(path)
 
 
-def get_incomplete_installer_cache_entries() -> list[dict]:
+def get_incomplete_installer_cache_entries() -> list[dict[str, object]]:
     """Return incomplete installer download artifacts in the installer cache."""
     cache_path = get_cache_path()
     if not os.path.isdir(cache_path):
         return []
 
-    entries = []
+    entries: list[dict[str, object]] = []
     seen_paths = set()
     for root, _dirs, files in os.walk(cache_path):
         for filename in sorted(files):
--- a/lutris/cache.py
+++ b/lutris/cache.py
@@ -2,12 +2,15 @@
 
 import os
 import shutil
+import time
 from gettext import gettext as _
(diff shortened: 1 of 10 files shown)


danieljohnson2:
It is odd that this is a "GOG installer" thing- I would have thought this would work on many sources, and that would be well, since GOG now has (in master) a 'depot based' installation which I think this doesn't integrate with. But that's just GOG, other sources don't have anything like that.

blakkguard:
I tried with Epic and it was able to cache the Epic installer, but when trying to install a game Epic takes over the process and download.

PR #6747, opened 2026-06-19
Fix filesystem detection for Btrfs subvolume paths

KogasaPls:
system.find_mount_point() uses os.path.ismount(), which can treat Btrfs subvolume roots as mount points even when findmnt does not list those roots as separate targets. When a Wine prefix was placed below such a subvolume, get_fs_type_for_path() looked for an exact findmnt target, found none, and returned None. Wine prefix creation then failed with: "Can't create the prefix on a file system that does not support Linux symbolic links."

Instead of matching a single mount point by exact path, choose the deepest findmnt target that contains the requested path. This resolves the backing filesystem for Btrfs subvolume paths and preserves the existing fuseblk type lookup.

Adds unit coverage for filesystem type detection across Btrfs subvolume paths, nested mount targets, partial path-prefix matches, and existing fuseblk handling.

### DIFF
--- a/lutris/util/linux.py
+++ b/lutris/util/linux.py
@@ -200,6 +200,23 @@ class LinuxSystem:  # pylint: disable=too-many-public-methods
         return [drive for drive in json.loads(findmnt_output)["filesystems"] if drive["fstype"] != "squashfs"]
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


danieljohnson2:
Looks right, I'll take this.

PR #6756, opened 2026-06-25
Add missing Wine/DXVK dependency detection and install dialog

ghostrunner42:
When a Wine game is launched, Lutris now checks whether the system has the required Vulkan and Mesa/NVIDIA packages installed. If any are missing, a dialog is shown that lists the packages, displays the install command, and allows the user to install them inline by entering their sudo password.

- lutris/util/system_packages.py: detects GPU vendor (AMD/NVIDIA/Intel), identifies the distro's package manager (apt/pacman/dnf/zypper), checks for missing packages via dpkg or ldconfig, and builds the exact install command for the user's system
- lutris/gui/dialogs/missing_deps.py: GTK dialog with package list, command preview with copy button, password entry, async install via sudo -S, and success/failure feedback with auto-close on success
- lutris/runners/wine.py: calls check_wine_dependencies() at the top of prelaunch() so the check runs before every Wine game launch

This fixes a long-standing silent failure mode where Wine games (especially launchers like Battle.net) would fail to start or show blank screens because 32-bit Vulkan or Mesa libraries were missing, with no actionable feedback.

### DIFF
--- a/lutris/gui/dialogs/missing_deps.py
+++ b/lutris/gui/dialogs/missing_deps.py
@@ -117,8 +117,6 @@ class MissingWineDepsDialog(Gtk.Dialog):
         password = self._password_entry.get_text()
         if not password:
             self._status_label.set_markup(_("<span foreground='red'>Please enter your password.</span>"))
-            # Stop dialog from closing
-            GLib.idle_add(lambda: self.run())
             return
 
         self._start_install(password)
@@ -196,5 +194,7 @@ class MissingWineDepsDialog(Gtk.Dialog):
         """Show the dialog and block until it is dismissed.
         Returns True if packages were successfully installed, False otherwise.
         """
-        self.run()
+        loop = GLib.MainLoop()
+        self.connect("destroy", lambda _: loop.quit())
+        loop.run()
         return self.install_succeeded
--- a/lutris/gui/dialogs/missing_deps.py
+++ b/lutris/gui/dialogs/missing_deps.py
@@ -1,11 +1,15 @@
 """Dialog for detecting and installing missing Wine/DXVK system dependencies."""
 
+import os
 import subprocess
 from gettext import gettext as _
 from typing import Optional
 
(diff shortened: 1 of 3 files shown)


danieljohnson2:
This is wild. This installs some of the dependencies of Wine explicitly with ``dnf`` or ``apt`` or whatnot. I don't think it's all of the dependencies though. Maybe we want this, maybe not.

But I wonder why we need the big distro list- you already have code to probe for the appropriate executable directly anyway.

ghostrunner42:
Good catch on the distro list. You're right that the binary probe already handles it cleanly. Simplified get_package_manager() to probe directly and dropped DISTRO_PACKAGE_MANAGERS entirely. Less maintenance surface, same behavior across the board. Thanks for the review!

ghostrunner42:
Looking at linux.py, I see LinuxSystem.get_missing_libs() already detects missing libraries via lib probing. Would it make more sense to hook into that as the detection source and map missing lib names to package names for the install command, rather than maintaining a parallel package list? Happy to refactor in that direction if so.

danieljohnson2:
I would think not- you are working with *packages* not libraries, and you are also doing different things for different GPUs. Does not seem like a great fit to me.

ghostrunner42:
Wanted to check in on the concept itself since the implementation questions seem settled. The original motivation was the Battle.net installation experience — deps were scattered across two different repos and the install flow was messy enough that it was hard to tell what was actually missing vs. misconfigured. That context probably explains why the package list is incomplete rather than comprehensive: it was scoped to what was visibly painful there, not a full Wine dep audit.

If the idea still has legs, happy to take direction on next steps — whether that means narrowing scope, expanding the package list more carefully, or something else entirely. And if it’s not the right fit for the project, that’s useful to know too. What’s your read?

danieljohnson2:
I don't know if we want this. I think it has attractions, but the boss will have to decide.

PR #6759, opened 2026-06-27
Fix OS name blank in system info on Mint and other distros

### LINKED ISSUE #6757: Lutris system infos: missing OS name and version + Wine default version
Hello,
concerning system info, on Mint, it misses OS name and version.
Example taken from a Mint user on our Discord support channel:

```
Vulkan support: YES
Esync support: YES
Fsync support: YES
Wine installed: YES
Gamescope: NO
Mangohud: NO
Gamemode: YES
Steam: NO
In Flatpak: YES
[System]
OS:
Arch: x86_64
Kernel: 6.17.0-35-generic
Lutris Version: 0.5.22
Python Version: 3.13.13 (main, Nov 10 2011, 15:00:00) [GCC 15.2.0]
Desktop: X-Cinnamon
Display Server: x11
[CPU]
Vendor: GenuineIntel
Model: Intel(R) Core(TM) i7-2600K CPU @ 3.40GHz
Physical cores: 4
Logical cores: 8
[Memory]
RAM: 31.3 GB
Swap: 2.0 GB
[Graphics]
Vendor: NVIDIA Corporation
OpenGL Renderer: NVIDIA GeForce GTX 1060 3GB/PCIe/SSE2
OpenGL Version: 4.6.0 NVIDIA 580.159.03
OpenGL Core: 4.6.0 NVIDIA 580.159.03
OpenGL ES: OpenGL ES 3.2 NVIDIA 580.159.03
Vulkan Version: 1.4.321
Vulkan Drivers: NVIDIA GeForce GTX 1060 3GB (1.4.312)
```

I don't know why Lutris can not obtain OS name and version on Mint.

Then, could it be possible to add this info?
Default Lutris Wine version set.

ghostrunner42:
distro.linux_distribution() has been deprecated since distro 1.6.0 and returns empty strings on newer versions for some distros (Linux Mint, etc.). The Flatpak path already used the correct modern API; align the non-Flatpak path to match.

Also filter empty strings from the OS join so distros without a codename don't produce trailing whitespace in the system info output.

Fixes #6757

### DIFF
--- a/lutris/util/linux.py
+++ b/lutris/util/linux.py
@@ -217,7 +217,7 @@ class LinuxSystem:  # pylint: disable=too-many-public-methods
         if self.is_flatpak():
             host_distro = distro.LinuxDistribution(root_dir="/run/host")
             return host_distro.name(), host_distro.version(), host_distro.codename()
-        return distro.linux_distribution()
+        return distro.name(), distro.version(), distro.codename()
 
     @staticmethod
     def get_arch() -> str | None:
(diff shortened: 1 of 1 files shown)


danieljohnson2:
It is not clear that this fixes the problem @legluondunet experienced- he didn't say it worked in a Flatpak and it's not clear to me why the current API would do anything different on Mint.

But moving off the deprecated method seems like a good idea, so I'll take it and ask @legluondunet to test + re-open if required.

PR #6761, opened 2026-06-28
Fix/tray exits on external game stop

ghostrunner42:
Fixes #6749

## Root cause

When a game is launched via a `lutris:` URI (which is what every Lutris
.desktop file does), `do_command_line()` unconditionally sets
`quit_on_game_exit = True`. This flag bypasses the tray icon check in
`_quit_if_hidden_and_idle()`, causing Lutris to quit when that game exits
— even when Lutris was already running with a tray icon and/or an open window.

## Fix

Only set `quit_on_game_exit = True` when Lutris has no persistent presence
at the time the URI is received — no tray icon and no visible window. This
preserves the original behaviour (launch-game-and-quit) for headless/scripted
use, while letting an already-running Lutris stay alive after a .desktop-launched
game closes.

PR #6760, opened 2026-06-27
Fix Wine runner arguments stripping quotes from Windows-style values

ghostrunner42:
Wine games commonly use arguments like INI="path with spaces" where the quotes must survive into the Wine command line so the Windows program receives them correctly. shlex.split() in POSIX mode (the default) strips the quotes after using them for grouping, breaking this pattern.

Adds keep_quotes=True to split_arguments() which uses a regex-based tokenizer that groups by quotes (preventing whitespace splitting within them) while preserving the quotes in the output. All other runners (Linux, Steam, ScummVM, MAME, etc.) keep the existing POSIX behavior since stripping quotes is correct for native Linux programs.

Fixes #4018

PR #6762, opened 2026-06-28
 Fix libretro cores not found when retroarch.cfg already exists

ghostrunner42:
`prelaunch()` only set `libretro_directory` (where RetroArch looks for cores)
when creating `retroarch.cfg` from scratch. If the config file already existed
— from a standalone RetroArch install, or from RetroArch's Flatpak defaults
pointing to `config/retroarch/cores` — the wrong directory was used at runtime.

Lutris installs cores to `RUNNER_DIR/retroarch/cores`
(`~/.var/app/net.lutris.Lutris/data/lutris/runners/retroarch/cores` on Flatpak)
but an existing config could point anywhere else, so no cores were found.

## Fix

On every launch, compare the current `libretro_directory` in the config against
the Lutris-managed cores directory. If they differ, update and save the config.
This is a no-op when already correct, and self-healing when mismatched.

### DIFF
--- a/lutris/runners/libretro.py
+++ b/lutris/runners/libretro.py
@@ -361,6 +361,13 @@ class libretro(Runner):
             retro_config.save()
         else:
             retro_config = RetroConfig(config_file)
+            # Always keep the cores path aligned with where Lutris installs them,
+            # even when loading an existing config (e.g. one created by a standalone
+            # RetroArch installation that may point to a different directory).
+            lutris_cores_dir = get_default_config_path("cores")
+            if retro_config.get("libretro_directory") != lutris_cores_dir:
+                retro_config["libretro_directory"] = lutris_cores_dir
+                retro_config.save()
 
         core = self.game_config.get("core")
         info_file = os.path.join(RETROARCH_DIR, f"info/{core}_libretro.info")


danieljohnson2:
This one seems to include PR #6761 and all the stuff it does. I'd rather merge *separate* PRs.

I think maybe what's happened is that you didn't do a branch per PR; you just did a bunch of PR's consecutively on your master branch, so each included the previous.

This means I'll have to merge them in order, if I can figure out the order.

danieljohnson2:
This also includes PR #6760 and I won't take that PR, so this needs to have that code removed.

Claude can probably sort this out- tell it you need each of your PRs to be a separate branch, so they don't share commits with each other.

danieljohnson2:
I think this PR may be good, but I am a little concerned that this is modifying the .cfg file - we'll be fighting with whoever it was installed that file that wasn't us. I don't know if we can avoid this. But is it desirable?

ghostrunner42:
Closing this in favor of `dj/libretro-core-paths` which covers this and more — reading both `libretro_directory` and `libretro_info_path` from retroarch.cfg, falling back to `~/.config/retroarch/cores` and `info`, and omitting `--libretro=` entirely when nothing resolves so RetroArch can use its own config. That approach is more complete and avoids any .cfg modification concern. Thanks for the review!

Issue #6755, opened 2026-06-25
Lutris [GoG] No installer found for...


Maximo0790:
### Bug description

I can't download linked to my Lutris account. When I proceed to install the game (any game) from the Lutris screen to the GoG account, it does not display any options.

### How to Reproduce

1. Go to 'GoG Sources'
2. Click on 'Every Games'
3. Do 'Install'
4. See 'Nothing'

### Expected behavior

Lutris doesn't release any error videos when I try to install a game on GoG

### Log output

```shell
[INFO:2026-06-25 16:52:47,941:application]: Starting Lutris 0.5.22
[INFO:2026-06-25 16:52:48,323:startup]: "card1" is Intel Iris Xe Graphics (8086:46a6 1025:159c i915) Driver 26.1.3
[INFO:2026-06-25 16:52:48,708:startup]: "card0" is NVIDIA GeForce RTX 4060 Laptop GPU (10de:28a0 1025:159c nvidia) Driver 580.159.03
[ERROR:2026-06-25 16:52:59,356:base]: No installer found for {'id': 102, 'service': 'gog', 'appid': '1790237640', 'name': '60 Minutes to Extinction: Escape Room', 'slug': '60_minutes_to_extinction_escape_room', 'icon': None, 'logo': None, 'url': None, 'details': '{"id": 1790237640, "availability": {"isAvailable": true, "isAvailableInAccount": true}, "title": "60 Minutes to Extinction: Escape Room", "image": "//images-1.gog-statics.com/c6e2d2635913aeca2753ad5a42ccc046283fa84b91f802cef963feb2b4528a22", "url": "/en/game/60_minutes_to_extinction_escape_room", "worksOn": {"Windows": false, "Mac": false, "Linux": false}, "category": "Adventure", "rating": 4, "isComingSoon": false, "isMovie": false, "isGame": true, "slug": "60_minutes_to_extinction_escape_room", "updates": 0, "isNew": true, "dlcCount": 0, "releaseDate": {"date": "2025-10-07 00:00:00.000000", "timezone_type": 1, "timezone": "+03:00"}, "isBaseProductMissing": false, "isHidingDisabled": false, "isInDevelopment": false, "extraInfo": [], "isGalaxyCompatible": true, "tags": [], "isHidden": false}', 'lutris_slug': '60-minutes-to-extinction-escape-room'}
[...]

PR #6767, opened 2026-06-29
created a workaround for issue #6755

gnarz:
after reading the entire product list from gog, for all products that have all worksOn entries set to False, we check with the url https://api.gog.com/v2/games/<id> which OSes are actually supported. This URL is from the gogdb website. It will not work for games that are in the inventory, but are no longer sold by gog. But until there is a proper solution, this will allow installation of games again.


danieljohnson2:
This seems important enough to take, but I'll have to make some fixes. It needs to be more resilient against variations in the JSON shape, and should log errors better. I'll take care of it.

PR #6770, opened 2026-06-29
feat(installer): support return_code: "*" to accept any exit code

### LINKED ISSUE #5806: [feature request?] Allow treating failed install as successful
From my experience, only game installs (made from Windows installers) ended with error code 0 are considered as successful.
ALL other error codes are treated as failed installs so no entry in Library is created and install directory wipe is offered.
Which is mildly inconvenient.
Because legacy installers of old games tend to fail on their own.
Because GOG's installers tend to die by timeout on weak hardware.
Because the game could be installed just fine but it's some unnecessary redistributable whose install failure fails whole process (mmmmm, old bundled PhysX versions, yum!)
Whatever hiccup occurred, it necessiates restarting WHOLE install process from scratch even if everything is Just Fine.

I as end user with hands mounted slightly above waist can decide if install was successful or not on my own.
Like, in case #5805 installed app is fully functional and I can cope with whatever hiccups it might have down the line.
It's good to me.

However, to have it my way I currently have to install some completely irrelevant app (to make sure that error code will be 0 so library entry appears) and only THEN try installing problematic software.

Which was fun only first half-a-hundred times and tolerable first few hundred times.

Maybe it's possible to streamline it somehow.

ghostrunner42:
Closes #5806

### Problem

Installer scripts can already set `return_code: 256` to accept a specific non-zero exit code. But some installers — particularly legacy Windows titles, GOG games on weak hardware (timeouts), and old bundled redistributables like PhysX — exit with unpredictable non-zero codes even when the install succeeded. There's no way to say "accept whatever exit code this produces."

### Fix

Accept `"*"` as a wildcard value for `return_code`:

```yaml
- task:
    name: wineexec
    executable: setup.exe
    return_code: "*"
```

One-line change in `_monitor_task` — if `accepted_return_code` is `"*"`, skip the exit code check entirely. Also updates the installer documentation to describe the wildcard.

This is strictly additive: existing `return_code: 256` behavior is unchanged, and the default (reject anything non-zero) is unchanged.

### DIFF
--- a/docs/installers.rst
+++ b/docs/installers.rst
@@ -632,7 +632,9 @@ possible to call functions from other runners by prefixing the task name with
 the runner's name (e.g., from a dosbox installer you can use the wineexec task
 with ``wine.wineexec`` as the task's ``name``)
 If the command you will run in the task doesn't exit with a return code of 0,
-you can specify an accepted return code like ``return_code: 256``
+you can specify an accepted return code like ``return_code: 256``, or use
+``return_code: "*"`` to accept any return code (useful for installers known
+to exit non-zero on success, such as some legacy or GOG installers).
 
 Currently, the following tasks are implemented:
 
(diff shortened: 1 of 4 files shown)


ItsAllAboutTheCode:
This PR contains other changes from closed Pull Request.
You need to remove commit 40c954c and 4b70f77 from this branch.

PR #6778, opened 2026-07-02
Fix Ubisoft Connect install when latest Wine (GE-Proton) is already installed

ghostrunner42:
**Bug:** When installing Ubisoft Connect via the service, Lutris prompts to install "Wine's latest". You have to click through the install step to proceed. If the latest Wine (currently GE-Proton via umu) is *already* installed, the runner installation step errors out with something like "The 'ge-proton' version of the 'wine' runner can't be downloaded."

This blocks completing the Ubisoft Connect installation even when you already have a working latest Wine setup.

**Root cause:**
- The "ubisoft-connect" installer (and similar launcher installers) rely on the Wine runner being present.
- `get_runners_to_install` / `is_installed_for` (with `use_api=True`) resolves the version to the current default (GE_PROTON_LATEST / "ge-proton").
- For GE_PROTON_LATEST, `is_installed` ultimately relies on `get_umu_path()` succeeding.
- If that check decides "not installed" (or in edge cases/version name resolution), `runner.install()` is called.
- Wine runner install for "ge-proton" has no "url" (it's managed on-demand by umu, not a fixed downloadable version), so it raises `RunnerInstallationError`.
- Even the `install_dialog` yes/no path calls `install()` unconditionally on "yes", without re-checking, leading to the error if present.

**Fix:**
- Add an early-return guard at the start of `Runner.install()`: if `is_installed(version=...)` (or without for the runner), log and invoke the `callback` (so installer chains continue) then return. Makes the install step idempotent/safe.
- Widen the except in `Wine.is_installed_for()` to also catch `MissingExecutableError` (plus log unexpected errors) so "not present" cases for ge-proton etc. cleanly return False instead of leaking exceptions.
- This ensures "continue if the latest wine is already installed" without erroring the Ubisoft (or other) installation flow.

The pre-checks should usually avoid unnecessary calls to `install()`, but the guard makes the step itself robust (handles races, name mismatches between "latest" snapshot and actual installed Proton dir, etc.).

**Testing notes:**
- Compile clean.
- Basic import/smoke checks for runners pass.
- For full verification: install/run the Ubisoft Connect service flow with latest Wine already present (umu + a GE-Proton) vs. not present. The prompt should only appear when truly needed, and clicking through (or auto) should not error when present.
- Recommend running `python3 -m pytest tests/ -q -k "wine or runner or installer or ubisoft"` locally.

This is a minimal, targeted fix as described. No behavior change for cases where Wine really needs installing.

(The cleanup work on this branch was separate; this is a new bugfix commit.)

### DIFF
--- a/lutris/runners/flatpak.py
+++ b/lutris/runners/flatpak.py
@@ -83,7 +83,7 @@ class flatpak(Runner):
         },
     ]
 
-    def is_installed(self, flatpak_allowed: bool = True) -> bool:
+    def is_installed(self, flatpak_allowed: bool = True, version: str | None = None) -> bool:
         return _flatpak.is_installed()
 
     def get_executable(self) -> str:
--- a/lutris/runners/linux.py
+++ b/lutris/runners/linux.py
@@ -115,7 +115,7 @@ class linux(Runner):
         """Linux programs should get individual shader caches if possible."""
         return self.game_path or self.shader_cache_dir
 
-    def is_installed(self, flatpak_allowed: bool = True) -> bool:
+    def is_installed(self, flatpak_allowed: bool = True, version: str | None = None) -> bool:
         """Well of course Linux is installed, you're using Linux right ?"""
         return True
 
--- a/lutris/runners/pico8.py
+++ b/lutris/runners/pico8.py
@@ -131,7 +131,7 @@ class pico8(Runner):
     def get_run_data(self):
         return {"command": self.launch_args, "env": self.get_env(os_env=False)}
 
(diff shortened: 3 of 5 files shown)


danieljohnson2:
Is this just for robustness, or there an issue that this actually fixes?

strycore:
> Is this just for robustness, or there an issue that this actually fixes?

This shouldn't be merged, ge-proton is not a proton build shipped by us, it's an indicator to use umu-launcher.

strycore:
this should fix the issue: https://github.com/lutris/lutris/pull/6780

danieljohnson2:
Better close this lest I merge it later.

Issue #5703, opened 2024-11-04
Can't download from moddb with Lutris flatpak - 403 Client Error: Forbidden 


eskay993:
### Bug description

When installing a yaml script with has a moddb file link,. the script fails with ` 403 Client Error: Forbidden ` just before the file selection screen.  There is no way to continue the install process and select the file manually, short of modifying the script to remove the modddb url.

This only seems to happen with the flatpak version of Lutris.  The version shipping with CachyOS does not have this issues, and the moddb file is downloaded successfully.

I don't know exactly when this started happening but I first noticed it yesterday. It was working fine before.

My system and all flatpaks are fully up to date. The version of Lutris reported by the flatpak is `lutris-0.5.17`.  The system version installed by pacman is `lutris-git 0.5.17.r385.g66d4170-1`

### How to Reproduce

1. Open flatpak version of Lutris
2. Go to `+` > `install from local install script`
3. Select an install script with a moddb file link. Here is one that I created - [resident-evil-gog.yaml](https://github.com/eskay993/gamefiles/blob/main/resident-evil/scripts/resident-evil-gog.yaml)
4. Click through the install process until you get to the file selection screen. The error will occur just before you get to that screen.

### Expected behavior

The file selection screen to appear and the moddb file to be downloaded successfully, as it was working previously.

Additionally, if there is an issue with moddb, I would expect the file selection screen to still appear, and any issues indicated on that screen. This would allow the user to manually select the file and continue the install when an issue with moddb occurs.

### Log output

```shell
ERROR    2024-11-04 17:11:23,686 [jobs.target:32]:Error while completing task <bound method LutrisInstaller.prepare_game_files of <lutris.installer.installer.LutrisInstaller object at 0x7faf87851590>>: <class 'requests.exceptions.HTTPError'> 403 Client Error: Forbidden for url: https://www.moddb.com/downloads/re1-classic-rebirth-high-quality-video-pack
  File "/app/lib/python3.11/site-packages/lutris/util/jobs.py", line 30, in target
    result = self.function(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/lib/python3.11/site-packages/lutris/installer/installer.py", line 165, in prepare_game_files
    file.set_url(ModDB().transform_url(file.url))
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #6782, opened 2026-07-04
installer: fall back to file selection when ModDB resolution fails

masonliiu:
ModDB URL resolution fails before the installer file-selection page is shown, displaying a 403 response from ModDB/Cloudflare. When that happens, the installer currently aborts and the user cannot continue with installation.

This pr keeps the existing automatic ModDB mirror resolution path unchanged if it succeeds. However, if ModDB resolution fails with a request error, Lutris now logs the original URL and falls back to the existing manual file-selection flow for that file, providing a link to the designated installation site as well.

Refs #5703

## Tests completed

- `python -m nose2 tests._test_installer tests._test_moddb_helper`
- `ruff format lutris/installer/installer.py tests/_test_installer.py --check`
- `ruff check lutris/installer/installer.py tests/_test_installer.py`
- `python -m compileall -q lutris/installer/installer.py tests/_test_installer.py`

### DIFF
--- a/lutris/installer/installer.py
+++ b/lutris/installer/installer.py
@@ -5,6 +5,8 @@ from functools import cached_property
 from gettext import gettext as _
 from pathlib import Path
 
+import requests
+
 from lutris.api import get_game_details
 from lutris.config import LutrisConfig, write_game_config
 from lutris.database.games import add_or_update, get_game_by_field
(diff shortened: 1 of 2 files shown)


danieljohnson2:
Better than nothing! Merging!

PR #6794, opened 2026-07-13
Update pre-commit

ford220102:
This PR fixes two logic bugs in `.hooks/pre-commit`:

1. `skip_static_analysis_checks` was computed with `bool(stdout)`, which
   is always True for any non-empty string — including the literal text
   "false\n" returned by `git config --type=bool`. This meant explicitly
   setting the config to false was silently treated as true, skipping
   checks when it shouldn't. Fixed by comparing the stripped output
   against "true" explicitly.

2. The "command is missing" error message interpolated `command_path`
   (which is always None in that branch — that's why we're reporting
   it as missing) instead of `command`, so it always printed "None
   command is missing" regardless of which tool was actually absent.
   Fixed to reference the actual command name.

### DIFF
--- a/.hooks/pre-commit
+++ b/.hooks/pre-commit
@@ -18,10 +18,6 @@ def main():
     comp_process = subprocess.run(
         ["git", "config", "--type=bool", "hooks.skip-static-analysis-checks"], capture_output=True
     )
-    # FIX: bool() on a non-empty string is always True in Python, so the
-    # original `bool(comp_process.stdout.decode("utf-8"))` treated the
-    # literal string "false\n" as truthy and skipped checks incorrectly.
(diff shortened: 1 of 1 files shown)


danieljohnson2:
You need to explain what you are doing in your PR message, not in a bunch of "FIX:" comments.

ItsAllAboutTheCode:
> You need to explain what you are doing in your PR message, not in a bunch of "FIX:" comments.

I agree about the PR message needing to explain what changes were made and there is no need for the FIX comments.
The changes themselves look good however.

ford220102 (commented):
Removed the inline FIX comments as requested — the explanation is now
only in the PR description. Let me know if anything else needs
adjusting.

danieljohnson2:
Alright, I'll merge it.

PR #6803, opened 2026-07-21
Fixed Unit Test builds when running on CI / Enabled util test

ItsAllAboutTheCode:
Removed the importlib machinery used in `_test_gog_dlcs.py` and `_test_collection_progress.py` to fix issues with due to test module discovery order causing bespoke mocked modules to be placed in the sys.modules.

Instead the builtin Python UnitTest Mock and Patch framework is used to properly patch the methods needed to run the UnitTest instead of manipulating the importlibs in a bespoke manner.

Added missing `__init__.py` to the `lutris/tests/util` directory to allow discovery of those UnitTest by the nose2 and the Python builtin UnitTest framework. Those test weren't not being run by either `nose2` nor the `UnitTest` framework.

Fixed the UnitTest that imports the pytest module. Replaced the usage of `pytest` inside of the `lutris/tests/util` directory to use the python UnitTest framework. Lutris does NOT specify a dependency on `pytest `and therefore when those test were actually enabled to be run, they would complain about the missing `pytest` import.

### DIFF
--- a/tests/services/_test_gog_dlcs.py
+++ b/tests/services/_test_gog_dlcs.py
@@ -9,153 +9,11 @@ Tests the ability to:
 5. Merge results into unified JSON
 """
 
-import importlib.util
-import os
-import sys
-import types
 import unittest
+from unittest.mock import MagicMock, patch
 
-REPO_ROOT = os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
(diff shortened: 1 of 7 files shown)


danieljohnson2:
Looks like this is just unit test fixes, so it it's big but it shouldn't be a problem. I'll merge it.

PR #6812, opened 2026-07-27
Prevent crashes from numeric Wine component versions

ford220102:
# Prevent crashes from numeric Wine component versions

## Summary

- Normalize Wine component version overrides to strings before they are used by
  the DLL managers.
- Normalize `dxvk_version` before checking Vulkan compatibility warnings.
- Document that numeric-looking YAML versions must be quoted.
- Add `dxvk_version` to the installer schema.

## Problem

YAML parses an unquoted value such as `dxvk_version: 2.6` as a `float`. A game
configuration loaded from disk can therefore reach the Wine runner as a number.
The current code calls string methods on that value:

```text
AttributeError: 'float' object has no attribute 'lower'
```

The same type mismatch can reach the DXVK compatibility warning path, where
`startswith()` is called on the version.

## Root Cause

Installer substitutions already produce strings, but persisted or manually
edited YAML is loaded without coercing version values. The DLL manager and the
DXVK warning path assumed the value was always a string.

## Fix

- Convert non-`None` DLL manager versions to strings in `DLLManager.__init__`.
- Convert a configured DXVK version to a string before using string methods.
- Make the schema and documentation identify the expected string type.

The patch does not silently reinterpret unsupported configuration nesting such
as a top-level `config:` key. That is not a Lutris game configuration section;
the executable and runner sections must remain at the YAML top level.

## Verification

- `PYTHONDONTWRITEBYTECODE=1 PYTHONPATH=. python3 -m nose2 --exclude-ci -v`
  - 469 tests passed.
- `PYTHONDONTWRITEBYTECODE=1 python3 -m compileall -q lutris tests`
- `git diff --check`
- `dpkg-buildpackage -us -uc -b`
  - produced `lutris_0.5.23_all.deb`.

### DIFF
--- a/docs/installers.rst
+++ b/docs/installers.rst
@@ -281,7 +281,9 @@ Example: ``version: staging-2.21-x86_64``
 
 ``dxvk``: Use this to disable DXVK if needed. (``dxvk: false``)
 
-``dxvk_version``: Use this to define a specific DXVK version. (``dxvk_version: 1.10.3``)
+``dxvk_version``: Use this to define a specific DXVK version. Version values
+must be quoted because YAML can parse numeric-looking versions as numbers.
+(``dxvk_version: "1.10.3"``)
 
 ``esync``: Use this to disable esync. (``esync: false``)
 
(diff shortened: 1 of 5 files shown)


danieljohnson2:
Looks legit. I will merge it!

PR #6821, opened 2026-08-01
Fix GOGs OS choosing algorithm

### LINKED ISSUE #6788: [REGRESSION] GOG wont download the windows files when the runner is linux
<s> Whenever a lutris attempts to download a game as specified by an installer it will check with the service provider the different versions that it offers. When the runner specified in said installer is either linux or wine, lutris will attempt to download the version corresponding to said OS. 
t'd useful to allow certain installers to download the game for OSes that are not the ones indicated in the runner.
Ideally the installer should control which version to download, either by a new child section to the file or by parsing what comes after the _N/A_.
Nevertheless, since this is mainly useful for games in where there is no native linux version but rather 3rd party source ports, it could be made so that when there exists no download option in the specified OS, the next OS in a list is downloaded. </s>

UPDATE: After further testing I've found it's a regression between 0.5.22 and git. It seems to be limited to the GOG service, more precisely to a partial rewrite in the _get_installers_ function in _lutris/services/gog.py_. It doesnt check if the installer has a linux version and therefore when a title doesn't, it doesn't replace it with the windows one.

Mineblock01:
Makes it so that the GOG service chooses the installer that makes sense by the runner, but allows it to fall back on other installers if such an installer isn't available.
Fixes https://github.com/lutris/lutris/issues/6788.

### DIFF
--- a/lutris/services/gog.py
+++ b/lutris/services/gog.py
@@ -444,28 +444,20 @@ class GOGService(OnlineService):
                     all_extras[product.get("title", "").strip()] = extras
         return all_extras
 
-    def get_installers(
-        self, downloads: dict[str, list[dict]], runner_name: str | None, language: str = "en"
-    ) -> list[dict]:
+    def get_installers(self, downloads: dict[str, list[dict]], runner: str, language: str = "en") -> list[dict]:
         """Return available installers for a GOG game"""
-        if runner_name:
-            allowed_os = {self.runner_to_os_dict[self._normalize_runner_name(runner_name)]}
-        else:
-            allowed_os = {"linux", "windows"}
-        all_installers = downloads.get("installers", [])
-        # Prefer native Linux installers if available, otherwise fall back to Windows
-        # (scripts may need Windows installers for data extraction, engine reimplementations, etc.)
-        selected_installers = sorted(
-            [installer for installer in all_installers if installer["os"] in allowed_os],
-            key=lambda i: i["os"] != "linux",
-        )
-        selected_language = self.determine_language_installer(selected_installers, language)
-        return [installer for installer in selected_installers if installer["language"] == selected_language]
-
-    def _normalize_runner_name(self, runner: str) -> str:
-        """Normalize a runner name to either 'linux' or 'wine'."""
-        return "linux" if runner == "linux" else "wine"
-
+        # Filter out Mac installers
+        gog_installers = [installer for installer in downloads.get("installers", []) if installer["os"] != "mac"]
+        filter_os = self.runner_to_os_dict.get(runner)
+        # Chechs if there are installers that pass the OS filters and selects those to install
+        filtered_installers = [installer for installer in gog_installers if installer["os"] == filter_os]
+        if filtered_installers:
+            gog_installers = filtered_installers
+        return [
+            installer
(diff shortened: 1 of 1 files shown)


danieljohnson2:
This needs some work. You've made the ``runner`` parameter mandatory, and your code will crash if None is passed here. The original would return games for Linux *or* Windows.

Mineblock01:
> You've made the runner parameter mandatory

It is already mandatory. If you try to pass an installer without it, lutris will error out, saying that there is field "runner" is missing. If you remove the check the code fails on other places.

> your code will crash if None is passed here.

If you pass None as the runner, the code works fine, it just doesn't filter by the runner.

> The original would return games for Linux or Windows.

Yes, but if you wanted to automatically download a Windows installer with the Linux runner or the other way around, then (under certain conditions) this code would be able to do it, while the previous one couldn't.

I would also like to mention that this code, almost in it's entirety was previously in the [lutris codebase](https://github.com/lutris/lutris/blob/b0d42dd23d502fb2553c81d8aeddcfcf061f2acc/lutris/services/gog.py).

danieljohnson2:
I thought it looked familiar! I don't like reverting so much as a "fix" but I think I understand the issue- every install script specifies a runner, and that means you can only *ever* get that version. If it does not exist you are out of luck.

Let me see if I can vibe up an alternative fix.

danieljohnson2:
OK, try [this branch](https://github.com/lutris/lutris/tree/dj/gog-installer-os-fallback). Does that address your issue?

It still applies the runner, still has other fixes in it, but if there's no installer for the correct runner, it just provides them all, Linux first. Maybe that'll do it?

Mineblock01:
> I don't like reverting so much as a "fix"

I understand, though the changes made in this function after the previous version were entirely over the OS issue, and so I thought it fine to revert.

> OK, try [this branch](https://github.com/lutris/lutris/tree/dj/gog-installer-os-fallback). Does that address your issue?

It seems like it would work, though I think it could be improved. First, the _normalize_runner_name function could be removed and replaced with the get() method. Second, there is no need to sort the all_installers list if any non wine runners prefer the linux runner.

Before seeing the branch, I made some improvements to my patch to prefer Linux over windows and use mac if nothing else is available. I think at least the mac change is preferable.

danieljohnson2:
I am unsure. Did you try my branch? I'd like to know if it even works.

I think ``_normalize_runner_name`` is all about not using the Mac installers- so the fundamentally issue is whether it's useful to provide Mac installers. So I ask- when do we want these? Is there a concrete case of this?

Mineblock01:
> I am unsure. Did you try my branch? I'd like to know if it even works.

I have now. It works as expected.

> I think _normalize_runner_name is all about not using the Mac installers

No, its about transforming the runner to the os. The already existing dict runner_to_os_dict already does that, and using the [get](https://docs.python.org/3/library/stdtypes.html#dict.get) method one can do the exact thing while being easier to update.

> the fundamentally issue is whether it's useful to provide Mac installers. So I ask- when do we want these? Is there a concrete case of this?

Probably when its a mac exclusive title. Or if [darling](https://darlinghq.org) starts playing games, if its the selected runner.

PR #6825, opened 2026-08-06
Don't prevent sleep by default

addmix:
This change prevents Lutris from interfering with system power management without the user's knowledge/request. Considering this is hidden with the "Advanced" options, many users will likely not know it is enabled by default.

Background:
I created a service to allow programs to request sleep inhibitors in Bazzite/SteamOS game mode: https://github.com/addmix/Bazzite-Deck-Sleep-Inhibitor
Naturally, programs inhibiting sleep without user knowledge/consent or an active reason to inhibit sleep (playing non-interactive content, downloading, background processing) becomes an issue, especially in the context of handheld devices. I believe that this new default should benefit the large majority of users, and hopefully avoid inconsistency with system sleep.

### DIFF
--- a/lutris/sysoptions.py
+++ b/lutris/sysoptions.py
@@ -206,7 +206,7 @@ system_options: list[dict[str, Any]] = [  # pylint: disable=invalid-name
         "option": "disable_screen_saver",
         "label": _("Prevent sleep"),
         "type": "bool",
-        "default": SCREEN_SAVER_INHIBITOR is not None,
+        "default": False,
         "advanced": True,
         "condition": SCREEN_SAVER_INHIBITOR is not None,
         "help": _("Prevents the computer from suspending when a game is running."),


addmix:
Any update on this? It's incredibly inconvenient that my HTPC setup is never allowed to sleep due to Lutris deciding that the computer shouldn't be allowed to sleep, and having to manually change the setting for every single title.

danieljohnson2:
I think it is a reasonable default; and I think you already know where to change this setting if you need a different behavior. I do not see that we should change anything here.

addmix:
I will admit, I hadn't seen the global setting prior to now. That definitely makes things easier.

However, I still say that the default should change, considering that Linux handhelds are a significant portion of the Lutris userbase, where power saving is a primary concern. It just adds another setting for users to troubleshoot/change when setting up their linux install.

strycore:
Note that this setting should only affect the experience *when a game is running*. If sleep is prevented just by having Lutris opened or after a game has exited, then it's a bug. 
The reason this setting was added is that controller input does not count toward inputs that prevent the machine from going to sleep like keyboard and mouse, and you end up with your computer going to sleep while you're playing. This is a much more disrupting problem than a machine not going to sleep while a game is running,

addmix:
Maybe you have some insight or experience that I don't have, but in my experience over the past few years has not included any instances of controller input not counting towards "user activity", causing my devices to fall asleep (both windows and Linux, primarily Bazzite). In that case, wouldn't it be an OS configuration issue that should be remedied elsewhere? It seems a bit overbearing to apply a workaround for a potential issue with a limited scope to all installations indiscriminately.

Issue #6231, opened 2025-11-14
[beta 0.5.20] Ubisoft Connect no game pictures


legluondunet:
### Bug description

Ubisoft COnnect sources, after login to my Ubisoft connect account and click on sync, I obtain my Ubisoft Connect games list in Lutris but without games pictures:
<img width="806" height="644" alt="Image" src="https://github.com/user-attachments/assets/4a9c810b-8b77-477c-86aa-6616ae65b9fa" />

### How to Reproduce

1) activate Ubisoft Connect sources
2) login to your Ubisoft connect account
3) click on sync icon 
--> Ubisoft Connect games list in Lutris have no picture

### Expected behavior

sync with pictures

### Log output

```shell
$ ./bin/lutris -d
2025-11-14 23:45:02,414: The Battle.net source is unavailable because Google protobuf could not be loaded: Descriptors cannot be created directly.
If this call came from a _pb2.py file, your generated code is out of date and must be regenerated with protoc >= 3.19.0.
If you cannot immediately regenerate your protos, some other possible workarounds are:
 1. Downgrade the protobuf package to 3.20.x or lower.
 2. Set PROTOCOL_BUFFERS_PYTHON_IMPLEMENTATION=python (but this will use pure-Python parsing and will be much slower).

More information: https://developers.google.com/protocol-buffers/docs/news/2022-05-06#python-updates
INFO     2025-11-14 23:45:02,606 [application.do_command_line:479]:Starting Lutris 0.5.20
INFO     2025-11-14 23:45:02,671 [startup.run_all_checks:120]:"card1" is AMD Radeon RX 6600M (1002:73ff 1002:0123 amdgpu) Driver 25.2.3
INFO     2025-11-14 23:45:02,733 [startup.run_all_checks:120]:"card0" is AMD Radeon Graphics (1002:1638 1002:0123 amdgpu) Driver 25.2.3
DEBUG    2025-11-14 23:45:02,791 [xrandr.get_resolutions:91]:Retrieving resolution list
DEBUG    2025-11-14 23:45:02,804 [xrandr._get_vidmodes:18]:Retrieving 71 video modes from XrandR
DEBUG    2025-11-14 23:45:02,904 [path_cache._update_missing_games:114]:Checking for missing games
DEBUG    2025-11-14 23:45:02,932 [api.check_stale_runtime_versions:52]:Modified at ven. 14 nov. 2025 23:27:52, will update after sam. 15 nov. 2025 05:27:52
[...]

Issue #6229, opened 2025-11-14
[beta 0.5.20] disable Steam sources do not hide Steam games in Lutris library


legluondunet:
### Bug description

When I disable Steam in Lutris Preferences/sources, Steam games still showing in Lutris library.

### How to Reproduce

1) enable Steam sources in Lutris preferences/Sources
2) sync your Steam libray with Lutris. Now all your Steam games showing in your Lutris library
3) Now disable Steam sources in Lutris preferences/Sources, Steam games are still showing in your Lutris library

### Expected behavior

When you disable Steam sources in Lutris preferences/Sources, Steam games should no more appearing in your Lutris library

### Log output

```shell
not necessary
```

### System Information

```shell
Vulkan support: YES
Esync support: YES
Fsync support: YES
Wine installed: YES
Gamescope: YES
Mangohud: YES
Gamemode: YES
Steam: YES
In Flatpak: NO
[System]
OS: Manjaro Linux 25.0.10 Zetar
Arch: x86_64
Kernel: 6.16.8-1-MANJARO
Lutris Version: 0.5.20
Desktop: XFCE
Display Server: x11
[CPU]
Vendor: AuthenticAMD
Model: AMD Ryzen 9 5900HX with Radeon Graphics
Physical cores: 8
Logical cores: 16
[Memory]
RAM: 30.8 GB
Swap: 16.2 GB
[Graphics]
Vendor: AMD
OpenGL Renderer: AMD Radeon RX 6600M (radeonsi, navi23, LLVM 20.1.8, DRM 3.64, 6.16.8-1-MANJARO)
OpenGL Version: 4.6 (Compatibility Profile) Mesa 25.2.3-arch1.2
OpenGL Core: 4.6 (Core Profile) Mesa 25.2.3-arch1.2
OpenGL ES: OpenGL ES 3.2 Mesa 25.2.3-arch1.2
Vulkan Version: 1.4.321
Vulkan Drivers: AMD Radeon RX 6600M (RADV NAVI23) (1.4.318), AMD Radeon Graphics (RADV RENOIR) (1.4.318)
```

danieljohnson2:
This seems like it might be helped by #6212, no?

legluondunet:
https://github.com/lutris/lutris/pull/6212 solution looks answer this issue.
Nice work!

Issue #6000, opened 2025-03-16
can't launch winetricks while using UMU-latest


ismi-ismi:
### Bug description

i click on the "winettricks" option, nothing happened, winetricks fail to launch

how to edit the lutris command to open Winetricks to not include '--gui' because that is what makes wine doesn't open

### How to Reproduce

1. Go to any application
2. Click on wine logo
3. click Winetricks
4. nothing happened (ERROR: Value is not a Winetricks verb: '--gui')

### Expected behavior

Lutris should launch Winetricks

### Log output

```shell
2025-03-17 03:50:56,757: The Battle.net source is unavailable because Google protobuf could not be loaded: No module named 'google'
2025-03-17 03:50:59,591: Starting Lutris 0.5.20
2025-03-17 03:50:59,671: "card1" is Intel UHD Graphics 620 (8086:5917 1043:14be i915) Driver None
2025-03-17 03:50:59,786: "card0" is NVIDIA GeForce 930MX (10de:134e 1043:14be nvidia) Driver 550.120
Started initial process 32829 from /usr/bin/umu-run winetricks --gui
Start monitoring process.
ERROR: Value is not a winetricks verb: '--gui'
Monitored process exited.
Initial process has exited (return code: 256)
All processes have quit
Exit with return code 256
```

### System Information

```shell
Vulkan support: YES
Esync support: YES
Fsync support: YES
Wine installed: YES
Gamescope: NO
Mangohud: NO
Gamemode: YES
Steam: NO
In Flatpak: NO
[System]
OS: Linux Mint 22.1 Xia
Arch: x86_64
Kernel: 6.8.0-54-generic
Lutris Version: 0.5.20
Desktop: X-Cinnamon
Display Server: x11
[CPU]
Vendor: GenuineIntel
Model: Intel(R) Core(TM) i5-8250U CPU @ 1.60GHz
Physical cores: 4
Logical cores: 8
[Memory]
RAM: 7.6 GB
Swap: 1.9 GB
[Graphics]
Vendor: NVIDIA Corporation
OpenGL Renderer: NVIDIA GeForce 930MX/PCIe/SSE2
OpenGL Version: 4.6.0 NVIDIA 550.120
OpenGL Core: 4.6.0 NVIDIA 550.120
OpenGL ES: OpenGL ES 3.2 NVIDIA 550.120
Vulkan Version: 1.3.275
[...]

PR #6826, opened 2026-08-06
Fix Winetricks UMU, Ubisoft Connect cover fallbacks, and Steam source filtering

ford220102:
This Pull Request resolves three client-side issues from the Lutris tracker:

1. **Fix Winetricks with UMU (#6000)**: Handle Winetricks differently under UMU by omitting the `--gui` argument, which causes `umu-run` to fail. Winetricks launched without arguments defaults to launching in GUI mode.
2. **Fix Ubisoft Connect no game pictures (#6231)**: Provide a robust fallback to `bannerUrl` and `backgroundUrl` when the API response lacks `coverUrl` and local config lacks `thumbImage`.
3. **Fix Steam source hiding (#6229)**: Exclude service games from the general games view if the source itself is disabled entirely in the settings (not just the individual source view checkbox).

### DIFF
--- a/.github/workflows/publish-appimage.yml
+++ b/.github/workflows/publish-appimage.yml
@@ -22,7 +22,7 @@ jobs:
       contents: write
     steps:
       - name: Checkout the repository
-        uses: actions/checkout@v6
+        uses: actions/checkout@v4
 
       - name: Build the AppImage
         env:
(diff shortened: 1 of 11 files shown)


danieljohnson2:
Sorry to take so long to get back to you on this.

I think this needs to be broken up into 3 PRs; it will be much easier to review that way. There also seem to be some extra stuff about workflow in here that does not seem relevant.

Issue #6828, opened 2026-08-09
Lutris does not start on Debian Forky


timconinx:
### Bug description

I also reported this to the [Debian bugtracking](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1143863), but as the git version suffers from the same issue, I'm also reporting it here.

After upgrading Gnome packages to the Debian Forky version, Lutris doesn't start due to a dependency(?) issue. The git version has the same error:
```
> bin/lutris 
Traceback (most recent call last):
  File "/home/tim/Projects/lutris/bin/lutris", line 68, in <module>
    from lutris.gui.application import LutrisApplication
  File "/home/tim/Projects/lutris/lutris/gui/application.py", line 38, in <module>
    from lutris.game import GAME_START, GAME_STOPPED, Game, export_game, import_game
  File "/home/tim/Projects/lutris/lutris/game.py", line 17, in <module>
    from lutris.config import LutrisConfig
  File "/home/tim/Projects/lutris/lutris/config.py", line 8, in <module>
    from lutris import settings, sysoptions
  File "/home/tim/Projects/lutris/lutris/sysoptions.py", line 11, in <module>
    from lutris.util.display import DISPLAY_MANAGER, SCREEN_SAVER_INHIBITOR, is_compositing_enabled, is_display_x11
  File "/home/tim/Projects/lutris/lutris/util/display.py", line 141, in <module>
    DISPLAY_MANAGER = get_display_manager()
  File "/home/tim/Projects/lutris/lutris/util/display.py", line 135, in get_display_manager
    return DisplayManager(screen)
  File "/home/tim/Projects/lutris/lutris/util/display.py", line 67, in __init__
    self.rr_screen = GnomeDesktop.RRScreen.new(screen)
                     ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/gi/module.py", line 123, in __getattr__
    raise AttributeError(f"{self.__name__!r} object has no attribute {name!r}")
AttributeError: 'gi.repository.GnomeDesktop' object has no attribute 'RRScreen'
``` 

### How to Reproduce

Start lutris on an up to date Debian Forky system

### Expected behavior

Start correctly without dependency issue

### Log output

```shell
> bin/lutris -d
Traceback (most recent call last):
  File "/usr/games/lutris", line 68, in <module>
    from lutris.gui.application import LutrisApplication
[...]

PR #6833, opened 2026-08-12
Deprecated/removed attributes in GnomeDesktop

timconinx:
This PR solves https://github.com/lutris/lutris/issues/6828 by checking whether the - previously deprecated - attributes of GnomeDesktop have actually been removed


danieljohnson2:
Looks good! Thanks!

PR #6835, opened 2026-08-14
Fix category search in sources

jmcs:
Category searches inside a source were failing because the filter compared local game IDs against service app IDs. This fixes that mismatch for source-scoped searches.


danieljohnson2:
Good catch, thanks for the fix!

PR #6848, opened 2026-08-22
build: fix readme path and avoid importing lutris in setup.py

ford220102:
Two build fixes:

1. `pyproject.toml` referenced `README.md` but the repository only contains `README.rst`, breaking PEP 621 builds (e.g. the AppImage build fails with `File '/src/README.md' cannot be found`).

2. `setup.py` did `from lutris import __version__`, which imports `lutris/__init__.py` and its `gi.require_version("Gtk", "3.0")` calls. That forces PyGObject/Gtk to be installed just to build an sdist. Read the version string from the file instead.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,7 +8,7 @@ description = """
 Lutris is a video game preservation platform that helps you install and play video games from all eras and from most gaming systems.
 By leveraging and combining existing emulators, engine re-implementations and compatibility layers, it gives you a central interface to launch all your games.
 """
-readme = "README.md"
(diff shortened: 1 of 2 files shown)


danieljohnson2:
I'm cribbing your fix to ``pyproject.toml``- thanks for spotting that.

But using a regular expression to parse data out of source code just makes my skin crawl.

I'd rather change Lutris so we can import the version without getting GTK too, or else use the ``ast`` module to parse through the code. Or just accept it as it is.

PR #6867, opened 2026-09-03
ensure unique slug on add_game to prevent collision

loucass:
Fixes duplicate `slug` rows.

Repro: add two games then i found out that both get `slug='undertail'`,
with one shared `coverart/undertail.png` which is unwanted behavior.

i changed `lutris/database/games.py`: `add_game()` now dedups via new
`get_unique_slug()` helper that generate something like
`undertail, undertail-2, ...` a while loop so taken `-2` skips to `-3`
and so on... Explicit `slug=` gets the same treatment. No rename change

before: `undertail.png` (shared)
after: `undertail.png`, `undertail-2.png`

added new tests `tests/test_duplicate_slug.py`

python -m pytest tests/test_duplicate_slug.py tests/_test_pga.py -v
5 + 16 passed

### DIFF
--- a/lutris/database/games.py
+++ b/lutris/database/games.py
@@ -166,11 +166,22 @@ def get_games_by_slug(slug: str) -> list[DbGameDict]:
     return _stringify_game_ids(sql.db_select(settings.DB_PATH, "games", condition=("slug", slug)))
 
 
+def get_unique_slug(base_slug: str) -> str:
+    """Return the first unused slug for a base: base, base-2, base-3, ..."""
+    slug = base_slug
+    counter = 2
+    while get_games_by_slug(slug):
+        slug = f"{base_slug}-{counter}"
+        counter += 1
+    return slug
+
+
 def add_game(**game_data: Any) -> str:
-    """Add a game to the database."""
+    """Add a game to the database, ensuring the slug is unique to avoid
+    coverart/config collisions (coverart/<slug>.png)."""
     game_data["installed_at"] = int(time.time())
-    if "slug" not in game_data:
-        game_data["slug"] = slugify(game_data["name"])
+    base_slug = game_data.get("slug") or slugify(game_data["name"])
+    game_data["slug"] = get_unique_slug(base_slug)
     return str(sql.db_insert(settings.DB_PATH, "games", game_data))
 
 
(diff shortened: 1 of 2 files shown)


danieljohnson2:
I don't really like this; the slug is not the key for the game in your local database (that's the ``id``) but it is used for looking up the game on Lutris.net- which provide default artwork. Claude thinks this messes with Lutris library sync as well.

If you actually want the same game with different artwork, the solution is to change how we handle custom artwork at a more basic level, I think.

But it is not clear to me what you are actually trying to do here. Why have two undertales?

loucass:
thanks for looking but quick clarify, I didn't add two Undertails.
What happened is that I added Undertail then create a duplicate and rename it to `Plants vs Zombie`. The new entry kept slug='undertail', so now I have:
Plants vs Zombie (slug=undertail) + Undertail (slug=undertail)
both pointing to same coverart/undertail.png — change one poster, other changes too.
Totally get your point about slug being for lutris.net lookup + library sync, so forcing undertail-2 would break that, I;m sorry my bad.
ig better fix is naing artwork by id, not slug. I'm going to change this approach and work on that instead.
store custom artwork per game with its id so same slug can still have different art, and default art from lutris.net still works via slug.
are u okay with that ?
@danieljohnson2

danieljohnson2:
OK, but what's the goal here? If you are duplicating Undertale to create Plants vs Zombies, shouldn't you update the slug of the duplicate to ``plants-versus-zombies`` or whatever it is?

loucass:
ok got u ig i will close it but would you be open to storing custom artwork by id instead of slug?
So same slug could still have different posters, and default art still comes via slug. I know that's bigger work with migration, so totally okay to refuse if you don't want that direction.

If not, I'll just re-add my games correctly locally and we're done. Thanks

loucass:
ok got u ig i will close it but would you be open to storing custom artwork by id instead of slug?
So same slug could still have different posters, and default art still comes via slug. I know that's bigger work with migration, so totally okay to refuse if you don't want that direction.

If not, I'll just re-add my games correctly locally and we're done. Thanks

danieljohnson2:
I'd look with favor on a rework of the media storage- what I would like to see is a separation between the custom art (by id) and the Lutris.net downloads (by slug).

But it is a bigger job, and I wouldn't merge it - it goes beyond a bug fix. The boss would have to decide.

loucass:
@danieljohnson2  ok Thanks.

My opinion is that separating custom artwork (by game ID) from Lutris.net artwork (by slug) is the cleaner long-term solution.

So I will work on this and do my best to do it.

Issue #6869, opened 2026-09-05
Can't install Xenia (404 downloading Github release)


GiovanH:
### Bug description

Lutris requires installing Xenia through Lutris's dedicated screen, but it's hardcoded a Github release URL that no longer works, so Xenia cannot be configured. Downloading Xenia manually and specifying a path does not work as a workaround; Lutris needs to manage the installation.

### How to Reproduce

On a fresh installation of Lutris, try to download and configure Xenia through Settings > Runners.

### Expected behavior

Lutris can automatically install and enable the Xenia runner. 

Alternatively, provide an option for the user to download the xenia_canary_windows zip themselves and provide it to Lutris without requiring a live internet connection or a pre-determined expected path for builds.

### Log output

```shell
[INFO:2026-09-04 19:12:42,598:application]: Starting Lutris 0.5.22
[ERROR:2026-09-04 19:12:42,604:system]: ['/usr/bin/vulkaninfo', '--summary'] command failed: [Errno 2] No such file or directory: '/usr/bin/vulkaninfo'
[INFO:2026-09-04 19:12:43,102:startup]: "card0" is No GPU (1002:163f 1002:0123 amdgpu) Driver 26.1.5
[INFO:2026-09-04 19:13:08,674:downloader]: https://github.com/xenia-canary/xenia-canary-releases/releases/latest/download/xenia_canary_windows.zip returned a 404 error
[ERROR:2026-09-04 19:13:08,674:downloader]: Download failed: 404 Client Error: Not Found for url: https://github.com/xenia-canary/xenia-canary-releases/releases/download/02d2cb5/xenia_canary_windows.zip
Traceback (most recent call last):
  File "/app/lib/python3.13/site-packages/lutris/util/downloader.py", line 149, in async_download
    response.raise_for_status()
    ~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/app/lib/python3.13/site-packages/requests/models.py", line 1021, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 404 Client Error: Not Found for url: https://github.com/xenia-canary/xenia-canary-releases/releases/download/02d2cb5/xenia_canary_windows.zip
[INFO:2026-09-04 19:15:40,126:runner]: Download canceled by the user.
```

### System Information

```shell
Vulkan support: YES
Esync support: YES
Fsync support: YES
Wine installed: YES
Gamescope: NO
Mangohud: NO
Gamemode: YES
[...]

PR #6870, opened 2026-09-05
WIP: Get latest Xenia Canary win build w/ GH API

GiovanH:
One approach to addressing https://github.com/lutris/lutris/issues/6869

Not yet verified.

### DIFF
--- a/lutris/runners/xenia.py
+++ b/lutris/runners/xenia.py
@@ -1,6 +1,7 @@
 """Xenia Runner (Wine-based)"""
 
 import os
+from functools import cached_property
 from gettext import gettext as _
 
 from lutris import settings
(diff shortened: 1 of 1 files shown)


danieljohnson2:
This is rather fancy, and we've already commits a fix here that just uses the url ``https://github.com/xenia-canary/xenia-canary/releases/latest/download/xenia_canary_windows.7z``.

Is that not enough?

GiovanH:
I think -- evidently -- the url isn't stable, and with no way for users to manually install a bundle this needs to be somewhat resilient to an upstream change. Especially with manual releases,

mihawk90:
Since this just came up in Discord too:
Is the download via the runner even necessary? There are (maintained) Windows and Linux installers available here:
https://lutris.net/games/xenia-emu/

Could the runner not just trigger those? If necessary those installers could be adjusted to move the binary where Lutris expects them. That way users don't need to wait on Lutris updates every time the Xenia team feels like changing the URL scheme again.

If users turn up again with the installer broken, the fix could be shipped on the fly via installer update, just like it did an hour ago.

FWIW they use the `latest` download link on their website too:
```html
<a href="https://github.com/xenia-canary/xenia-canary/releases/latest/download/xenia_canary_windows.7z">canary</a>

<a href="https://github.com/xenia-canary/xenia-canary/releases/latest/download/xenia_canary_linux.AppImage">canary</a>
```

PR #6886, opened 2026-09-10
docs: Mounting directory as a drive/disc

mihawk90:
2 small corrections:

* the config path was replaced in 1d40df3cf2eb4d8595eed151c77ccc9a497035ad, but that's not where game config files live (unless that's just because I'm on an older/migrated install?)
* `delete_registry_key` was seemingly a copy-paste mistake and listing unrecognised parameters that would lead to an error

1 new block:

Mounting directories as a drive has been done in various installers now, but there is no documentation on doing it. It's mostly useful when extracting ISOs or when users already copied the data off a disc and have it in a plain directory structure.
Forcing CD-ROM mode is needed by some (weak?) DRMs that check for a "physical" disc instead of a directory structure.
Some stronger DRMs will still refuse to work because they require access to the physical hardware and not just a drive mapping (for reading sectors outside the filesystem), but that is not something we can solve.

### DIFF
--- a/docs/installers.rst
+++ b/docs/installers.rst
@@ -448,6 +448,67 @@ Example::
     - insert-disc:
         requires: diablosetup.exe
 
+Mounting a directory as a drive or disc
+---------------------------------------
+
+Wine (and by extension Proton) recognises entries in ``<prefix>/dosdevices`` and will
+add them as drives automatically. So in `most` cases simply creating a symlink like
+this is enough::
+
+    - execute:
+        command: cd "$GAMEDIR/dosdevices" ; ln -s "../CD_CONTENT_DIRECTORY" "i:"
+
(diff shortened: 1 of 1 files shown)


mihawk90:
Note that 329c0ff depends on #6887 being merged or this won't work and only pass an empty string.

Technically what this deletes is what Lutris calls a `key` in `set_regedit`, which is actually the "value name" in Regedit.

But calling it "key" would conflict with the parenthesis since `delete_registry_key` actually does delete a "key" in Regedit terminology.

If you want I can document this discrepancy in terminology between Lutris and the regedit GUI. Changing this in Lutris would break existing installers using `set_regedit`, unless you'd want to keep the current parameter names for backward compatibility, which is a whole different mess.

Issue #4765, opened 2023-03-05
Wine prefix created before checking if exe exists


Zykino:
### Bug description

The EGS disappeared from my PC and I see no way to re-install it.
```sh
$ ls /mnt/Data/Games/Lutris/epic-games-store/drive_c/Program\ Files\ \(x86\)/
'Common Files'  'Internet Explorer'  'Windows Media Player'  'Windows NT'
```
Apparently lutris try to start EGS from `/mnt/Data/Games/Lutris/epic-games-store/drive_c/Program\ Files\ \(x86\)/Epic Games/Launcher/Portal/Binaries/Win32/EpicGamesLauncher.exe`, but as seen above, there is no "Epic Games" folder there.
I do not find a way to re-install EGS. Starting a game that was installed say that the binary is not present/no permission to start it.
And trying to install a new EGS game give the same kind of error.

_Notes:_
* /mnt/Data is a different hard drive than / (which contains /home)
* in the `lutris -d` output I first try to start a game that is shown as installed, then I try installing a game
* **Auxiliary bug:** the logs from `lutris -d` are in french and any accentuated letter is missing, ex: êéèàù
> Le programme n'a pas pu tre lanc, ou aucun programme n'est associ au fichier spcifi.
ShellExecuteEx a chou: Chemin introuvable.

should be:
> Le programme n'a pas pu **ê**tre lanc**é**, ou aucun programme n'est associ**é** au fichier sp**é**cifi**é**.
ShellExecuteEx a **é**chou**é**: Chemin introuvable.

### How to Reproduce

Steps to reproduce the behavior (I think it is what removed it):
1. Install an EGS game (install the launcher, install the game inside, leave all, start the game, play, repeat)
2. a. Try installing a new game (in my case First class Trouble) that fail the installation.
2. b. Cancel the install when you see 
> Monitored process exited.
Initial process has exited (return code: 256)
Exit with return code 256
2. c. See the popup asking you to cancel the install. (I think I checked the box "Delete the game files". I don’t really remember)
3. Try starting a game already installed -> EGS.exe not found
4. Try installing a new EGS game -> EGS.exe not found

### Expected behavior

Canceling an install should remove the game’s files, not the launcher. Even if the checkbox is checked (or the French translation is non-obvious to me).
If the launcher is uninstalled, there should be a way to re-install it.

### Log output

```shell
INFO     2023-03-05 11:06:13,028 [startup.init_lutris:173]:Starting Lutris 0.5.12
[...]

PR #6894, opened 2026-09-14
Validate Wine game executable before prelaunch

Joes156:
This patch validates the configured Wine executable before prelaunch and prevents prefix creation when the game path is unavailable. It adds a default validate_game hook to Runner, calls it before prelaunch in Game.launch, and implements the check in the Wine runner. Tests, mypy, Ruff, compile checks, and manual reboot testing passed. Related to #4765.


mihawk90:
Maybe I'm having a moment, but why would the game.exe exist before creation of the prefix, when the game is only installed after the prefix is created?

Normally an install does the prefix creation, then runs the installer in the prefix - which creates the game.exe - then eventually finishes and lets you launch with the specified game.exe. So how would the game.exe exist before prefix creation?

danieljohnson2:
As I read this, it checks when starting the game, not when installing- so this is all after normal prefix creation anyway.

This seems to just move the check we have to an earlier point. Seems okay, not a lot value. Shouldn't break installers at least.

Joes156:
Thanks both — yes, this applies to launching an already installed game, not to installation.

The purpose of moving the existing check earlier is to avoid Wine prelaunch side effects when the configured executable is unavailable, for example because the drive containing the game is not mounted. Currently, prelaunch can attempt prefix creation or modification before the executable check in play() reports the missing file.

With this change, that launch attempt reports the existing MissingGameExecutableError before entering prelaunch. In my manual missing-path test, the early check was reached and the missing prefix path was not recreated.

The scope is deliberately small: it does not change installer ordering, mount drives, or resolve every cause of games appearing as missing.

mihawk90:
OK so this:

> prevents prefix creation

would be more accurate to call *re*creation I guess, that makes more sense then.

